PR #8, opened 2018-12-20
Sortable member UI -- with webpacker added to project

jrochkind:


jrochkind:
Needs staging server to have `yarn` before merge.

PR #15, opened 2019-01-10
Front end tests (wip)

eddierubeiz:


Gemfile
```diff
@@ -68,6 +68,7 @@ gem 'bootstrap4-kaminari-views'
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'pry-byebug', platforms: [:mri, :mingw, :x64_mingw]
+  gem 'equivalent-xml'
```
jrochkind: is this gem being used?

spec/rails_helper.rb
```diff
@@ -22,6 +22,22 @@
 #
 # Dir[Rails.root.join('spec', 'support', '**', '*.rb')].each { |f| require f }
 
+#Eddie adding
+require 'webmock/rspec'
```
jrochkind: what motivated adding webmock stuff in this PR?

jrochkind:
nice job adding FactoryBot, thanks!

Gemfile
eddierubeiz: Good catch! It isn't.

PR #33, opened 2019-02-01
A couple metadata updates

eddierubeiz:
- Adding "Figurines" and "Money" to the list of genres
- Adding "editor" to the list of creator categories
- Alphabetizing and updating the list of exhibitions


jrochkind:
thanks!

PR #34, opened 2019-02-04
Add "Attributed to"; alphabetize maker metadata slugs.

eddierubeiz:
Trivial change; I just missed this one piece of metadata.
Also alphabetizing while we're in the file.


jrochkind:
yeah, I guess the order they are listed here is the order they show up in the menu. I'm not sure if that was intentional, but it's fine. We could also have the menu alphabetize regardless of order here.

Issue #61, opened 2019-02-20
Work validation bug: "undated", with a blank date, should be accepted.


eddierubeiz:
However, "undated" should *always* be combined with a blank date.
Add tests with this change.
The validations are in the date_of_work.rb model.

eddierubeiz:
Fixed in master and in import branch

PR #66, opened 2019-02-22
WIP Enforce blank dates for "undated" items; reject them otherwise.

eddierubeiz:
Don't merge this before we have tests for the feature.

### DIFF
--- a/app/models/work/date_of_work.rb
+++ b/app/models/work/date_of_work.rb
@@ -7,7 +7,14 @@ class Work
 
     validates :start_qualifier, inclusion: { in: START_QUALIFIERS, allow_blank: true }
     validates :finish_qualifier, inclusion: { in: FINISH_QUALIFIERS, allow_blank: true }
-    validates_format_of :start, with: /\A\d{4}(-\d{2}(-\d{2})?)?\Z/, message: "must be of format YYYY[-MM-DD]"
+    validates_format_of :start, with: /\A\d{4}(-\d{2}(-\d{2})?)?\z/,
+      message: "must be of format YYYY[-MM-DD]",
+      unless: Proc.new { |d| d.start_qualifier == 'undated' }
+
+    validates_format_of :start, with: /\A\z/,
+      message: "should be left blank if you specify 'undated'.",
+      if: Proc.new { |d| d.start_qualifier == 'undated' }
+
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Fixes #61 .

jrochkind:
Looks good. Do you want me to write the tests?  Are you thinking model tests on Work and/or `DateOfWork` class, or integration tests? I think model tests are probably easier and fine. 

Let's get this done and merged, if you don't want to write the tests now I will!

eddierubeiz:
Yes -- if you don't mind writing the test, I'd appreciate it. I would suggest a model test on the DateOfWork class.
We should assert:
a) blank start date w/ "undated" should now be legal
b) non-blank start date w/ "undated" should now be illegal
Thanks much.

app/models/work/date_of_work.rb
```diff
@@ -7,7 +7,14 @@ class DateOfWork
 
     validates :start_qualifier, inclusion: { in: START_QUALIFIERS, allow_blank: true }
     validates :finish_qualifier, inclusion: { in: FINISH_QUALIFIERS, allow_blank: true }
-    validates_format_of :start, with: /\A\d{4}(-\d{2}(-\d{2})?)?\Z/, message: "must be of format YYYY[-MM-DD]"
+    validates_format_of :start, with: /\A\d{4}(-\d{2}(-\d{2})?)?\z/,
```
eddierubeiz: Note: the marker for the end of a line is \z, not \Z, so I changed it this too.

jrochkind: I think \z and \Z actually mean subtly different things, not sure which is correct, let me google... 

https://stackoverflow.com/questions/2707870/whats-the-difference-between-z-and-z-in-a-regular-expression-and-when-and-how

> \Z  The end of the input but for the final terminator, if any
> \z  The end of the input

Weird, not sure which of those is 'right', but `\z` seems likely, thanks!

eddierubeiz: I now realize it makes zero difference in this case actually, but \z is theoretically more correct.

jrochkind:
@eddierubeiz think I should add another one for if `undated` in `start`, `finish` must be blank?

app/models/work/date_of_work.rb
jrochkind: I guess it could make a difference if a newline somehow erroneously made it into the string, and your correction is correct!

eddierubeiz:
Totally! I'll amend the validation code accordingly.

PR #97, opened 2019-03-12
blank out arrays before adding to them

jrochkind:
I had noted some cases in kithe staging that seemed to have _double_ external IDs.

I noticed that the previous "empty_ararys" method neglected to blank out external IDs. Still, I couldn't actually reproduce the problem
in the test. Nevertheless, it seems better to blank out the array in every method that is about to set an array-model value, together
with that value, instead of an "empty_arrays" method. Keeping em together seems more clear when you forget one.

### DIFF
--- a/app/importers/importers/generic_work_importer.rb
+++ b/app/importers/importers/generic_work_importer.rb
@@ -13,7 +13,6 @@ module Importers
     end
 
     def populate()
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Makes a lot of sense! Thanks.

PR #110, opened 2019-04-15
New passenger restart using systemctl

sanfordd:
This updates the restart command for passenger to using the `systemctl restart passenger` command. Right now with our passenger setup running outside apache and not just placing all the files in `/tmp` using the normal passenger command fails unless we set the `PassengerInstanceRegistryDir`.

The downside to just setting it in capistrano is any future changes in Ansible will cause cap deploys to break. By using the slightly less elegant `systemctl` call we only need to make sure that Ansible builds the systemctl.service file for passenger correctly and any changes will happen in one place.

The change required adding a tweak to the sudoers to allow our deploying user to restart passenger.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -19,6 +19,8 @@ set :deploytag_utc, false
 
 # use 'passenger-config restart-app' to restart passenger
 set :passenger_restart_with_touch, false
+set :passenger_restart_command, 'sudo systemctl restart passenger'
+set :passenger_restart_options, ""
 
 # send some data to whenever
 #set :whenever_identifier, ->{ "#{fetch(:application)}_#{fetch(:stage)}" }


jrochkind:
Works for me if you're okay with giving the unix account deploying with cap that sudo permissions. 

Do you think that's better than just setting `PassengerInstanceRegistryDir` though? 

I'm curious what command it was using before, to document it here... looks like by default it was doing `passenger-config restart-app`. https://github.com/capistrano/passenger  I got no problem with it using `sudo systemctl restart passenger` if you think that's the right command for it to use to restart passenger.

jrochkind:
@sanfordd Is this going to break other passenger admin commands we might want to use?  If so, will just setting `PassengerInstanceRegistryDir` fix all of them too, and might we want to do that instead? I don't know, just asking!

jrochkind:
Cap deploy worked fine with passenger restart with this code.

jrochkind:
Note the relevant informative message from passenger on the original failure, with the alternate workaround:

> passenger-config stderr: *** ERROR: Phusion Passenger doesn't seem to be running. If you are sure that it is running, then the causes of this problem could be one of:

>  1. You customized the instance registry directory using Apache's
    PassengerInstanceRegistryDir option, Nginx's
    passenger_instance_registry_dir option, or Phusion Passenger Standalone's
    --instance-registry-dir command line argument. If so, please set the
    environment variable PASSENGER_INSTANCE_REGISTRY_DIR to that directory
    and run this command again.

PR #112, opened 2019-04-19
PDF derivatives

eddierubeiz:
The code is simpler due to the much more sane way deriv generation is organized in New Thing, but the functionality is exactly the same as in Sufia.

Includes simple tests.


app/derivative_transformers/kithe/vips_cli_pdf_to_jpeg.rb
```diff
@@ -0,0 +1,120 @@
...
+    def call(original)
+      files = (0..2).map {|x| Tempfile.new(["kithe_vips_pdf", ".jpg"])}
+      # Make a slightly smaller thumbnail to accomodate the border we're adding.
+      width_minus_borders = @max_width - (2 * @border_pixels)
+      # Create and sharpen a thumbnail, then add a black border to it.
+      thumbnail( width_minus_borders, original, files[0])
+      sharpen(   width_minus_borders, files[0], files[1])
+      add_border(width_minus_borders, files[1], files[2])
+      # The width of the thumbnail, including the border,
+      # is now exactly @max_width.
```
jrochkind: while Tempfile is theoretically cleaned up by ruby when it gets Garbage Collected, if we know we aren't going to need files[0] or files[1] again, we can make things easier on ruby by calling `files[0].unlink` and `files[1].unlink`

https://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html#method-i-unlink

Also, nit-picking, but as long as I'm here, it might be less confusing to use individually named local variables. Cleaning up that `map` a bit too...

thumbnail, sharpened_thumb, with_borders_thumb = 3.times.map { Tempfile.new(["kithe_vips_pdf", ".jpg"]) }

jrochkind:
one nitpicky comment. Looks awesome, nice job!

jrochkind:
@eddierubeiz actually also can you add some comments at the top of the class explaining what it does (extract page 1 as jpg, sharpen, add border), and that it is intended/used for PDFs whose page 1 is a text title page. 

(And if you still have any URL references around that you used for reference in figuring out the sharpening etc would be useful to add em, but no need to look for em if you don't have em around still)

PR #114, opened 2019-04-23
Ask travis to cache solr_wrapper solr download directory

jrochkind:
Among other things hoping to avoid getting rate-limited by apache.org for downloading it on every CI build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ cache:
   bundler: true
   directories:
     - /home/travis/.rvm/
+    - ./tmp/solr_dist
 before_install:
   - gem install bundler
 before_script:
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -20,7 +20,7 @@ class WorkIndexer < Kithe::Indexer
 
     to_field "text4_tesim", obj_extract("description")
 
-    to_field ["text_no_boost_tesim", "langauge_facet"], obj_extract("language")
+    to_field ["text_no_boost_tesim", "language_facet"], obj_extract("language")
     to_field "text_no_boost_tesim", obj_extract("external_id", "value")
     to_field "text_no_boost_tesim", obj_extract("related_url")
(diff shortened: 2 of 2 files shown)


jrochkind:
It's failing with an error unrelated to this change. It's a Capybara error saying when it tried to click on something, it couldn't, because the something was covered by the admin navbar `<div class="admin-header sticky-top d-flex justify-content-between p-2 mb-4 mt-1 bg-light border">...</div>`. 

OK.. but why is this applying only to this PR, which didn't touch anything related, and not to others with the same HTML?

jrochkind:
Aha, it has no started affecting other PR's. Dont' know what changed, but time to debug it!

jrochkind:
https://github.com/titusfortner/webdrivers/issues/73

jrochkind:
Aha, I did figure out how to figure out what versions of chrome/chromedriver are being used. Indeed on the succesful runs, it was Chrome/chromedriver 73, on the failed runs, it was Chrome/chromedriver 74. 

Chrome is being installed by travis the same way, but this afternoon travis decided to start installing Chrome 74. (Which I think Google Chrome says is still in beta?)  My Mac doesn't even seem to offer me the option of upgrading to Chrome 74. 

On my local Mac, I have Chrome 73. So the difference in Chrome version could be the failure. 

I can try forcing travis to install Chrome 73 instead, if I can figure out how. But we don't want to stay on an old version of Chrome forever, we like that it's always tested on the recent Chrome. (And on your Mac, it will always be tested on whatever version of Chrome you have, with Chrome auto-updating, so it should be recent). 

I'm not sure why the travis "addons: Chrome: stable" directive started installing Chrome 74. I will file a ticket with travis.

jrochkind:
AND I got my MacOS chrome to update to version 74, and indeed it's failing locally for me. 

Something with headless chromedriver API changed between 73 and 74, and broke one of our tests. 

Not an easy thing to debug still, but at least I'm reproducing locally.

jrochkind:
Worked around in #115

jrochkind:
Might as well try it to see if it helps with our travis build times.

PR #115, opened 2019-04-24
Fix chromedriver 74.0's failure to scroll to reveal something covered by a position:fixed element

jrochkind:
As of Chrome/chromedriver 74.0, chromedriver will refuse to click on something
if it's covered up by something with "position: sticky", even if scrolling
could uncover it. Chromedriver will normally scroll to reveal something to click
on it, and chromedriver 73.0 handled it fine.

I think this is probably a chromedriver bug, but don't know how/where
to report it. For now, this is a workaround, we occasionally hackily scroll
to top so chromedriver can find whatever we wanted to click on, at random places
where it was failing otherwise.

### DIFF
--- a/spec/system/work/new_work_spec.rb
+++ b/spec/system/work/new_work_spec.rb
@@ -2,12 +2,22 @@ require 'rails_helper'
 require 'pp'
 
 RSpec.describe "New Work form", :logged_in_user, type: :system, js: true do
+
+  # As of Chrome/chromedriver 74.0, chromedriver will refuse to click on something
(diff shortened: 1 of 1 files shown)


jrochkind:
I did try to alert the chromedriver people, although I'm not really positive it's a chromedriver bug or issue. 

https://groups.google.com/forum/#!topic/chromedriver-users/0Pu-TPvlu18

Don't know how that listserv works, might not get any response, but felt like it was my duty to try to tell someone!

PR #117, opened 2019-04-24
Initial Blacklight integration

jrochkind:
So this PR integrates Blacklight into our app in a _basically_ working fashion, but certainly not even close to an end-user usable UI yet, or styled for our brand. 

There were some tricks integrating BL into an already existing app; and trying to _limit_ BL's effect on our app to only the parts we actually want to use, in the places we want to use them.  We pretty much only want to use a `CatalogController`, and whatever else it insists upon to work. We already were not using "bookmarks", "saved searches", and other interactive functionality included with Blacklight on top of search/results behavior. 

## The install process

The BL installer wants to add gems to our `Gemfile` that were already there, which then makes `bundler` complain about listing the same gem twice, and keeps the installer from moving on to it's next steps. So I had to comment out some gems, get the installer to complete, then manually reconcile the Gemfile. 

I maybe should have run BL installer without `devise`, but I ran it with the `devise` option, but we already had devise, so had to deal with it trying to add conflicting routes and other devise things, similar to above. 

The things the BL installer tried to do to our existing `User` model totally messed it up, again had to manually reconcile. 

BL tried to add `solr_wrapper` rake tasks to our app, but we already found that `solr_wrapper` rake tasks were not useful, so added our own instead. In #109. Eliminated solr_wrapper rake tasks. 

The way that BL tries to set up your application.js and application.css for it's JS/CSS assets was also conflicting with our current stuff (and/or leading to double-imports of things like bootstrap), so I reorganized there. 
* popper.js was gene
* BL tries to give you the twitter-typeahead-rails gem -- I think this is only used for built-in BL "suggest" feature which we aren't currently using. and twitter-typeahead-rails is an unsupported project. We are trying commenting this out of Gemfile, as well as the include of it's js in our sprockets manifest. 

## Post-install, trying to minimize Blacklight's impact on our app

**We don't plan to use the bookmarks feature**. It looks like without using this feature, you can avoid quite a few places Blacklight ordinarily injects things into apps. 
*  Eliminating things generated in `config/routes.rb` to route to a BookmarksController.
* We also provide a `render_bookmarks_control?` method in CatalogController that returns `false`, trying to keep bookmarks out of our app. 
* No `Blacklight::User` mixin in our User class, it seemed to be only bookmarks-related. 
* Don't seem to need the `devise_guests` gem at all. It, and a related `current_or_guest_user` method, seem to only be used by bookmarks features. (This should avoid having "guest" users added to our `users` table as well). 
* Also lets us avoid several db migrations adding new tables/columns to our app (see below). 

In routing, also couldn't figure out why it was wanting to `mount Blacklight::Engine => '/'` at all -- it has separate instructions to route to CatalogController. Eliminated mount of the whole engine at route, to avoid interfering with the rest of our app. 

Not adding the `search_history_controller.rb` that BL generates, our sufia app has never used it. 

Blacklight generates an `include Blacklight::Controller` in ApplicationController. We don't want this infecting our app, so took it from there and put it in _just_ our local CatalogController. Similarly, Blacklight put in a custom way to choose the Rails layout based on BL configuration in ApplicationController; we just use normal Rails code to explicitly choose layout. 

## Avoiding migrations

We did need blacklight's `create_searches` migration, to create a `searches` table that BL is insisting on for our basic search. 

But we are trying to _avoid_ other Blacklight db migrations that may make changes to our db we don't need. `create_bookmarks`, and `add_polymorphic_type_to_bookmarks` -- we are not using bookmarks functionality, so hope we can get away without the migrations to add a `bookmarks` table. 

`add_devise_guests_to_users` -- adds a boolean "guest" column to your users table. BL seems to be working okay without this, since we aren't using bookmarks, and seem to be able to not use `devise_guests` at all. 

## TODO finish up

- [x] can we get away without using devise-guests?
- [x] simple rspec smoke test of search results page
- [x] need whenever/crontab for cleaning up BL search table
- [x] should create separate tickets for additional SearchBuilder customizations from sufia
- [x] ticket for add bL range limit
- [x] note on javascript/css in README


jrochkind:
@eddierubeiz this is nearing readiness to merge into master (although it's still _bare bones_ and nowhere *close* to a usable end-user interface). 

Would you like to begin reviewing the code? And maybe try it out locally?  

Here's what you could do to try it out locally:

0. Optional, but I like having the large 'real' corpus in my dev copy. Get one of those directories of import/export json, and then run `DISABLE_BYTESTREAM_IMPORT=true ./bin/rake scihist_digicoll:import` to import _without_ the actual asset bytestreams. (Should take about 20 minutes only I think?)

1. Start a dev solr server if one isn't started already with `./bin/rake solr:start`

2. Index everything you got in your dev instance with `./bin/rake scihist:solr:reindex` (shouldn't take more than ~30 seconds)

3. Start your Rails server and look at the really bare bones Solr interface at eg `http://localhost:3000/catalog`. You can see the "hits" are _not_ being displayed in a useable way. But basic search is there, with most of our facets, our sort fields, etc.

eddierubeiz:
Getting closer! I can see the interface at /catalog, but postgres is 'searches' table. Shouldn't you include a db migration to add that?

jrochkind:
What does "but postgres is 'searches' table" mean?

It means you get an error that the table doesn't exist?

Yeah, I was trying to see if I could avoid the migrations, and thought I had successfully done so, but I guess I must have run them locally and forgotten about it, without checking them in. Thanks for testing it out, I'll look into it further. (By rebuilding my DB from scratch, and seeing what migrations we need. BL is trying to put some in that we _don't_ need/want, I think, have to figure out the minimum).

eddierubeiz:
Oh, sorry, what I meant to type is "postgres is missing a 'searches' table". OK -- keep me posted.

jrochkind:
My spec isn't working for some reason, but I think the branch should work for you now @eddierubeiz

eddierubeiz:
Great! It works -- I see the basic functionality, with searches and faceting.

PR #121, opened 2019-04-25
Use aws_sdk to discover servers with specific tags and automatically use their IP addresses for deployment.

sanfordd:
This is an initial working change to allow us to stop recording IP addresses or domain names in our deploy files.

Currently it uses the `aws_sdk` gem to scan all of our EC2 instances. It filters on the Role tag and Service_level tag to find all machines we've tagged with a value and passes the IP address to Capistrano via the server syntax.

The current requirement it has that is not in the Gemfile or something we programmatically add is the need for your aws/config or aws/credentials file to have your IAM account details listed as [default]. I'd like to eventually have it look for a specific profile like deploy and fall back to default if deploy is missing but that's for the future. 

There is a filter (instance-state-code) that will only target running instances which should avoid problems in trying to deploy to servers which are stopped, in the process of being shut down, or recently terminated.

## note

If we had discovered it before we started writing this, we might have considered using https://github.com/fernandocarletti/capistrano-aws

We would have had to add filtering on `instance-state-code` to just running instances, because it lacks that feature. And maybe other stuff we wanted. 

We could still consider switching to that in the future.


jrochkind:
`aws-sdk` is kind of a "master" gem that has a LOT of `aws-sdk-*` gems as it's own dependencies, and installs all of them. 

Can we identify what part of aws-sdk we actually need, and only depend on that?  That will keep our build times and sizes smaller. 

For instance, some things that used to depend on `aws-sdk` too now depend on just `aws-sdk-s3` cause that was all they used.

sanfordd:
That makes sense. We should only need `aws-sdk-ec2` so I'll fix that.

README.md
```diff
@@ -18,6 +18,7 @@ Prequisites:
 * Postgres installed and running -- on MacOS, I like https://postgresapp.com/
 * `yarn` installed for webpacker -- on MacOS, `brew install yarn`.
 * vips installed --  on MacOS `brew install vips`
+* Have your AWS credentials put in the default profile at ~/.aws/config
```
jrochkind: i guess running a cap deploy didn't AWS credentials at all before?

This is a standard place to put AWS credentials?

For a developer running an app locally, we keep AWS credentials somewhere else (local_env.yml or your bash ENV). Not sure if having two places is gonna be a problem... especially because the AWS credentials we use to run a dev app may _not_ have permission to do what's needed here! May need more thought, my head is swimming. 

But I already have a `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID` -- how does that interact with `~/.aws/config`?

config/deploy/staging.rb
```diff
@@ -1,10 +1,19 @@
...
-server 'kithe.sciencehistory.org', user: 'digcol', roles: [:web, :app, :db, :jobs, :solr, :cron]
-#server '34.227.26.131', user: 'hydep', roles: [:solr]
-#server '54.204.60.1', user: 'hydep', roles: [:jobs, :cron] # dzi bg job box
-
+#Edit ec2 for region changes, right now we only use one region.
+ec2 = Aws::EC2::Resource.new(region:'us-east-1')
+#Server role keys should be the Role tag (assigned by ansible) that you want to deploy to. The array value is the list of capistrano roles that the server needs.
+server_roles = {"kithe"=>[':web', ':app', ':db', ':jobs', ':solr', ':cron']}
+server_roles.each do |key, value|
+  current_server_roles = "#{value}"
```
jrochkind: I don't think there's a reason to take `value` and put it in a new variable as `"#{value}"`?  If you want to call the variable `current_server_roles`, just call it that above in `do |key, current_server_roles|`, I think?  (although I think ti's always _singular_ so should maybe be `current_server_role`?)

config/deploy/staging.rb
```diff
@@ -1,10 +1,19 @@
...
+#Service level is set manually here, maybe make it a variable further up to be easy to spot when making new stages?
+  ec2.instances({filters: [{name:'instance-state-code', values:["16"]},{name: 'tag:Role', values: ["#{key}"]},{name: 'tag:Service_level', values: ['stage']}]}).each do |ip|
```
jrochkind: what does the magic string `"16"` mean? Explain in comment?

sanfordd:
Correct, prior versions did not need those credentials. The AWS_ACCESS_KEY and SECRET_KEY are what is stored in the credentials (along with default regions and other env options). The `~/.aws/config` is the default location such values are stored in. If we can better leave them as env_vars I'm more than fine with that.  You'll need the same values that correspond to your user account on AWS, which has access to all the needed permissions. 

The magic 16 is the value for a running aws instance, I'll add it to the comment.

With the #{value} to variable, I wanted to make sure it passed down to the loop below, that might be an artifact of a much more complicated idea I had earlier. I can try and simplify it.

jrochkind:
I merged master into this branch, to get #115, which should fix test failure.

config/deploy/staging.rb
```diff
@@ -1,9 +1,38 @@
...
+  creds= YAML.load_file(secret_location)
+  Aws.config[:credentials] = Aws::Credentials.new(creds['AccessKeyId'],creds['SecretAccessKey'])
+  ec2 = Aws::EC2::Resource.new(region:"#{aws_region}")
+#Server role keys should be the Role tag (assigned by ansible) that you want to deploy to. The array value is the list of capistrano roles that the server needs.
+  server_roles.each do |key, value|
+#Service level is set manually here, maybe make it a variable further up to be easy to spot when making new stages? 
+#The instance-state-code of 16 is a value from Amazon's docs for a running server. See: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
+    ec2.instances({filters: [{name:'instance-state-code', values:["16"]},{name: 'tag:Role', values: ["#{key}"]},{name: 'tag:Service_level', values: ["#{service_level}"]}]}).each do |aws_server|
+#Search across the tags and find the one labeled capistrano_roles, tags are hashes with 2 values, key for tag name and value for tag value.
```
jrochkind: @sanfordd Did we talk about putting in another tag that limited to the "kithe" application, like `tag:application` maybe?  In case we start using this system for other applications, we only want to pull back EC2 hosts for _this_ one.  Or is that there and I'm missing it? (although I'd rather not call the value `kithe`, that is not the name of this application. We could call call it `digicol2` maybe?)

jrochkind: Oh wait, i see the `key` variable is that. Can you change the name to `tag:Application` instead of `tag:Role`, and use a value that describes our application rather than `kithe` which is just one dependency of it (which could hypothetically change)?

I also notice you are using uppercase-first-letter for some tags `tag:Role`, `tag:Service`, but lowercase for "capistrano_roles" below. Make em consistent?

sanfordd: I'd prefer not to use numbers in the application name since we're also using them for the specific server identification digcol2-1, 2-2, 2-3 seems likely to introduce some confusion versus name-1, name-2. digcol (or digicoll as we current have the repo named) would be fine.

jrochkind: Makes sense, I was just worried about confusing it with our current chf_sufia app which is also "digital collections", this new one is "version 2". We could come up with a snazzy internal codename for it, like Alberta calls theirs "jupiter" but that doens't seem likely to reduce confusion. I just don't like "kithe", becuase it's an implementation detail which could change. If you think `digicoll` is fine, fine with me. 

Also rename "Role" to "Application" makes sense for how we're now going to use it?

sanfordd: I'll add an application section, we do want to keep Role (or a tag akin to it) so we can distinguish between a solr server, web server, jobs server, or other specific purpose server in group of servers running the new application.

jrochkind: > so we can distinguish between a solr server, web server, jobs server, or other specific purpose server in group of servers running the new application.

As far as capistrano is concerned, that's what `capistrano_role` defines, no?  If you need the role for other purposes not related to capistrano, that's cool (although I don't understand what a role "kithe" means, but I don't know if I need to) -- but I don't think capistrano needs to know about it. Capistrano needs to know about all the servers in the "application" of concern, at the "service_level" of concern -- and then it needs to know what capistrano_roles they have, because that's how it knows how/what to deploy to each one. I think that's all capistrano needs to know about, not this additional "Tag:Role".  Am I missing anything?

sanfordd: The Application tag has been added and tested, and all tags are now following the convention of a capitalized first letter. Switching from prod/stage to production/staging is still in progress. There's a number of things that use those values.

sanfordd: @jrochkind If want to get this out sooner, the service_level tag should be easy enough to adjust in the future by swapping out a variable. Otherwise if there's no time pressure I can keep adjusting service_level naming schemes.

sanfordd: I'll work on adjusting it for staging to test with, due to work on the new server setup this will be delayed until next week.

sanfordd: The renaming worked and we're using the full production/staging terminology.

PR #128, opened 2019-04-30
Admin-only Search Fields

### LINKED ISSUE #124: SearchBuilder extension: only search restricted fields if logged in
Similar to https://github.com/sciencehistory/chf-sufia/blob/master/app/models/search_builder/restrict_admin_search_fields.rb , although we'll probably switch it to be a positive add instead of a negative subtract.

jrochkind:
Closes #124 

The Blacklight "SearchBuilder" is the object that creates the Solr query. 

We want it to include certain Solr fields as search targets only if there is a logged in user, to make sure non-staff isn't searching over staff-only fields. 

To do this, we needed features missing from BL 7, but added in BL 7.1.0.alpha, which we now use (hopefully 7.1.0 final soon). 

We tell Blacklight that `current_user` should be included in context that becomes available to our SearchBuilder. 

And we give the SearchBuilder a very simple extension to, based on presence of current_user, add in search fields.

Tests: We test with system tests (end-to-end browser tests). While these are slow... it's difficult to set up Blacklight-related tests otherwise, especially with confidence they are *really* doing what's expected and not false positives due to wrong assumptions in test setup (or assumptions that become wrong). This is the easiest way to write tests that reliably let us know it's doing what's intended -- making sure only admins can search admin-protected fields.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ ruby File.read(".ruby-version").chomp
 gem 'rails', '~> 5.2.1'
 gem 'webpacker', '~> 3.5'
 
-gem "blacklight", "~> 7.0"
+gem "blacklight", "~> 7.0", ">= 7.1.0.alpha"
 
 # Use postgresql as the database for Active Record
 gem 'pg', '>= 0.18', '< 2.0'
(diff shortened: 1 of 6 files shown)


jrochkind:
Done and ready for review/merge whenever @eddierubeiz , assuming tests are green!

PR #129, opened 2019-05-01
Tell kithe to index friendlier_id to Solr id field, instead of pk id

jrochkind:
After you have an app with this change (including on staging), must run:
    ./bin/rake scihist:solr:delete_all scihist:solr:reindex

(should take under 30 seconds)


jrochkind:
reindexed on kithe.sciencehistory.org

PR #132, opened 2019-05-02
Provenance metadata field.

eddierubeiz:
- Provenance field
Still to do:
- [x] Remove provenance_notes field -- turns out we don't really need it.
- [x] Since our work display page isn't ready yet, create a new issue to ensure the display of provenance, along with its notes, will be consistent with the corresponding features in Sufia. We will want to reuse some of the unit tests in Sufia.


jrochkind:
Do porting over description field formatting (assuming it's not here already? I forget) as a separate PR of it's own please, with just that.

jrochkind:
Actually maybe hold off on porting over the formatting of description fields in general. I'm still working out the patterns we'll want to use for view/helper-layer stuff. It can wait until we have a front-end, and we can do it for all of em at once.

jrochkind:
Hey @eddierubeiz , I actually think tiny PR's are _great_ and preferable to omnibus PR's. Smaller PR's are easier to review, less likely to cause problems, keeps separated concerns sepearated instead. And merging changes as they are done means they can show up for stakeholders to see as they are done, instead of having to wait for 'all' of them. 

Having an "omnibus" _issue_ to keep track of the work makes sense maybe? But what do you think about maybe doing that but still PR'ing and merging work as it's done?  Like if this PR has work that is done, I'd love to merge it now, I don't see any reason to wait until there's _more_ work that needs to be reviewed too.

jrochkind:
Oh wait, I just realized this was _scihist_digicol_, where we're trying to _match_ what we did in chf_sufia. 

NEVERMIND. I get it now, this makes sense thanks @eddierubeiz !

jrochkind:
Nicely done. I don't love how many places there are you to have to add a new field -- and especially that stuff in the BatchCreateController and WorksController. But I guess the trade-off is always some kind of over-complicated abstraction. And it works, and you seem to have found all the places even though you didn't write the code (good job on BatchCreateController, I think i would have forgotten that one!), so!

PR #138, opened 2019-05-14
ViewModel approach experiment

jrochkind:
We use draper to create "view models" (not sure if this is the right term we want to use) that
are ruby objects that can render HTML, such as in this case: 

   WorkResultDisplay.new(work).display # => renders a work 'hit' in search results
   CollectionResultDisplay.new(collection).display # => renders a collection hit in results

We use polymorphism, so both WorkResultDisplay and CollectionResultDisplay use the same ERB view template, but backed by the different 'view model' objects, providing different values to go in the template slots. 

The document_list template uses another kind of polymorphism, with a hetereogenous list of results that can be Works or Collections, to decide which 'view model' class to use to render a given hit. 

The view models can use Rails helper methods (including 'render', 'content_tag', and 'safe_join') by virtue of magic from the [draper](https://github.com/drapergem/draper) gem. Draper's magic also sets up test environment so we can write isolated tests on the 'view models' that Just Work.

We aren't doing too much to 'extend' Draper, we are just using Draper in a somewhat different way than the README guides you to -- not as a decorator that adds methods to a model, but as a 'view object', 'display object' or 'view model' object that takes a model (or other object) and renders it as HTML on a page. There can be multiple view models associated with a given kind of model (like work), to render different things based on it, there is no assumed 'main' one for a given model class. There are parts of Draper we simply don't use, like the automatic '#decorate' method, or decorating associations. It's a pretty simple use of Draper mainly for it's magic giving access to Rails helpers in the "view_context", in live app as well as testing. 

This is a work in progress demonstrating the architecture, for displaying search results. It's not done. In addition to having no CSS styling, it's still missing some parts of search results, such as thumbnail, "Num Items" count, description, and metadata 'list' values.

* Not sure if "view model" is the right terminology. One place this effects is what directory we put the objects and their tests in. Right now `app/view_models` and `spec/view_models`. Could be "presenters", "renderers", "displayers", I dunno! All these words end up being used differently by different people even within Rails use.


jrochkind:
Hmm, especially if we are putting non-draper-based objects-that-format-for-presentation in the same place... and thus calling them by the same name (?)... I wonder if a better name than "view models" is preferable. 

Renderers?  Displayers?  I think "renderers", except we can't call the main method in draper-based ones "#render", because it conflicts with the Rails `render` method. 

"Presenters" might make sense, but in Rails land that word is often used for "decorator"-style use  of draper. I dunno. 

"Displayers"? Any opinions?

jrochkind:
OK, here's what I'm thinking about naming. 

I think "presenter" is actually the right word for these things -- both ones based on Draper so they can call Rails helper methods and render ERB templates. And "pure ruby object" ones. 

See for instance [this blog post](https://nithinbekal.com/posts/rails-presenters/) one of the first things you get googling "Rails presenters", which is doing a very similar thing. 

I _think_ it _does_ make sense to keep draper-based and non-draper-based ones in the same directories. So `app/presenters` for the actual objects, and `spec/presenters` for the tests. They're pretty much the same thing, whether draper/rails helpers are used is an implementation detail, the caller still does `Something.new(something).some_method` either way.  

But it might make sense to use a different name for the draper-based ones instead. Maybe we leave the draper based ones called "view models" (this is a term the draper gem itself uses among other things, it says it's "Decorators/View-Models for Rails Applications"). I think the way "View Model" is used generally is closer to when we use an ERB template, which we can only do with our draper-based ones. 

So the base class will remain `view_model`, and we'll still keep templates for draper-based view models in `app/views/view_models`. Our draper-based "View Models" are, thus, one particular kind of "presenter". 

Naming, one of the hardest things in programming!

What do you think @eddierubeiz ?

eddierubeiz:
So. Taxonomically, makes perfect sense. Yes, the base class should be `view_model`. However, in practice, if you're gonna put the **objects** in `app/presenters` and the **tests** in `spec/presenters`, consider putting the **templates** in `app/views/presenters` rather than `app/views/view_models`. Yes, any presenter with a template must necessarily be a `view_model`. But in practice, nine times out of ten we're going to be dealing with these three directories in parallel, so I would argue for naming them `presenters`.

PR #139, opened 2019-05-14
port date-formatting for display code from chf_sufia

jrochkind:
Tests not currently passing... becuase it turns out the tests were buggy and testing nothing in current chf_sufia. **fixed**

commit https://github.com/sciencehistory/chf-sufia/commit/fb3b73881a6361567b7d1064326c775282a837ac

Accidentally makes test be a no-op, setting 'expected_dates = correct_results', then checking to make sure expected_dates matches correct_results. At some point after this, code was accidentaly broken to do capitalization that does not match what the tests expect, and the broken test did not warn us. Have not tracked down that commit yet. 

I guess we should fix the code to match specs including expected capitalization, rather than fix the specs to accept accidentally broken capitalization. 

Trivial fixes to capitalization did not work. The code is a bit convoluted, could perhaps be simplified generally. 

Alternately, we can cange the specs to allow 'faulty' behavior, as it's been in production for a while apparently...

### DIFF
--- a/app/view_models/date_display_formatter.rb
+++ b/app/view_models/date_display_formatter.rb
@@ -1,7 +1,7 @@
 # Takes an array of `Work::Date` objects (generally from `some_work.date_of_work`),
 # and formats them for display to the user.
 #
-#     DateDisplayFormatter(work.date_of_work).display_dates
+#     DateDisplayFormatter.new(work.date_of_work).display_dates
 #     # => Array of strings, each being a human presentable date/range, like "1901", or "after 1950 -- circa 1990"
 #
 # Copied from old app chf-sufia/app/presenters/chf/dates_of_work_for_display.rb
--- a/app/view_models/date_display_formatter.rb
+++ b/app/view_models/date_display_formatter.rb
@@ -40,6 +40,11 @@ class DateDisplayFormatter
       date_string = "#{date_string} (#{date_of_work.note})"
     end
 
+    # Want to capitalize first letter only, and only if it's in position 0
+    if date_string && date_string[0] =~ /\w/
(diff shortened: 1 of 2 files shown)


jrochkind:
Note also there is a `capitalize` method that is never used -- I imagine it was intended to be. 

But note also the `capitalize` method with special logic to capitalize just the first letter of the string -- is unnecessary. The built-in ruby 'capitalize' method already does this. `"some string with words".capitalize == "Some string with words"` `capitalize` already just capitalizes the first letter. (unlike `upcase`). No need for a special logic with conditions on length of string.

jrochkind:
also original tests are missing a test case for 'undated', one should be added (and I think may initially fail)

jrochkind:
I identified where the 'test bug' was introduced, making the test no longer test anything. But I believe the bug that changed capitalization from what the test expected was actually introduced sometime later, and have not found that. It might (or might not) be useful to look at history of the relevant class in chf_sufia to see where the bug that changed capitalization was introduced.

jrochkind:
relevant line at one point:

https://github.com/sciencehistory/chf-sufia/commit/36971b6b4569477690eea44dbb8e154cdedb48cc#diff-a0c3713d2c218b8d1b1a2c0c4653c4c8R37

I see why it wasn't just using "capitalize".  `"(circa).capitalize == "(Circa)"` which was not desired, is why it was trying to slice. ruby capitalize built in capitalizes the _first letter found_, even if it's not at position 0.  At some point later perhaps fixing a bug involving nil or short strings, the capitalize was dropped altogether maybe. May be an easy fix, introducing that line again.

PR #144, opened 2019-05-29
Description display

eddierubeiz:
This is a port of chf-sufia/app/helpers/description_formatter_helper.rb, with tests.

Note: Just as in Sufia, this code creates links out of any bare URLs it finds, adding external-link icons to them, but leaves untouched any link *tags* entered into metadata.

This is an artifact of
 - the way things worked *before* link tags were allowed in the Sufia description field
 - obsolete communications guidelines.

We may later decide to overhaul the content such that descriptions and provenances contain no bare links, and thus do not need to be processed like this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -80,7 +80,7 @@ gem 'devise', "~> 4.5" # user accounts and login
 gem 'access-granted', "~> 1.0" # authorization
 
 # decorating and truncating html
-gem "rails_autolink"
+gem "rails_autolink", '~> 1.0'
 gem 'html_aware_truncation', '~> 1.0'
 
 group :development, :test do
--- a/Gemfile
+++ b/Gemfile
@@ -79,6 +79,10 @@ gem 'bootstrap4-kaminari-views'
 gem 'devise', "~> 4.5" # user accounts and login
 gem 'access-granted', "~> 1.0" # authorization
 
+# decorating and truncating html
+gem "rails_autolink"
+gem 'html_aware_truncation', '~> 1.0'
+
 group :development, :test do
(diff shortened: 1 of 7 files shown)


jrochkind:
- [ ] Description metadata can be `nil`, so there should be a test for that. (I guess it should return `nil`? Or empty string?) (Maybe also a test for empty string `""` as input, which should definitely return empty string). 

- [ ] Now that we've merged the new feature in ViewModel superclass, use `valid_model_type_names [String, NilClass]`?

I can do either or both of these things, or you can @eddierubeiz !

Gemfile
```diff
@@ -79,6 +79,10 @@ gem 'bootstrap4-kaminari-views'
 gem 'devise', "~> 4.5" # user accounts and login
 gem 'access-granted', "~> 1.0" # authorization
 
+# decorating and truncating html
+gem "rails_autolink"
```
jrochkind: Current rails_autolink version is `1.1.6`, so this should be `gem "rails_autolink", '~> 1.0'` to limit to `1.x` version (not that I expect a 2.0 to ever be released, it's pretty much abandoned, but this is good Gemfile hygeine). 

The only cases I don't do it are in some cases where gem only exists in pre 1.0, so it's unclear what the version specification should be, although even in those cases I'm leaning toward `'< 2.0'` in favor of entirely unrestricted.

jrochkind:
@eddierubeiz Looking good, Okay, just one more (I think! :) ). 

If this PR is going to contain change to app/views/presenters/_index_result.html.erb , as it does, to make description actually show up on results page...

Please edit our single basic "smoke test" of results in catalog_search_spec.rb. It does not need to test all the edge cases or really anything except the description is _there_ on the page.

PR #145, opened 2019-05-30
Correct "initializer" to "initialize" in the view_model class

eddierubeiz:
So the model validation code can run.


eddierubeiz:
This PR somehow ended up including other files I didn't want, so I'm closing it.

PR #148, opened 2019-05-30
Add "N items" to search results page

jrochkind:
For works, this is number of 'members'. For collections, it's number of 'contains' assoc destinations (we use different AR associations for work children (1-N) and collection belonging (N-N). 

The naive approach would end up having a separate SQL query executed for each item on results page, to fetch it's "child count for display". That would be bad. (SQL N+1 problem). 

So we provide a ChildCountDisplayFetcher that can make ONE SQL query to fetch all the counts for everything in result set. (Actually it potentially does two, one for collections#contains and one for works#members, but they are triggered lazily on demand so only if both collections and works are in result set). 
* It uses a bit of fancy ActiveRecord to do a 'group by' query to fetch counts.

Now we have to make this available somehow at view time... to our relevant view models probably. We need to initialize a ChildCountDisplayFetcher with the IDs of everything in the current results page (so it can batch fetch them)... so it winds up in the controller, where that info is easily available. This does introduce some unfortunate coupling between controller and our 'view models' (need to mock the helper method we add to provide a ChildCountDisplayFetcher in specs for the view models). But it works, and seems good enough for now. 
* I did think of some rather clever ways to use a Blacklight SearchService here instead. But I only thought of it after this apporach was complete; I think it would actually result in _more_ lines of code, although better encapsulation/loose coupling/right things knowing about each other... but this is done and working and not *too* bad, so I'm thinking this is good enough for now.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -109,6 +109,7 @@ group :test do
   gem 'factory_bot_rails'
   gem "database_cleaner", "~> 1.7"
   gem "webmock", "~> 3.5"
+  gem "db-query-matchers", "< 2.0"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -177,6 +177,9 @@ GEM
       safe_yaml (~> 1.0.0)
     crass (1.0.4)
     database_cleaner (1.7.0)
+    db-query-matchers (0.9.0)
+      activesupport (>= 4.0, <= 6.0)
+      rspec (~> 3.0)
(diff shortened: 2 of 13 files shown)


eddierubeiz:
OK this makes a lot of sense. Good to see you can do complex join / group-by stuff in the Rails ORM; not going to spend a ton of time investigating the details but I get the general idea.

jrochkind:
What made this one especially tricky was it was a _self-join_, members/contains are both relations from `kithe_models` *to* `kithe_models`, a self-join. Still made it work, woo.

PR #150, opened 2019-05-31
Make ChildCountDisplayFetcher dependency explicit in initializer

jrochkind:
Small refactor of that display num children stuff.

The Work/Collection ViewModels already had a 'dependency' on a ChildCountDisplayFetcher being available. (To efficiently access child count numbers). But it was a kind of "implicit" dependency/requirement, they required a helper method called `child_count` to be available in the current Rails "view context", which we had the Rails controller provide.

This gave me a bad feeling from the start, introduced coupling between the view models and the controller, in a not entirely obvious way, tried to get around that with copious documentation. Required mocking in the unit tests for the view models to "pretend" the helper was available, which required relying on some odd rspec features and draper-provided test support.

I realized we can instead just make the view models require a ChildCountDisplayFetcher to be passed in to their initializers. Now the dependency/requirement is totally clear, and actually enforced by ruby. Doing this in a spec becomes trivial. And we can have the _document_list partial we were already overriding construct the batched ChildCountDisplayFetcher, and pass it in to all the view models.

While this requires logic in the _document_list partial, when ideally you don't want logic like this in the view... overall I think it's an improvement, keeping the view models nice and self-contained. (I suppose the controller _could_ still provide it in helper method that the _document_list partial _uses_ to pass to the view models... the view models would still be nice and self-contained. I think this is probably good enough how it is, but if you understand what I'm saying and think it would be an improvement I can do that.)

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -45,12 +45,6 @@ class CatalogController < ApplicationController
   end
   helper_method :view_model_class_for
 
-  # Used by our ViewModels for displaying results
-  def child_counter
-    @child_counter ||= ChildCountDisplayFetcher.new(@response.documents.collect(&:id))
-  end
-  helper_method :child_counter
-
   configure_blacklight do |config|
     ## Class for sending and receiving requests from a search index
     # config.repository_class = Blacklight::Solr::Repository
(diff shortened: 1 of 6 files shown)


eddierubeiz:
I think app/views/catalog/_document_list.html.erb is still perfectly readable; I'd rather have the overall structure be simpler overall.

PR #151, opened 2019-05-31
Provenance display

eddierubeiz:
A view_model for provenance display, along with a simple template.
The template isn't used anywhere yet but we can drop it into the work view page whenever that's ready.
Includes a test for the view_model.


app/presenters/work_provenance_display.rb
```diff
@@ -0,0 +1,33 @@
...
+
+
+  # To split the provenance field into regular provenance info and a set of notes,
+  # use as a delimiter the string NOTES, in all caps, followed by an optional colon.
+  #
+  # The regular expression also captures extra blank space
+  # before and after the match, as well as before and after the word "NOTES".
+
+  def split_provenance
+    raise ArgumentError.new("Don't call this on a blank string.") if model.provenance.blank?
```
jrochkind: I think this argument check in unneccessary, especially on a method that should be only called "internally", it's not meant as public as API.  (Could it be marked `private` instead? I guess the reason it's not is becuase you want to test it?)

app/presenters/work_provenance_display.rb
```diff
@@ -0,0 +1,33 @@
...
+    DescriptionDisplayFormatter.new(@provenance_notes).format
+  end
+
+
+  # To split the provenance field into regular provenance info and a set of notes,
+  # use as a delimiter the string NOTES, in all caps, followed by an optional colon.
+  #
+  # The regular expression also captures extra blank space
+  # before and after the match, as well as before and after the word "NOTES".
+
```
jrochkind: eliminate empty line between comments and method. In addition to being consistent style, the empty line would prevent the `yard` automatic documentation generator from realizing the comments are documenting the method.

app/views/presenters/_work_provenance.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    <div class="collapse" id="collapseProvenanceNotes">
+      <%= view.provenance_notes %>
+    </div>
+<% end %>
+
+<%#
+  TODO: add css when we include this template.
+  #collapseProvenanceNotes {
+    margin-top: 1em;
+  }
```
jrochkind: The CSS should be added in the same PR, no reason to leave a note explaining they need to be added instead of adding them. It's true that we don't have a CSS file for it yet; we can pick one now, can always change it later. Maybe `show_page.css` or `work_show.css` or `show_item.css`? I dunno, can always be changed, but we should get the CSS in now.

spec/presenters/work_provenance_display_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails_helper'
+
```
jrochkind: There are only tests of the `split_provenance` method, which isn't actually the public API. 

We need at least minimal test(s) of the `display` method that is actually the public API.  1. Test that for a typical "happy path" case, it produces the HTML we expect. 2. Test that for `nil` input, it returns an empty string. 

See existing view model tests for how to use Nokogiri to parse the output, and then use capybara testers `have_selector` or `have_content` to test output. Or you can just use nokogiri manually. We can discuss more if unclear. 

An alternate approach would be trying to use the actual `#display` output for _all_ tests. Less convenient cause you have to deal with the HTML when you really just want to test "did it split the summary from the notes right?", but avoiding testing on what's really an internal method. 

Or perhaps a medium ground, your existing tests could be written in terms of the `provenance_summary` and `provenance_notes`  methods to test exactly what you want without having to deal with HTML apparatus, but not use the more internal `split` method -- which could then be marked private. Right now, `provenance_summary` and `provenance_notes` are only _available_ after you call `display`, so writing the tests this way might require some refactoring there, so calling those methods would trigger splitting and population of the ivars. 

```ruby
# lazily make sure split happened and ivars are filled
def populate_parts
   unless @populate_parts
      @provenance_summary, @provenance_notes = split_provenance
      @populate_parths = true
   end
end

  def provenance_summary
    populate_parts
    DescriptionDisplayFormatter.new(@provenance_summary).format
  end
```

That would also have the benefit of allowing this class to be easily used by a caller that wants to split provenance notes and summary for some non-HTML purpose, and won't be calling display. 

But these are just an ideas, arguably existing tests can stand as they are -- but still needs at least minimal tests of `#display`

The non-nil `#display` test should definitely use *real sample data* that looks like real data, not placeholder text like `"provenance metadata"`.

spec/presenters/work_provenance_display_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  it "correctly splits the provenance" do
+    expect {
+        WorkProvenanceDisplay.new(work_0).split_provenance
+    }.to raise_error(ArgumentError)
+
+    #no notes
+    expect(WorkProvenanceDisplay.new(work).split_provenance).
+      to contain_exactly("provenance metadata")
+
+    work.provenance = "provenance metadata \n\n NOTES: \n\n notes"
```
jrochkind: I believe a space before and after `\n\n` is _atypical_ data?  Someone entering or copying or pasting isn't going to wind up with a space before or -- especially -- after.  A space after the `\n\n` would mean a space before the newline:

```
some provenance summary

NOTES:

 there's a space at the beginning of this line

but maybe not before this one
```

Am I right that's atypical data?  If it's atypical, it shouldn't be the pattern used in all the tests, most of the tests should have no space before/after `\n\n`, so we're testing typical data. If you want a test or two testing the atypical edge case  with spaces to make sure that's handled too, that's cool too, but it shouldn't be every test. 

In fact, ideally, these tests should be written with _real sample data_.  At least one/some, if not all of them, should be actual data from our sample records, not placeholder "provenance metadata".  Testing with real data instead of placeholder data makes it more likely that the test data is representative, and not dissimilar in some way that would cause a break on real data but not test data.

jrochkind:
@eddierubeiz In addition to comments I left with GH "review" feature, I want to discuss one more change. 

The ERB template is relatively small. We have the option of generating the HTML tags inline programmatically, instead of using an ERB template. I think this is small enough that it might make sense to do that (although any more complex than this it probably wouldn't). 

It might look something like this:

```
def html_content    
  parts = []
  parts << provenance_summary
  if provenance_notes.present?
     parts << content_tag("a", "Show notes", href: "#collapseProvenanceNotes", role: "button", :"aria-expanded" => "false", :"aria-controls" => "collapseProvenanceNotes")
     parts << content_tag("div", provenance_notes, class: "collapse", id: "collapseProvenanceNotes")
  end
  safe_join(parts)
end
```

Thoughts on this approach? It puts all the stuff right in the file, you don't have to go find the partial to look at. It is also somewhat more efficient for Rails -- partial templates can have performance issues, especially in development-mode, although we still use them where we need for code clarity. On the minus side, maybe it's more confusing to deal with, rather than the customary ERB. What do you think?

spec/presenters/work_provenance_display_spec.rb
eddierubeiz: Good call. I used nokogiri to parse the template output. Added a couple span tags to make testing easier.

eddierubeiz:
Hm... I defer to you of course, but my instinct here is to follow the MVC patter as long as the difference in efficiency is negligible in production. Separate templates may waste some CPU time in development, but they are easier to read, which makes us, as developers, more efficient. My two cents. :)

jrochkind:
Looking good. Looks like you still need a test on `nil` input?  since `work.provenance` can be nil/empty string.

jrochkind:
Also please click 'resolve' on the GH feedback comments from me that are now resolved!

app/assets/stylesheets/application.scss
```diff
@@ -28,6 +28,7 @@
 @import "local/member_sort";
 @import "local/error_pages";
 @import "local/member_list";
+@import "local/work_show";
```
jrochkind: Did you forget to add this file to the repo?

eddierubeiz: Yup. Oops!

eddierubeiz:
No, we already test for [nil input](https://github.com/sciencehistory/scihist_digicoll/blob/d5d603b30c750d364c3274ddc597741831f6772c/spec/presenters/work_provenance_display_spec.rb#L8). The decorator returns an empty string.

PR #154, opened 2019-06-03
Fix two small bugs in view_model and adding a rudimentary test

eddierubeiz:
Fixes a couple typos in the code handling constructor argument validation. Also adds a very simple test (we will add more soon but it's better than nothing).

### DIFF
--- a/app/presenters/view_model.rb
+++ b/app/presenters/view_model.rb
@@ -125,10 +125,10 @@ class ViewModel < Draper::Decorator
   # do include all Rails helpers in the ViewModel with:
   include Draper::LazyHelpers
 
-  def initializer(model, *other_args)
+  def initialize(model, *other_args)
     if self.class.valid_model_type_names
       unless self.class.valid_model_type_names.any? { |t| model.kind_of?(t.constantize) }
-        raise ArgumentError.new("#{self.class.name} arg must be of one of type #{self.class.valid_model_arg_types.inspect}")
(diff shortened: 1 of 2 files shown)


jrochkind:
cool cool. 

I'm not sure it's safe to just define classes like that in an rspec `describe` block -- I think they'll "leak" to the rest of the app. Like another spec will also have a `WidgetFormatter` defined. (this could be tested manually, wanna try?). And if another spec forgot/didn't know that and tried to define it's own `WidgetFormatter`, whichever ended second would have a mutant frankenstein `WidgetFormatter` with methods from both. 

Googling "rspec temporary class", I get [this](https://makandracards.com/makandra/47189-rspec-how-to-define-classes-for-specs), which suggests the same, but suggests the workaround is to do:

```ruby
describe Notifier do
  class self::TestRecord
    # ...
  end
  
  let(:record) { self.class::TestRecord.new }
  
  it { ... }
end
```

Wanna try that? 

Alternately, I did this weird `temporary_class` helper to try to get around this, especially when I was testing things where the class _couldn't_ be called `self::Whatever` without messing up ActiveRecord behavior that inferred things from class names. But that `temporary_class` stuff I did is weird and fragile, so probably best not used if the `self::Wahtever` works.

eddierubeiz:
Thanks for the research. As long as we don't define a view_model_spec_spec.rb I'm open to whatever really. I think I'll go with the self:: recipe -- stay tuned.

jrochkind:
Yeah, I'd try just doing what that blog post recommends with `self::` if you don't mind? The blog post explains why and what.

PR #155, opened 2019-06-03
index actual postgres pk in solr

jrochkind:
we are now using friendlier_id as the solr 'id', so don't need to index that separately. But will index actual postgres pk id separately, which would now not otherwise be in solr. helpful for debugging if nothing else.


eddierubeiz:
We'll use it all the time.

PR #158, opened 2019-06-06
switch solr restart to systemctl command

sanfordd:
Moved to the systemctl command to match what we're using for passenger and also be more forward prepared with any other systemd changes that might affect the service command in the future.


jrochkind:
Tested it with that branch and `solr_restart=true cap staging deploy`. 

It worked. 

```
00:52 chf:restart_or_reload_solr
      01 sudo /bin/systemctl restart solr
    ✔ 01 digcol@kithe.sciencehistory.org 7.464s
```

PR #162, opened 2019-06-11
add font-awesome

jrochkind:
Adding it with the same rails gem we used in chf_sufia. This is bootstrap v4, not the newer v5.

Decided it was not in our brief to spend time on figuring out v5 upgrade at this point. 

## Finding font-awesome icons

We are using v4, not newest v5. So a good directory of the v4 icons is at:

https://fontawesome.com/v4.7.0/icons/

## Notes on font-awesome and accessibility

It turns out that ALL uses of font-awesome icons via CSS classes should really have `aria-hidden="true"` on them. Eg:

    <i class="fa fa-address-book" aria-hidden="true"></i>

This prevents a screen-reader from trying to read the unicode codepoint that font-awesome is hijacking to be it's own icon. 

We normally try to use icons that are only decorative or supplemental to words already there, so it's okay if a non-visual browser simply skips it. 

If you additionally would want a non-visual browser to have some text in place of the icon, there are some patterns. See:

* https://fontawesome.com/v4.7.0/accessibility/
* https://fontawesome.com/v5.9.0/how-to-use/on-the-web/other-topics/accessibility  (for v5, but basic patterns still applicable)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,8 @@ gem 'sprockets-rails', '>= 2.3.2' # bootstrap gem requirement
 
 gem 'jquery-rails', "~> 4.3"
 
+gem 'font-awesome-rails', '~> 4.7'
+
 # temporary kithe indexing branch, for scihist_digicoll indexing branch, do not
 # intend to merge to master like this.
 gem 'kithe', git: "https://github.com/sciencehistory/kithe.git"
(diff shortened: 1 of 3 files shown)


eddierubeiz:
Thanks for the notes above. Good to have.

PR #165, opened 2019-06-14
Delete conflicting items of other types.

eddierubeiz:
Summary:

This will report (and, if you request it, delete) pre-existing items of *other* types if the pre-existing item's friendlier_id conflicts with that of the one we're trying to import.

For instance:
We want to import an `asset` with `friendlier_id` `a1b2c3` but we discover there is already a `work` with that same `friendlier_id` in the destination site. We remove the `work` to make room for the `asset`, consistent with our current policy that an import wipes conflicting information.
To report, but not delete, run as usual:
```scripts/import/import.sh```
Once you've moved the conflicting items out of the way, or just determined you don't want to keep them, you can tell the import to remove them automatically:
```KEEP_CONFLICTING_ITEMS=false scripts/import/import.sh```

*****

Details by file:

**In importer.rb:**

- added / refreshed the method documentation
- reordered the three main methods -- `import`, `preexisting_item?` and `target_item` so they appear (to me at any rate) in their logical order. I can imagine refactoring these three methods, but decided against further changes for now.
- updated the `target_item` method to correctly handle conflicting items. If it finds a conflicting item and the `KEEP_CONFLICTING_ITEMS` flag (`true` by default) is in fact `true`, it sets the `@conflicting_item_was_kept` flag and returns from the method. If it finds a conflicting item and the `KEEP_CONFLICTING_ITEMS` is `false`, it deletes the conflicting item and continues as usual.
- added a `check_for_items_with_same_id` method to isolate the process of figuring out what items have the same friendlier_id
- added a new ivar `@conflicting_item_was_kept` as a flag. It gets set if a conflicting item was found AND NOT DELETED, so that the import process knows to skip this item.
- renamed boolean ivar `@preexisting_item` to `@preexisting_item_was_found` -- this was confusing me to no end because I kept imagining it referred to the preexisting item itself.

**In the tests:**

- Added tests to assert that the default options were indeed set to the correct defaults
- Tested that the import behaves as expected when a conflicting item is present, depending on how the `KEEP_CONFLICTING_ITEMS` flag is set.

**In the import.rake task:**

- Checked for said flag and pass it as an option to FileSetImporter, GenericWorkImporter and CollectionImporter.

### DIFF
--- a/app/importers/importers/importer.rb
+++ b/app/importers/importers/importer.rb
@@ -25,10 +25,45 @@ module Importers
     # options:
     # keep_conflicting_items: default true, if false will delete conflicting items to make room for incoming items.
     def initialize(metadata, options = {})
-      #raise ArgumentError unless target_item.is_a? self.class.exportee
       @metadata = metadata
       @errors ||= []
-      @keep_conflicting_items = !!options[:keep_conflicting_items]
+      @conflicting_item_was_kept = false
+      @keep_conflicting_items = options[:keep_conflicting_items].nil? ? true : options[:keep_conflicting_items]
+    end
+
+
+
+
+
+    # This is the only method called on this class
+    # by the rake task after instantiation.
+    # It reads metadata from file, creates or finds
+    # an item based on it, blanks out some of its metadata for freshening up from import if needed,
+    # then saves it to the database.
+    def import
+      self.class.without_auto_timestamps do
+        if preexisting_item?
+          blank_out_for_reimport(target_item)
+        end
+        return if @conflicting_item_was_kept
+        common_populate
+        populate
+        save_target_item
+      end
+    end
+
+    # Is there a preexisting item with the same friendlier_id and the same type?
+    def preexisting_item?
+      # Calling target_item:
+      #   * checks for any conflicting items
+      #       * deletes them if it is allowed to,
+      #       * sets @conflicting_item_was_kept otherwise.
+      #   * creates or sets @target_item (without blanking it out -- that's done in the import method.)
+      #   * sets boolean ivar @preexisting_item_was_found
+      target_item
+      # If a conflicting item still exists, return false.
+      return false if @conflicting_item_was_kept
+      @preexisting_item_was_found
     end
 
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jrochkind** [comment]
It's a bit surprising to me this one is needed. 

Do you have a hypothesis about what causes it?

I'm guessing that an item was 'manually' (not via an export) created in the new app, that used a friendlier_id that conflicts with an item in old app (the item in old app could have been already created or created after the one in new app, the important thing is it wans't imported to new app at the time the 'manual' item was created in new app). 

I wasn't expecting this to be an issue, because friendlier_id's in new app are created randomly, and from a pretty large 'key space', so I was thinking conflicts should be rare. 

But I wonder if the random generator in new app is somehow the same one in old app with the same seed, making conflicts actually expected. Although that seems surprising to me. 

The downside of this workaround is that -- I was hoping that the import mechanism could be counted on to _update_ items in new app that represent the *same thing* as an item in old app (possibly losing any local customizations), but not to *delete* any independent work you had done in new app. So it would for instance be safe to be doing real work in new things in new app, and keep syncing things that exist in both. 

But with this change that's no longer safe -- any work you do in new app, even on unique items not imported from old app -- could be erased/lost on an import. 

So i'm a bit troubled, and not sure what to do. We could just give up on this goal, and say the switch from old app to new app has to be complete and clean. We could do more investigation to make sure we understand exactly what's going on. We could change this workaround to *change* the friendlier id of a conflicting item, rather than delete a conflicting item. (All you have to do to change a friendlier ID is update the attribute in the usual Rails way, the only trick is writing code to find a non-conflicting new value).

**Comment by eddierubeiz** [comment]
The conflicting items have to be manually created in the new app. Otherwise, the conflicts would already have existed in the old app: there would have to have been e.g. a `GenericWork` and a `FileSet` with the same NOID in Sufia, which doesn't make sense to me.

I think we can actually have it both ways. We could just tell the import to skip (and report) all items where it sees a conflict. Then we add a `DESTROY_CONFLICTING_ITEMS` option to force the destruction of conflicting items.

**Comment by jrochkind** [comment]
OK. I don't understand how the conflicting item created in the new app ends up _conflicting_ though. Especially with such regularity. Since friendlier_id's assigned in the new app are randomly chosen from a very large space, i'd expect conflicts to be rare. 

So ideally I'd like to understand more what's going on. Like verify *for sure* that when there's a conflict: a) the item in new app *really was* manually created (not from an import) *after* the time of last improt, and b) the conflicting item in old app *really was* created after time of last import too. 

Because even though we are assuming this is the case (because we can't think of any explanation) -- I'm a bit worried that it *could* be some other bug in the import/export, so would ideally like to get more evidence to support our hypothesis. 

The skip-and-report is certainly less dangerous. We might not even need a "DESTROY_CONFLICTING_ITEMS" flag. Rather, when there are conflicts, we manually review them, and then manually either delete them in new app or change their friendlier_ids in new app, and then run the import again. Might be safer to require manual conflict resolution, with individual attention to conflicts.

**Comment by jrochkind** [comment]
And actually, here's a problem with our hypothesis:

friendlier_id's automatically created for new items in new app are actually a *different number of characters* than the ID's from old app!

This was kind of a mistake (I originally meant them to be the same number of characters), but actually I kind of like it, because it should make this kind of conflict *impossible*. 

Example, in new app I just created a new work, it winds up with friendlier_id: `h1ynm5r` -- 7 chars. ID's in old app are actually all have 9 chars, eg `nv935386s`. 

So I believe the kind of conflict we are hypothesizing *ought* to be impossible. A friendlier_id assigned for a newly created item in new app should not be able to conflict with a NOID id from sufia, because they should be a different number of chars!

(New app does not enforce any particular character length of friendlier_ids, just that they be unique. But the ones it automatically generates will be of particular character length). 

So I'm a bit disturbed that we may not really understand what's going on, and it could be an as yet not understood bug in export/import.

File: app/importers/importers/importer.rb [hunk 1]
```diff
@@ -192,5 +206,8 @@ def self.without_auto_timestamps
       Kithe::Model.record_timestamps = original
     end
 
+    def same_friendlier_id_different_type
+      Kithe::Model.where(friendlier_id:metadata['id']).where.not(type:self.class.destination_class.to_s).first
+    end
```
jrochkind: I don't like that this is doing an extra SQL fetch for every single imported record. 

It seems like there should be a way to handle this (unusual) case without an extra fetch for every single object to import. 

Two possible approaches:

1. If the current code has a problem where it gets an error on _save_, because of the conflicting friendlier_id -- rescue that error, and recover from it *there*. So the extra Database work only happens when there really is a conflict. 

OR

2. If right now when looking for an existing work, we're doing `Work.find_by_friendlier_id`, you can change that code to `Kithe::Model.find_by_friendlier_id` -- the object will *still come back* as class `Work` if it is one. Then, at _that_ point, you can check `object_that_came_back.kind_of?(ClassIExpect)`, and if it's _not_, do the extra work there again (to skip or delete it whatever). 

Either of these two approaches should make it possible to handle conflicts (including by respecting a choice of either "skip" or "delete" strategies), without having to do an extra SQL fetch for every single import target.

eddierubeiz: Yeah, good point. Option 2 is nice; will do that.

app/importers/importers/importer.rb
```diff
@@ -22,63 +22,104 @@ class Importer
...
+
+    # Is there a preexisting item with the same friendlier_id *and* the same type?
+    def preexisting_item?
+      # Calling target_item:
+      #   * checks for any conflicting items
+      #       * deletes them if it is allowed to,
+      #       * sets @conflicting_item_was_kept otherwise.
+      #   * creates or sets @target_item (without blanking it out -- that's done in the import method.)
+      #   * sets boolean ivar @preexisting_item_was_found
+      target_item
```
jrochkind: This all seems overly complex and indirect, that you need to document here what some _other_ method does seems a sign that the given methods are not responsible for clear single things. It makes the code very hard to follow. It is also better design when methods return as return values the things they are supposed to return, instead of setting ivars as side-effects -- makes it hard to keep track of where the ivars are being set (could be multiple places), and who consumes them (could be multiple places again). 

All this code just seems overly complex. But it's not obvious to me what the solution is, and it's not clear if it's worth it for the import script. I'm not really sure what to do.

app/importers/importers/importer.rb
```diff
@@ -22,63 +22,104 @@ class Importer
...
     # but we're refactoring here.)
+    #
+    # If there are conflicting items, either removes them or sets @conflicting_item_was_kept.
     def target_item
       @target_item ||= begin
-        model_object = self.class.destination_class.where(friendlier_id:@metadata['id']).first
-        if (model_object)
-          @preexisting_item = true
+        items_with_same_id = check_for_items_with_same_id
+
```
jrochkind: This is also really confusing, what's going to be returned here, what we can expect. It's a hash? Which has one key, either :conflicting, or :preexisting?  And the value is... a model?  Or sometimes it returns nil? Or an empty hash?

I can figure this out following it all through the code, but it's a baroque API.

jrochkind:
@eddierubeiz Do you have the original version of this that did the 'extra' check?

That was reasonably simple. I feel like there ought to be a way to start from there, avoid the 'extra' check, but still keep things almost as simple without all this extra code. 

But I'm definitely feeling a bit burnt out on this stuff, and am not sure how much it's worth it for code we don't plan to use long-term. But I don't like all of the complexity being added here, lots of new methods, with unexpected return values and side effects (which are dependencies on things being run in the right order so the iVars are present).  Not really sure what to do about it. 

But one idea would be taking a look at the original simpler but extra-fetch code, and then working on it together.

jrochkind:
If supporting `KEEP_CONFLICTING_ITEMS` flag is making things more complicated, then I would tend back to my original suggestion, we don't need to have this as a flag, we can just _always_ keep conflicting items and handle them manually. Coding for features we don't really need at the cost of code complexity is not a good bargain. 

But if you kept it, you must think it's more important/useful to auto-delete them then i do, like it's going to take you lots of extra time to deal with on our interim periodic imports, without this flag?

It's also possible we can support the flag fine with simpler code. 

Here are my just brainstormed I will admit might not actually be practical ideas for making the code simpler, how I would have tried to handle it:

1. The code that fetches an existing item could simply be willing to fetch any Kithe::Model
2. Then "def import" just has to check `preexisting_item? && ! target_item.kind_of?(expected_class)` to see if it has a problem case. Then it just either deletes it (and re-sets target_item to a new item, ideally just by using the target_item method); or skips it and aborts. 

I feel like this should be do-able with just a handful of code, and no additional iVars. But there could be devils in the practical details I am not aware of. But on initial view, this code seems awfully complex and hard to follow/reason about/maintain.

jrochkind:
I believe you intended this to be closed in favor of #175 , @eddierubeiz ?  Can you confirm and if so close this?

eddierubeiz:
Yes-- we no longer need this PR. Keeping the branch, though, as it has some useful stuff including tests we might be able to use.

PR #175, opened 2019-06-20
Don't import if there's a conflicting item of other type.

eddierubeiz:
This is a simpler and better version of https://github.com/sciencehistory/scihist_digicoll/pull/165/commits/448e4e49fad6bbdc265f73ad19a895774ee7be81 .

### DIFF
--- a/app/importers/importers/importer.rb
+++ b/app/importers/importers/importer.rb
@@ -67,9 +67,12 @@ module Importers
     def import
       self.class.without_auto_timestamps do
         if preexisting_item?
+          if !target_item.is_a? self.class.destination_class
+            add_error("Skipping this item; found a conflicting #{target_item.type}")
+            return
+          end
           blank_out_for_reimport(target_item)
         end
-
         common_populate
         populate
         save_target_item


jrochkind:
Simple enough there's no reason not to merge!

I think we _could_ implement "delete conflicting" behavior too without too much trouble. But especially now that we think we know what causes this, I think we have more important fish to fry. 

* We need to figure out how we're going to wipe out all data for our actual live migration. The only hard part is the S3 stuff. Not sure if we should just wipe it all, or develop a script to 'delete orphans'. The latter will probably be cheaper (as well as faster). both have S3 expenses. 

* At some point in advance of final go-live, we should switch to importing from production probably. Not sure if we should just do this now? Not sure if the export from chf_sufia production will cause problems for production performance when occuring?  Do you have a sense?
  * Once we wipe the db for re-import, and switch to importing from production -- we don't expect to run into the duplicates problem, which we think is an artifact of importing items from staging that get wiped, and  then have the same ID's used for other items in production.

eddierubeiz:
Cool, thanks! re: importing from production, I've given this a bit of thought -- let's discuss in person.

PR #176, opened 2019-06-20
Adding school_of

eddierubeiz:
Note: the import audit explicitly checks for *all* the items in the Work::Creator::CATEGORY_VALUES array, so if an import file specifies a `school_of` and the resulting imported item has none or a different value, the audit will catch it.


jrochkind:
that's awesome and encouraging that the audit would already be catching it, and reporting problems prior to this PR.

jrochkind:
Does it *really* only require this one file to be edited at present? I think.. you are right?!?  Solr indexer will still index it, I think, yes? Wow!

eddierubeiz:
Yeah you know what else is awesome? It's a one-file change.

PR #183, opened 2019-06-26
Work show page

jrochkind:
Goal: Not yet taking account of audio, not yet including viewer or downloads, but basic page layout.

### DIFF
--- a/app/assets/stylesheets/_mixins.scss
+++ b/app/assets/stylesheets/_mixins.scss
@@ -33,7 +33,7 @@
 %text-font {
   font-family: $brand-serif;
   // 16px in abril is really small, 17px looks more like 16px in other fonts.
-  font-size: 17px;
+  font-size: 1.0625rem; // 17px if 16px==1rem;
   line-height: $paragraph-line-height;
 }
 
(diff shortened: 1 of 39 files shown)


jrochkind:
Looks like travis works with draft PR's now, as far as running checks? 

I'm not sure why it says "This branch is out-of-date with the base branch". It is not out of date with master as far as I can tell, trying to merge master locally says "Everything up-to-date".

jrochkind:
Haven't totally figured out how to do good tests for a major view like this with lots of functionality. 

Especially when capybara/browser tests are so slow and you don't want to do tons of them. 

This might suggest putting more stuff in easier to test separated ViewModels. 

But I only have so much time for architecture astronauting here. The tests will better than they were in chf_sufia, but probably won't get every detail of display. 

The most important thing is tests that make us confident no 500 will be raised.

jrochkind:
I'm going to call this one ready for review and merge @eddierubeiz . It's big enough already, I'll do any other work in additional PRs. 

Feedback or questions welcome!

eddierubeiz:
OK taking a look...

PR #185, opened 2019-06-27
Citable attributes port, with tests.

eddierubeiz:
A basic port of the citable_attributes object, along with all the tests.
Includes a default test setting for :web_hostname.
Note: does not include a port of app/models/chf/citable_attributes/renderer.rb. This will go in a separate PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -144,3 +144,9 @@ gem 'aws-sdk-core'
   # gem 'devise-guests', '~> 0.6'
 # end BL generated
 
+# we use for data structures for citation models, and for generating citations
+gem "citeproc-ruby", '~> 1.0'
+gem 'csl-styles', '~> 1.0' # Need to load the styles so we can use chicago
+# On MRI <= 2.3, citeproc-ruby insists upon `unicode` or `unicode_utils` gem. :(
+# https://github.com/inukshuk/citeproc/commit/c14d3cd272698dd4aa52625dd140864b7a7bd6cb
+gem 'unicode'
--- a/Gemfile
+++ b/Gemfile
@@ -145,8 +145,4 @@ gem 'aws-sdk-core'
 # end BL generated
 
 # we use for data structures for citation models, and for generating citations
-gem "citeproc-ruby", '~> 1.0'
-gem 'csl-styles', '~> 1.0' # Need to load the styles so we can use chicago
-# On MRI <= 2.3, citeproc-ruby insists upon `unicode` or `unicode_utils` gem. :(
-# https://github.com/inukshuk/citeproc/commit/c14d3cd272698dd4aa52625dd140864b7a7bd6cb
-gem 'unicode'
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -144,3 +144,9 @@ gem 'aws-sdk-core'
   # gem 'devise-guests', '~> 0.6'
 # end BL generated
 
+# we use for data structures for citation models, and for generating citations
+gem "citeproc-ruby", '~> 1.0'
+gem 'csl-styles', '~> 1.0' # Need to load the styles so we can use chicago
+# On MRI <= 2.3, citeproc-ruby insists upon `unicode` or `unicode_utils` gem. :(
+# https://github.com/inukshuk/citeproc/commit/c14d3cd272698dd4aa52625dd140864b7a7bd6cb
+gem 'unicode'
```
jrochkind: we won't be using MRI <= 2.3, so shouldn't need this. Try it without it?  I don't think this particular PR is actually using citeproc-ruby or csl-styles at all yet? so we won't know if it works without it or not -- that's fine, we can leave it for the PR that uses it. so remove ALL of these, or just remove `unicode` now cause we don't think we need it.

app/presenters/citable_attributes.rb
```diff
@@ -0,0 +1,569 @@
...
+  # work.place_of_publication
+  # In this code, use instead work_lookup(work, "place", "place_of_publication")
+  # work_lookup("place", "place_of_publication")
+  # => returns ["New York (State)--New York"]
+  # Returns empty array if main_attribute or sub_attribute are not present.
+  def self.work_lookup(work, main_attribute, sub_attribute)
+    work.send(main_attribute).
+      select{|p| p.category == sub_attribute }.
+      map { |y| y.value}
+  end
```
jrochkind: this works!

spec/factories/work_factory.rb
```diff
@@ -153,5 +156,37 @@
       end
     end
 
+    trait :oral_history_work do
+      title { "Oral history interview with William John Bailey" }
```
jrochkind: FactoryGirl lets you have an additional factory (that inherits from the base work) in addition to traits. I'm never really sure when to use one with the other -- the main thing traits give you is the ability to combine them, but we'd never want to combine "oral history work" trait with, say, a "museum object" trait, so it _probably_ would work best as an additional factory instead of a trait. 

But not a big deal, and not a blocker to merge if you don't feel like doing it or it seems too confusing.

Gemfile
eddierubeiz: Oh, we do need citeproc-ruby to parse names, but can def. get rid of csl-styles for now. Unicode also not needed, as you noted.

jrochkind: cool fine to leave csl-styles in, no big deal, we know we're gonna need it, the important thing is not adding dependencies we don't plan on needing!

PR #187, opened 2019-06-27
Lazyloading JS with lazyload.js added to webpacker

jrochkind:
OK, I added the lazyload.js dependency using the [webpacker](https://github.com/rails/webpacker) system (via yarn), instead of using sprockets. 

Since this is somewhat new to me (and to many), and I ran into some troubles, I wanted to document exactly what I did. 

Also **note** you will probably have to locally run `yarn install` on your dev machine when using these commits, such as after they are merged to master. 

## 1. Completely update all yarn dependencies

You *shouldn't* have to do this, but I was having trouble where any yarn operations were erroring out with errors that were incomprehensible to me, and I couldn't figure out why or how to fix it -- until I tried completely updating all dependencies from scratch, which worked. 

* removed the `yarn.lock` file, and ran `yarn install`, which generated a new `yarn.lock`. 

This is comparable to removing `Gemfile.lock` and running `bundle install`. It ignores whatever previous dependency versions you had pinned, and regenerates all your pinned dependency versions based on top-level requirements in your `package.json` (compare to `Gemfile`). 

This could be a "dangerous" thing to do, as you are abandoning your known working dependency tree, and some dependencies might not actually work together even though their requirements say they ought to. 

However, we aren't production yet, and it *is* actually a good thing to do periodically (update all your dependencies to get bug fixes security patches etc), and it fixed my problem and let me proceed, so seemed reasonable to do here. 

With that out of the way, the rest is pretty standard for Webpacker. 

## 2. Add lazysizes to our yarn dependencies

1. `yarn add lazysizes`. This makes a change to package.json. (Analagous to `Gemfile` for gems, but `lazysizes` is a JS npm package). 
2. `yarn install`. This modifies `yarn.lock` based on your new expressed dependency on lazysizes, resolving all requirements and dependencies to a consistent state. 
3. Commit changes to package.json and yarn.lock to repo. Now the project has a JS yarn/webpacker dependency on lazysizes. 

## 3. Create a new 'pack' file for lazysizes

A "pack" file will generally result in one separate JS file, compiled from multiple sources, that can be delivered to browser. 

 Lazy sizes `import` _could_ just be added to our existing `application` "pack" file. But lazysizes documentation suggested that while you might want much of your JS to be "late loading" by the browser (to optimize performance in page load), it is best to have lazysizes be "Early loading". 

So I decided to make it it's own "pack" file, that can be loaded separately by the browser. Called the pack file "lazysizes". 

## 4. Make our Rails layouts *load* this new pack file

With `javascript_pack_tag 'lazysizes'`. I added that to both our 'admin' layout, and our general public-facing 'application' layout. 

While we don't currently use lazysizes in admin pages, we might in the future. 

We do at the moment have two separate Rails "layouts" that don't really share common code, so it requires doing things in both places sometimes. For now, that's how it is. 

## 5. Make HTML use lazysizes

Lazysizes is just triggered by a class and data- attributes on an `img` tag. Our ThumbDisplay class was already written to generate these, just had to tell it to do so in the relevant place.


eddierubeiz:
Thanks for the docs above. Merging.

PR #188, opened 2019-06-27
use eager-loading on work member representatives/derivatives for display

jrochkind:
Without this, was doing a separate SQL query to get each members derivatives and/or representative image (and it's derivatives), which was adding ~800ms to 600-member ramelli, that this one line trims.

Should probably add as a feature to kithe, we need this particular `includes` all the time.


jrochkind:
@eddierubeiz review/merge please? Fairly small PR.

PR #189, opened 2019-06-28
Citation display

eddierubeiz:
Show citations (for now, without RIS download) on the public-facing work show item.


jrochkind:
Oh, I actually thought you were going to do the RIS converter *first*. Did human citation rendering first instead? OK!

jrochkind:
Or are they both together in this branch? I was really wanting them to be separate branches, trying to keep everything small and focused one thing at a time. I guess I wasn't clear though, no big deal!

app/presenters/citable_attributes.rb
```diff
@@ -15,6 +15,10 @@
 #     # => A hash of attributes.
 #
 # Ported from Sufia: app/models/chf/citable_attributes.rb
+#
+# A citable_attributes instance can be passed to a CitationDisplay as follows:
+#
+# CitationDisplay.new(citable_attributes).display
```
jrochkind: Do you think it would be a more sensible API to have `CitationDisplay.new(work).display`?  Let the `CitationDisplay` instantiate a CitableAttributes itself. I know this is different than how it worked in chf_sufia. What do you think?

app/presenters/citable_attributes.rb
```diff
@@ -195,7 +199,7 @@ def csl_type
     end
 
     def abstract
-      work.description.present? ? ActionView::Base.full_sanitizer.sanitize(work.description.join(" ")) : nil
+      work.description.present? ? ActionView::Base.full_sanitizer.sanitize(work.description) : nil
```
jrochkind: you changed this from returning a string to returning an array? Comment doc on the method as to what it returns?

app/presenters/citable_attributes.rb
```diff
@@ -507,6 +511,8 @@ class TreatAsOralHistory < StandardTreatment
     def title
       interviewee = CitableAttributes::work_lookup(work, "creator", "interviewee")
       interviewer = CitableAttributes::work_lookup(work, "creator", "interviewer")
+      interviewer = [""] if interviewer == [nil]
```
jrochkind: While it works, this is an unusual unfavored syntax for calling a class method. `CitableAttributes::work_lookup`. Instead, do `CitableAttributes.work_lookup`. Or, `self.class.work_lookup`. 

Also, rather than put in a special guard for `[""] if interviewer == [nil]`, I wonder if it makes more sense to pay attention to the return value of `work_lookup`?  What does it return? Always an array? But sometimes an array including `nil`?  Returning `[nil]` doesn't seem like a useful thing for `work_lookup` to return, I can't think of why you'd ever want it. Should it be returning an empty array (`[]`) instead?  

I think we can make this a lot fewer lines of code by paying attention to what makes sense to return in the first place, instead of returning something not useful and then guarding on it. Note that the empty array (`[]`) is already `blank?`.

app/presenters/citable_attributes.rb
jrochkind: Ah, what I said may actually *be( the API of CitationDisplay below, this comment is just wrong?

eddierubeiz: Oh yeah, we had the same idea and I refactored but forgot to change that comment. Will fix.

eddierubeiz:
> Oh, I actually thought you were going to do the RIS converter _first_. Did human citation rendering first instead? OK!

No, the ris converter relies on the citable_attributes object, so it made sense to do things in this order. I've made some progress on the ris_converter but I have that work on a separate branch and won't even post the PR until this one is merged: less confusion that way.

eddierubeiz:
> Or are they both together in this branch? I was really wanting them to be separate branches, trying to keep everything small and focused one thing at a time. I guess I wasn't clear though, no big deal!

No, tney'll be on separate branches.

jrochkind:
Oh, okay, cool. RIS depends on CitableAttributes, but does not depend on CitationRenderer. And CitableAttributes was already merged in another PR, right?

So RIS and CitationRenderer are independent, neither depends on each other, and they can be done in any order. I was thinking RIS would be first -- but since they are independent and can be done in any order, it doens't really matter, so this is fine! It is good to keep them in separate PR's (and both separate from CitableAttributes, which we already merged, so they are, hooray!)

app/assets/stylesheets/local/work_show.scss
```diff
@@ -33,5 +33,15 @@
...
 
+  .show-sub-head-body {
+    border-top: 3px solid #ddd;
+    padding: 5px;
+  }
+  .citation {
+      line-height: 1.5 !important;
+      word-break: break-word;
+  }
+
```
jrochkind: i wanna look at the CSS and see if there's a way to make it more consistent with existing CSS, perhaps using existing SASS variables from bootstrap, or other styles we already have. But there might not be. It's confusing to get a handle on the CSS, for me too. So just want to look at it a bit.

eddierubeiz:
> Oh, okay, cool. RIS depends on CitableAttributes, but does not depend on CitationRenderer. And CitableAttributes was already merged in another PR, right?
> 
> So RIS and CitationRenderer are independent, neither depends on each other, and they can be done in any order. I was thinking RIS would be first -- but since they are independent and can be done in any order, it doens't really matter, so this is fine! It is good to keep them in separate PR's (and both separate from CitableAttributes, which we already merged, so they are, hooray!)

Correct!

app/assets/stylesheets/local/work_show.scss
eddierubeiz: Oh yeah absolutely

app/presenters/citation_display.rb
```diff
@@ -0,0 +1,35 @@
...
+  def display
+    attributes = CitableAttributes.new(model)
+    csl_data = attributes.as_csl_json.stringify_keys
+    citation_item = CiteProc::CitationItem.new(id: csl_data["id"] || "id") do |c|
+      c.data = CiteProc::Item.new(csl_data)
+    end
+    renderer = CiteProc::Ruby::Renderer.new :format => CiteProc::Ruby::Formats::Html.new,
+      :locale => self.class.csl_en_us_locale
+    renderer.render(citation_item, self.class.csl_chicago_style.bibliography).html_safe
+  end
```
jrochkind: So this might be "bike shedding", talking about a minor detail that doesn't matter in the end. But humor me for a moment, although it doesn't necessarily need to lead to any code changes. 

If we imagine that we might instantiate a CitationDisplay, and then call `display` on it more than once. (In fact, we have no plans to, which is part of what makes this is a side conversation). 

All this work in this method would be done each time `display` is called -- and it's actually somewhat expensive work. But it will get the exact same result each time, because the result is a function of the work/model passed in, and you can't change the work attribute of a CitationDisplay. So it'll be doing unneeded expensive work more than it needs to. 

The solution is `memoizing`.  Ie, lazily storing in an instance variable. In the most basic way, you could wrap the whole method body in:

```
@display ||= begin
   # existing implementation
end
```. 

Now, the first time it's called, since the @display instance variable is nil, it executes the thing after the `||=`, the begin/end block, and assigns the result to @display. And then returns @display from the `display` method, since it's the last (actually only) line of the method body.  The next time it is called, since @display is not nil, it just returns @display.  (There is a hint about the gotcha here -- if the result of the block _can be nil_, it will never recognize it as already calcualted, and re-calculated it each time. Not sure if that's true here, sometimes it doesn't really matter).   

Now we've lazy cached/memoized the expensive stuff, so it only gets done once no matter how many times the method is called. 

Or, you could go further and extract things like `citeproc_citation_item` and `renderer` to methods, which are themselves memoized. 

```ruby
  def display
    @display ||= renderer.render(citation_item, self.class.csl_chicago_style.bibliography).html_safe
  end

  def citable_attributes
     @citable_attributes ||= CitableAttributes.new(model)
  end

  def csl_citation_item
    @csl_citation_item ||= begin
       csl_data = citable_attributes.as_csl_json.stringify_keys
       citation_item = CiteProc::CitationItem.new(id: csl_data["id"] || "id") do |c|
          c.data = CiteProc::Item.new(csl_data)
       end
    end

    def renderer
       @renderer ||=  CiteProc::Ruby::Renderer.new :format => CiteProc::Ruby::Formats::Html.new,
            :locale => self.class.csl_en_us_locale
     end
```

More readable/maintainable? Less readable/maintainable? Just shuffling deck chairs on the titanic? I think more readable/maintainable (as well as adding memoization), but I could totally be wrong, and this could all just be irrelevant.

jrochkind:
The point of `< ViewModel` is to get to use Rails helper methods. 

Does this class actually use any rails helper methods? I'm not sure, if not, could it get by without `< ViewModel`, as a plain old ruby object? Avoids the overhead and potential bugs/complexity of subclassing ViewModel if not needed.

app/presenters/citation_display.rb
eddierubeiz: Yeah, I think memoization is perfectly reasonable. There's a lot of processing here. I'm also getting more used to the idiom. I think this is a great suggestion.

app/presenters/citable_attributes.rb
eddierubeiz: No, work.description is a string in schihist_digicoll, so the method still returns a string or `nil`. Will add a comment though; this did throw me for a loop on Friday.

eddierubeiz: But yeah, good catch!

app/presenters/citable_attributes.rb
eddierubeiz: Very good suggestion! I think a strategically placed .compact can make this all go away... stay tuned.

jrochkind: And perhaps the strategically placed compact goes in the `work_lookup` implementation (which I'm not looking at at the moment).

app/presenters/citation_display.rb
eddierubeiz: Memoized it.

app/presenters/citable_attributes.rb
eddierubeiz: Done.

app/presenters/citable_attributes.rb
eddierubeiz: Yup! Worked like a charm. Thanks for catching this.

app/presenters/citation_display.rb
eddierubeiz: Done!

eddierubeiz:
> The point of `< ViewModel` is to get to use Rails helper methods.
> 
> Does this class actually use any rails helper methods? I'm not sure, if not, could it get by without `< ViewModel`, as a plain old ruby object? Avoids the overhead and potential bugs/complexity of subclassing ViewModel if not needed.

Removed the < ViewModel. Works just fine. Good call!

jrochkind:
@eddierubeiz looking good. I'll figure out the CSS and make a commit?

One more thing -- while we don't want to test the CSL gem's job, I think we should probably have some limited spec of the citation_display object. 

Like, just one or two specs, that may not do anything more than give it a work as input, and make sure it produces some non-empty string (implied, without raising). Maybe the bare-bones work factory, and a work with complete metadata and a collection etc.

jrochkind:
Or maybe we don't need any tests, with both the insides (CitableAttributes etc) and outsides (the overall work page integration test) already tested. I dunno, what's your feeling?

app/assets/stylesheets/local/work_show.scss
eddierubeiz: Thanks for the CSS!

eddierubeiz:
Yeah, I made the edits and then apparently committed the wrong file. Travis tests caught it of course. Oops. Fixed, sorry.

jrochkind:
somehow this is still showing changes from my other branch that should already be in master in the diff. Which is making it hard to review. 

Something odd is going on perhaps in the interaction of how I'm using git and how you're using git, maybe involving one or both of us doing rebases or other history rewriting in ways we shouldn't. 

But we've looked at this PR enough, I'm ready to move on, I'm just going to merge it.

jrochkind:
It also weirdly isn't giving me the option to delete the citable_attributes_display branch used here, like it usually does in the Github PR, which is probably also about something being confused in history. I dunno.

PR #193, opened 2019-07-01
Social media share icons, and rights icon, on work page

jrochkind:
Also includes meta tags relevant to social media share, such as opengraph tags, on the works page. (When you share to various social media sites, they scrape the page and use these meta tags to provide their preview).


eddierubeiz:
Looks great to me. Loving these presenters.

PR #194, opened 2019-07-01
WIP ris_serializer

eddierubeiz:
This is just a port; does not contain any front-end stuff; adding it to the actual UI can be a separate (and very simple) PR.
Note that this relies on features in another PR not yet merged into master, so the list of changes is misleading. I'm basically just adding `ris_serializer.rb` and `ris_serializer_spec.rb`.


jrochkind:
So this has all the commits in it from #189 (because you branched off there to create this branch), so we actuallly have to merge that one first, or else all those commits will come in as part of this one. 

After we merge that one, we can switch this one's base to something else and back to master to get it's diff right.

eddierubeiz:
Yes, exactly. I realized the same thing as you were typing :)

jrochkind:
Two different commits that both say "Initial commit for a port of ris_serializer." is not ideal (how can they both be?), nor is one that just says "initial commit". 

But I'm getting picky becuase overall looking good.

jrochkind:
Looks like you didn't create this branch off `citable_attributes_display`, you actually intentionally merged it into here? Cause one way to clean this all up would to rebase it... on master? on citable_attributes_display cause you need stuff from there? Not sure.

eddierubeiz:
Closing as github refuses to perform the diff correctly against master.

PR #195, opened 2019-07-02
Use :with_representative_derivatives scope now in kithe

jrochkind:
Update kithe to get new :with_representative_derivatives scope, and then use it: 

To pre-load all representatives and their derivatives from a hetereogenous member list of Assets and (child) Works. Avoiding n+1 queries, using Rails eager loading to do bulk queries on result set. A bit tricky how to do this, so we provided a scope in kithe to do it, now we replace our explicit code doing it with use of the scope from kithe.

Part of what makes it tricky is that when you have a hetereogenous list of members that are Assets and Works. Assets don't really have their own 'leaf_representative', they need :derivatives pre-loaded directly. But child works need their :leaf_representatives loaded, and then for those :leaf_representatives need :derivatives loaded on them. So we load *both* to catch em all, [:derivatives, leaf_representative: :derivatives].

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,9 +7,9 @@ GIT
 
 GIT
   remote: https://github.com/sciencehistory/kithe.git
-  revision: 3c90736d2ace4dad1091fec4ed0d214790e46ef4
+  revision: 7ac64ad9af19d9cae026b27aa146e578349134bc
   specs:
-    kithe (0.2.0)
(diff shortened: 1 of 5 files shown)


eddierubeiz:
The corresponding kithe commit is:

https://github.com/sciencehistory/kithe/commit/d1684f31bd53cb4f0ea45ebfc4381fba61385f1c

PR #197, opened 2019-07-02
Adding a ris download feature to the public facing WorksController.

eddierubeiz:
With tests. (also adding smoke test for the regular public facing work controller.)

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -8,7 +8,7 @@ class WorksController < ApplicationController
       format.html
       format.ris {
         send_data RisSerializer.new(@work).to_ris,
-          disposition: 'inline',
+          disposition: 'attachment',
           type: "application/x-research-info-systems"
       }
     end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -7,7 +7,9 @@ class WorksController < ApplicationController
     respond_to do |format|
       format.html
       format.ris {
-        send_data RisSerializer.new(@work).to_ris, disposition: 'inline'
+        send_data RisSerializer.new(@work).to_ris,
+          disposition: 'inline',
+          type: "application/x-research-info-systems"
       }
     end
   end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -3,9 +3,13 @@
 class WorksController < ApplicationController
   before_action :set_work, :check_auth
 
-
   def show
-
+    respond_to do |format|
+      format.html
+      format.ris {
+        send_data RisSerializer.new(@work).to_ris, disposition: 'inline'
+      }
+    end
   end
 
   private
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -9,7 +9,7 @@ class WorksController < ApplicationController
       format.ris {
         send_data RisSerializer.new(@work).to_ris,
           disposition: 'attachment',
-          type: "application/x-research-info-systems"
+          type: :ris
       }
     end
   end
--- a/app/views/works/_show_info.html.erb
+++ b/app/views/works/_show_info.html.erb
@@ -185,9 +185,12 @@
      <%=CitationDisplay.new(@work).display %>
   </p>
   <p class="btn-group" role="group">
-    <a class="btn btn-sm btn-primary export-citation" href="/works/<%= @work.friendlier_id %>.ris">
-      <i class="fa fa-address-card"></i>&nbsp;&nbsp;Export citation (RIS)
-    </a>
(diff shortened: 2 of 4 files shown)


spec/controllers/works_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+        get :show, params: { id: work.friendlier_id }, as: :html
+        expect(response.body).to include work.title
+      end
+
+      it 'allows user to download RIS citation' do
+        get :show, params: { id: work.friendlier_id }, as: :ris
+        expect(response.status).to eq(200)
+        expect(response.body).to include "TI  - Test title"
+        expect(response.body).to include "M2  - Courtesy of Science History Institute."
+        expect(response.body).to include "M2  - Courtesy of Science History Institute."
```
eddierubeiz: Duplicate line; will fix in the nex edit.

jrochkind: cool, makes sense to test the actual content very limited if at all, since that's already tested in the RIS serializer. 

I guess we could do something to test that it looks like a valid RIS file at all, a bunch of lines matching `/[A-Z0-9]{2}.*\r\n/` or whatever. 

But I'm not worried about it, this is just fine!

app/views/works/_show_info.html.erb
```diff
@@ -184,11 +184,8 @@
   <p class="citation">
      <%=CitationDisplay.new(@work).display %>
   </p>
-
-
-  <%#
   <p class="btn-group" role="group">
-    <a class="btn btn-sm btn-primary export-citation" href="/works/< % = @work.friendlier_id % >.ris">
+    <a class="btn btn-sm btn-primary export-citation" href="/works/<%= @work.friendlier_id %>.ris">
```
jrochkind: `href="/works/<%= @work.friendlier_id %>.ris"` is the wrong way to do this. We want to use Rails routing helpers!

To generate the URL: 

    work_path(@work, format: :ris)

To generate the link nicely without having to put ERB tags in an attribute, let's use Rails `link_to`. 

    link_to(work_path(@work, format: :ris), class: "btn btn-sm btn-primary export-citation") do
       <i class="fa fa-address-card"></i>&nbsp;&nbsp;Export citation (RIS)
     end

Just typed that in without trying it, may have typos or errors, but that's the idea.

app/controllers/works_controller.rb
```diff
@@ -3,9 +3,15 @@
...
 
-
   def show
-
+    respond_to do |format|
+      format.html
+      format.ris {
+        send_data RisSerializer.new(@work).to_ris,
+          disposition: 'inline',
+          type: "application/x-research-info-systems"
```
jrochkind: So I don't believe this should be disposition 'inline' -- that tells the browser to try to display it in a  browser tab if it can. Probably no browsers can, but still not actually what we want to be requesting. 

The code in chf_sufia is [here](https://github.com/sciencehistory/chf-sufia/blob/a103c5908af90c9076cb826140527213ba2e645f/app/controllers/curation_concerns/generic_works_controller.rb#L31-L34)

One thing we can see it is doing, is it IS specifying a nice filename to the browser to use for 'save as', instead of just guessing from the URL (which will end up the friendly_id as the filename). 

We haven't conquered that yet in this app, so don't need to in order to merge this PR -- but how about you leave us a GH issue reminding us that we need to fix the filename of the RIS download, and linking to the relevant code location to start with in chf_sufia as above?

If we look through further to the [encoding_safe_content_disposition method](https://github.com/sciencehistory/chf-sufia/blob/master/app/helpers/application_helper.rb#L31) , we see it is setting content-disposition to "attachment", not "inline". So let's do that here now, since that's easy. 

(Don't try to copy the logic for escaping or creating filename from chf_sufia -- there's a better way to do it, but we don't have to do it now to merge this PR! Let's leave an Issue reminding us to do it later as above).

jrochkind: I also believe you don't have to supply `type` -- I think the fact that `format.ris` is there is enough to have Rails look up the right type we already registered for `ris`, no need to specify it again non-DRY. 

I could be wrong, but you can try it and the tests should tell us if it worked!

eddierubeiz: Looks like you do need type -- it serves it as the default (application/octet-stream) otherwise.

eddierubeiz:
> OK, closer but still not there. Putting a string with `html_safe` is an unusual way to do this in ERB, usual way is the way I put an example of in the former comment, using a block arg to `link_to` for the tag content, which can just be in ERB.

Oh duh! Got it, stay tuned.

app/controllers/works_controller.rb
jrochkind: That's annoying. Hmm, what if you put `type: :ris`? Or we could research Rails more, I feel like there's a way to do this where it uses the type you specified as responding to.

eddierubeiz: That worked!

PR #199, opened 2019-07-02
Better type error message from ViewModel

jrochkind:
Tell us the class of the non-allowed thing in the error message, makes it much easier to spot the bug.


eddierubeiz:
Good call.

Issue #210, opened 2019-07-03
URLs in Citations don't include protocol, arne't real URLs


jrochkind:
In chf_sufia, when looking at the citaiton under "Cite As", you get a URL like:

    https://staging.digital.sciencehistory.org/works/zk51vh89m.

In scihist_digicoll staging, the URL looks like:

    kithe.sciencehistory.org/works/fb4949523

That's not actually a URL! It needs the https. 

Same in RIS export, so it's an issue in CitableAttributes. In scihist_digicoll:

    UR  - kithe.sciencehistory.org/works/fb4949523

Needs the http://!

You could look at how this was done in chf_sufia (probably hackily) and do it that way, or we could figure out a 'righter' way to do it. 

There should probably be a test for CitableAttributes ensuring it's a full URL with protocol and host and everything.

eddierubeiz:
Done!

PR #215, opened 2019-07-08
Add protocol to the citable_attributes URL method.

eddierubeiz:
The production value of :web_hostname is currently set to kithe.sciencehistory.org, without the protocol. We're now supplying the `https://` in the code.
Also modified the test settings and added better documentation.

Note:
This is implemented in Sufia at https://github.com/sciencehistory/chf-sufia/blob/a103c5908af90c9076cb826140527213ba2e645f/lib/chf/env.rb#L158 , and works differently.

### DIFF
--- a/app/lib/scihist_digicoll/env.rb
+++ b/app/lib/scihist_digicoll/env.rb
@@ -50,10 +50,22 @@ module ScihistDigicoll
 
     define_key :aws_region, default: "us-east-1"
 
-    # where our web app is located, https://WHAT
+    # Where our web app is hosted.
+    #
+    # This setting should not include the protocol, as this
+    # is supplied in the code ("https://").
+    #
+    # The production value is set in config/local_env.yml.
+    # This default value is used in the development and testing environments;
+    # if you choose, you can also override it in
+    # your own local config/local_env.yml file.
+    #
+    # The equivalent setting in Sufia is called app_url_base
+    # and is defined at
+    # https://github.com/sciencehistory/chf-sufia/blob/a103c5908af90c9076cb826140527213ba2e645f/lib/chf/env.rb#L158 .
     define_key :web_hostname, default: -> {
       unless Rails.env.production?
-        "https://localhost"
+        "localhost"
       end
     }
 
--- a/app/presenters/citable_attributes.rb
+++ b/app/presenters/citable_attributes.rb
@@ -300,7 +300,7 @@ class CitableAttributes
     end
 
     def url
-      "#{ScihistDigicoll::Env.lookup!(:web_hostname)}/works/#{work.friendlier_id}"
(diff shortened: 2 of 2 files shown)


app/lib/scihist_digicoll/env.rb
```diff
@@ -50,10 +50,22 @@ class Env < Kithe::ConfigBase
...
+    # if you choose, you can also override it in
+    # your own local config/local_env.yml file.
+    #
+    # The equivalent setting in Sufia is called app_url_base
+    # and is defined at
+    # https://github.com/sciencehistory/chf-sufia/blob/a103c5908af90c9076cb826140527213ba2e645f/lib/chf/env.rb#L158 .
     define_key :web_hostname, default: -> {
       unless Rails.env.production?
-        "https://localhost"
+        "localhost"
```
jrochkind: Ah, i see it wasn't being provided in production with `https`, so we were not testing the true production conditions. 

Should have caught that the first time!

jrochkind:
Please say in the title and/or description what you are fixing, not just "fix citable_attributes URL method". Fix what aspect of it?

jrochkind:
I wonder if we should change the Env/local_env value to actually be parallel to chf_sufia local_env_base, and have the scheme and port in it. 

Since we will sometimes need this. 

But we can leave that for another time.

eddierubeiz:
Yeah I think porting the Sufia architecture over and using :app_url_base (which includes the protocol) instead of :web_hostname is a really good idea. The code is currently simple enough that it should be an easy port.

jrochkind:
@eddierubeiz do you want to work with @sanfordd to change the local_env.yml? Maybe make a ticket. 

I'm thinking:

1. local_env.yml should have an app_url_base like chf_sufia, with scheme and port. Eg "https://kithe.sciencehistory.org". 

2. Our Env class should implement it, with defaults like in chf_sufia (can be done concurrently with 1). 

3. We're going to remove all references to Env `web_hostname` in current app. But some things still use the hostname. chf_sufia [hackily extracts](https://github.com/sciencehistory/chf-sufia/blob/a103c5908af90c9076cb826140527213ba2e645f/lib/chf/env.rb#L170-L175) it from `app_url_base` as another Env variable; we could do it that way (maybe even keep calling it `web_hostname` --but the citation stuff should use the new app_url_base instead!)

4. Then web_hostname can be removed from local_env/ansible. 

Actually I'll go ahead and make the ticket with this -- for now I'll assign it to you and daniel.

sanfordd:
Do we want to move the environment to production so we can start better testing those cases?

I'm happy to work on getting that local_env.yml updated.

jrochkind:
> Do we want to move the environment to production so we can start better testing those cases?

Unneccesary, we can wait until we actually have a production/staging. 

Our code pretty more or less treats production and staging the same, very few differences.

jrochkind:
Closes #210

PR #218, opened 2019-07-08
Public collections list page

### LINKED ISSUE #202: Collection list page
Equivalent of https://digital.sciencehistory.org/collections , and at the same path `/collections`. 

While typical Rails would have this as an "#index" action in a `CollectionsController` that also has, say, a `#show` action for showing an individual collection, and `update` and `destroy` actions etc -- we've done some things that make that not a thing. 1) Using a separate admin collections controller for CRUD, 2) Our public collection show page is the `#index` action in CollectionShowController, because it needed to include Blacklight-based search-within-the-collection, and it's hard to get Blacklight to do that from any action but `#index`. (Or at least it has been in the past and I didn't want to try this time). 

So I believe this will be a new controller that _only_ has an #index action, let's call it `CollectionsListController`. 

You can route to it in config/routes.rb with `get '/collections', to: "collection_list#index", as: :collections`

The `as: :collections` part should mean Rails will create a `collections_path` route helper, that returns the URL path. 

Other than that, it should be relatively standard Rails to create a controller action and view. 

You can use our existing ThumbDisplay view model to displays the collection thumbs (as high-res srcset-having img tags).  A Collection's thumb is an Asset available off `some_collection.representative`. 

You will want to make sure to list only *published* collections, but only fetching published collections when you fetch the list of collections to display. 

You will need to do some Rails eager-loading to make sure all of the collection representative derivatives are fetched in only one (or a fixed number) of SQL fetches, instead of going to the db with an SQL query for each collection. That is a bit tricky, we can leave it for last, it will be easy to add in top, as long as remember to. (Not totally sure if the new kithe `with_representative_derivatives` scope is suitable for that, since it was really designed for work/members, rather than the weird way we're using representative for collection thumb. But it might be!)

eddierubeiz:
Still to do:
- [x] Show only published collections
- [x] Clean up css (some redundancy; may include unnecessary definitions)
- [x] Try using $dark in place of $brand-inverse in css
- [x] Rails eager-loading
- [x] Add tests
- [x] Sort collections alphabetically by name.

Fixes #202


app/assets/stylesheets/_variables.scss
```diff
@@ -91,6 +91,9 @@ $warning: $brand-yellow;
 $primary: $brand-blue;
 $dark: $brand-dark-blue;
 
+
+$brand-inverse: $brand-dark-blue;
+
```
jrochkind: use existing `$dark` above, instead of defining a `$brand-inverse`. What Bootstrap 3 called `$brand-inverse`, Bootstrap 4 calls `$dark`.

app/views/collections/index.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <% @collections.each do |col| %>
+        <div class="collection">
+            <%= link_to collection_path(col.friendlier_id) do %>
+            <div class="collection-thumb">
+                <%= ThumbDisplay.new(
+                    col.leaf_representative, thumb_size: :collection_page, placeholder_img_url: asset_path("default_collection.svg")).display %>
+                </div>
+                <div class="collection-title-wrapper">
+                  <p class="collection-title"><%= col.title %></p>
+                </div>
```
jrochkind: fix indentation on all this stuff to all line up, close tags at same indentation as open tags.

jrochkind: oh actually I think the div on line 7 is _missing_ it's close tag maybe?

app/assets/stylesheets/local/collections.scss
```diff
@@ -0,0 +1,71 @@
+$collection-index-box-width: 266px;
+$collection-index-box-height: 266px;
+$collection-index-box-margin: 17px;
+
```
jrochkind: there's probably a better way to do this CSS wihtout those ugly variables, maybe without the "spacers" -- but if this is what was ported from chf_sufia... it's not too crazy CSS relatively workable, doesn't seem necessary to refactor it now. Just wanted to point it out heh.

eddierubeiz: Fair enough. I'm more than happy to spend some time on it of course.

app/views/collections/index.html.erb
eddierubeiz: Done!

app/assets/stylesheets/_variables.scss
eddierubeiz: Cool that did work. (I was trying to do $brand-dark before).

app/controllers/collections_list_controller.rb
```diff
@@ -0,0 +1,10 @@
+# Front-end controller to list public collections.
+class CollectionsListController < ApplicationController
+  def index
+    @collections = Collection.
+      where("published = true").
+      order(:title).
+      with_representative_derivatives
+    render :template => "collections/index"
```
jrochkind: This is an unusual way to do this.  More typical ways to do this:

* `render "collections/index"` -- you don't need the `template:`, either that's the default, or maybe it even does a slightly different thing that isn't right, not sure. Normally it's done like this. EXCEPT. 

* `render "index"` or `render :index` -- it'll look up in `collections/` automatically, because that's the controller we're in. If you were using controller superclasses, if not found in `collections/` it will *also* look in a superclass, up the chain. You break this if you say `render "collections/index"` specifically.  We're not doing this at present, but avoid surprises by doing things conventionally. 

* In fact, `index` is the default since we are in an action method `def index`, so you can just say `render`, or even leave off the `render` entirely. 

The last is most conventional -- rendering a standard template with the same name as the action method, in the same folder as the controller -- leave off the `render` entirely. 

I will make the commit myself to make this change.

jrochkind: Although actually you didn't have the view in the standard place. Since the controller is CollectionsListController, the view should be in `collections_list/index.html.erb`, not `collections/`. I will fix that too.

jrochkind:
Oops, I thought this was ready for review/merge, but I see it has no spec yet? 

I was going to do the things I noted above so I could go ahead and merge it without waiting for you to get back, but since you're still working on it, I'll let you do em!

app/controllers/collections_list_controller.rb
eddierubeiz: I moved the controller and got rid of the 'render'. Thanks for the advice there!

eddierubeiz:
I actually had made a test but never committed the file it was in. Oops!

PR #219, opened 2019-07-08
DownloadsController, with content-disposition filenames

jrochkind:
This provides a DownloadsController that we will use to deliver downloads to the user -- a url like `/downloads/#{asset_friendly_id}` will _redirect_ to a signed S3 url (with directive to S3 on content-disposition with chosen filename). 

Can also deliver derivatives, expected to be signed derivatives. 

See header comments on DownloadsController. 

This PR also has the somewhat complex code for determining desired filenames (including audio variants). This logic is implemented as just static 'functions' (not OO) to try avoid the problems we got into in chf_sufia with an architecture that was just wrong for what we actually needed to to, and hopefully make it easy to refactor if we realize what we need to do is different. (Still not quite sure filename suffixes are right -- maybe it should be trusting the suffixes on the original files, which it isn't now). 

It also has some somewhat unpleasantly complex code for setting up Assets and Derivatives in *specs*, to try to make it as quick as possible with mocked (not necessarily entirely accurate) files and metadata and derivatives. 

It does *not* include code for actually wiring up the 'download' buttons, it's got enough already!


jrochkind:
I realize trying to review this myself I really should have put the "new way to setup cheap Assets for testing" stuff in a different PR, to keep them separate. Sorry!

eddierubeiz:
Thanks for the docs as always. I realize it's not perfect but it's much easier to read and understand than what we ended up with in Sufia. Agreed: we can tweak as needed after launch.

PR #223, opened 2019-07-09
upgrade webpack-dev-server to 3.x instead of 2.x

jrochkind:
This seems to get rid of lots of warnings about unmaintained dependencies, as well as a potential vulnerability warning about one dependency. (That since it's a dependency of webpack-dev-server, probably didn't actually affect our production app).

A newly generated rails 5.2.3 app with webpacker uses `"@rails/webpacker": "^4.0.7"` and `"webpack-dev-server": "^3.7.2"`.

This PR will make this app use webpack-dev-server ^3.7.2, but does NOT ugprade webpacker itself, our webpacker is still 3.5. Trying to upgrade webpacker itself was not a no-changes-necessary upgrade, it had errors, and would take some hacking I don't have time for right now.

But can this new webpack-dev-server work with the old webpacker? I'm not totally sure. Nor can I figure out how to figure it out. I think neither of us are actually USING webpack-dev-server anyway. So it may not matter.

I don't like I don't know what I'm doing, but this should get rid of vulnerability warning and lots of other warnings, so we'll go with it for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,6 @@
     "mdn-polyfills": "^5.15.0"
   },
   "devDependencies": {
-    "webpack-dev-server": "^2.11.5"
+    "webpack-dev-server": "^3.7.2"
   }
 }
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Sounds good. Will take a look then merge.

PR #227, opened 2019-07-11
Add a Digitization Funder metadata item.

eddierubeiz:
Note this is an array in Sufia, and a string in ScihistDigicoll.


jrochkind:
Still takes a lot of files to edit to add a field, oh well!

Issue #217, opened 2019-07-08
Have an app_url_base in local_env.yml (like chf_sufia) instead of web_hostname. 


jrochkind:
Thinking we could do it this new way was a mistake, we need to know the scheme and (unlikely in production but in dev/test) port number too. 

1. local_env.yml should have an app_url_base like chf_sufia, with scheme and port. Eg "https://kithe.sciencehistory.org". 

2. Our Env class should implement it, with defaults like in chf_sufia (can be done concurrently with 1). 

3. We're going to remove all references to Env `web_hostname` in current app. But some things still use the hostname. chf_sufia [hackily extracts](https://github.com/sciencehistory/chf-sufia/blob/a103c5908af90c9076cb826140527213ba2e645f/lib/chf/env.rb#L170-L175) it from `app_url_base` as another Env variable; we could do it that way (maybe even keep calling it `web_hostname` --but the citation stuff should use the new app_url_base instead!)

4. Then web_hostname can be removed from local_env/ansible. 

(For now assigning to Eddie, but @eddierubeiz you don't need to interrupt your work on other stuff, we can decide if you or I do the Rails app part later).

sanfordd:
I've added `app_url_base` to the local_env file. It's currently https.

jrochkind:
closed by #230

PR #230, opened 2019-07-15
Env use app_base_url instead of web_hostname

jrochkind:


jrochkind:
Closes #217

eddierubeiz:
Thanks for doing this. Like the pre-parsed version being available from config, too.

app/lib/scihist_digicoll/env.rb
```diff
@@ -50,25 +55,25 @@ class Env < Kithe::ConfigBase
...
+      elsif Rails.env.development?
+        "http://localhost:3000"
       end
     }
 
+    # Sometimes we need the individual parts of the app_base_url, returns a URI object
+    # they can be obtained from.
+    def self.app_url_base_parsed
+      @app_url_base_parsed = URI.parse(lookup!(:app_url_base))
+    end
```
jrochkind: oops, this should have been `||=` to cache/memoize it.

PR #234, opened 2019-07-16
Stub for homepage.

eddierubeiz:
Includes a Rails route, a new controller, a template with some placeholder text on it, and a basic stylesheet.

### DIFF
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -1,4 +1,3 @@
-<% provide :page_title, "Science History Institute Digital Collections" %>
 <% provide :suppress_controls, true %>
 <% content_for(:suppress_masthead_search, true) %>
 <div>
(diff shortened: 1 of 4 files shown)


app/views/homepage/index.html.erb
```diff
@@ -0,0 +1,7 @@
+<% provide :page_title, "Science History Institute Digital Collections" %>
```
jrochkind: I _think_ you shouldn't have to do this, the title should default to that if you don't provide one. Did you find different?

config/routes.rb
```diff
@@ -13,13 +13,14 @@ def self.matches?(request)
...
-      "/login"
-    end
-  }
+  # root to: redirect { |path_params, req|
+  #   if LoggedInConstraint.matches?(req)
+  #     "/admin"
+  #   else
+  #     "/login"
+  #   end
+  # }
```
jrochkind: just delete the commented out code

eddierubeiz: Cool!

PR #235, opened 2019-07-16
Hero image and search form

eddierubeiz:
With proper bootstrap 4 media queries.

PR #236, opened 2019-07-16
Recent items

eddierubeiz:
Recent items. Should look and function like the ones in chf-sufia.
Please merge https://github.com/sciencehistory/scihist_digicoll/pull/235 first, btw.


eddierubeiz:
Adjust the following:
1) Get rid of $collage_style in the css and hardcode it so $collage_style is always true
2) Move the helper class into our presenters directory and call it recent_items.rb

Issue #237, opened 2019-07-17
Two quick changes to recent items feature.


eddierubeiz:
Adjust the following:
1) Get rid of $collage_style in the css and hardcode it so $collage_style is always true
2) Move the helper class into our presenters directory and call it recent_items.rb

_Originally posted by @eddierubeiz in https://github.com/sciencehistory/scihist_digicoll/pull/236#issuecomment-512447149_

jrochkind:
oh did we merge without doing these things? I thought we were going to do them before merge? But ok!

eddierubeiz:
Yeah so did I! Then you merged the PR and I thought fine, I'll do it next week :)

eddierubeiz:
#1 is fixed in https://github.com/sciencehistory/scihist_digicoll/pull/238. Still need to do the refactor.

PR #238, opened 2019-07-17
Featured topics on homepage

eddierubeiz:
Featured topics, displayed on the homepage. Note: no links to the featured topic page yet, as we consider that page a separate project and will build it later.


app/assets/stylesheets/local/site_wide.scss
```diff
@@ -112,14 +112,14 @@ hr.brand {
 // We set 'p' tags within a .branded-body-font -- with some exceptions -- to use our serif font, and have
 // appropriate max-width and spacer for that font.
 .branded-body-font, .scihist-main-layout {
-  p:not(.collection-title):not(.sans-serif), dl:not(.sans-serif) dd, table:not(.sans-serif) td, table:not(.sans-serif) th, .serif {
+  p:not(.featured-topic-title):not(.sans-serif), dl:not(.sans-serif) dd, table:not(.sans-serif) td, table:not(.sans-serif) th, .serif {
     @extend %text-font;
```
jrochkind: Hmm, you *removed* `not(".collection-title")` -- but do we need that for another part of the app? Is it safe to remove?

What this is doing is saying that all `p` (paragraphs) will be styled with our "text" styling, _except_ for certain ones -- could we also just make  the relevant collection-title not a `p`?  If it's a title, perhaps it should be a hN tag of some kind? Or just a div?

eddierubeiz: Nah, this is a mistake on my part, but now that I'm looking at it again, I see that I can and should refactor this in such a way that site_wide.scss doesn't need to change. Let me give this another shot. Yeah, we can have this be an h3.

eddierubeiz: OK - now an h3, and the rules that apply to it are stated explicitly in homepage.scss .

app/assets/stylesheets/local/homepage.scss
```diff
@@ -171,4 +171,64 @@
             }
         }
     }
+
+    $featured-topics-box-width: 21.875rem;
+    $featured-topics-box-height: 21.875rem;
+    $featured-topics-box-margin: 3.6875rem;
```
jrochkind: So I know I said to do everything in `rem`... but these are actually sized based on _our thumb image sizes_. 

I wonder if they should actually still be in px?  At least the height/width, maybe not the margin.  Especially cause width is used as a media query -- I havne't totally wrapped my head around what media queries in terms of `rem` do, if they actually work (and on all browsers). I think they may not work on some. 

I'm actually not totally sure, we could just go ahead and merge... but how about we change these back to `px`? 

And then feel free to go ahead and merge this PR yourself please, rest looks good!

eddierubeiz: Changed back to px! no problem.

eddierubeiz:
Fixes #237 too.

Issue #262, opened 2019-07-30
Mismatch in count of public items


eddierubeiz:
This page: https://kithe.sciencehistory.org/  says 7,204 items

But this page: https://kithe.sciencehistory.org/catalog?utf8=%E2%9C%93&search_field=all_fields&q=&range%5Byear_facet_isim%5D%5Bbegin%5D=&range%5Byear_facet_isim%5D%5Bend%5D=  says there are  6,752 items

eddierubeiz:
This is done!

PR #264, opened 2019-07-30
Fixing the item count on the homepage so it matches a blank Solr search

eddierubeiz:
If you run a Solr search, you get an item count which is exactly `Kithe::Model.where(:published => true, :type => ['Work', 'Collection']).count`. So... use that expression on the home page.

### DIFF
--- a/app/views/homepage/_hero_image_and_search_form.html.erb
+++ b/app/views/homepage/_hero_image_and_search_form.html.erb
@@ -17,7 +17,8 @@
                 <div class="digital-description col-12 col-md-6 col-lg-12">
                   <div class="pre-heading">Science History Institute</div>
                     <h2 class="brand-alt-h2">Digital Collections</h2>
-                    <p>Search <%= number_with_delimiter(Work.count) %> of our digitized items: artifacts, photographs, advertisements, letters, rare books, and more. You can also learn more about the rest of our <a href="https://sciencehistory.org/collections">collections.</a></p>
+                    <% published_item_count = Kithe::Model.where(:published => true, :type => ['Work', 'Collection']).count %>
+                    <p>Search <%= number_with_delimiter(published_item_count) %> of our digitized items: artifacts, photographs, advertisements, letters, rare books, and more. You can also learn more about the rest of our <a href="https://sciencehistory.org/collections">collections.</a></p>
                 </div>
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Fixes https://github.com/sciencehistory/scihist_digicoll/issues/262

app/views/homepage/_hero_image_and_search_form.html.erb
```diff
@@ -17,7 +17,8 @@
                 <div class="digital-description col-12 col-md-6 col-lg-12">
                   <div class="pre-heading">Science History Institute</div>
                     <h2 class="brand-alt-h2">Digital Collections</h2>
-                    <p>Search <%= number_with_delimiter(Work.count) %> of our digitized items: artifacts, photographs, advertisements, letters, rare books, and more. You can also learn more about the rest of our <a href="https://sciencehistory.org/collections">collections.</a></p>
+                    <% published_item_count = Kithe::Model.where(:published => true, :type => ['Work', 'Collection']).count %>
+                    <p>Search <%= number_with_delimiter(published_item_count) %> of our digitized items: artifacts, photographs, advertisements, letters, rare books, and more. You can also learn more about the rest of our <a href="https://sciencehistory.org/collections">collections.</a></p>
```
jrochkind: In an idea world we wouldn't put local variable declarations in our ERB templates. (As part of a general desire to keep complex logic out of ERB altogether, but local variable declarations seem particularly galling). 

Part of the problem with Rails is that it lacks a good place to pull this logic out to (which we solve with our "ViewModels", but we aren't using that here, and in general haven't ever used one for a top-level template rather than a partial). 

The standard Railss answer is to put them in "helper methods", but these end up global and are terrible. 

Another thing you can do is put them in the *controller*, with a `helper_method` declaration in the controller to make it available to the view. So a method `published_item_count` maybe. 

Or, in this case, i think it would be fine to just shove it inline `number_with_delimiter(Kithe::Model.where(:published => true, :type => ['Work', 'Collection']).count)` -- not great, but still better than a local variable in ERB I think, and good enough.

eddierubeiz: Cool. Inline seems simpler - it's a simple enough calculation.

PR #266, opened 2019-07-31
Tweaks to the cookie banner

### LINKED ISSUE #212: cookie banner
The one on current website, the color is sometimes same color as background it's over. Find new color? Add border? 

Make 'accept' button full height, even when banner gets higher on narrow screens due to wrapping?

### LINKED ISSUE #261: Tweak cookie banner
Make button full-height when cookie banner ends up larger: https://www.dropbox.com/s/a0sgkorkq6r8rpj/Screenshot%202019-07-30%2012.06.23.png?dl=0

Do something about when page bg is same color as banner bg . maybe just a 1px solid grey (probably a brand gray) top border. https://www.dropbox.com/s/l8rdmowktpxt6cq/Screenshot%202019-07-30%2012.06.56.png?dl=0

eddierubeiz:
Several changes:
 - Adding a grey border to the top. (same size and color as the grey border we use for metadata display.)
 - The columns are directly under the nav: no container, no row (per bootstrap 4 examples).
Fixes #261 
Fixes #212

### DIFF
--- a/app/assets/stylesheets/local/footer.scss
+++ b/app/assets/stylesheets/local/footer.scss
@@ -89,45 +89,44 @@
 
 }
 
-$accept-cookies-banner-tall: 5rem;
-$accept-cookies-banner-reg:  3rem;
 nav.accept-cookies-banner-nav {
-  border-top: 2px solid lightgray;
+  border-top: 3px solid #dee2e6;
   min-height: 1px;
-  background-color: white;
   padding: 0;
+  display: flex;
+  align-items: stretch;
+  background-color: white;
   div.i-accept-copy {
-    line-height: 1.4;
+    line-height: 1.3;
     padding-top: 0.05rem;
     color: $text-muted;
   }
   div.i-accept-button-div {
-    background-color: green;
     padding: 0;
     margin: 0;
     a.i-accept-link {
+      display: block;
       color: white;
       background-color: $brand-red;
       &:hover { background-color: $brand-blue; }
       text-decoration: none;
       font-size: .9em;
       font-weight: 700;
-      display: block;
-      width: 100%;
       border: 0;
       white-space: nowrap;
-
-      height: $accept-cookies-banner-tall;
-      line-height: $accept-cookies-banner-tall;
-      @include media-breakpoint-only(xs) {
-          height:       $accept-cookies-banner-reg;
-          line-height:  $accept-cookies-banner-reg;
+      line-height: 3;
+      @include media-breakpoint-between(sm, lg) {
+         line-height: 6;
       }
-      @include media-breakpoint-only(xl) {
-          height:       $accept-cookies-banner-reg;
-          line-height:  $accept-cookies-banner-reg;
+      @include media-breakpoint-between(xs, sm) {
+        margin-top:0.5rem;
       }
-
     }
   }
+
+
+
+
+
+
 }
--- a/app/assets/stylesheets/local/footer.scss
+++ b/app/assets/stylesheets/local/footer.scss
@@ -87,39 +87,3 @@
     }
   }
 }
-
-nav.accept-cookies-banner-nav {
-  border-top: 3px solid $table-border-color;
-  min-height: 1px;
-  padding: 0;
-  display: flex;
-  align-items: stretch;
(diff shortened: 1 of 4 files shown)


jrochkind:
> The nav is taller: 5rem, except at widths xs and xl. Before, at medium widths, the text was wrapping to three lines and overflowing the former height of 2.8rem.

@eddierubeiz Hmm, I didn't realize the nav height was being specified explicitly. 

This seems wrong to me. But it's maybe what we inherited from general sciencehistory stylesheet?

Is there a way to avoid specifically setting an explicit height, and just let it be as high as it needs to be for the content, as HTML blocks normally are? (It's possible the fixed positioning makes that harder or impossible?)

If possible, then we wouldn't need to worry about the 'wrap', and wouldn't get it any *higher* than it needs to be.  With explicit, it is sometimes higher than it needs to be (depending on wrap), in order to always be no shorter than it needs to be -- but this cookie banner is gross, and we don't want to make it any higher than it needs to be I think.

If this is impossible or too difficult no big deal, but thought I'd mention it. Thanks for the good PR description which led me to consider it.

jrochkind:
Also does this fix both of the problems mentioned in the corresponding issue (and can you reference to corresponding issue to link/auto-close?)

app/assets/stylesheets/local/footer.scss
```diff
@@ -89,42 +89,45 @@
 
 }
 
-$accept-cookies-banner-container-height: 2.8125rem;
-
+$accept-cookies-banner-tall: 5rem;
+$accept-cookies-banner-reg:  3rem;
 nav.accept-cookies-banner-nav {
+  border-top: 2px solid lightgray;
```
jrochkind: Can you use one of our brand grey colors contained in the existing scss variables, instead of an arbitrary grey, so we stick to brand colors? For consistency and policy.

eddierubeiz: Good call!

eddierubeiz:
Added the "closes" clauses -- thanks for reminder.

app/assets/stylesheets/local/footer.scss
```diff
@@ -89,45 +89,44 @@
 
 }
 
-$accept-cookies-banner-tall: 5rem;
-$accept-cookies-banner-reg:  3rem;
 nav.accept-cookies-banner-nav {
-  border-top: 2px solid lightgray;
+  border-top: 3px solid #dee2e6;
```
jrochkind: this is still a a literal copy of the brand color. We have our brand colors in SCSS variables. can you use the existing SCSS variable as previously suggested?

eddierubeiz: Yeah, of course, stay tuned...

jrochkind:
1. How about putting the border only over the text cell, not the "I accept" button? (So I guess CSS *not* on entire banner footer, but just that section). It looks a bit cheesier on the "I Accept" button when over a dark-but-not-the-same background. And the red "I accept" doesn't need it: 

Screenshot 2019-07-31 16 19 38


2. A one-pixel border still seems better to me. I think it looks okay over white background which is where we needed the border 

example of 1px border over white bg, seems fine to me:  Screenshot 2019-07-31 16 20 34


while looking less bad over dark-but-not-the-same bg, where the border kind of introduces an unpleasant extra color, but with one px at least more subtle:

1px:  Screenshot 2019-07-31 16 21 18
   

2px: Screenshot 2019-07-31 16 21 32


3. At certain screen sizes, the text and "I accept" become over/under -- but the "I Accept" is way higher than it needs to be. At this screen size (eg 730px wide), I think "I accept" should probably be only as high as it needs to be for it's content, no higher. (just as it is at 575px window and smaller). Bad example: 
Screenshot 2019-07-31 16 23 09

jrochkind:
I wonder if using our darker grey brand color might actually work more universally without clashing or creating an odd optical effect with whatever other non-white color it happens to be over, while still doing the job of separating from a white bg. 

Another option would be experimenting with CSS semi-transparent drop shadow effect instead of a solid border. I forget the way (or ways?) to do that in CSS, but we've done it on our drop-down with search options from the navbar search.

Issue #274, opened 2019-08-02
Add pagination to collection show page.


eddierubeiz:
Compare https://kithe.sciencehistory.org/collections/wh246s128 with the corresponding page on the old site. Pagination is missing!

eddierubeiz:
Gonna do this now...

PR #275, opened 2019-08-05
Add pagination on "show collection" page

eddierubeiz:
It was missing for whatever reason.
Fixes https://github.com/sciencehistory/scihist_digicoll/issues/274 .

### DIFF
--- a/app/views/collection_show/index.html.erb
+++ b/app/views/collection_show/index.html.erb
@@ -103,6 +103,16 @@
       <%= render 'search_header' %>
 
       <%= render "document_list", documents: @response.documents %>
+      <% if show_pagination? and @response.total_pages > 1 %>
+        <%# expanded will be shown by CSS at large screen sizes, else compact %>
+        <div class="pagination pagination-alt-expanded">
+          <%= paginate @response, :window => 3, :outer_window => 1, :theme => 'local' %>
+        </div>
+        <div class="pagination pagination-alt-compact">
+          <%= paginate @response, :page_entries_info => page_entries_info(@response), :theme => 'blacklight_compact' %>
+        </div>
+      <% end %>
+
     </div>
   </div>
 


jrochkind:
@eddierubeiz do you got to do this in 'featured topics' too?

app/views/collection_show/index.html.erb
```diff
@@ -103,6 +103,16 @@
...
       <%= render "document_list", documents: @response.documents %>
+      <% if show_pagination? and @response.total_pages > 1 %>
+        <%# expanded will be shown by CSS at large screen sizes, else compact %>
+        <div class="pagination pagination-alt-expanded">
+          <%= paginate @response, :window => 3, :outer_window => 1, :theme => 'local' %>
+        </div>
+        <div class="pagination pagination-alt-compact">
+          <%= paginate @response, :page_entries_info => page_entries_info(@response), :theme => 'blacklight_compact' %>
+        </div>
+      <% end %>
```
jrochkind: If we're gonna be repeating the logic (the conditional, the double pagination for small and big pages) on multiple pages... should we extract it into a partial view template, so it can be re-used without copy and paste? Thoughts @eddierubeiz ?

eddierubeiz: We can totally do this and eliminate the duplication. Probably simpler to do this as a separate PR though, after the two PRs that contain the duplication are merged. I'll create an issue for it.

eddierubeiz: By the way, to answer your earlier question, yes -- the pagination code is already in the featured topics PR.

jrochkind:
@eddierubeiz our search results have pagination at top and bottom of results (displayed somewhat differently). 

Did within-collection search in chf_sufia do this too, or decide to leave out the top-of-results pagination?   If both, are both present after this PR? (Like maybe the top one was already there?)

PR #277, opened 2019-08-05
rearranging some webpacker dependencies

jrochkind:
updating uppy to 1.x (still loaded via script tag to CDN at present), 

~and registers uppy with webpacker 'externals' for what I think is better uppy practices~ (stopped doing that, it may have been better webpacker practice, but it meant we would have to load uppy on EVERY page since we have only one webpacker pack, and with externals it would complain if uppy wasn't available. It works just using uppy from window.Uppy in our local JS; we can fix it to be 'right' later, at the same time we figure out how to load uppy via webpacker). 

- [x] move some more stuff to webpacker?
- [x] move admin-only webpacker stuff to an 'admin' subdir, to keep straight what's admin only to make it easier to make an admin-only pack if desired?  (the viewer will require webpacker on public end). 

Note updated README explaining what's up with the JS.


jrochkind:
Can't figure out why travis is failing, and in just one of it's tests. Seems to be failing trying to start Solr for the tests, but I don't know why. I'm going to try merging master in, which should make the two travis tests identical at least.

jrochkind:
Could be a travis problem. https://www.traviscistatus.com/incidents/hl4vqb7hvv5n

Issue #278, opened 2019-08-05
Featured topics need tests.


eddierubeiz:
A system test and probably also a controller test.

eddierubeiz:
This is done!

PR #281, opened 2019-08-06
Featured topic tests

eddierubeiz:
Just adding a controller and system test for the featured topics.


eddierubeiz:
Fixes #278

jrochkind:
I wonder if the controller test is necessary, because I don't think it tests anything that isn't already being tested by the feature test (which tests more) too. Every test slows down our test suite, and is code to maintain. 

but I'm going to go ahead and merge, just leaving this as thought for the future.

PR #283, opened 2019-08-06
Featured topic title css adjustment

eddierubeiz:
Move the title down by a small fraction of a rem, just to completely cover the bottom edge of the thumbnail in e.g. Safari.
This shouldn't be necessary but it is.


jrochkind:
For this purpose would pixel make more sense than rem? I guess it depends on the nature of the layout bug, which we don't really know. So good enough!  .05rem is actually... 0.8 pixels at 16px==1rem. 🤷‍♂

PR #285, opened 2019-08-06
Download menu fix

eddierubeiz:
The download menu presenter iterates through all the siblings of the item being shown, to see if it needs to offer a whole-work PDF download option.
If one of the siblings' leaf representative asset lacks a file, the code fails.

If you run an import without files (the current default), e.g. in dev, you're certain to run into this error.

### DIFF
--- a/app/presenters/download_dropdown_display.rb
+++ b/app/presenters/download_dropdown_display.rb
@@ -158,9 +158,9 @@ class DownloadDropdownDisplay < ViewModel
   def has_work_download_options?
     display_parent_work &&
       display_parent_work.members.length > 1 &&
-      display_parent_work.members.all? do |member|
-        member.leaf_representative && member.leaf_representative.content_type.start_with?("image/")
-      end
+       display_parent_work.members.all? do |member|
+         member.leaf_representative && member.leaf_representative.file_data && member.leaf_representative.content_type.start_with?("image/")
+       end
   end
 
 end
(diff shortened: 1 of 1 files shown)


jrochkind:
Good catch and fix, thanks. 

I'm not totally sure whether it makes more sense to check `member.leaf_representative.file_data` or `member.leaf_representative.file`. 

`file_data` is the raw db column (a json-serialized hash). 

`file` is the attribute presented by shrine that will be a Shrine::Attachment object (or something like that) -- it should still be `nil` if `file_data` is nil, I think. 

But I'm not sure, so `file_data` is probably fine, just providing some more context/options. Will merge as is.

jrochkind:
(And it doens't just apply to weirdly-imported items, it is definitely possible to have an Asset that doesn't currently/yet have an attached file, and important to check for it. 

Although... actually. The reason this fails, is because `member.leaf_representative.content_type.start_with?` will fail when `...content_type` is nil. 

To be precise, what we *really* want to check is that `member.leaf_representative.content_type` is non-nil -- relies less on the implementation details of knowing this is associated with `file_data` -- it *is*, but maybe it won't be forever, and it's less clear. 

So maybe change to a check for `member.leaf_representative.content_type.present?`

OR, use the fancy new rails `&.` operator might work... `member&.leaf_representative&.content_type&.start_with?("image/")`. 

What do you think @eddierubeiz ?

eddierubeiz:
Caution: https://github.com/sciencehistory/scihist_digicoll/pull/289 has `download_menu_fix` merged into it already. The good news is there's only one line that needs to be changed, but I just want to make a note of it here.

eddierubeiz:
Yup, went with your suggestion. Much more readable.

PR #287, opened 2019-08-06
WIP display audio on the member admin screens

eddierubeiz:
Just to open the discussion, here's probably the simplest possible thing that could work: adding a dumb audio scrubber to the member list table.
No tests, no css, just a proof of concept for now.


jrochkind:
Gonna close this for now, don't need it right away, and we wouldn't probably put it in quite this place if we did.

PR #289, opened 2019-08-07
Audio front end

eddierubeiz:
Shows an audio player on the front end, similar to the one in Sufia.
Subclasses the WorkShowDecorator with audio-specific methods, but uses the same `show.html.erb` template.
The template instantiates a playlist ViewModel, which takes an array of track Assets as its "model". The array of tracks is calculated by the decorator.
Modified download menu ViewModel so it works within an unordered list.
The section containing title and dates is moved out to a separate ViewModel, to reduce repetition.


app/assets/javascripts/playlist.js
```diff
@@ -0,0 +1,68 @@
+/*
```
jrochkind: since we've moved all things to webpacker, this needs to go in `./app/javascript/src/js`, and be imported in `./app/javascript/packs/application.js`.

app/controllers/works_controller.rb
```diff
@@ -17,6 +19,17 @@ def show
 
   private
 
+  def set_decorator_and_template
+    if AudioWorkShowDecorator.show_playlist?(@work)
+      then
```
jrochkind: you don't use `then` in ruby `if` style, I didn't even know it was possible!

app/controllers/works_controller.rb
```diff
@@ -1,11 +1,13 @@
 # FRONT-END controller for public Works views -- mostly just the #show action.
 # Not to be confused with admin/works_controller for staff/management/edit views.
 class WorksController < ApplicationController
-  before_action :set_work, :check_auth
+  before_action :set_work, :set_decorator_and_template, :check_auth
```
jrochkind: rather than a before action, if you are using a different decorator for the different cases anyway, make this `decorator.view_template`. 

If you were not using decorators, I would still make this a `view_template_name` method in in controller, that returns the right one. 

I know Rails uses instance variables sometimes, but too many of them get confusing, where was the state set by whom what is it now where can it be changed? A method with `||=`-style memoization is better.

app/controllers/works_controller.rb
```diff
@@ -26,7 +39,7 @@ def check_auth
   end
 
   def decorator
-    @decorator = WorkShowDecorator.new(@work)
+    @decorator
```
jrochkind: Instead of making this an iVar set in a `before_action`, just make it conditional here and memoized. (And I see there was a bug before, even in the before-change version, that `=` should have been `||=`. 

```ruby
@decorator ||= if something
    WorkShowDecorator.new(@work)
else
    AudioWorkShowDecorator(@work)
end
```

app/presenters/download_dropdown_display.rb
```diff
@@ -100,14 +102,24 @@ def button
...
 
+  def link
+    content_tag("a",
+                "<i class='fa fa-download' aria-hidden='true'></i> Download".html_safe,
+                class: "dropdown-toggle download-link",
+                id: menu_button_id,
+                "data-toggle" => "dropdown",
+                "aria-haspopup" => "true",
+                "aria-expanded" => "false")
+  end
```
jrochkind: Is `link` exactly the same code as `button` but for the first parameter of `content_tag`? Maybe one or two more? But a lot in common?

I wonder if it would be better to have `link_or_button` (or some other name), that just picks it's arguments to `content_tag` that vary appropriately, while leaving the ones that don't vary alone. 

That way we don't have to copy things like `data-toggle` and `aria-haspopup` in two places, and make sure to change them in two places if they change.

app/views/works/_playlist.html.erb
```diff
@@ -0,0 +1,34 @@
+<div class="show-page-audio-playlist-wrapper" data-role="audio-playlist-wrapper">
```
jrochkind: We talked about making this a ViewModel, so it can have it's own helpers that go with it, rather than just a template. What do you think?

app/views/works/_title_and_dates.html.erb
```diff
@@ -0,0 +1,49 @@
+<header class="show-title">
```
jrochkind: we talked about making this a ViewModel so it has it's own helpers that go with it (and they don't need to be in the decorator), rather than just a template. And thus make the decorator(s) have to have a lot less stuff. What do you think?

app/controllers/works_controller.rb
eddierubeiz: OK. It still feels kinda wrong this way. I'll try `decorator.view_template`; that sounds less ugly.

app/controllers/works_controller.rb
eddierubeiz: Yup -- thanks!

app/presenters/download_dropdown_display.rb
eddierubeiz: Yup - thanks for catching this.

app/views/works/_title_and_dates.html.erb
eddierubeiz: Done. It only really removes one method from the decorator, but it's somewhat better.

app/views/works/_playlist.html.erb
eddierubeiz: So I gave this a shot, but it ended up giving me more complexity and not less.

Seems to me a `ViewModel` is best suited for a self-contained chunk of HTML, possibly with its own template, and definitely with its own helper methods. The code that splits the members of a work into `audio_members` and `non_audio_members` doesn't pertain to the playlist alone: it affects the whole page, because you want all the rejected `non_audio_members` to show up in the non-playlist part of the page. And you don't want a ViewModel making decisions about stuff that's outside of its purview. So I'm leaving the methods in the decorator for now, where they all make sense together.

jrochkind: I mean making the show-page-audio-playlist-wrapper into a ViewModel. The component that shows the player. 

The list of audio children to show could be passed into it. Although I also think it might be fine for it to determine that itself -- it would be deciding what the appropriate things for it to show are, not necessarily making decisions for the page. (As we previously had a version where audio things were repeated at the bottom -- what to show in the playlist, and what to show in other places, are to some extent separate decisions)

eddierubeiz: If they can be separate decisions (and very good point about how the Sufia page *did* repeat some of the members on the page) then that's more manageable, yes. I'm going to give it another shot, then.

jrochkind: I think it would be fine to pass into the ViewModel the list of audio Assets to display too. Definitely as a first step. The idea is just that putting it in a ViewModel keeps it more encapsulated, and re-usable somewhere else. ANY helper methods it needs, come with it. (Are there no other helper methods it needs? I confess I'm not looking at the code right now).

eddierubeiz: Yup. There's really only one helper method - figure out whether a member has playable audio or not.

jrochkind: i suppose there could become more if we factor logic out of the ERB, or if the complexity of requirements change. I dunno, ViewModels seems good to me for something like this we might want to re-use on another page. Even for one helper method might be enough justification.

eddierubeiz: And if we pass an already-filtered list of assets, that's not even a helper method actually; that can stay in the decorator.

eddierubeiz: This ended up working nicely on the second try.

app/views/works/show.html.erb
```diff
@@ -6,58 +6,10 @@
...
-              </li>
-            <% end %>
-          </ul>
-      </div>
-    <% end %>
-  </header>
+  <%= WorkTitleAndDates.new(@work).display %>
 
   <div class="show-hero">
+    <%= Playlist.new(decorator.audio_members).display if decorator.audio_members.present? %>
```
jrochkind: So the downside of only having one template is that your decorator for the non-audio still needs that `audio_members` method. And does that mean the decorator for the non-audio will refuse to include audio in the "ordinary" list of members?

If there IS going to be only one template... there should possibly be only one decorator too, since any method the template calls needs to be available in the decorator, so both decorators need the same methods. (Are they even implemented any different in the two decorators?)

But I'm still kinda liking the idea of two templates -- I think it makes them (and their decorators) both _simpler_, with less conditional logic. And also opens up the idea of choosing to show any Work with either template, instead of it being baked into the code in several places that any work with at least one audio file _always_ displays like our current "oral history" template. With two templates/decorators, it'd just be a question of changing the logic for picking which template/decorator, and you could always show any Work with either one. 

But let's talk about this more in a voice chat?

app/decorators/work_show_decorator.rb
```diff
@@ -101,12 +106,7 @@ def public_collections
...
   def member_list_for_display
     @member_list_display ||= begin
-      members = model.members.
-        with_representative_derivatives.
-        where(published: true).
-        order(:position).
-        to_a
-
+      members = ordered_public_members
       members.delete_at(0) if members[0] == representative_member
```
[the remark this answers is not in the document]
eddierubeiz: Good call. In practice it's not a big deal because member_list_for_display is itself memoized, so it's impossible to run the core of its code more than once. But still, oops. I'm adding a .dup . Yes, ordered_public members is used in a couple places.

app/decorators/audio_work_show_decorator.rb
```diff
@@ -0,0 +1,36 @@
...
+  # Used in works controller to figure out whether
+  # it's appropriate to show an audio playlist
+  # for a particular work.
+  # Doesn't care about the order of the members,
+  # just whether any of the published ones have playable audio.
+  def self.show_playlist?(some_work)
+    some_work.members.
+      where(published: true).
+      any? { | x| self.has_playable_audio_derivatives?(x) }
+  end
```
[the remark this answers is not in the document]
eddierubeiz: No, this is a class method. I'm running it from the controller on *all* works (audio or not) to determine which decorator to use. It looks at all the derivatives and stops as soon as it finds *one* that's audio. `audio_members` isn't available in this context -- `audio_members` is an instance method to list *all* the audio members for an item that we've already determined that it has at least one playable audio track.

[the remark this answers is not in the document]
eddierubeiz: Will update the comment! :)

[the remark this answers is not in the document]
jrochkind: hmm, I guess I'm not sure why it's a class method! Maybe there's a larger refactor to be done, not sure. Bigger discussion maybe.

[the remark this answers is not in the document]
jrochkind: It is doing an extra SQL fetch that I think doesn't have to and is best avoided, may require some refactoring to avoid, not sure, need to look at the code holistically and think about it.

jrochkind:
@eddierubeiz can you do a merge master and resolve conflicts? Or let me know if you want help with that.

eddierubeiz:
Yeah hang on -- merging now

app/views/works/show.html.erb
eddierubeiz: Moved it back to two templates!

PR #290, opened 2019-08-08
DZI file generation and management

jrochkind:
Control of creation/deletion/access to DZI files for viewer.


jrochkind:
@eddierubeiz ready for review.

PR #300, opened 2019-08-20
oai-pmh

### LINKED ISSUE #304: oai-pmh
Will want to ask PA Digital to test with our new software, and make sure it still works. 

It should be pretty compatible/identical.

jrochkind:
- dc:format with IANA content types (eg application/pdf). chf_sufia did that. This one is not currently doing that, cause have to figure out how to avoid n+1 queries while dealing with the existing architecture of the oai gem, it's not easy. Emailed PA Digital today to ask if this info was actually being used by the or DPLA. To help us decide how important it is to do this now. 
   * decided not to do this for now based on feedback from PA Digital that the metadata is not currently being used. See comment below

- [x] Use `friendlier_id` as our oai-pmh exposed IDs, instead of the  UUID pk in `id`. While it might not really matter much, it seems nice to leave our oai-pmh feed exactly compatible with the feed in chf_sufia, including re-using ids from there. Have a PR to oai gem to make this possible, waiting on merge of that. https://github.com/code4lib/ruby-oai/pull/76

Closes #304


app/presenters/description_display_formatter.rb
```diff
@@ -1,26 +1,32 @@
...
-
-# DescriptionDisplayFormatter.new(work.description).format
-# DescriptionDisplayFormatter.new(work.description, truncate:true).format
-
+#
+#     DescriptionDisplayFormatter.new(work.description).format
+#     DescriptionDisplayFormatter.new(work.description, truncate:true).format
+#
+# Or for a plain-text description, with html tags stripped:
+#     DescriptionDisplayFormatter.new(work.description, truncate:true).format_plain
```
eddierubeiz: Interesting -- two different output methods on the same formatter object. I kinda like this better than passing two different parameters to the same method.

eddierubeiz:
Strongly agree with keeping `friendlier_id` as the identifier.

Issue #253, opened 2019-07-24
allow for editing/deleting comments within digitization queue item view


yonyitz:
Please allow for editing/deleting comments within digitization queue item view.  Editing/deleting should only be permissible by author.

PR #303, opened 2019-08-20
WIP: Delete digitization cue comments.

eddierubeiz:
Fixes https://github.com/sciencehistory/scihist_digicoll/issues/253

TODO:
- [ ] Consider using cancancan permissions - currently explicitly checking.
- [ ] DRYer routes.rb
- [ ] Add tests


jrochkind:
We're actually using the similar [access-granted](https://github.com/chaps-io/access-granted) rather than cancancan. 

But agree we should be doing that before this is merged.

Issue #232, opened 2019-07-15
fixity checking


jrochkind:
Outstanding things:

- [x] Report of fixity checks for an individual asset on asset admin page
- [x] ability to request "fixity check now/soon" for a specific asset
- [x] Fixity check summary "dashboard" 
- [x] Email/slack notify on any fixity check failure
- ~import old fixity check data from chf_sufia?~ (will not do)

eddierubeiz:
Basic idea:
```
Asset.find_each do |asset|
   asset.sha512 == calculated_512(asset)
end
```
Only run checks on items where `asset.stored?==true` because if it’s not finished ingesting, it won’t have a sha512 yet

If asset.sha512 is null, report that.

jrochkind:
Other hints/ideas (some maybe wrong):

The basic way fixity checking was done before involves creating a database record (it was called something weird before, but we could call it... `FixityCheckLog`) that represents a logged fixity check. It has success(boolean), the ID of the Asset checked (we might want to store our UUID actual pk/fk for easy rails relationships, along with the friendlier_id for easy reporting), and the actual/expected checksums. 

One tricky thing is we don't want to keep these log files forever, so after creating a new one, we had some clever SQL to delete ones older than most recent N, unless they were error logs, in which case we do want to keep them forever, or the FIRST log on record, which we also want to keep forever (so we know earliest verified fixity). That tricky "clean up old log records" can be left for last. 

we want to use sha512 checksums here, although chf_sufia may have used md5. md5 is deprecated in general (in the whole world), sha512 is the newness. 

Another tricky thing is how you get the remote file (via shrine) to calculate it's checksum. And ideally doing it without having to read the whole thing into memory, or write the whole thing to disk -- check streaming. 

`asset.file.open(rewindable: false)` should give you an "IO"-like object.

One way to calculate checksum from it... I got this from looking into shrine source and seeing that maybe we can re-use how shrine does it:

I think something like this (some of this I got from looking :

```ruby
io_object = asset.file.open(rewindable: false)
sha_512_calculated = Shrine::Signature::SignatureCalculator.new(:sha512, :hex).call(io_object)
```

* In addition to batch, there should be some kind of "service" class to fixity check (and log to db) an individual asset. Cause we may want a "fixity check this now" button on Asset admin page. Should probably be an ActiveJob for individual fixity check to be scheduled in jobs system.

eddierubeiz:
[...]

PR #315, opened 2019-08-27
Fixity Checking

eddierubeiz:
A model, with its own database table, that can check the fixity of an Asset and record the result.

TODO:
- [x] Tests on the model.
- [x] Rake task for running checks on all assets.
- [x] Rake task for pruning stale logs of checks
- Reports (this might be in another PR). (different PR)
- Info re: an asset's checks on the asset page (also maybe in a separate PR). (different PR)

Ref #232

### DIFF
--- a/spec/models/fixity_check_spec.rb
+++ b/spec/models/fixity_check_spec.rb
@@ -55,7 +55,7 @@ describe FixityCheck do
   let(:good_asset) { build(:asset_image_with_correct_sha512) }
   let!(:twenty_checks) do
     (0..19).to_a.map do |x|
-      FixityCheck.create!(
+      create(:fixity_check,
         asset: good_asset,
         checked_uri: good_asset.file.url,
         created_at: Time.now() - 100000 * x,
--- a/db/migrate/20190912200533_add_fixity_check_non_null.rb
+++ b/db/migrate/20190912200533_add_fixity_check_non_null.rb
@@ -1,0 +1,17 @@
+class AddFixityCheckNonNull < ActiveRecord::Migration[5.2]
+  def up
+    change_column_null :fixity_checks, :actual_result, false, "[missing]"
+    change_column_null :fixity_checks, :expected_result, false, "[missing]"
+    change_column_null :fixity_checks, :hash_function, false, "[missing]"
+    change_column_null :fixity_checks, :checked_uri, false, "[missing]"
+    change_column_null :fixity_checks, :passed, false, false
+  end
+
+  def down
+    change_column_null :fixity_checks, :passed, true
+    change_column_null :fixity_checks, :checked_uri, true
+    change_column_null :fixity_checks, :hash_function, true
+    change_column_null :fixity_checks, :expected_result, true
+    change_column_null :fixity_checks, :actual_result, true
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -9,34 +9,6 @@ SET client_min_messages = warning;
 SET row_security = off;
 
 --
--- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
---
-
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
-
-
---
--- Name: EXTENSION plpgsql; Type: COMMENT; Schema: -; Owner: -
---
-
-COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
-
-
---
--- Name: pg_trgm; Type: EXTENSION; Schema: -; Owner: -
---
-
-CREATE EXTENSION IF NOT EXISTS pg_trgm WITH SCHEMA public;
-
-
---
--- Name: EXTENSION pg_trgm; Type: COMMENT; Schema: -; Owner: -
---
-
-COMMENT ON EXTENSION pg_trgm IS 'text similarity measurement and index searching based on trigrams';
-
-
---
 -- Name: pgcrypto; Type: EXTENSION; Schema: -; Owner: -
 --
 
@@ -162,10 +134,10 @@ ALTER SEQUENCE public.digitization_queue_items_id_seq OWNED BY public.digitizati
 CREATE TABLE public.fixity_checks (
     id bigint NOT NULL,
     asset_id uuid NOT NULL,
-    passed boolean,
-    expected_result character varying,
-    actual_result character varying,
-    checked_uri character varying,
+    passed boolean NOT NULL,
+    expected_result character varying NOT NULL,
+    actual_result character varying NOT NULL,
+    checked_uri character varying NOT NULL,
     hash_function character varying DEFAULT 'SHA-512'::character varying NOT NULL,
     created_at timestamp without time zone NOT NULL,
     updated_at timestamp without time zone NOT NULL
@@ -664,13 +636,6 @@ CREATE UNIQUE INDEX index_users_on_reset_password_token ON public.users USING bt
 
 
 --
--- Name: trgm_idx_kithe_models_title; Type: INDEX; Schema: public; Owner: -
---
-
-CREATE INDEX trgm_idx_kithe_models_title ON public.kithe_models USING gin (title public.gin_trgm_ops);
-
-
---
 -- Name: kithe_model_contains fk_rails_091010187b; Type: FK CONSTRAINT; Schema: public; Owner: -
 --
 
@@ -764,7 +729,6 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20181211182457'),
 ('20190107205722'),
 ('20190107222521'),
-('20190109000356'),
 ('20190110154359'),
 ('20190219225344'),
 ('20190226135744'),
@@ -775,4 +739,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20190422201311'),
 ('20190716180327'),
 ('20190827124516'),
-('20190910160148');
+('20190910160148'),
+('20190912200533');
+
+
--- a/spec/factories/fixity_check_factory.rb
+++ b/spec/factories/fixity_check_factory.rb
@@ -1,0 +1,9 @@
+FactoryBot.define do
+  factory :fixity_check, class: FixityCheck do
+    actual_result { "fake_actual_digest" }
+    expected_result { "fake_expected_digest" }
+    hash_function { "SHA-512" }
+    checked_uri { "http://example.com/fake/location" }
+    passed { false }
+  end
+end
--- a/spec/models/assets_needing_fixity_checks_spec.rb
+++ b/spec/models/assets_needing_fixity_checks_spec.rb
@@ -35,13 +35,24 @@ describe ScihistDigicoll::AssetsNeedingFixityChecks do
   end
 
   describe "#assets_to_check" do
+    # Work around FactoryBot not liking explicit created_at setting.
+    around do |example|
+      original = FixityCheck.record_timestamps
+      FixityCheck.record_timestamps = false
+      example.run
+      FixityCheck.record_timestamps = original
+    end
+
+
     let!(:blank_one) { create(:asset) }
-    let!(:old_one) { create(:asset, fixity_checks: [FixityCheck.new(created_at: 1.year.ago)]) }
+    let!(:old_one) { create(:asset, fixity_checks: [
+      build(:fixity_check, created_at: 1.year.ago, updated_at: 1.year.ago)
+    ]) }
 
     before do
       # some newer ones
       4.times do |i|
-        create(:asset, fixity_checks: [FixityCheck.new(created_at: i.days.ago)])
+        create(:asset, fixity_checks: [build(:fixity_check, created_at: i.days.ago, updated_at: i.days.ago)])
       end
     end
 


eddierubeiz:
Looks like I've inadvertently made some unwelcome changes to db./structure.sql in this branch, including removing `pg_trgm` and `trgm_idx_kithe_models_title`. I'm assuming I can just start on a fresh copy of the master branch and rerun the `CreateFixityCheckLogs` migration and this won't happen.

jrochkind:
@eddierubeiz You shouldn't have to start on a fresh copy of master branch. 

Try just running `./bin/rake db:structure:dump`to create a fresh structure.sql file from the db schema. 

This may result in the same thing you are considering "unwelcome" though. But at least you'll see that quicker!

What are the unwelcome changes? I can debug them with you. 

Since structure.sql is dumping your pg schema, if you have changes you don't want _in your pg schema_, even "start on a fresh copy of master branch" probably wouldn't work, you'd still wind up with the same thing too. No need to spend time juggling a fresh copy of master branch, just do db:structure:dump and if there are still problems, let's diagnose and fix them ourselves.

eddierubeiz:
Thanks for the tip. If you look at https://github.com/sciencehistory/scihist_digicoll/pull/315/files#diff-f9ae8314082d8229b38da27dd6c2642d you'll see there's a bunch of stuff there that's unrelated to what I'm trying to do in this PR, namely add a new table and a couple new indices.

I tried `db:structure:dump` and it didn't change anything, so I'm going to try and delete and recreate the DB from scratch in dev.

eddierubeiz:
Some notes:
 - OK to rename the AR model FixityCheck instead of FixityCheckLog
 - JR suggests moving the actual fixity checking logic out of the AR model into another class

jrochkind:
One thought: Should we save it in the db not as the row digest/checksum, but prefixed with the algorithm. To make it transparent what it was, and to allow changing algorithm while retaining historical records. 

There is some standard for how you do this, but I can't find it/remember what it is, but pretty sure it's just `sha512:xxxxx` or `md5:xxxx` or whatever.

eddierubeiz:
That could work, or we could just have type field in the table, right?

app/services/fixity_checker.rb
```diff
@@ -0,0 +1,117 @@
...
+  #FixityChecker.new(asset).check
+  def check(sieve_integer = 0)
+    return unless sift(sieve_integer)
+    return nil if @asset.nil?
+    return nil if @asset.file.nil?
+    the_expected_checksum = expected_checksum
+    # Note: @asset.file.sha512 can be nil on e.g.
+    # items imported with DISABLE_BYTESTREAM_IMPORT=true
+    # For now, we are ignoring these.
+    return nil if the_expected_checksum.nil?
```
jrochkind: I think you should raise if you can't fixity check cause don't have a fixity on file, instead of silently ignoring/failing.

jrochkind: I think it should raise an exception if it can't fixity check, rather than silently ignore/fail. Fixity checking too important to make it that easy to not know you didn't do one. The caller would be responsible for not calling `check` unless it can be checked. 

I think that applies to the @asset.nil? and @asset.file.nil? checks above too -- they should just raise. Maybe actually check the @asset argument in the initializer, and raise argument error there if you got nil or an asset that can't be fixity checked?

Ah, I see the problem though, then the rake task has to know if an asset can be fixity checked or not before trying... But in fact, any asset that is `stored?` (has had ingest finalized) ought to be able to be checked -- if something that is `stored?` can't be checked, we want to find out about it one way or another, it's a software problem, we don't want it to be silently skipped and not checked. So rake task could just check `stored?` maybe.

app/services/fixity_checker.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  # @param work [Work] Work object, it's members will be put into a zip
+  # @param callback [proc], proc taking keyword arguments progress_i: and progress_total:, can
+  #   be used to update a progress UI.
+  def initialize(asset)
+    raise ArgumentError.new("Please pass in an Asset.") unless asset.is_a? Asset
+    @asset = asset
+  end
+
+  #FixityChecker.new(asset).check
```
jrochkind: Needs more docs saying what it will do (create a FixityCheck log record in db? And/or delete old ones?), and what the `sieve_integer` arg does.

I'm not sure about your approach to `sieving` -- although it is quite smart (nice job), I am not sure if we want a way to avoid even _fetching_ the things that we are not going to be checking. The "sifting"/"sieve" stuff needs to be better-documented if we're going to keep it, but let's talk about it first and see if we want to do it this way or not. 

if we keep doing it this way, I think the sifting logic should not be in `check` -- if you call `check`, you should get a check. The sifting stuff should be lifted to somehow be outside of `check`, to decide if you want to call `check`. 

But let's talk through the overall sifting approach before you spend time making any changes to it.

app/services/fixity_checker.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  private
+
+  # Returns an array of checks that we don't want to keep.
+  # The current rules are:
+  # Never throw out FAILED checks
+  # Never throw out the earliest PASSED check
+  # Always keep N recent PASSED checks
+  def checks_its_ok_to_delete
```
jrochkind: Method name `checks_to_delete` slightly better I think. May not matter (bike shedding), so ignore if you want.

app/services/fixity_checker.rb
```diff
@@ -0,0 +1,117 @@
...
+    new_check.save!
+  end
+
+  #FixityChecker.new(asset).prune_checks
+  # After pruning, for each asset - URI combination, you will be left with:
+  # up to NUMBER_OF_RECENT_PASSED_CHECKS_TO_KEEP recent passed checks,
+  # plus all the failed checks,
+  # plus one initial check (to remember the file that was intially uploaded.)
+  def prune_checks(sieve_integer = 0)
+    return unless sift(sieve_integer)
```
jrochkind: I don't understand what's going on with `sift`. If later `checks_its_okay_to_delete` returns an empty array, then none will be deleted. What is the extra `sift` for? Ah, I see, it's about trying to create a 'cycle' -- I am not sure this is doing it a good way, I need you to talk me through what it's doing.

app/services/fixity_checker.rb
```diff
@@ -0,0 +1,117 @@
...
+      raise ArgumentError.new(
+        "Expected a positive int between  1 and #{CHECK_CYCLE_LENGTH}. Got #{sieve_integer}"
+      )
+    end
+  end
+
+  def actual_checksum
+    sha512 = Digest::SHA512.new
+    @asset.file.open(rewindable:false) do |io|
+      if io.is_a? File
```
jrochkind: Curious in what cases you were getting a `File` here?  Does `each_chunk` not work on a `File`? I see it does not! Annoying that we need two different paths here depending on what we get from `file.open`, I'm going to bring it up with janko from shrine.

lib/tasks/check_fixity.rake
```diff
@@ -0,0 +1,35 @@
...
+    if asset_count == 0
+      abort ("No assets found to check.")
+    end
+    progress_bar = ProgressBar.create(total: asset_count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
+    progress_bar.log("INFO: pruning the logs of asset fixity checks")
+    Asset.all.each do |asset|
+      FixityChecker.new(asset).prune_checks
+      progress_bar.increment
+    end
+  end
```
jrochkind: I don't think there should be a separate prune_fixity_checks task. Rather, in `check_fixity`, you should just prune after checking. Maybe this works?

```ruby
checker = FixityChecker.new(asset)
checker.check
checker.prune_checks
```

Is there any reason you wouldn't want to prune checks after checking in the rake task?

I think in chf_sufia, it even automatically pruned as _part_ of the `check` method, you got a prune every time you called check. Which I think a case could be made for -- is there any case at all we'd want to check without pruning? Say, we have a "check this asset now" button in admin UI -- you'd want to prune after adding a new check, and making it a separate method makes it easy to forget to do that. So maybe it should just be automatically called by `check` method -- but i'm not sure, it's okay to leave it separate for now, but I am feeling confident the `check_fixity` rake task should call it for each thing it checks. 

(Ah, I see how this interacts with the "sifting" stuff maybe? Yeah, let's talk about sifting stuff).

spec/factories/asset_factory.rb
```diff
@@ -87,6 +91,14 @@
...
       end
 
+      trait :mp3_with_correct_sha512 do
+        faked_file { File.open((Rails.root + "spec/test_support/audio/ice_cubes.mp3")) }
+        faked_sha512 { Digest::SHA512.file(Rails.root + "spec/test_support/audio/ice_cubes.mp3").to_s }
+        faked_content_type { "audio/mpeg" }
+        faked_height { nil }
+        faked_width { nil }
+      end
+
```
jrochkind: I think _maybe_ you could do with this our existing traits/factories, but it's confusing adding a new one is fine! But I think this should be a sub-factory not a trait. You can't really combine it with other traits in ways that makes sense, just make it a sub-factory.

app/services/fixity_checker.rb
```diff
@@ -0,0 +1,117 @@
...
+    # items imported with DISABLE_BYTESTREAM_IMPORT=true
+    # For now, we are ignoring these.
+    return nil if the_expected_checksum.nil?
+
+    new_check = FixityCheck.new(asset: @asset)
+    new_check.expected_result = the_expected_checksum
+    new_check.checked_uri = @asset.file.url
+    new_check.actual_result = actual_checksum
+    new_check.passed= (the_expected_checksum==actual_checksum)
+    new_check.save!
```
jrochkind: Let's do this all in one statement like this:

```ruby
FixityCheck.create!(
  asset: @asset,
  checked_uri: @asset.file.url,
  actual_result: actual_checksum,
  passed: (the_expected_checksum==actual_checksum)
)
```

But that makes me think of a couple things. 

1. Should we be storing the expected checksum too in the FixityCheck log, as an added check, in case it gets corrupted or disappears in the 'main' place it is stored? So we have a record of exactly what we expected -- in case of passing, that's the same as "actual", but in case of failing it might be different, and we might want it recorded here. 

2. Saving @asset.file.url is a bit odd. For instance, if it's on S3 and shrine doesn't think it's a public ACL, it'll be a time-limited signed S3 URL which won't actually be accessible in the far future. So saving URL isn't quite right. We could try to come up with the 'right' thing to save... but I wonder if we need it at all, the asset ID (along with _expected_ as well as actual checksum) might be enough to preserve?

3. As discussed, let's store that the checksums are SHA512 somewhere. Another column, or a prefix "sha512:[checksum]", not sure which is best, another column is fine if you think so. 

Also, how about calling the method `expected_checksum` to match `actual_checksum` instead of `the_expected_checksum`, `the` in method name is unnecessary and mismatched. 
)

And let's make the method names match the column/attribute names in the FixityCheck model. Either `actual_checksum` for both or `actual_result` for both, not `checksum` for one and `result` for the other.

app/services/fixity_checker.rb
```diff
@@ -0,0 +1,117 @@
...
+    if !(sieve_integer.is_a? Integer)        ||
+      sieve_integer < 1                      ||
+      sieve_integer > CHECK_CYCLE_LENGTH
+      raise ArgumentError.new(
+        "Expected a positive int between  1 and #{CHECK_CYCLE_LENGTH}. Got #{sieve_integer}"
+      )
+    end
+  end
+
+  def actual_checksum
```
jrochkind: This calculates a checksum every time you call it, it doens't have memoization. 

So if you call it twice, you'll calculate it twice. And for instance, you do call it twice when you create the FixityCheck record, once for setting as the value of `actual_result`, and once for checking `(expected_checksum == the_actual_checksum)` -- so you're calculating checksum twice for every check, unnecessarily! And it is a fairly expensive operation, the most expensive thing `check` is doing. 

So either memoize it in the `actual_checksum` method (which I think would be fine), or note in comments that it gets calculated on every call, _and_ adjust the `check` method (and anywhere else) to make sure to put the result in a local var and re-use it, instead of calculating twice. I think memoizing is fine, but either would work.

app/services/fixity_checker.rb
jrochkind: Okay, here's how shrine actually does calculation:  https://github.com/shrinerb/shrine/blob/master/lib/shrine/plugins/signature.rb#L101-L106

So, if we do it shrine's way a) I trust that janko knows what he's doing it and it's efficient, and b) it should work with File or Down::ChunkedIO without requiring a separate path.

It might look like this:

```ruby
@asset.file.open(rewindable:false) do |io|
   digest = Digest::SHA512.new
   digest.update(io.read(16*1024, buffer ||= String.new)) until io.eof?
   digest.hexdigest
end
```

(Tested in a console, that worked. Changed shrine's `.digest` to `.hexdigest` to get the hexadecimal one we want instead of binary bytes.)


We could leave a comment linking to shrine's code saying we copied that. On the other hand, this is weird code, maybe your simpler version is better?

lib/tasks/check_fixity.rake
jrochkind: Or maybe add a #check_and_prune method that just calls both? And the rake task (And most other places) can just call it. Header docs can suggest that's the one you want.

app/services/fixity_checker.rb
jrochkind: If you do want to get a reliable non-time-limited non-signed URL from S3, you can call `@asset.file.url(public: true)` to ensure it. 

It may give you an access denied if you actually try to access it and it wasn't public, but it will identify the file location in a clear way.

Shrine storage locations that aren't S3 will just ignore the `public:true`... I think. Unless they raise on it. 

Shrine's back-end storage agnosticism ends up making it a bit difficult to get the "really, where was this thing actually stored" in a clear way. Not totally sure if we need to try to store that or not.

app/services/fixity_checker.rb
eddierubeiz: Great advice.

spec/factories/asset_factory.rb
eddierubeiz: Sub-factories! What a concept!

eddierubeiz: Fixed

lib/tasks/check_fixity.rake
eddierubeiz:
```checker = FixityChecker.new(asset)
checker.check
checker.prune_checks``` is perfect, with some memoization.

**Comment by eddierubeiz** [comment]
> One thought: Should we save it in the db not as the row digest/checksum, but prefixed with the algorithm. To make it transparent what it was, and to allow changing algorithm while retaining historical records.
> 
> There is some standard for how you do this, but I can't find it/remember what it is, but pretty sure it's just `sha512:xxxxx` or `md5:xxxx` or whatever.

Actually went with a separate column in the table. (Currently it's a dumb string type; we can change to an `enum` if we feel like it and save some (possibly a ton of) space.)

File: app/services/fixity_checker.rb (hunk 1 above)
Comment (eddierubeiz) [reply app/services/fixity_checker.rb]: Modified rake task to check whether the asset is `stored?` before checking it, and added corresponding exceptions.

File: app/services/fixity_checker.rb (hunk 4 above)
Comment (eddierubeiz) [reply app/services/fixity_checker.rb]: Moved the sift function out to the rake task where it makes more sense to put it, and also rewrote it and commented it a bit better.

File: app/services/fixity_checker.rb (hunk 5 above)
Comment (eddierubeiz) [reply app/services/fixity_checker.rb]: Used the shrine recipe-- looks great to me.

File: app/services/fixity_checker.rb (hunk 8 above)
Comment (eddierubeiz) [reply app/services/fixity_checker.rb]: Using the public:true -- I definitely don't want these URIs expiring on us. The idea is to make sure everything works in case an asset gets a different file some day. (Kind of an edge case.)

File: app/services/fixity_checker.rb (hunk 9 above)
Comment (eddierubeiz) [reply app/services/fixity_checker.rb]: Memoized!

**Comment by jrochkind** [comment]
Using pg enums with ActiveRecord properly takes some third-party extensions to AR. But it could be done. I have no strong opinions that it's necessary now.

File: spec/models/fixity_check_spec.rb [hunk 10]
```diff
@@ -0,0 +1,77 @@
...
+
+describe FixityCheck do
+
+  let(:good_asset)    { build(:asset_image_with_correct_sha512) }
+
+  let(:good_asset_2)  { build(:asset_mp3_with_correct_sha512) }
+  # :asset_with_faked_file, by factory default,comes with a random fake sha512.
+  let(:corrupt_asset) { build(:asset_with_faked_file) }
+
+  it "can check an asset's fixity sha512" do
```
jrochkind: So, normally rspec recommends you split things into separate indpendent tests for readability and so when one fails it's clear _what_ failed. 

Sometimes we violate that when the test setup is so expensive and we don't want it to be run for each test. 

That _might_ be true here. 

But think it may not be worth it here regardless, this is an awful lot  in one `it` example, it makes it hard to figure out what it's checking, and if one of the tests fail in the future it may make it confusing what functionality broke. Split into more tests I think?

Issue #249, opened 2019-07-23
language autocomplete not functioning


yonyitz:
Language autocomplete on metadata edit form does not function as it does in current app.  Please utilize current app's word bank.

jrochkind:
Related to #170, similar technical mechanisms will be developed to handle both, but good for the explicit record that it needs to be slightly different for language, thanks!

jrochkind:
This was accomplished in #323

PR #323, opened 2019-08-29
Autocomplete fields on admin work edit

### LINKED ISSUE #170: auto-complete subject entry
Not yet implemented in new app like it is in chf_sufia, is needed.

jrochkind:
Subject, Place, Creator (From FAST), and Language (from local controlled vocab). 

Using the `qa` gem, which provides lookups from vocabs and HTTP actions returning JSON results. (chf_sufia used this). 

And local custom Javascript for autocomplete based on the devbridge-autocomplete jquery plugin. (Which requires some hacks to get it working well).  (This is different than what chf_sufia used). See docs at top of qa_autocomplete.js file for some overview of what's going on JS-wise. 

Had to kind of hack into the internals of devbridge-autocomplete to get it working how we wanted -- but it works! I did pretty extensive research looking for a JS autocomplete/autosuggest that was working, did what we needed, had relatively small delivery file size, and was not _entirely_ abandoned -- this was the best I could find. (Maybe future project to fork it to our own autocomplete, which doesn't depend on jquery and has the features we need. devbridge-autocompelte is maintained for bugfixes, but no new features are being added -- which also means reaching into undocumented internal API should be pretty safe, since it's not going to change!)

Closes #170 and #249


app/javascript/src/js/admin/qa_autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+// -- although really we wanted to disable "show suggestions onFocus" altogether to match
+// chf_sufia behavior.
+//
+// We fix things up mostly with a few kind of hacky private API interventions into the plugin --
+// but hey, this still isn't much code, and the plugin isn't getting much development so
+// it probably won't break.
+//
+// We could try to add features we need upstream (https://github.com/devbridge/jQuery-Autocomplete/issues/769),
+// we could consider forking this (it's not actually much code), or we could in the future provide a
+// better alternate tool for a JS auto-suggest UI.
```
eddierubeiz: Noted.

PR #324, opened 2019-08-29
when running tests, keep sitemap_generator output from messing up our output

jrochkind:


eddierubeiz:
You'll have to explain what this fixes later, but I'll go ahead and merge it.

PR #325, opened 2019-08-29
fix related items styling on work show

### LINKED ISSUE #322: Related Items CSS styling not quite right
See

https://kithe.sciencehistory.org/works/8623hz248

vs 

https://digital.sciencehistory.org/works/8623hz248

jrochkind:
Mostly copied from chf_sufia. Closes #322


eddierubeiz:
Never noticed this was missing.

PR #334, opened 2019-09-04
avoiding expensive work in DownloadDropdownDisplay

jrochkind:
The `DownloadDropdownDisplay` renders the list of download links for an individual asset. So on a work show page, it is rendered once each for every child image shown. 

It needs to decide whether to include links for our multi-image "whole-work" PDF/zip downloads. At present, we want to include those only for works whose children are _all_ images (and which have more than one). So `DownloadDropdownDisplay#has_work_download_options?` simply checked, before this PR:

```
      display_parent_work.members.all? do |member|
        member.leaf_representative&.content_type&.start_with?("image/")
      end
```

I knew this would be a problem if it resulted in SQL queries (on `.members` or `some_member.leaf_representative` -- the "n+1 query problem"). But it was not doing so, I made sure all data was efficiently eager-loaded. So I figured this would not be a perf problem, so long as it was not triggering SQL. 

But it was! If we think about it, Ramelli has ~700 members. So `DownloadDropdownDisplay` will be rendered 700 times. And each time it will check `start_with?("image/")` for _all 700 members_ of Ramelli. So that's 700x700==490,000 `start_with?` queries. While 500k of everything is not peanuts, I'm _still_ surprised that it added as much time as it did -- it may be because of inefficiencies in shrine in accessing shrine-serialized attributes (which content_type is), which may have been fixed in upcoming shrine 3.0. 

But wherever it's coming from, I confirmed that _avoiding_ this repeated check brought Ramelli from 10s to 4-5s on deploy server. 

The solution is checking whether the parent work's members are all images _once_, and letting all the `DownloadDropdownDisplay` re-use this calculation -- since they all have the same parent work, and the answer is the same for all of them. 

The trick is it's unclear how to get this calculation to all executions of `DownloadDropdownDisplay`, they are sometimes called through a chain of ViewModels, and a different one in different contexts. So I just added it as a memoized method on `Work` itself. I don't love that architecture, but it works, it's simple for now, and it saves us like 50% of response time, so. (A universal "Work decorator" might be another solution, but I am reluctant to add another architectural concept into our views). 

This PR also changes rights statement rendering to be from the `display_parent_work` passed in, rather than the direct `asset.parent`. These two works will usually be the same thing, they may not be in case of intermediate child works, but even then we expect the rights statement to be the same, and I just wanted to avoid any potential for accidental n+1.


jrochkind:
Very mysterious why test is failing... I managed to reproduce in development only by creating _exactly_ the same data as in the test, but now I have!

I think it's probably a pre-existing edge case bug that the new code exposed, but we'll see!

eddierubeiz:
Makes a lot of sense. Checks are failing but feel free to merge at your leasure. I read the description and studied the code.

PR #337, opened 2019-09-05
Fix copying of static asset pages on initial deploy

### LINKED ISSUE #336: Can't precompile assets on fresh production server
The bug is in my custom rake task to try and copy the static error pages to a non-fingerprinted URL location in public/. 

It first has to get the fingerprinted location generated by sprockets. On a fresh server, when it tries to do that, it is given a nil location, the fingerprinted location doesn't exist yet. 

I think this means even on a non-fingerprinted server, we may be copying the _previous_ static error pages, not the ones generated with this deploy. Which would be wrong too. 

We may be able to fix by changing the rake task hook at which our custom copy code runs, although it seems like it should be right now, very odd. 

----

[scihist_digicoll-EXPERIMENTAL/staging] TypeError: no implicit conversion of nil into String
## Backtrace

    line 36 of [PROJECT_ROOT]/lib/tasks/assets_compile.rake: block (3 levels) in <main>
    line 34 of [PROJECT_ROOT]/lib/tasks/assets_compile.rake: each
    line 34 of [PROJECT_ROOT]/lib/tasks/assets_compile.rake: block (2 levels) in <main>

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/54442034)

jrochkind:
We have a custom extension to rake assets:precompile that copies our statically compiled error html pages to un-fingerprinted (eg) public/404.html, so it'll be used by Rails.

I believe the routine for looking up the sprockets-compiled digest-named error pages was using a cached value and finding the one compiled _previous_ to this deploy. So it was the wrong one if there were changes in this deploy. And on _first_ deploy to a new server, it was nil, which also produced an uncaught exception with it not expected to be nil.

Rooting around in the sprockets and rails source, I found a different way of looking up compiled asset path. Which seems to find the right current one, not previous one. It is not nil even on initial deploy, so can work on initial deploy.

I also fixed source so if it _can't_ find it, it fails gracefully with an error message instead of halting deploy. Which it was meant to do all along, but it had tried to access the `digest_relative_path` variable prior to the conditional that checked for it being set. That is now fixed.

Closes #336

### DIFF
--- a/lib/tasks/assets_compile.rake
+++ b/lib/tasks/assets_compile.rake
@@ -32,11 +32,12 @@ namespace :scihist do
   #   https://stackoverflow.com/questions/17536023/rake-assetsprecompilenodigest-in-rails-4/26049002
   task :create_non_digest_assets => :"assets:environment"  do
     %w{404.html 500.html}.each do |file|
-      digest_relative_path = Rails.application.assets_manifest.assets[file]
-      digest_path          = Rails.root + 'public/assets' + digest_relative_path
-      destination_path     = Rails.root + 'public/' + file
+      digest_relative_path = Rails.application.assets.find_asset("404.html")&.digest_path
 
       if digest_relative_path
+        digest_path          = Rails.root + 'public/assets' + digest_relative_path
+        destination_path     = Rails.root + 'public/' + file
+
         logger.info "(Local asset_compile.rake) Copying to #{destination_path}"
 
         # Use FileUtils.copy_file with true third argument to copy


lib/tasks/assets_compile.rake
```diff
@@ -32,11 +32,12 @@ namespace :scihist do
   #   https://stackoverflow.com/questions/17536023/rake-assetsprecompilenodigest-in-rails-4/26049002
   task :create_non_digest_assets => :"assets:environment"  do
     %w{404.html 500.html}.each do |file|
-      digest_relative_path = Rails.application.assets_manifest.assets[file]
-      digest_path          = Rails.root + 'public/assets' + digest_relative_path
-      destination_path     = Rails.root + 'public/' + file
+      digest_relative_path = Rails.application.assets.find_asset("404.html")&.digest_path
```
eddierubeiz: You probably mean:
```digest_relative_path = Rails.application.assets.find_asset(file)&.digest_path```

eddierubeiz: Oh no wait, that's not a mistake -- you're just assuming both files are in the same path, so you're using one arbitrarily.

jrochkind: No, you actually caught a bug! Although you are right that it probably would work anyway for the reason you said, that's not what I meant to do. 

Since you've already merged, I'll prepare another PR!

jrochkind: (And thank you, this is why I appreciate actual review! :) )

jrochkind: actually that was a real bug, it probably resulted in the 500.html page being a copy of the 404 instead of it's own thing! I happen to know how to force a 500 on our current app due to unrelated bug... oh weird, no I don't, that bug went away! But yeah, it should be fixed.

Issue #296, opened 2019-08-15
Performance on many-membered pages bad


jrochkind:
At one point our favorite test page Ramelli was rendering pretty fast. But it is now far too slow. 

Just now, trying Ramelli on scihist_digicoll staging, takes **6s** to load. This is way too slow. The `viewer_images_info` action that returns JSON info (including URLs) for a work is another good example, as it has almost no 'view' logic, just json representation of data. It took 1.21s to load (still too slow). 

* This is still faster than current chf_sufia production. Although I am positive we used to have chf_sufia production faster than it is now, after much work performance tuning, it appears to have had some performance regression since then. Testing just now on chf_sufia, ramelli took `10s` to load, which is _entirely_ unacceptable. (And the viewer_images_info took ~4s). I suspect some n+1 has crept in. :(

We could make Ramelli HTML faster by _not_ loading all children at once (say, having JS that auto-loads more if you scroll down to actually try to view them).  But that technique wouldn't work for the `viewer_images_info` that needs to return JSON for everything to the viewer (unless we improved the viewer to have much more sophisticated logic where it fetches only info for images it needs, but this would get complicated). 

Ramelli on scihist_digicoll used to be much faster, but at some point got slow as we added functionality. 

*ONE* hypothesis, backed up by some data, for where the slowdown is, is how we load _all_ derivatives for all (leaf-representatives of all)  members.  Just loading those AR models is some slowdown -- but there is even *more* slowdown from accessing the shrine file info for a derivative (requires a JSON parse and instantiation of lots of objects for shrine), and even *more than that* for asking for the S3 URL for a derivative. 

On my macbook, testing straight loads of derivatives (2080 deriv records):
* just fetching them all: 30ms
* fetching, and asking for shrine to create an UploadedFile for each one: 77ms
* fetching, asking for UploadedFile, then asking the UploadedFile for a (public) s3 URL: 200ms (this shoudln't take that long for public non-signed S3 URLs, but it's the amazon aws_sdk ruby gem that's doing it...)

So we can see it getting much longer for each step. 600 members on ramelli, each with ~10+ derivatives, when our code fetches, looks at shrine info, and asks for URLs for many of them... can be a lot of derivatives. 

It may be enough to explain the ~1.2s `viewer_images_info` load (still too long)... but then there's another 5s for the web page, not entirely sure where that's going. 

[...]

PR #338, opened 2019-09-05
DownloadDropdownDisplay no longer uses derivative models.

jrochkind:
Instead it just creates blind links based on what derivatives it knows ought to exist for a given content type.

This means we can no longer supply file size of download to user with link (we would get that from deriv model). It also means we no longer check to make sure a derivative really exists before making download link -- if a user clicks on a download link for a deriv that doesn't actually exist, they'll just get a 404. These should be the only UI changes/degradatins.

This is an attempt at performance improvement. Accessing shrine json metadata (for file size, width, and height) seems to be a mysteriously expensive operation (may be fixed in shrine 2).

It also means that as far as DownloadDropdownDisplay is concerned, we COULD stop fetching/pre-loading derivative models at all. Which would be additional performance improvements. At present we are still preload fetching them for use as thumbnail src though. We will benchmark how much of an improvement we get from this step, without that.

ref: #296

### DIFF
--- a/app/presenters/download_dropdown_display.rb
+++ b/app/presenters/download_dropdown_display.rb
@@ -84,6 +84,7 @@ class DownloadDropdownDisplay < ViewModel
     end
   end
 
+
   private
 
   def viewer_template_mode?
@@ -221,6 +222,7 @@ class DownloadDropdownDisplay < ViewModel
       url: download_path(asset),
       analyticsAction: "download_original",
       content_type: asset.content_type,
+      size: asset.size,
       width: asset.width,
       height: asset.height)
   end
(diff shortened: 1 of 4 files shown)


jrochkind:
This change alone (while still pre-loading derivatives models and using them for thumbs seems to get us to):

* ramelli html: 2.7-4s (was 4-5s)
* ramelli viewer_images_info 1.7-2s (was ~2.5s)

So some non-trivial improvement even from this, but interested to see how much more we can get by avoiding derivatives for thumbs and avoiding fetching them at all too.

jrochkind:
Experimented with removing display of file size even for original, on the the theory that ANY access of shrine metadata is dangerous (when we're doing it 700 times). 

Not sure if that resulted in additional time savings, or it's just statistical noise. It may have taken it down to 2.3-3.7s. (Not entirely sure why we still get so much distribution in timing).

jrochkind:
Will not be doing, kithe/shrine upgrade should let us get similar benefit without giving up functionality.

PR #340, opened 2019-09-05
update slackistrano gem

jrochkind:
Slack updates from capistrano. Just because why not, and we had a mysterious warning message from the old version we're hoping to be rid of.

And fix URL for github diff compare. It was old chf_sufia url.


lib/scihist_digicoll/slackistrano_messaging.rb
```diff
@@ -27,7 +27,7 @@ def payload_for_updated
             payload[:attachments] ||= []
             payload[:attachments] << {
                 title: "Github Diff",
-                title_link: "https://github.com/chemheritage/chf-sufia/compare/#{CGI.escape previous}...#{CGI.escape current}",
```
eddierubeiz: Nostalgic, but yeah, time to change that.

PR #345, opened 2019-09-09
try yet again to figure out the current compiled asset-with-digest filename to create static 404.html and 500.html

jrochkind:
This is entirely too much reverse-engineering of crazy sprockets to get this to work, it's not reliable,
but not sure what other approach to use for reliable static 404 and 500 with our real stylesheets.

At some point we should probably escape the sunk cost fallacy, and just not do things this way. But this seems to work for now.


eddierubeiz:
Got it.

PR #347, opened 2019-09-10
Fixity reporting for one asset

eddierubeiz:
Simplest possible thing that could work.
Not using a ViewModel in this case; we can change our minds later if this gets much more complicated and farm this out to its own object.
Note: trying a different GitHub pattern here; the base branch is fixity_checks, since this really is a sub-feature.

### DIFF
--- a/app/services/fixity_checker.rb
+++ b/app/services/fixity_checker.rb
@@ -12,14 +12,17 @@ class FixityChecker
 
   # Creating a FixityChecker.new(asset) doesn't actually do anything.
   def initialize(asset)
+
+    raise ArgumentError.new("Please pass in an Asset.") if asset.nil?
+
     raise ArgumentError.new("Please pass in an Asset.") unless asset.is_a? Asset
 
     raise ArgumentError.new(
-        "#{@asset.friendlier_id} has no file associated with it, so we can't run a check on it."
+        "#{asset.friendlier_id} has no file associated with it, so we can't run a check on it."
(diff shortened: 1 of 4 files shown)


app/services/fixity_checker.rb
```diff
@@ -46,6 +49,15 @@ def prune_checks
...
   end
 
+  # A memoized list of checks for this asset's current file.
+  # Used for reporting on a particular asset on the asset view page
+  # without fetching the info more than once.
+  def checks_for_this_uri
+    return [] if @asset.file.nil?
+    return [] if @asset.file.url.nil?
+    @checks_for_this_uri ||= FixityCheck.checks_for(@asset, permanent_url)
+  end
```
jrochkind: believe this method is no longer needed in this class.

app/views/admin/assets/show.html.erb
```diff
@@ -66,6 +66,76 @@
...
+                  <% else %>
+                    <i class="fa fa-times" aria-hidden="true" style="color:red;"></i> <span class="sr-only">Failed</span>
+                  <% end %>
+                  <%= l ch.created_at, format: :admin %>
+                </div>
+              <% end %>
+            </dd>
+          <% end %>
+        </dl>
+      </dd>
```
jrochkind: This is a lotta lines, I dunno, what do you think about extracting it to a partial template? Not really sure, could go either way, what do you think?

PR #348, opened 2019-09-10
Additional search filters for Admin Work search

### LINKED ISSUE #342: More staff search filter options
After discussion, changed ticket to:

Staff Work search needs filtering options:

* published/not
* department
* genre
* format
----
old: 

This is so [embargoed content ](https://digital.sciencehistory.org/catalog?f%5Bsubject_sim%5D%5B%5D=Nobel+Prize+winners&f%5Bvisibility_ssi%5D%5B%5D=restricted)can be separated from staff-only content. Can visibility faceting on the back end be included so staff can view staff-only and private works (let me know if that should be a separate GitHub ticket)?  Thanks!

jrochkind:
1. I noticed that while our 'published' column is a postgres boolean, we were allowing SQL "null" values, so it was actually a tri-valued field. Which made things confusing in querying. Added a migration to not allow nulls, and default false. (Changing any existing nulls in db to false). 

2. Add on UI elements and logic for limiting by additional fields. Since some of the additional fields are inside our JSON, we use specialty postgres json operators in constructing our SQL conditions. (Especially odd for things like genre, where we're looking for one field in a JSON _array_ at a certain path). We had been using a gem called "ransack" to construct the query based on UI choices with little code (including supporting the sort-by-column-of-choice controls really easily), but it doesn't do custom JSON operators well. So now it's a weird frankenstein combo, but it owrks. (Looked at removing ransack entirely, wasn't worth rewriting everything that would have to be rewritten I don't think). 

3. I am not sure if these postgres JSON-operartor queries are actually using PG indexes, or just scanning all rows. I think probably scanning. There are ways to set up PG indexes for such queries, but I don't think we've done them. However, with the actual size of our db, the searches still tested quite fast enough on staging, so good enough for now. 

4. Not sure if this UI will end up confusing for staff users, we'll have to see.

Screenshot 2019-09-10 16 33 53

Closes #342

### DIFF
--- a/app/assets/stylesheets/local/admin/admin.scss
+++ b/app/assets/stylesheets/local/admin/admin.scss
@@ -3,3 +3,17 @@ dl.admin-metadata {
     @extend .ml-3;
   }
 }
+
+.admin-filter {
+  display: inline-block;
+  margin-right: 1rem;
+  margin-bottom: 1rem;
+
+  label {
+    display: block;
+  }
+  .custom-select {
+    max-width: 10em;
+    width: auto;
+  }
+}
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -10,15 +10,8 @@ class Admin::WorksController < AdminController
   # GET /admin/works
   # GET /admin/works.json
   def index
-    # weird ransack param, we want it to default to true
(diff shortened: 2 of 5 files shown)


eddierubeiz:
Not familiar with Ransack, but everything looks sane to me. Merging.

PR #351, opened 2019-09-11
We weren't actually properly creating faked/mocked/quicker Shrine::UploadedFiles in test

jrochkind:
Fixing that revealed a conflict with kithe, where we were trying to create derivatives in tests in a hacky but quicker way, but kithe was deciding to delete the derivatives attached pre-promotion after the file was promoted (on the theory that the original had changed, so derivatives were no longer applicable). Updated to a newer kithe that has a fix to allow us to do what we wanted: https://github.com/sciencehistory/kithe/pull/63

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,7 +7,7 @@ GIT
 
 GIT
   remote: https://github.com/sciencehistory/kithe.git
-  revision: 28b237e5cc77b0bd7a0d997b48ea0003b76c57c1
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Took a look at https://github.com/sciencehistory/kithe/commit/d6c1c8654cc74cf3722ee358f5fdeb5853f63ed5

-- the spec explains exactly what you're fixing here.

PR #352, opened 2019-09-11
Check for has_audio_member? with more efficient SQL

jrochkind:
Previously, has_audio_member? was:

1. Fetching all public members in, and then just forgetting them -- our "decorators" would fetch them all in _again_
2. Checking leaf_representative for each member, _without_ having eager-loaded them, causing n+1 query problem

After this change, we use some _really fancy_ ActiveRecord to execute more targetted SQL to get just what has_audio_member? needs.

I am not thrilled with:

* the complexity of the code to do this fancy SQL (put in comments trying to provide a guide)
* That it still needs to be two SQL queries, one to get content types from 'direct' members, and one from indirect leaf_representatives
* That the decorators would have had the info they needed to answer this question but we need it BEFORE we have decorators, so still need to double-fetch it, but no clear way to refactor to avoid it.

BUT this is good enough for now and an improvement.

This actually does NOT seem to be a signficant response time advantage for Ramelli. But just trying to get anything that _might_ be out of the way, to make my perf analysis more clear and rule it out.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -38,9 +38,31 @@ class WorksController < ApplicationController
   helper_method :decorator
 
   def has_audio_member?
-    @work.members.
-      where(published: true).
-      any? { | x| x.leaf_representative&.content_type&.start_with?("audio/") }
+    # some pg JSON operators in our WHERE clause to pull out actually just
+    # what we want.
+    #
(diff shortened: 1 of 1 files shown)


jrochkind:
In my tests, this returns ramelli in like 3.5-4.5, I think my tests are not reliable cause they vary too much, and this is basically no change, but oh well, let's do it anyway, hard to say and shouldn't hurt.

PR #360, opened 2019-09-16
update honeybadger to more recent version

jrochkind:
Which can now support 'breadcrumbs' feature, which we also turn on. 

https://docs.honeybadger.io/lib/ruby/getting-started/breadcrumbs.html


eddierubeiz:
Skimmed breadcrumb docs. Will be interesting to see this in action.

PR #362, opened 2019-09-16
add autocomplete plugin to inputs on every focus

### LINKED ISSUE #359: Bug: auto-complete only works on first entry in field
Bug related to issue #323: on the edit work form, auto-complete only works for the first entry in the field. I encountered this bug in both the creator and subject fields while editing metadata for work thru9n2. Similarly, if you edit metadata on an existing work (I used work krwaibj to test this), auto-complete does not work when selecting "add another creator" or "add another subject."

jrochkind:
only if not already added. But, in contrast to adding on page load, this way we catch suitable input elements added to page after load dynamically -- for instance in response to a 'add another link.

Closes #359

### DIFF
--- a/app/javascript/src/js/admin/qa_autocomplete.js
+++ b/app/javascript/src/js/admin/qa_autocomplete.js
@@ -46,15 +46,11 @@ import 'jquery'
 // https://github.com/devbridge/jQuery-Autocomplete
 import 'devbridge-autocomplete';
 
-// Uses jquery, assumes it exists in `window` for now.
+var errorContainer =  $('<div class="autocomplete-no-suggestion"></div>')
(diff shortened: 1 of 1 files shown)


app/javascript/src/js/admin/qa_autocomplete.js
```diff
@@ -112,5 +108,20 @@ jQuery( document ).ready(function() {
     // WANT to autocomplete on focus, plus it's broken with 'disabled' state,
     // that keeps us from working around re-suggest on select
     input_el.autocomplete().onFocus = function() {};
```
eddierubeiz: Could use `jQuery.noop` here if you wanted. https://api.jquery.com/jQuery.noop/

eddierubeiz: But whatever. Merging now - looks good to me.

PR #363, opened 2019-09-16
Sort certain fields on display

### LINKED ISSUE #330: Alphabetically sort values on display (for subject and/or other fields)
Could we consider adding drag-and-drop reordering to the subject fields, or possibly all the multiple-value fields?

By way of justification, the following scenario: a curator enters four subjects to describe a work. The order is meaningful: the most important subject needs to be first, and so on. Once these are entered, if the curator changes his/her mind re: the relative importance of the subjects, s/he needs to cut and paste, rather than simply drag-and-drop.

jrochkind:
Closes #330

### DIFF
--- a/app/presenters/attribute_table/row_display.rb
+++ b/app/presenters/attribute_table/row_display.rb
@@ -15,10 +15,12 @@ module AttributeTable
   class RowDisplay < ViewModel
     attr_reader :label_sym, :values, :link_to_facet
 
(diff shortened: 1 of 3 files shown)


eddierubeiz:
Cleverly done!

jrochkind:
Yep, I think it's confirmation of the benefits of this "ViewModel" architecture, a clear place to put the "THESE values we want sorted" logic.

PR #365, opened 2019-09-16
can we show viewer immediately even if it takes a while to load json data, with progress bar

jrochkind:


jrochkind:
oops premature didn't mean to PR this yet

PR #369, opened 2019-09-17
Check fixity for one asset

eddierubeiz:
There should really be a front-end test for this -- not sure if it's worth merging before we do that.


jrochkind:
nice job

PR #370, opened 2019-09-17
keep old query URLs working on new app

### LINKED ISSUE #135: Facets in URLs from old app to new app -- keep them from breaking?
[applies to sort_fields too]

We had intended to keep URLs from old app working as-is in new app. 

But I realized some Solr configuration changes have broken some in new app. Because Blacklight, when applying limits from facets, exposes the Solr field names in browser URLs.

In old app, all our facets were in solr fields ending in `_sim`, which was defined in the solr schema (that came from sufia) as dynamic fields which were `type="string" stored="false" indexed="true" multiValued="true"`. ("sim" => s=string, i=indexed, m=multivalued)

In the new app, I had created a new dynamic field definition in Solr for `*_facet`, defined appropriately for facets. It was multi-valued, and defined in term of a Solr field definition I created `string_dv`, which is equivalent to the old definition _except_ I turn on Solr "docValues", because that is appropriate for best performance in Solr facetting, even though Sufia was not using it. 

So at first I thought I should switch all our facet fields to be `*_sim` fields to match sufia behavior and keep old URLs working. But that would not have docValues, missing out on best Solr configuration. I could redefine "*_sim" to have docValues on, but that would be confusing with the system used to define dynamic field definitions with those suffixes meaning certain things (what if you wanted a string/indexed/multivalued field that did not have docValues on?)

(Blacklight 7 suggests some dynamic field definitions too, which used to be just samvera doing that. And it does use docValues in some places, but not actually for any string fields. But we know docValues on a string field used for facet is appropriate. https://github.com/projectblacklight/blacklight/blob/master/lib/generators/blacklight/templates/solr/conf/schema.xml)

So I dunno, now I'm thinking maybe we need to define some redirects in the new app that just catch the old URLs and do an HTTP 301 ("moved permanently") redirect to a URL using the new "_facet" fields. 

Any thoughts @eddierubeiz ?

jrochkind:
Blacklight query URLs by default include literal solr field names in the URL for facet fields and sort fields. That ties your URLs to your Solr implementaton in a way that makes it hard to keep them working -- especially since Blacklight's solr field naming schema embeds the solr field type in the field name, but best practices for types change.

In the new app, they indeed changed. In this PR, we change BL configuration to keep solr field name out of URLs for sorting. For facets, the field names are still in the URLs, but we have chosen more 'generic' solr field names, like 'subject_facet'. 

And this PR also catches all the old URLs and redirects them with HTTP status 301 moved permanently.

[Cool URIs don't change](https://www.w3.org/Provider/Style/URI.html)

Bonus: Working for searches on a Collection page too (within a collection), by virtue of CollectionShowController inheriting from CatalogController. (Although the first implementation I had of transforming query params didn't work, didn't preserve the :collection_id that is in the URL _path_). 

Closes #135

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -353,7 +353,6 @@ class CatalogController < ApplicationController
     "system_modified_dtsi asc" => "date_modified_asc"
   }
 
-  # TODO date?
   LEGACY_FACET_REDIRECTS = {
     "subject_sim" => "subject_facet",
     "maker_facet_sim" => "creator_facet",
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -353,10 +353,35 @@ class CatalogController < ApplicationController
     "system_modified_dtsi asc" => "date_modified_asc"
   }
 
+  # TODO date?
+  LEGACY_FACET_REDIRECTS = {
+    "subject_sim" => "subject_facet",
+    "maker_facet_sim" => "creator_facet",
+    "genre_string_sim" => "genre_facet",
+    "resource_type_sim" => "format_facet",
+    "medium_sim" => "medium_facet",
+    "place_facet_sim" => "place_facet",
+    "language_sim" => "language_facet",
+    "rights_sim" => "rights_facet",
+    "division_sim" => "department_facet",
+    "exhibition_sim" => "exhibition_facet"
+  }
+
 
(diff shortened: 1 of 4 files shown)


jrochkind:
Our date range facet uses solr field `year_facet_isim`, and that _is_ exposed in the URL... but it hasn't changed between old app and new app at present, so we won't worry about changing it for now.

PR #376, opened 2019-09-18
slightly less confusing to put AR callback methods in their own module

jrochkind:
Pretty minor refactor, but I kept finding it a bit confusing what was going on, and I wrote the thing. Maybe this will be a little bit less confusing.


eddierubeiz:
Let's actually go over this one together. Happy to merge it, but not super familiar with AR callbacks to begin with. I think I get the general idea though.

PR #377, opened 2019-09-19
Fixity report

eddierubeiz:
Fixity report, with tests.
Only hits the DB 3 times.
Loads in about a second; we could optimize it a lot but fine for a page that's seldom looked at.


app/helpers/fixity_report_helper.rb
```diff
@@ -0,0 +1,172 @@
...
+module FixityReportHelper
+  class FixityReportHelper
+    # All assets, including collection thumbnail assets.
+    def asset_count
+      @asset_count ||= asset_stats.count
+    end
+
+    # All assets with stored files.
+    def stored_files
+      @stored_files ||= count_asset_stats ({stored_file: true})
```
eddierubeiz: remove space before parens

app/helpers/fixity_report_helper.rb
```diff
@@ -0,0 +1,172 @@
+# This class digs up a bunch of stats about fixity
+# (ensuring files haven't changed since we last looked at them)
+# and pulls them into a report hash. This is then displayed in
+# fixity_report.html.erb.
+# The class is used in controllers/admin/assets_controller.rb .
+module FixityReportHelper
```
eddierubeiz: Move this to models.

jrochkind:
@jrochkind will:

1. Try to do real SQL cause bad_assets cause why not
2. Create a constant to be used for our fixity period, and use the same one here, so 'stale' always matches our actual fiixty period

jrochkind:
On _production_, report says:

* No assets currently failing fixity
* No fully ingested assets without fresh fixity check

**BUT** 4 assets older than a day that don't have attached files!

That is unexpected, so will investigate further... although I guess taht's really a separate ticket. But we don't expect production in that state -- it could be an import problem, *or* it could just mean there were some files in that state in chf_sufia!

jrochkind:
@eddierubeiz okay, after doing a lot more, I'm done with this now, phew. 

The diff(s) might be confusing, you might want to check out the branch and look at it to review.

eddierubeiz:
This is great. Not confusing at all--I've stared at it more than enough. Great work on the oldest most recent check. Merging!

Issue #367, opened 2019-09-17
Add `cd` to rails app to .bash_profile for dev accounts


jrochkind:
While we would like to never manually ssh to an EC2 machine, we still do it sometimes. 

Almost always the first thing me and @eddierubeiz (and maybe dan too?) have to do is `cd /opt/scihist_digicoll/current`. 

I've added that to my `~/.bash_profile` on staging jobs -- but of course, that will only last until the next time the machine is (re)built. 

Would it make sense to have ansible create a .bash_profile in the individual accounts it creates that just has `cd /opt/scihist_digicoll/current` in it? 

Not a high priority item.

jrochkind:
Not do for now

PR #378, opened 2019-09-19
"capistrano tasks for ssh to a jobs server

jrochkind:
$ cap staging jobs_ssh

Opens up an ssh session on a jobs server (an arbitrary one if there are more than one, in our current setup. Technically the "first" one listed, but that's arbitrary. Or one listed to cap with "primary: true" but we don't do that. `cd`s into app `current` dir.

    $ cap production jobs_ssh:console

Opens up an ssh session which has you cd'd into app current dir AND drop you right into `rails console`. (Exiting rails console will immediately exit your ssh session). Also outputs the hostname to terminal before opening up rails console, since you won't get the hostname in the shell prompt, to try and avoid confusion on wrong host staging/production.

This is convenient with our "auto-discover" EC2 hostnames that have only numeric IPs, cause it avoids you having to look up "Wait, what's the IP of a current jobs server" to ssh there.

In an ideal world, we wouldn't manually ssh to deploy servers much/at all, but we still do plenty at present so don't need to pretend we don't. :)

Both of these tasks ssh in as `digcol` user (ie, the ordinary capistrano deploy user). Rather than using our personal accounts. I think that's actually fine? But if we wanted to, we could make it take an arg (actually an ENV variable) for specifying user, and or automatically trying the same user as your local `username`.

Curious to have @sanfordd's review of this approach.

Implementation taken from hints at https://stuff-things.net/2017/03/22/capistrano-ssh/

May make #367 irrelevant, if we mostly end up using this instead of manual ssh.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -86,6 +86,31 @@ desc "list auto-disocvered EC2 servers"
 task :list_ec2_servers
 
 
+# https://stuff-things.net/2017/03/22/capistrano-ssh/
+desc "ssh to a jobs server"
+task :jobs_ssh do
+  on primary(:jobs) do |host|
+    command = "cd #{fetch(:deploy_to)}/current && exec $SHELL -l"
+    ssh_command = "ssh -l #{host.user} #{host.hostname} -p #{host.port || 22} -t '#{command}'"
+    puts ssh_command # if fetch(:log_level) == :debug
+    exec ssh_command
+  end
+end
+
+namespace :jobs_ssh do
+  desc "ssh to a jobs server and open up a rails console"
+  task :console do
+    on primary(:jobs) do |host|
+      command = "hostname && echo && cd #{fetch(:deploy_to)}/current && bundle exec rails console -e #{fetch(:rails_env)}"
+      ssh_command = "ssh -l #{host.user} #{host.hostname} -p #{host.port || 22} -t '#{command}'"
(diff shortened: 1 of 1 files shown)


jrochkind:
Actually maybe I should make the task `ssh:jobs`, to leave open the possibiilty of later adding other `ssh:$ROLE` tasks for other roles? Thoughts?

jrochkind:
@sanfordd have you had a chance to see what we're doing, does this make sense?

sanfordd:
I did a quick test of it and looked over the code and it seems like a nice tool. Leaving the room open for other servers would make it easier in the long run to get into solr, though I don't know how much the capistrano user can do there.

jrochkind:
Thanks @sanfordd ! We'll prob merge then and see how it goes. 

the other thing to note is that if/when we're using this, we're all going to be logging in as `digcol` instead of our personal accounts. Which might not be ideal? We can write more code to let you specify what account to log into and/or try to log in with the same account name as on your local machine, but if you don't think it's a disaster, we'll not worry about it for now.

sanfordd:
It's not perfect if we need to track a problem down but with 3 people I think it's a fairly minor issue. There's not that many people to check with if you're trying to figure out who ran something.

jrochkind:
If it becomes confusing, then I'll write the code to insist you use your personal account. :)

PR #379, opened 2019-09-19
Tweaks to individual-asset fixity check report

jrochkind:
1. Clean up display a bit, hopefully to make it more readable. 

Before:

Screenshot 2019-09-19 11 42 23

After:

Screenshot 2019-09-19 16 34 50

(Fixity check history is now in ASC order, among other more obvious graphic design changes)

2. The "checked file URI" in a "currently failing details" read out. is now a _link_ to the object in AWS S3 Console for further debugging. A bit hacky and subject to breakage if Amazon changes their URLs, but handy enough to give it a try. 

I'm still not totally sure about the `checked_file_uri` thing -- what if it _doesn't_ match the presently expected file URI? That could happen in a historical FixityCheck if the file has changed... or a current one if it also changed after FixityCheck was done?  

I guess maybe we should at least warn people if they don't match? Maybe I'll go add that.

### DIFF
--- a/app/controllers/admin/assets_controller.rb
+++ b/app/controllers/admin/assets_controller.rb
@@ -3,9 +3,9 @@ class Admin::AssetsController < AdminController
   def show
     @asset = Asset.find_by_friendlier_id!(params[:id])
     if @asset.stored?
-      @checks = @asset.fixity_checks.order('created_at desc')
-      @latest_check   = @checks.first
-      @earliest_check = @checks.last
+      @checks = @asset.fixity_checks.order('created_at asc')
+      @latest_check   = @checks.last
+      @earliest_check = @checks.first
     end
   end
 
--- a/app/models/fixity_check.rb
+++ b/app/models/fixity_check.rb
@@ -38,13 +38,15 @@ class FixityCheck < ApplicationRecord
   private
 
   # returns false if we do not think checked_uri looks like S3.
-  # (If we later use a custon CNAME for S3 buckets, may have to adjust this!)
+  # (NOTE If we later use a custon CNAME for S3 buckets, may have to adjust this, it assumes
+  # it can recognize S3 as *.s3.amazonaws.com)
   #
   # returns a hash with keys :bucket, :key_path, :end_path
   # Used for #checked_uri_in_s3_console to link admins directly to a AWS S3 console.
-  #
   def checked_uri_on_s3_components
     parsed = URI.parse(checked_uri)
+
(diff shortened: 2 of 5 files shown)


app/controllers/admin/assets_controller.rb
```diff
@@ -3,9 +3,9 @@ class Admin::AssetsController < AdminController
   def show
     @asset = Asset.find_by_friendlier_id!(params[:id])
     if @asset.stored?
-      @checks = @asset.fixity_checks.order('created_at desc')
-      @latest_check   = @checks.first
-      @earliest_check = @checks.last
+      @checks = @asset.fixity_checks.order('created_at asc')
+      @latest_check   = @checks.last
+      @earliest_check = @checks.first
```
eddierubeiz: Fine by me!

app/models/fixity_check.rb
```diff
@@ -20,4 +20,50 @@ def passed?
...
+    {
+      bucket: host_parts.first,
+      key_path: key_path,
+      end_key: end_key
+    }
+  rescue URI::InvalidURIError
+    false
+  end
+
+end
```
eddierubeiz: I think this could be handy, yeah. One less minute of panic is worth a lot in the case of an actual fixity check failure.

app/views/admin/assets/_fixity.html.erb
```diff
@@ -51,28 +34,44 @@
...
+    </dl>
+  <% end %>
+
+  <%# Check the asset now: %>
+  <% if @asset.stored? %>
+    <div class="row col-sm-12">
+      <%= link_to "Schedule a check now", admin_check_fixity_path(@asset.friendlier_id), method: "post", class: "btn btn-primary mt-3 mb-3"%>
+    </div>
+  <% end %>
+</dd>
```
eddierubeiz: Looks a lot better. I used the handy Boostrap css classes in the all-asset fixity report too btw - thanks for the tip.

jrochkind:
@jrochkind will put something relevant to expected checksum in the History list, so you can see if expected changed at any point in history.

jrochkind:
Realized putting _actual_ checksum in the fixity History list is better -- if it's success, it will match expected of course; whether it matches _current_ depends on if the file was changed. 

Actual tells you what more of what you need to know about all of it, including confidence that the "passing" history really did calculate a checksum matching current value. 

May also put some colors in depending on if it matches current value. So "passing" but NOT matching current value would mean -- it changed!

PR #384, opened 2019-09-24
Upload DZI files to S3 using concurrent threads

jrochkind:
Using API from ruby-concurrent, which is already a rails dependency. http://ruby-concurrency.github.io/concurrent-ruby/master/file.promises.out.html

We use a custom thread pool to limit concurrency to 8 threads, to avoid accidentally using lots of resources or whatever. Experimentation showed more than 8 doesn't give much advantage.

Since there can be a LOT of tiles, using 8 concurrent threads to upload was observed to result in major performance improvement, as opposed to waiting for each to upload one by one serially. This works on MRI even with the GIL, becuase uploading is mostly waiting on IO, and MRI threads can work concurrently when busy on IO. The limit is probably when the network to S3 is saturated, and/or local disk IO is saturated.

## Simple non-accurate benchmark
(just ran a `create_dzi_for` task once for one file, just rough check not reliable profiling)

Before this PR, full processing of a ~100MB file DZI, including uploading to s3:

* On my macbook, ~33s
* on staging jobs EC2: 1m17s (!)

After this PR:

Around 14s (!) on both my macbook and staging jobs EC2. Can't explain why the times differ so much pre-PR, but are abourt the same post-PR. 

@eddierubeiz after merging this, if our bulk DZI creation isn't complete, it's probably worth cancelling and starting again with this merged in, should make it go quite a bit faster.

### DIFF
--- a/app/models/dzi_files.rb
+++ b/app/models/dzi_files.rb
@@ -113,17 +113,34 @@ class DziFiles
     end
   end
 
+
+
   # Just takes all files in tmp_output_dir passed in, and uploads them at exactly the
   # path they have (with input arg as base) to the remote storage.
+  #
+  # Uses concurrent-ruby to do the uploads concurrently, which speeds things up
+  # significantly, currently with a thread pool of 8 threads.
+  # http://ruby-concurrency.github.io/concurrent-ruby/master/file.promises.out.html
   def upload(tmp_output_dir)
-    Dir.glob("#{tmp_output_dir}/**/*", base: "#{tmp_output_dir}/").each do |path|
-      next if File.directory?(path)
-      upload_path = path.delete_prefix("#{tmp_output_dir}/")
+    thread_pool = Concurrent::FixedThreadPool.new(8, auto_terminate: false)
+    futures = []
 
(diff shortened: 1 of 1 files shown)


app/models/dzi_files.rb
```diff
@@ -113,17 +113,34 @@ def create_and_yield
...
+
   # Just takes all files in tmp_output_dir passed in, and uploads them at exactly the
   # path they have (with input arg as base) to the remote storage.
+  #
+  # Uses concurrent-ruby to do the uploads concurrently, which speeds things up
+  # significantly, currently with a thread pool of 8 threads.
+  # http://ruby-concurrency.github.io/concurrent-ruby/master/file.promises.out.html
   def upload(tmp_output_dir)
-    Dir.glob("#{tmp_output_dir}/**/*", base: "#{tmp_output_dir}/").each do |path|
+    thread_pool = Concurrent::FixedThreadPool.new(8, auto_terminate: false)
```
eddierubeiz: Awwww, yeah.

eddierubeiz:
Cool! This makes sense to me. Merging.

PR #385, opened 2019-09-24
Last-known-good and first-known-good fixity checks: don't throw them out.

### LINKED ISSUE #382: fixity pruning: ensure keeping the correct ones
We actually want to keep:

1. earliest check
2. most recent N checks (whether passing OR failing)
3. Any failed checks
3a. ALSO the passing checks on either side of any failing checks (and there could be more than one failing check in a row) -- we want to know when it was known good just before the failure, as well as just after, for history, to know the maximum possible time it might have been failing. 

I think it is not doing 3a right now. Not sure about the others. Note we don't exactly want to keep most recent N _passing_ checks, we want ot keep most recent N checks period (plus all failing checks). 

Might be helpful to look at chf_sufia code for doing this, let's see if I can find it... Yeah, looks like `N` is just `1` for the existing chf_sufia code, which is probably fine, we probably don't really need to keep most recent N, just the one most recent, plus all the others. 

Here's the code in chf_sufia, may just be able to copy and paste it here, make sure existing tests pass, add a new test for 3a. 

https://github.com/sciencehistory/chf-sufia/blob/master/patches/fixity_checks/checksum_audit_log.rb#L36-L51

eddierubeiz:
Keep passing fixity checks if they're preceded or followed by a failing check. Why? Because they're useful for guessing the time window when something bad happened to a file, or when a bad file was fixed.

Fixes #382 .

### DIFF
--- a/app/services/fixity_checker.rb
+++ b/app/services/fixity_checker.rb
@@ -44,56 +44,27 @@ class FixityChecker
     )
   end
 
-  #FixityChecker.new(asset).prune_checks
-  # After pruning, for each asset - URI combination, you will be left with:
-  # up to NUMBER_OF_RECENT_PASSED_CHECKS_TO_KEEP recent passed checks,
-  # plus all the failed checks,
-  # plus one initial check (to remember the file that was intially uploaded.)
-  def prune_checks
-    checks_to_delete.map(&:destroy)
-  end
-
-  private
-
-  # Returns an array of checks that we don't want to keep.
-  # The current rules are:
+  # FixityChecker.new(asset).prune_checks
   # Never throw out FAILED checks
   # Never throw out the earliest PASSED check
   # Always keep a PASSED check if it is preceded or followed by a failing check.
-  # Always keep N recent PASSED checks
-  def checks_to_delete
-    checks = FixityCheck.checks_for(@asset, permanent_url)
-    return [] if checks.empty?
-
-    # Throw all the passed checks INTO the trash.
-    trash = checks.select { | ch | ch.passed? }
-
-    # But then, pop the earliest passed check back OUT of the trash.
-    earliest_passed_check = trash.pop
-
-    # Shift the most recent N passed checks back OUT of the trash.
-    recent_passed_checks = trash.shift(NUMBER_OF_RECENT_PASSED_CHECKS_TO_KEEP)
-
-    # One more thing:
-    # Certain passing checks are important for
-    # guessing the time window when something bad happened
-    # to a file, or when a bad file was fixed:
-    checks.each_with_index do | ch, i |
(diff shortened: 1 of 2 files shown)


app/services/fixity_checker.rb
```diff
@@ -59,16 +59,39 @@ def prune_checks
...
+      next unless trash.include?(ch)
+      # OK, we're looking at a passed check we were
+      # about to delete.
+      if checks[i+1]&.failed? || checks[i-1]&.failed?
+        # This check passed, BUT it's either
+        # preceded or followed by a failing check.
+        # Let's remove it from the trash, so it
+        # won't get deleted.
+        trash.delete(ch)
+      end
```
jrochkind: the implementation already in chf_sufia that doesn't rely on a 'trash' seems simpler to me. It just goes through the ordered log entries one by one, and decides whether to delete each one. What if I try putting that imp in here and see if specs still pass?

eddierubeiz: Yeah, I did take a look. Only difference is NUMBER_OF_RECENT_PASSED_CHECKS_TO_KEEP is hardcoded in sufia, I believe.

jrochkind: Hm, what  I meant is that the sufia logic doesn't use a `trash` array, does it? It has less steps, it doesn't put everything in a `trash` array first, it just goes through the things and deletes them immediately.

eddierubeiz: I'm happy to do this, btw -- let me know. It shouldn't take long.

eddierubeiz: By "only difference" I mean the only difference to pay attention to when porting the code.

jrochkind: Yeah, really all I mean is "get rid of the temporary `trash` array" and do it more like chf_sufia in that regard. So if you wanna do that. 

but you can also say "Dude, it doesn't matter, this is micro-critiquing the code in a way that just doesn't matter", and I'll take that too.

eddierubeiz: Nah, the Sufia way is simpler and better. Will refactor tomorrow; should be easy.

eddierubeiz:
Adapted the code from https://github.com/sciencehistory/chf-sufia/blob/master/patches/fixity_checks/checksum_audit_log.rb#L36-L51 
As of this writing, the rules are:
 - Never throw out FAILED checks
 - Never throw out the earliest PASSED check (note: not in the Sufia code)
 - Always keep a PASSED check if it is preceded or followed by a FAILED check.
 - Always keep N most recent checks, whether PASSED or FAILED.

app/services/fixity_checker.rb
eddierubeiz: Didn't get to it Friday, but done now.

PR #386, opened 2019-09-25
We can delete a prefix from S3 using just batch_delete!

jrochkind:
Turns out we were re-inventing a method already built into S3 sdk API. Let's just use it.


eddierubeiz:
Awesome.

PR #387, opened 2019-09-25
WIP: Notify via email and honeybadger if a fixity check fails.

eddierubeiz:
This is just waiting on:
    https://github.com/sciencehistory/scihist_digicoll/issues/394 . After that is taken care of it will be OK to merge / push this.

Ref #232

### DIFF
--- a/app/services/fixity_check_failure_service.rb
+++ b/app/services/fixity_check_failure_service.rb
@@ -1,0 +1,75 @@
+class FixityCheckFailureService
+  include ActionView::Helpers::UrlHelper
+
+  attr_reader :fixity_check
+
+  def initialize(fixity_check)
+    raise ArgumentError.new("Fixity check passed was nil.") if fixity_check.nil?
+    raise ArgumentError.new("Please pass in a FixityCheck.") unless fixity_check.is_a? FixityCheck
+    @fixity_check = fixity_check
+    @asset = fixity_check.asset
+  end
+
+  def send
+    ActionMailer::Base.mail(from: "no-reply@sciencehistory.org",
+                            to: "digital-tech@sciencehistory.org",
+                            subject: subject,
+                            content_type: "text/html",
+                            body: message).deliver_now
+    if defined? Honeybadger
+      Honeybadger.notify("Fixity check failure", context: {
+        asset: @asset.inspect,
+        asset_url: asset_url,
+        fixity_check: @fixity_check.inspect
+        })
+    end
+  end
+
+  def subject
+    "FIXITY CHECK FAILURE: #{ScihistDigicoll::Env.lookup!(:app_url_base)}, \"#{@asset.title}\" (asset #{@asset.friendlier_id})"
+  end
+
+  def message
+    <<-EOF
+    <p>A fixity check for asset #{@asset.friendlier_id} on #{hostname} just failed at #{ date }.</p>
+    <dt>Asset:</dt><dd>#{asset_message}</dd>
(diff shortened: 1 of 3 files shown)


app/services/fixity_check_failure_service.rb
```diff
@@ -0,0 +1,75 @@
...
+  def initialize(fixity_check)
+    raise ArgumentError.new("Fixity check passed was nil.") if fixity_check.nil?
+    raise ArgumentError.new("Please pass in a FixityCheck.") unless fixity_check.is_a? FixityCheck
+    @fixity_check = fixity_check
+    @asset = fixity_check.asset
+  end
+
+  def send
+    ActionMailer::Base.mail(from: "no-reply@sciencehistory.org",
+                            to: "digital-tech@sciencehistory.org",
```
jrochkind: Ideally email addressees woudln't be hard-coded in, but would be from local_env.yml somewhere. Not sure if we already have keys for these though, not that big a deal to hard-code them in.

jrochkind:
I'm not totally sure why I didn't in chf_sufia do things the 'normal' way with an ActionMailer sub-class and and template file. Instead putting the 'template' inline in a string. 

Maybe cause I thought I was gonna reuse the plain text template for more than one delivery, not just email? Or cause it seemed so simple there was no need to put it in a separate template file?

Here it is non-trivial and is only used for email... it kind of seems like it should be a normal ActionMailer sub-class. but I dunno, maybe it's good enough like this, not worth fixing? Could run into unexpected problems doing it the 'normal' way, and it is working now, so. What do you think?

eddierubeiz:
I like the idea of standardizing this so it's more in line with Rails best practices. But I'd probably vote for making that refactor a GitHub issue and putting that in our post-launch to-do list.

eddierubeiz:
I put the refactor in a separate ticket so we can merge this PR after I fix the hard-coded emails.

app/services/fixity_check_failure_service.rb
eddierubeiz: Fixed!

jrochkind:
Eddie writes in Slack:

> Eddie Rubeiz 3:00 PM
> Added the email addresses for fixity check email failures to the staging branch version of the local_env.yml template in Ansible

> https://bitbucket.org/ChemicalHeritageFoundation/ansible-inventory/commits/8034aa85e7cdaa17eea7a8da7cd88a56f1b458e7/raw

That seems to be a link to a specific commit, not a PR, in bitbucket. Not sure if bitbucket uses PR's or what. 

Is it okay to merge this PR #387 before that ansible change? Or will deploying this code without the ansible change result in errors?

I am also not sure where we are with the situation where emails do not actually arrive. @eddierubeiz  could you update on that here?  We can still merge the code even if emails never arrive, but we can't close the issue without either getting stakeholder sign-off on email alerts not working, or making them work. Can you keep track of that and create necessary issues and such to track and bring them up with stakeholders, @eddierubeiz ?

eddierubeiz:
Good call! Updated the ticket with the two preconditions for merge above. Basically the jobs server should be able to send emails, and the to addresses (no longer hardcoded) need to be added by Ansible to the `local_env.rb` settings file.

jrochkind:
@eddierubeiz will move error emails to a bg job (ie use `deliver_later`), and then we can safely deploy this even though email sending currently raises an error. (Pending #394)

eddierubeiz:
Now using `deliver_later`; @jrochkind, you may merge at your convenience.

PR #390, opened 2019-10-01
Check files, assets, works and files against Fedora.

eddierubeiz:
Just a proof of concept for now.
Note from Jonathan: asset.file.exists guarantees that a file exists in s3 without actually downloading it. Along with a recent fixity check, this is great for checking originals against Fedora without actually downloading them from Fedora.


app/services/fedora_checker.rb
```diff
@@ -0,0 +1,130 @@
+require "byebug"
```
jrochkind: remove

app/services/fedora_checker.rb
```diff
@@ -0,0 +1,130 @@
...
+
+class FedoraChecker
+  def initialize(options:)
+    @options = options
+    File.open(options[:metadata_path], 'r') do |f|
+      @data = Yajl::Parser.new.parse(f)
```
jrochkind: Let's try just `JSON.parse` without the extra dependency, since we aren't using streaming. If it gives you any trouble don't worry about it just use yajl, the Gemfile comment could be even more explicit that remove this when we're done with fedora.

app/services/fedora_item_checker.rb
```diff
@@ -0,0 +1,515 @@
...
+    new_vals = @collection.contains.
+      pluck(:friendlier_id).sort
+    confirm(compare(old_val, new_val), "collection members", old_val, new_val)
+  end
+
+
+
+  def check_generic_work
+    @work = @local_item
+    check_scalar_attributes
```
eddierubeiz: relabel these methods to specify that they're just for works.

app/services/fedora_item_checker.rb
```diff
@@ -0,0 +1,515 @@
...
+  # Nil-tolerant, as it uses present? to determine
+  # whether something exists.
+  # Handles both scalars and arrays.
+  # In the case of arrays,
+  # pass in a mapping method if you want,
+  # and it will be applied to a before comparing.
+  def compare(a, b, the_method=nil, order_matters:true)
+
+    # In source, not destination:
+    return false if (!a.present? &&  b.present?)
```
eddierubeiz: Just convert to null before comparing.

app/services/fedora_item_checker.rb
```diff
@@ -0,0 +1,515 @@
...
+  end
+
+  # Tests for equivalency between a and b.
+  # Nil-tolerant, as it uses present? to determine
+  # whether something exists.
+  # Handles both scalars and arrays.
+  # In the case of arrays,
+  # pass in a mapping method if you want,
+  # and it will be applied to a before comparing.
+  def compare(a, b, the_method=nil, order_matters:true)
```
eddierubeiz: comparison_method

jrochkind:
Put all the files in namespace `FedoraChecker` in `./app/services/fedora_checker`, so we keep it isolated.

jrochkind:
This code is a little bit hard to read and maintain, but since we only plan to execute it 3-5 times in the next two weeks, then abandon fedora and this code forever, it's okay.

app/services/fedora_checker.rb
eddierubeiz: No evidence that yajl is faster than plain old JSON.parse. Getting rid of yajl.

PR #402, opened 2019-10-04
Access control has to be set to true or false

eddierubeiz:
Nulls are no longer allowed in the DB.
This sets items' `published` flag explicitly to either true or false.

### DIFF
--- a/app/importers/importers/importer.rb
+++ b/app/importers/importers/importer.rb
@@ -176,6 +176,8 @@ module Importers
         add_error("missing 'access_control'")
       elsif metadata["access_control"] == "public"
         target_item.published = true
+      else
+        target_item.published = false
       end
     end
 


jrochkind:
Hmm, AR should know that the default is false, it should not be necessary to explicitly set it to false. 

Did you run into a case where it was necessary, you were running into an exception or something? Can you describe that case (or... even make a test case?)

eddierubeiz:
Yes, the latest export from Sufia generated a few files that did not have an explicit `access_control` key in their metadata. This resulted in the import trying to save an `asset` with a `published` value of `nil` (rather than either `true` or `false`).

jrochkind:
Ah, I see, if you explicitly try to set it to `nil`, then teh default won't matter. 

I missed on first read this change was to importer code. Merging.

Issue #401, opened 2019-10-03
Deleting an Asset that doesn't have an MD5 (say because it has no attached file at all), causes error


jrochkind:
## Backtrace

    line 37 of [PROJECT_ROOT]/app/models/dzi_files.rb: initialize
    line 72 of [PROJECT_ROOT]/app/models/asset.rb: new
    line 72 of [PROJECT_ROOT]/app/models/asset.rb: dzi_file

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/55595707)

PR #406, opened 2019-10-07
Make sure DZI lifecycle doesn't get in the way of dealing with Assets without attached files

jrochkind:
Our app doesn't mean to allow for an asset without an attached file, but it can happen in odd edge cases. The DZI lifecycle was getting confused not taking accoun this could happen, and raising trying to delete DZI's when there was no attached file so no md5 so couldn't figure out where associated DZI's might be. Ref #401


eddierubeiz:
OK -- this makes sense. Thanks for fixing.

PR #409, opened 2019-10-08
Detect and delete orphaned files on S3

jrochkind:
A bunch of rake tasks. "report" and "delete" variants of "originals", "derivatives", and "dzis". 

```
./bin/rake scihist:orphans:report:originals
./bin/rake scihist:orphans:delete:originals

./bin/rake scihist:orphans:report:derivatives
./bin/rake scihist:orphans:delete:derivatives

./bin/rake scihist:orphans:report:dzi
./bin/rake scihist:orphans:delete:dzi
```

The "report" variants will give you a report of any original/derivative/dzi sets that do not appear to be referenced/corresponding to an Asset record in the postgres database. And print out a human-readable log of any such found. The "report" variants are basically dry runs for the "delete" variants. 

The "delete" variants still check all files (so still take as long to run), but instead of just reporting details on "orphaned" files found, will delete them (still printing out some basic summary info of what was done). 

Both variants will use a progress_bar unless you set `PROGRESS_BAR=false` shell env. 

---

## technical notes

I had a LOT more trouble with this than expected, not sure why. It was hard to get the code readable/maintainable, and it's still a bit obtuse, but the best I could do. 

* One interesting thing is that I _stepped back_ from trying to have a common superclass for the various orphan-finders. Fairly early on. There was just too much that differed, it wasn't making the code easier to understand or change. So I ended up with just some limited shared functionality in the `S3PathIterator` class. There is some duplication (non-DRY) between orphan-finder classes, oh well. 
* I did a "monkey patch" to `shrine` to add `delete_prefixed` functionality. This will show up in a future version of shrine, (https://github.com/shrinerb/shrine/pull/413), I _think_ 3.0 but maybe not until 3.1?  Used `SanePatch` gem to warn us if we upgrade to shrine 3.0 to check to see if patch is still necessary. 
  * This patch also allowed to greatly simplify some hacky code I had in the `DziFiles` class to try to do this in a much more complicated way -- now it just uses `delete_prefixed`. 
* The file layout we are using for DZI files requires more time (and AWS cost) to look for "orphaned" DZI than if we were using the same layout as on chf_sufia, with one level less of "directories" in the key path. But it's down to ~12 minutes and ~20 cents per orphan check, and rearranging the DZI's is hard, so this is good enough for now. (Maybe should make a different issue to rearrange the DZI's later?)
  * Note the DZI checking code will only check the main ".dzi" file "manifests" -- if it finds an "orphaned" one, it will delete all it's corresponding tiles. But it won't check random files that don't go with any ".dzi" file in the expected way, so it can still leave behind unreferenced files, if some like that show up. It would be more complicated/expensive/slow to check every possible file, when there are _so many_ DZI tiles expected.


eddierubeiz:
Thanks for the comments and the notes above. Makes sense to me.

PR #412, opened 2019-10-14
Add 'production' or 'staging' note to admin sidebar menu

jrochkind:
"service_level" is what we call this as set in local_env.yml and in ansible.

If it's set, "staging" or "production" will show up in admin sidebar, to help keep them clear, since otherwise they look identical but for browser location bar. "staging" will be with a highlighted yellow caution background, per stakeholder feedback.

If `service_level` is not set at all, this line won't show up at all. In development envs, it is currently not set at all (is nil).  for better or worse, didn't htink too hard about that case.


eddierubeiz:
This will be very helpful. Thanks.

Issue #395, opened 2019-10-02
update FAQ, to not promise machine-readable access formats we don't have yet


jrochkind:
But more might be coming

Issue #201, opened 2019-07-02
consider "API" use


jrochkind:
Our current [FAQ](https://digital.sciencehistory.org/faq) says:

> Can I freely download or extract your metadata?

>We strive to make our linked open data freely available, and it can currently be accessed in JSON and RDF. Please note that our data are formatted for internal use and have not been cleaned up for distribution. In addition, our data model and API are currently undocumented. If you wish to use the data in their current form, you can get the RDF data via Turtle by adding “.ttl” to the end of a record’s URL. You can get the JSON data by adding “.json” to the end of a record’s URL. You can do the same for search results: just make sure to add “.ttl” or “.json” to the end of the URL before the query string. In general the RDF results give more complete data: more metadata fields are returned with search results, and better item membership information is returned with record results.

I'm not totally sure this is actually true even of present chf_sufia at present -- it might be, but I haven't verified lately, and don't know what metadata might be missing from those serializations, or if they still work. 

We are also unaware of anyone using any of these as machine-readable interfaces, from our app. 

It won't, exactly as written, be true of our new app. In part, because we are not doing things based on linked data/RDF, so will not have a RDF/Turtle representation available (unless we do other work to make it so, just becuase we want to say we have it). 

In general, one of the things we lose by not using sufia, and highly customizing Blacklight, is some of those 'automatic' machine-readable interfaces. (But again, I'm not totally confident they worked as is). 

However, we can create some machine readable interfaces. 

We will definitely provide an OAI-PMH endpoint (already in our requirements, for DPLA). Which necessitates making an OAI-DC XML serialization. 

We could provide that available as a download for an individual record. We will also have an RIS serialiation available. 

We could/probably should provide a .json serialization (should it include child IDs and representative IDs, or only direct metadata? Should it provide _links_ to URLs for downloading child/representative members? According to some standard way to do this in JSON? (not RDF/LOD))?

We should provide html `link` tags in the work show page, for discovery of alternate formats available (RIS, oai_dc+xml, json, etc). 

We could provide an Atom and/or RSS feed of search results. It could be fancy and support asking for any of our available formats (json, RIS, oaidc+xml etc), and/or provide link tags to advertise them. 

(Blacklight theoretically provides Atom/RSS feeds, but unsure how useful they are, and they are likely not *currently* working in our app. See #190 ). 

All of those things would be *reasonably* easy to do (easier than RDF probably), and provide pretty good machine-readable access. 

[...]

PR #422, opened 2019-10-17
revise FAQ on extract/download metadata to be accurate

jrochkind:
Ref #395

There is an FAQ "Can I freely download or extract your metadata?"

Previously it made reference to "linked open data" -- we no longer traffic in "linked data", so I changed it to just "open data". 

Previously it referenced metadata formats and access methods that are not aviailable in new app, including both RDF and other. I'm not totally sure if the previous FAQ was 100% accurate even for chf_sufia. However, it is true that we currently offer much less access to extracted/downloadable metadata than previously. We have a ticket to consider adding more at #201. 

But in the meantime, fix the FAQ so it is accurate when we go live. 

Previous text:

> We strive to make our linked open data freely available, and it can currently be accessed in JSON and RDF. Please note that our data are formatted for internal use and have not been cleaned up for distribution. In addition, our data model and API are currently undocumented. If you wish to use the data in their current form, you can get the RDF data via Turtle by adding “.ttl” to the end of a record’s URL. You can get the JSON data by adding “.json” to the end of a record’s URL. You can do the same for search results: just make sure to add “.ttl” or “.json” to the end of the URL before the query string. In general the RDF results give more complete data: more metadata fields are returned with search results, and better item membership information is returned with record results.

New:

> We strive to make our open data freely available, but the options we provide for machine-readable metadata access are currently limited. If you have a project that could benefit from more convenient machine-accessible APIs for metadata access, please get in touch to share your use case.

> Currently our metadata can be accessed in an XML format designed for OAI-PMH access which includes Dublin Core as well as a few additional fields, but does not include all internal metadata or any membership/association information.

> * You can bulk harvest via an OAI-PMH 2.0 endpoint at http://localhost:3000/oai
> * You can get an oai-dc XML representation for any record by adding `.xml` to the end of a record's URL. For instance, http://localhost:3000/works/vt150j62m.xml.

(Both with some internal hyperlinks not shown here).

### DIFF
--- a/app/views/static/faq.html.erb
+++ b/app/views/static/faq.html.erb
@@ -32,7 +32,7 @@
     <dt>Can I freely download or extract your metadata?</dt>
     <dd><p>We strive to make our open data freely available, but the options we provide for machine-readable metadata access are currently limited. If you  have a project that could benefit from more convenient machine-accessible APIs for metadata access, please <%= link_to "get in touch", "mailto:#{ ScihistDigicoll::Env.lookup!(:admin_email)}" %> to share your use case.
 
-        <p>Currently our metadata can be accessed in an XML format designed for <a href="https://www.openarchives.org/pmh/">OAI-PMH</a> access which includes Dublin Core as well as a few additional fields, but does not include all internal metadata or any membership/association information.
+        <p>Currently our metadata can be accessed in an XML format designed for <a href="https://www.openarchives.org/pmh/">OAI-PMH</a> access which includes Dublin Core as well as a few additional fields, but does not include all internal, administrative, and relational metadata.
         <ul>
             <li>You can bulk harvest via an OAI-PMH 2.0 endpoint at <code><%= oai_provider_url %></code></li>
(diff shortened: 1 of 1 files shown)


jrochkind:
@HKativa Please sign-off on this one too, or feel free to suggest edits or concerns, thanks!

HKativa:
@jrochkind overall this looks good! Though I will confess, not sure I understand what this refers to: "...but does not include all internal metadata or any membership/association information." Specifically, I'm drawing a blank on "membership/association information."

jrochkind:
@HKativa Thanks for reviewing!

Right now, the machine-readable metadata we offer is OAI-DC of works. There's no way to see anything about 'children' of a work (constituent assets or child works), there's no way to see if a given work has a parent (_is_ a child work), there's no way to see collection membership. 

Lack of exposed info about parent/children and collection membership is what I meant by "membership/association information" perhaps I should be more explicit/less abstract... not sure how to do so without taking a paragraph trying to explain our data model, heh.

HKativa:
@jrochkind thanks for the additional info and clarification. That's not where my mind went at all! 

I think it would be sufficient to say "...but does not include all internal or administrative metadata" and leave it at that. The main point is for people to understand that what they're getting isn't comprehensive and that they can get in touch with us for additional info.

PR #424, opened 2019-10-21
Base title of RIS download file on work title, not on representative.

eddierubeiz:
Fixes https://github.com/sciencehistory/scihist_digicoll/issues/423 .

### DIFF
--- a/app/presenters/download_filename_helper.rb
+++ b/app/presenters/download_filename_helper.rb
@@ -52,7 +52,7 @@ class DownloadFilenameHelper
   end
 
   def self.ris_download_name(work)
-    "#{first_three_words(work.title)}.ris"
+    "#{first_three_words(work.title)}_#{work.friendlier_id}.ris"
   end
 
   # Pass in a string, get the first three words separated by underscores, stripping punctuation.
--- a/app/presenters/download_filename_helper.rb
+++ b/app/presenters/download_filename_helper.rb
@@ -49,12 +49,13 @@ class DownloadFilenameHelper
     base = if work.representative.nil?
       first_three_words(work.title)
     else
-      self.filename_base_for_asset(work.representative)
+      self.filename_base_for_asset(work.leaf_representative)
     end
     "#{base}.ris"
   end
 
   def self.filename_base_for_asset(asset)
+    raise ArgumentError, 'Pass in an asset.' unless asset.is_a? Asset
     if asset.content_type && asset.content_type.start_with?("audio/")
       return asset.title
     end
--- a/app/presenters/download_filename_helper.rb
+++ b/app/presenters/download_filename_helper.rb
@@ -43,23 +43,16 @@ class DownloadFilenameHelper
     self.filename_with_suffix(base, content_type: content_type)
   end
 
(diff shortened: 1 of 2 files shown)


jrochkind:
This fixes the bug that causes an uncaught exception... but yeah, the existing logic looks weird to me, I'm not sure I understand what it's trying to do, or what it's doing. (Acknowledging I may have written it the first time, I don't know!)

Why would the RIS download of a work be based on the representative/leaf representative asset title anyway?  I think it should probably always be based on the work name?

eddierubeiz:
Agreed.

jrochkind:
While we still have the old app to compare, let's document (by testing out some real things?) how RIS download file title is chosen in chf_sufia? We probably want to just copy that logic, or even if not want to know what it is to decide not to just copy it. The intent was to copy that logic, so it's basically a bug if it's not doing so currently. 

@eddierubeiz since you were looking for stuff to do, you feel like doing that, documenting what chf_sufia is doing for RIS download titles? (I'd just look at the actual app by use, rather than try to look at the code, myself, although it could be a combo)

eddierubeiz:
Absolutely. Stay tuned. We can use this same PR.

eddierubeiz:
In Sufia, the RIS download name is based on the first three words of the work's title.
See e.g. [Exterior views of Charles Lennig and Company plant in Bridesburg](https://digital.sciencehistory.org/works/rr171z13c) where the downloaded file is `exterior_views_of_rr171z13c.ris` .

jrochkind:
yeah, we should prob just do that here? I think it was a bug/mistake to use representative asset title for RIS downloads, which are always metadata about the work. 

(When you are downloading an asset itself and discussing filename of downloaded asset, somewhat different considerations, including different logic for differnet content types). 

@eddierubeiz do you want to just make that so in this PR?

eddierubeiz:
The rationale, as stated in the comments for the broken function, was: "[Try to keep the ris download filename similar to the one we use for originals: filename_for_asset(asset)](https://github.com/sciencehistory/scihist_digicoll/blob/fbb523d7b86795dcebb3a5b481c149fdeb272d83/app/presenters/download_filename_helper.rb#L46)". That is actually not a good idea, since the citation pertains to the parent work.

app/presenters/download_filename_helper.rb
```diff
@@ -43,24 +43,18 @@ def self.filename_for_asset(asset, derivative: nil)
...
+    raise ArgumentError, 'Pass in an asset.' unless asset.is_a? Asset
     if asset.content_type && asset.content_type.start_with?("audio/")
       return asset.title
     end
     self.filename_base_from_parent(asset)
   end
 
+  def self.ris_download_name(work)
+    "#{first_three_words(work.title)}.ris"
+  end
```
jrochkind: For other downloads, in addition to first three words, we put the friendlier_id in, just as a unique ID/disambiguation in case other things have same first three words. 

Based on your report-back of what chf_sufia does, it looks like chf_sufia does that too? So keep this consistent for now, just porting behavior over, and add friendlier_id too?

PR #436, opened 2019-10-24
Adjustments to .gitignore

eddierubeiz:
Per our discussion in Slack:
`personal_bin` is a great place to put temporary developer scripts that can be used with `watch` to monitor the output of a command in response to changes in the code (a test, a grep or find command, or a simple rake task)
`public` contains a variety of files don't belong in the codebase, but also a few that do.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,4 +38,12 @@ scripts/import/fedora_creds.sh
 # Ignore the aws credential file
 /cap_aws_credentials.yml
 
-/public/sitemap
+public/__sitemaps/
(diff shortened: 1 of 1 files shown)


jrochkind:
Huh, `/public/packs` and `/public/packs-test` are new to me! I don't get those in my `public`, and haven't heard of them before. 

By the name, I'd assume they have to do with `webpacker`, but I don't expect `webpacker` to create them. 

Kind of mysterious, but still doesn't hurt to add them to .gitignore.

PR #438, opened 2019-10-25
Add alternate OAI link to audio template

eddierubeiz:
Adding the link tag for auto-discovery of oai-dc representation to the parallel audio template.
( see https://github.com/sciencehistory/scihist_digicoll/commit/17543a77e640457a055585e22c868faf0d485a00 , which adds it to  `app/views/works/show.html.erb`  .)


jrochkind:
Wow, how'd you even notice this was missing? 

I wonder if we should have tests, but for now good enough.

Issue #432, opened 2019-10-23
Public view needs non-public items for logged in users


jrochkind:
We were trying to avoid this, but it's just inconvenient for various things staff wants to do, and also really a necessary pre-requisite for #171 batch edit needs -- really need to be able to select items for batch editing using the full facetting power of the public UI. 

So. 

- [x] For logged in users only, non-public items should show up in front-end search. They should still not show up for non-logged in users. 
- [x] non public items should be clearly identified as "private" in UI, when they do show up
- [x] logged-in users need a facet/limit for public/non-public #457
- [x] _child items_ that are private need to be shown _only_ for logged-in users (see #449 )...
- [x] ... but also labeled as private (see #489)
- [x] All items need an "Edit" button next to them for staff users (this was intended to be so, but keeps not working, we need to add a test and make it work).

eddierubeiz:
All of the above are ready in production. See e.g. 
https://digital.sciencehistory.org/works/vcs937d

PR #443, opened 2019-10-29
Unpublished items are included in public search results for logged in users.

eddierubeiz:
Logged-users can see unpublished items in the public-facing search results (e.g. at `/catalog`).
Includes tests.
Ref https://github.com/sciencehistory/scihist_digicoll/issues/432.


jrochkind:
Checked out branch, verified non-published things already do have a "private" flag in search results and public item detail page. Good, avoid confusion!

eddierubeiz:
Yeah. You will want this working before finishing work on 457. (Which I'm happy to do too, btw.)

PR #444, opened 2019-10-29
fix skip __sitemap

jrochkind:


jrochkind:
mistake didn't mean to PR yet

Issue #451, opened 2019-10-30
"View in library catalog" link displays twice


HKativa:
Link to the OPAC record for the collection associated with this work displays twice in public view on the new app:  https://kithe.sciencehistory.org/works/x346d4432

HKativa:
Note: I think this is affecting older records, where the OPAC link was manually copied and pasted into the "Related URL" field, before the functionality was added to render the "view in library catalog" link using the Bib number in the External ID field. See also:

https://kithe.sciencehistory.org/works/w66343603 
https://kithe.sciencehistory.org/works/8w32r5659
https://kithe.sciencehistory.org/works/pc289j53n

jrochkind:
Thanks for finding and reporting! We can get that fixed.

HKativa:
Update: this issue has been resolved for all of the records cited in the original issue. However, additional examples were found:

https://kithe.sciencehistory.org/works/t722h939f
https://kithe.sciencehistory.org/works/gm80hv33v

jrochkind:
https://digital.sciencehistory.org/works/fb494841c with four link to catalog links displaying,
Has one BibID entered in metadata, for `b1071974`. 

But has FOUR sierra OPAC  URLs entered as "related urls"
* http://othmerlib.sciencehistory.org/record=b1071974
* http://othmerlib.sciencehistory.org/record=b1043344
* http://othmerlib.sciencehistory.org/record=b1069453
* http://othmerlib.sciencehistory.org/record=b1043343

Only one of which matches the BibID.  What does this mean, and is it intentional?

It looks like the old logic in chf_sufia ignored _any_ othmerlib sierra OPAC URLs, if any bibID was entered. So we could make that so in the new app too... but it is ignoring unique content there, related urls to other OPAC records. 

I didn't know that was a use case we might have, putting multiple different OPAC links in there. How would we _want_ them to show up, if at all?

jrochkind:
https://digital.sciencehistory.org/works/t722h939f has two links to catalog showing up. 

It's metadata has bib number `b1069527`, but related URL `http://othmerlib.sciencehistory.org/record=b1069527~S5` -- that weird extra `~S5` on the end, which we don't really understand what it means to Sierra, it's optional in the URL.  I think staff usually tries to enter a URL *without* that `~S5` on the end, but I guess sometimes they slip through. 

In this case, then the code for de-duplication doesn't realize it matches. I can fix to truncate to first 8 chars of bibID in URL. 

[...]

PR #452, opened 2019-10-30
avoid double display of same link to opac

jrochkind:
When a bib number is entered as an identifier, PLUS a link to opac is included as a related_url (which happens in some of our older data), don't show the same link to opac twice. The solution is just putting a 'uniq' in the right place. With a unit test.

Ref #451

### DIFF
--- a/app/presenters/work_show_info.rb
+++ b/app/presenters/work_show_info.rb
@@ -44,10 +44,11 @@ class WorkShowInfo < ViewModel
         external_id.category == "bib"
       end.map(&:value)
 
-      (bib_ids + related_url_filter.opac_ids).map do |bib_id|
+      (bib_ids + related_url_filter.opac_ids).uniq.map do |bib_id|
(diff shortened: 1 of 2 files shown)


spec/system/digitization_queue_spec.rb
```diff
@@ -42,4 +42,12 @@
     expect(work.extent).to eq(["test-dimensions"])
     expect(work.medium).to eq(["test-materials"])
   end
+
+  context "auto-submit AJAX status form on index" do
+    let!(:queue_item) { create(:digitization_queue_item) }
+
+    it "properly changes value" do
+
```
eddierubeiz: I don't understand why this block is blank.

jrochkind: Cause I ended up committing something I didn't mean to! Ended up not writing a "system" test, but just a unit test instead. Thanks for catch, will fix.

PR #453, opened 2019-10-30
Auto-submitting digitization queue item select menu, in a couple more places

jrochkind:
The Queue List (index) page already had a select menu for item status for each item listed. It was in it's own form (one per item), and used the very poorly documented rails-ujs `date-remote`-triggered JS functions to do an "AJAX" submit, so when you saved the form, it was done via AJAX, didn't create a page reload. 

Previously, the results of the AJAX form submit were pretty much ignored. 

This PR adds some features to the JS. Still building off rails-ujs, which is again very poorly documented, and I had to google to figure out how to do a lot of things. But for now, seems better than rewriting from scratch, it does take care of a lot of things for us. 

* Auto-submit the form on change of select menu, so you don't actually need to press save ==> user-testing revealed this is what people expected, they were highly likely to not know/forget to submit the form as a separate click. 
* Better "in progress" spinner, using a fontawesome spinner icon that replaces a "floppy save" icon, and disables the form while in progress. This uses the `data-disable-with` attribute feature of rails-ujs remote forms. 
* Catch errors -- if the back-end app says there is an error, print out diagnostics to console, warn the user with an ugly JS alert, and revert the select menu back to what it was before the save, since the changed didn't actually get persisted (or at least we don't know if it did). 

Additionally, the HTML code for the auto-submitting-form-with-select was extracted to a "ViewModel". AND then this auto-submitting form was put in a couple other places where observation of staff users led me to think it would be convenient and save clicks. 

I tried to write a capybara JS test to test that all this fancy JS actually works -- but I couldn't get it to work, changing the select menu with Capybara/Chromekit JS somehow does not trigger the JS "change" event.  After spending some non-trivial time on it, I gave up, sorry no tests. 🤷‍♂

### DIFF
--- a/app/javascript/packs/admin.js
+++ b/app/javascript/packs/admin.js
@@ -11,3 +11,4 @@ import '../src/js/admin/tab_selection_in_anchor';
 import '../src/js/admin/simple_uppy_file_input';
 import '../src/js/admin/uppy_dashboard.js';
 import '../src/js/admin/qa_autocomplete.js';
+import '../src/js/admin/queue_status_submit.js';
--- a/app/javascript/src/js/admin/queue_status_submit.js
+++ b/app/javascript/src/js/admin/queue_status_submit.js
@@ -23,8 +23,19 @@ domready(function() {
   }
 
 
-  // The "ajax:" events are custom events from rails-ujs
+  // On digitization queue item status select menu change, auto-submit form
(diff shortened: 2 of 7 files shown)


eddierubeiz:
OK -- I've stared at the code and messed with the UI. Looks great to me. I'm going to merge and push to production.

PR #457, opened 2019-10-31
Adding a "published ?" facet

eddierubeiz:
Adding a facet on whether items are published or not.
The new facet is only visible to logged-in users.


eddierubeiz:
Ref #432

jrochkind:
I think the facet shoudl be called `Visibility (Staff-only)`  matching chf_sufia (the "Staff-only" is a reminder that the public can't see this facet), and the values should be "published" and "private" to match the language we use elsehwere in the admin screens. 

I'll go ahead and make commits myself to do these things. Looking at chf_sufia to see how you customize the facet labels.

jrochkind:
Non-published things are of course not showing up in the facet yet, because we don't have them showing up in search results at all yet. Which makes this a bit confusing to test (manually or automatically), and will be a bit confusing to users when there's a facet but it only ever shows 'published' things. 

Don't think we should merge until we have non-published things showing up in search.

eddierubeiz:
Yes, that makes sense.

Issue #458, opened 2019-10-31
upgrade to sprockets 4.0


jrochkind:
Not technically necessary even to get to Rails 6, but a good idea to stay current, and theoretically get new features like sourcemaps for dev debugging. 

https://bibwild.wordpress.com/2019/10/15/sprockets-4-and-your-rails-app/

There are some files that get accidentally compiled as top-level targets under new sprockets defaults, causing errors. Will have to move them and/or customize top-level target config. 

Will have to make sure our html files  for custom error pages (compiled with sprockets) are targetted as top-level targets.

jrochkind:
See also https://github.com/rails/sprockets/blob/master/UPGRADING.md

Some things don't seem to be working as I expect in sprockets 4, will spend some more time investigating and possibly reporting issues to sprockets.

PR #459, opened 2019-10-31
Lock to sprockets 3

jrochkind:
Upgrading to sprockets 4 breaks our app until we tweak it to work. Putting this in Gemfile will keep any 'bundle update's of other dependencies from accidentally updating sprockets 4 too, before we're ready. When we're ready, we change or removee this line in Gemfile. We did create an issue to upgrade to sprockets 4 eventually, #458


eddierubeiz:
Noted. Merging!

PR #465, opened 2019-11-04
Show the most recent items on the home page under "recent items".

eddierubeiz:
Rather than the *least* recent items, how about it?

### DIFF
--- a/app/presenters/recent_items.rb
+++ b/app/presenters/recent_items.rb
@@ -39,7 +39,7 @@ class RecentItems
 
   def fetch_bag
     @@bag ||= Work.where('published = true').
-      order('updated_at').
+      order('updated_at desc').
       limit(@how_many_works_in_bag)
   end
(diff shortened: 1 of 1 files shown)


eddierubeiz:
#ref 462

jrochkind:
OK, this is clearly a bug fix, but I wonder if there is another  bugs in here too. 

* why are we searching on `updated_at` instead of `created_at`? What did chf_sufia do?

I'm going to go look...

jrochkind:
OK, existing chf_sufia code used the equivalent of `updated_at` for some reason too. 🤷‍♀ 

https://github.com/sciencehistory/chf-sufia/blob/master/app/helpers/recent_items_helper.rb

This is probably another consequence of our lack of tracking `published_at`, which we might do in the future, which would make both "recent items" and "sort by recently added" less of an "estimate".

Issue #403, opened 2019-10-06
Certain nonstandard mp3 files are not correctly characterized as audio


jrochkind:
https://kithe.sciencehistory.org/works/zw12z6384
( post-launch: https://digital.sciencehistory.org/works/zw12z6384 )

eddierubeiz:
I'll be working on this Monday 10/7

eddierubeiz:
This may be a duplicate of https://github.com/sciencehistory/scihist_digicoll/issues/328 .

All three items are characterized as `application/octet-stream`:

https://kithe.sciencehistory.org/admin/asset_files/bc386k398
https://kithe.sciencehistory.org/admin/asset_files/k930bz06p
https://kithe.sciencehistory.org/admin/asset_files/5h73px50k

eddierubeiz:
These do appear to be mp3s of some sort. When I download the three files onto my mac, they aren't recognized by `file` as mp3s, but on the other hand `ffprobe` gives:

```[mp3 @ 0x7f986900b400] Skipping 480 bytes of junk at 0.
[mp3 @ 0x7f986900b400] Estimating duration from bitrate, this may be inaccurate
Input #0, mp3, from '5h73px50k.mp3':
  Duration: 00:32:09.07, start: 0.000000, bitrate: 160 kb/s
    Stream #0:0: Audio: mp3, 48000 Hz, stereo, fltp, 160 kb/s```


```[mp3 @ 0x7fbb2d008200] Skipping 480 bytes of junk at 0.
[mp3 @ 0x7fbb2d008200] Estimating duration from bitrate, this may be inaccurate
Input #0, mp3, from 'bc386k398.mp3':
  Duration: 00:02:15.10, start: 0.000000, bitrate: 160 kb/s
    Stream #0:0: Audio: mp3, 48000 Hz, stereo, fltp, 160 kb/s```


```[mp3 @ 0x7f9364808200] Skipping 480 bytes of junk at 0.
[mp3 @ 0x7f9364808200] Estimating duration from bitrate, this may be inaccurate
Input #0, mp3, from 'k930bz06p.mp3':
  Duration: 00:30:51.74, start: 0.000000, bitrate: 160 kb/s
    Stream #0:0: Audio: mp3, 48000 Hz, stereo, fltp, 160 kb/s```

jrochkind:
@eddierubeiz OK, the question was why they show up in new thing prod, but not chf_sufia prod. 

[...]

PR #474, opened 2019-11-07
use newer kithe with mediainfo-based fallback content-type detection

jrochkind:
Ref #403, should fix, but we need to fix existing uncharacterized files too

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,7 @@ gem 'font-awesome-rails', '~> 4.7'
 
 # temporary kithe indexing branch, for scihist_digicoll indexing branch, do not
 # intend to merge to master like this.
-gem 'kithe', git: "https://github.com/sciencehistory/kithe.git"
+gem 'kithe', git: "https://github.com/sciencehistory/kithe.git", branch: "mediainfo_characterization"
 
 # temporary git master, we should get on an attr_json release once we're settled down
 gem "attr_json", git: "https://github.com/jrochkind/attr_json" #path: "../attr_json"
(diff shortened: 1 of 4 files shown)


jrochkind:
Once we have characterization working properly, we need to fix existing files. 

We _could_ just force their content-type to `audio/mpeg`, like this:

```ruby
asset.file.metadata["mime_type"] = "audio/mpeg"
asset.save!
```

But better to use the (improved) automated code for detecting mime-type, to ensure we're not setting something that wouldn't have been calculated by the new improved code, and to avoid typos. 

It's a _bit_ cumbersome to do that in shrine 2.x (improved in shrine 3.x which we aren't on yet). 

```ruby
file = asset.file
file.refresh_metadata!
asset.file_data = file.data
asset.save!
```

Or, for the three that are mis-characterized, to re-detect content-type, and then generate derivatives based on newly corrected content-type:

```ruby
["5h73px50k", "bc386k398", "k930bz06p"].each do |friendlier_id|
  asset = Asset.find_by_friendlier_id(friendlier_id)
  file = asset.file
  asset.file_data = file.data
  asset.save!
  asset.create_derivatives
end
```

eddierubeiz:
Noted!

jrochkind:
having trouble getting those items fixed on production, not sure why, will go back to staging to play around with it!

jrochkind:
Ah, had a missing line in my script for fixing the assets -- the all important `file.refresh_metadata!` line. 

Fixed in script above. Run. All assets in https://digital.sciencehistory.org/works/zw12z6384 should be fixed, they are now recognized as audio/mpeg, should have the required derivatives, should be displaying in public UI correctly.

Issue #470, opened 2019-11-05
Clicking on "format" "Physical object" hyperlink in record broken


jrochkind:
When I go to:

https://digital.sciencehistory.org/works/z029p5724

And under "Format", click on "Physical object", I get "No results found for your search", at this URL: https://digital.sciencehistory.org/catalog?f%5Bformat_facet%5D%5B%5D=Physical+object

It ought to have results (like at minimum this very record, but of course many more). Clicking on "Image" works to give results.  None of it matters if you are logged in or not. 

Starting from a fresh search and choosing "Physical object" from "Format" in sidebar results in  this _working_ URL: https://digital.sciencehistory.org/catalog?f%5Bformat_facet%5D%5B%5D=Physical+Object&q=&range%5Byear_facet_isim%5D%5Bbegin%5D=&range%5Byear_facet_isim%5D%5Bend%5D=&search_field=all_fields

Looks like the difference is lowercase `o` vs uppercase `O`. I'm not sure which one is actually saved in the db, and what is choosing to change case where.  Also would be good to compare to chf_sufia app.

PR #475, opened 2019-11-07
transform 'format' for display with 'titleize', the same as it's transformed in work indexer for indexing

jrochkind:
This makes click on hyperlink for facet limit work right, #470


eddierubeiz:
Got it. Merging this and pushing now.

Issue #449, opened 2019-10-30
Public view for private (unpublished) works only displays 1st/representative image for multi-file works


HKativa:
When looking at an unpublished work with multiple files in the public view, only the first image is visible. For example, see https://kithe.sciencehistory.org/works/r494vm45w (compare to https://digital.sciencehistory.org/works/r494vm45w in current app). 

For cataloging workflow, it's helpful to have all files associated with a work visible and available for reference.

eddierubeiz:
Fixed in production. See e.g. https://digital.sciencehistory.org/works/vcs937d

PR #479, opened 2019-11-08
Tweak to member visibility rules on public work view.

eddierubeiz:
If a user is logged in, it's OK to show non-published member items on the public view of the work.
Includes tests for both audio and regular decorators.

Ref #432
Ref #449

Note: PR #489 depends on this one. #489 adds "Private" labels to the non-published member items now being displayed to logged-in users, to make the interface less confusing. Please merge this PR first, as it will simplify #489.

### DIFF
--- a/app/decorators/audio_work_show_decorator.rb
+++ b/app/decorators/audio_work_show_decorator.rb
@@ -1,5 +1,6 @@
 class AudioWorkShowDecorator < Draper::Decorator
   delegate_all
+  include Draper::LazyHelpers
 
   # This is called by works_controller#show.
   def view_template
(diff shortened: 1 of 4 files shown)


jrochkind:
This is looking good. I want to make sure we already had a spec that ensures we _don't_ show non-published items to non-loggedin-people. 

Also, if we're showing non-published things to logged-in-staff -- I think they need some kind of an _indicator_ on-screen that they are "private" items. It can be very confusing for a staff user to realize what they are looking at is not necessarily what the public is looking at, and a 'private' indicator can help.  (I thought I mentioned this somewhere in the issue description, but don't see it now!)

chf_sufia did this, by using some fancy CSS to put a translucent "private" badge ABOVE the thumbnail. That seems like a good way to copy it. I can do it, or you can do it @eddierubeiz , let me know your inclination!

jrochkind:
(Verified that there are specs that non-published things don't get shown to non-logged-in-users)

Issue #482, opened 2019-11-11
Promoting representative image to child work removes "representative" status


HKativa:
While editing hx11xf71x, I found that promoting the representative image of a multi-file work to a child work strips that image of its representative status, leaving the parent work without a thumbnail. Editing the parent work to re-select a representative image does resolve the issue.

eddierubeiz:
Good catch! Never thought of that!

eddierubeiz:
It occurs to me we could simply change the rules to prevent the user from promoting a representative asset to a child work. Instead, we display an error message along the lines of "You can't promote this asset, because it's currently the representative for Work X. Please change the representative to another member before promoting this asset."

jrochkind:
Nah, if that's something staff wants to do, we should let them do it, and make it work right, not put barriers in their way. 

Often you are going to want the promoted child still be the representative. No reason to make people switch the representative to something else, promote, then switch back. No reason to write special case code telling them they can't do what they actually want to do, instead of special case code to make it work.

eddierubeiz:
Ah, so the default behavior is to transfer the promoted asset's representative status to the newly-created child work. Duh!

jrochkind:
Taht's not what it's doing now -- haven't investigated to figure out exactly whats' going on now -- I suspect the "represntative_id" foriegn key isn't being touched at all, but becomes invalid after promotion. 

But yeah, we'll make it so.

PR #486, opened 2019-11-11
Preserve parent represntative on asset/child work conversion

jrochkind:
Both directions. 

Ref #482


eddierubeiz:
This all makes sense.

Issue #190, opened 2019-07-01
Requests for catalog.atom or catalog.xml results in uncaught exception


jrochkind:
We are getting these exceptions reported regularly by honeybadger, probably bots are asking for them. 

~But we might want atom feed to actually work. See #201. ~ (not for now)

See also #350, not sure which backtrace (or both?) the app is currently doing. 

## Backtrace

    line 24 of [PROJECT_ROOT]/app/controllers/application_controller.rb: block in batch_kithe_indexable
    line 23 of [PROJECT_ROOT]/app/controllers/application_controller.rb: batch_kithe_indexable

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/51385653)

jrochkind:
Per planning meeting today, we are not going to fix Atom/RSS feeds to work for now. 

Instead, we want to make it so the functions aren't avaialble at all:
* Aren't advertised in any HTML "auto discover" tags, so bots are less likely to find and try it
* Trying the .atom or .rss feeds just gets you a 404, not an uncaught software error. 

This may require some hacking to figure out how to override Blacklight like this. (Or wait, rather than mess with the BL-integrated controller, we could just put something in routing to "take over" the .atom and .rss links, and route them to a 404?)

jrochkind:
unsupported formats that result in an error page:

https://digital.sciencehistory.org/catalog.atom
https://digital.sciencehistory.org/catalog.rss

We might have meta tags in our html advertising these, get rid of them. 

https://digital.sciencehistory.org/catalog.xml => doesn't necessarily work, but doesn't give me a 500 error so not a priority.

PR #497, opened 2019-11-15
Turn off rss, json and atom search results. 

eddierubeiz:
1) short-circuit the offending URLs in rails.rb

2) override the three methods that were providing the alternate links to the template so they return nil.

Ref #190


jrochkind:
Nice, short and sweet! Well done. 

I will mark the honeybadger error resolved once this is deployed, so we'll know if it comes back unexpectedly. 

Overriding those weird methods in CatalogController is a bit weird, but seems like that is the cleanest way to intervene in the Blacklight stuff! Good code archeology to find that.

PR #500, opened 2019-11-19
honeybadger notify fixity failure, don't group

jrochkind:
Only group by asset id, so different failures become different honeybadger groups.

Also add a tag fixity in case we need to treat them differnetly in honeybadger UI or API.

https://docs.honeybadger.io/lib/ruby/getting-started/reporting-errors.html
https://docs.honeybadger.io/lib/ruby/getting-started/customizing-error-grouping.html
https://docs.honeybadger.io/lib/ruby/getting-started/tagging-errors.html


eddierubeiz:
Works for me!

Issue #448, opened 2019-10-30
Image viewer does not open on public view for private works 


HKativa:
In the current app, the image viewer opens in the public view for works whose visibility is set to "staff-only." For example, see https://digital.sciencehistory.org/works/4b29b7216

Compare to the new app, where the image viewer does not open for works whose visibility is set to "private" (the equivalent of "staff-only" in the new app: https://kithe.sciencehistory.org/works/4b29b7216

PR #505, opened 2019-11-19
Allow deep zoom image viewer for unpublished items if user is logged in.

eddierubeiz:
If the user is logged in, describe all members in viewer_member_info_serializer .


eddierubeiz:
Ref #448

eddierubeiz:
This is verified as working in Staging:
https://staging-digital.sciencehistory.org/works/40rytt7
I ingested the file; it has never been public, and you can still access the DZI viewer.

jrochkind:
nice

Issue #408, opened 2019-10-07
Small SOLR bug w/ department search


eddierubeiz:
Just logging a small bug w/ solr faceting. Start with a blank search and then open the "Department" facet.
4 items with no department show up like this:

![Department facet](https://user-images.githubusercontent.com/94723/66343519-ccbe7980-e919-11e9-8bb8-9f90f45dc101.png).

HKativa:
@eddierubeiz is there a way to see what these items without a department are? There shouldn't be any items without a department designation, unless these are test records that shouldn't be visible in any case.

jrochkind:
There's no way in the app UI, either in chf_sufia or the new app. But we can run a manual report by writing some code. I will do so shortly in both chf_sufia and new app, to see what they are and make sure they match in both. (This is easier to do in a more reliable way in the new app, but we'll want to reassure ourselves that it matches in old app, so I'll do my best to get it out of old app too).

jrochkind:
OK, in the new app production I found works with no department set like this:

    Work.where("json_attributes ->> 'department' is null OR json_attributes ->> 'department' = ''").collect(&:friendlier_id)

I actually found 7 IDs: ["pn89d691t", "dv13zv52d", "r494vk60s", "ft848q73q", "v405sb22w", "tt44pn900", "mp48sd662"]

 However, three of them are test records:   `r494vk60s`, `v405sb22w`, `mp48sd662`. (Confirmed all three are non-public test records in chf_sufia too). 

One of them is is non-public in both chf_sufia and new app production: `tt44pn900`

That leaves three that appear to be public and erroneously missing a department (not sure why this is 3 instead of the 4 suggested by the weird UI at top of this ticket; would take more investigation to figure out what's going on there):  `pn89d691t`, `dv13zv52d`, `ft848q73q`

(Note that we have a ticket suggesting feature making it impossible to mark things public without expected metadata, #372)

Now we're going to try to find all records missing department in chf_sufia, to make sure it doesn't include any additional records.  In chf_sufia, the only way to do this is via the Solr index. `http://3.93.216.251:8983/solr/collection1/select?df=*&fl=id&fq=-division_tesim:[*%20TO%20*]&fq=has_model_ssim:%22GenericWork%22&indent=on&q=*:*&wt=json`

Found 7:  "ft848q73q", "pn89d691t","r494vk60s", "mp48sd662", "dv13zv52d","tt44pn900","v405sb22w"

Matches! Same list.

[...]

PR #506, opened 2019-11-19
WIP Indexing blank values in Rights and Department

eddierubeiz:
Changes to work_indexer.rb to account for (unlikely) blank values in fields that are a) not supposed to be blank in the first place and b) picked from a limited vocabulary. Practically speaking, this means "Department" and "Rights".

We still want to facet these so they can be easily found (ref #408 )

As a proof of concept, I'm indexing these, respectively, under:
`No Department` and `http://rightsstatements.org/vocab/CNE/1.0/` , which translates as `Copyright Not Evaluated` on the front end.

No tests yet.

### DIFF
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -26,10 +26,10 @@ class WorkIndexer < Kithe::Indexer
     to_field "text_no_boost_tesim", obj_extract("related_url")
     to_field ["text_no_boost_tesim", "place_facet"], obj_extract("place", "value")
     to_field "text_no_boost_tesim", obj_extract("related_url")
-    to_field ["text_no_boost_tesim", "department_facet"], obj_extract("department")
+    to_field ["text_no_boost_tesim", "department_facet"], obj_extract("department"), transform(->(v) { v.empty? ? "No Department" : v })
     to_field ["text_no_boost_tesim", "medium_facet"], obj_extract("medium")
     to_field ["text_no_boost_tesim", "format_facet"], obj_extract("format"), transform(->(v) { v.titleize })
-    to_field ["text_no_boost_tesim", "rights_facet"], obj_extract("rights") # URL id
+    to_field ["text_no_boost_tesim", "rights_facet"], obj_extract("rights"), transform(->(v) { v.empty? ? "http://rightsstatements.org/vocab/CNE/1.0/" : v }) # URL id
     to_field ["text_no_boost_tesim"], obj_extract("rights"), transform(->(v) { RightsTerms.label_for(v) }) # human label
     to_field "text_no_boost_tesim", obj_extract("rights_holder")
     to_field "text_no_boost_tesim", obj_extract("series_arrangement")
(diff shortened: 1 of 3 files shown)


jrochkind:
OK, what I'd like to suggest is  something different than you are doing here. 

You are sending "No Department" to the Solr index when the value is nil or blank. This works, but takes up space in the Solr index for it. 

It ought to instead be possible to send _nothing_ to the Solr index in this case. Recognize that it's the empty string, and treat it just like `nil` instead --- don't index a value for this Solr field at all. 

Then, as I [mentioned in #408](https://github.com/sciencehistory/scihist_digicoll/issues/408#issuecomment-555558656), we should be able to use the Solr config for `facet.missing` to tell Solr to give us a facet "bucket" for things with no value.

eddierubeiz:
Back in 2015  stated [Blacklight does not support Solr's facet.missing option at all](https://github.com/projectblacklight/blacklight/issues/1130#issuecomment-77307264).  [ScholarSphere's approach](https://github.com/psu-stewardship/scholarsphere/pull/1544) could work, but might be overkill.

jrochkind:
Huh, okay! I could swear I remembered using it in a Blacklight project, and there's some way to make it work... but maybe not? Shall we just stick to this way then I guess? What do you think?

jrochkind:
Wanna ask in Blacklight slack just to see if anyone has anything to say? Or want me to?

eddierubeiz:
I posted a note on the Blacklight slack -- seems like it couldn't hurt to ask.

eddierubeiz:
OK -- I posted on Blacklight Slack. David Kinzer over at Temple kindly pointed out that Blacklight *does* support `facet.missing`, as of cbeer's merging the PR quoted above. In a separate branch in dev, I've been able to turn on facet.missing for the two fields in question and reindex. This does indeed lead to a resulting `[missing]` field in the results for those fields, but it is clearly not working the way it should.

( If you want to experiment, @jrochkind : I just added the following to `solrconfig.xml`, 
```
<str name="f.department_facet.facet.missing">true</str>
<str name="f.rights_facet.facet.missing">true</str>
```
... then ran clean, stop, start and reindex . )

I'm going to recommend going with the code in this PR instead.

jrochkind:
works for me, good looking into the Blacklight feature, not worth fighting with it.

eddierubeiz:
OK, I've done some more research.

**Facts:**

- The facets are not actually for works, they're for (works or collections).
- Published collections are *supposed* to have null departments and rights fields.
- Using facet.missing works in terms of counting nulls (rather than blanks) but would need some more work in order to get the facet link to actually list the items.
- However, in practice, even if we got the link working, the link would just point to a list of our 52 collections, PLUS any stray works with nil works or department. This is not what we want per the use case described in the issue.

**Recommendation:**

Consider the following rules:

- Collections should ALWAYS have null departments and rights fields.
- Works should NEVER have null departments and rights fields
- UNPUBLISHED Works CAN have empty-string departments and rights fields.
- PUBLISHED works CANNOT (but this is a validation issue for another PR).

If and when these rules are true, the fact that null items don't get faceted means collections don't show up in the facets for these fields. Which is great, because we don't want them in those facets: the whole point of these facets was to identify works (not collections) with missing rights or department.

Good news! These rules are **already** true, except for 26 unpublished works with `nil` (rather than `""`) rights. We change these 26 and then the rules are all true.

Whether we use the code in this PR is a matter of debate; we probably don't need it if we make a couple modifications to the form and validation so that the rules *stay* true. But it wouldn't hurt either, I guess.

jrochkind:
Not going to do this for now, rely on kithe feature instead. See #408

PR #512, opened 2019-11-22
Show all items to logged-in users on featured-items and collections pages.

### LINKED ISSUE #509: Make non-public items visible to logged-in users on collection landing pages
For example, see https://digital.sciencehistory.org/collections/b2773v77g: as a logged-in user, I can see that this collection currently has 88 public items out of 158 total items (the remaining 70 are set to "private" while awaiting metadata). However, from the collection page, there is no way to see the private items that are awaiting cataloging. Alternately, I can find these items in the public search (as a logged-in user) by limiting the visibility to "private" and searching by the accession number or another identifier, but it would be more convenient to locate these items directly from the collection page.

eddierubeiz:
Fixes #509 .

`current_user` wasn't being provided to the `AccessControlFilter` because the two search controllers were clobbering the superclass' `search_service_context` instead of extending it.

Better now.

### DIFF
--- a/app/controllers/collection_show_controller.rb
+++ b/app/controllers/collection_show_controller.rb
@@ -44,7 +44,7 @@ class CollectionShowController < CatalogController
 
   # Our custom SearchBuilder needs to know collection id (UUID)
   def search_service_context
-    { collection_id: collection.id }
+    super.merge!(collection_id: collection.id)
   end
 
   def check_auth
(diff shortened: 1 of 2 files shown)


jrochkind:
Nicely done @eddierubeiz , thanks!

jrochkind:
We could have added tests (of some/any kind, controller, system, whatever) that failed before the fix, which is what someone rigorous about testing would say you should do fixing almost any bug, but merged anyhow. We use our judgement about when we need tests to have adequate confidence in the software.

Issue #299, opened 2019-08-16
upgrade to Rails 6.0


jrochkind:
Rails 6.0 is out. https://weblog.rubyonrails.org/2019/8/15/Rails-6-0-final-release/

Most of our app should be pretty straightforward to upgrade. 

The one challenge might be Blacklight (further raising question of Blacklight's cost/benefit value). 

Also, we'll need to get sprockets-for-JS back in there, for our dependencies only available via sprockets-for-JS from gems. 

Not necessary before launch, but we should stay on top of it. 

(kithe and attr_json are both already passing their own tests on Rails 6). 

**note** will probably want to do #233 first to avoid confusion.

jrochkind:
Blacklight 7.3.0 released yesterday should support Rails 6, so one less barrier to update. 

https://github.com/projectblacklight/blacklight/releases/tag/v7.3.0

jrochkind:
While blacklight now supports Rails 6, the `qa` gem (aka "questioning_authority") sadly does not. https://github.com/samvera/questioning_authority/issues/287

jrochkind:
Note, as part of upgrading to Rails, we are also upgrading ALL dependencies to get more recent versions released that meet dependency tree requirements. (Ie, by doing a `bundle update`).

PR #513, opened 2019-11-26
upgrading to rails 6

jrochkind:
Ref #299 

## Changes...

Gemfile.lock of course changes to update a lot of dependencies. And Gemfile to ask for Rails6, and related dependency updates. 

Most file changes are just updating some 'standard' config files to have the config or logic that would have been generated with Rails6, to stay up to date. 

Additionally, some `response.content_type` changed to `response.media_type` for Rails 6. https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#actiondispatch-response-content-type-now-returned-content-type-header-as-it-is

## Many dependencies updated

As of this moment,  `bundle update` can be run regularly to update all dependencies to the latest allowed by dependency tree restrictions, getting any new versions with bugfixes, security fixes, etc.  The Gemfile is such that `bundle update` will not update anything known to break. And this PR includes regular `bundle updates` so when merged we'll be getting latest versions of lots of dependencies. 

## Some dependencies not updated

Originally I had hoped to upgrade some other dependencies in this PR too, just get all our dependencies really spiffy, with a really tiny output in `bundle outdated`. 

However, browse-everything is still limiting us to some old dependencies -- `bundle update` won't get to latest version of some dependencies because browse-everything's gemspec restricts it. 

 An upcoming release of browse-everything might fix this, but at the moment we can't upgrade browse-everything because it refuses to let us use sprockets 4. This may be resolved subsequently https://github.com/samvera/browse-everything/issues/310 https://github.com/samvera/browse-everything/pull/345


jrochkind:
Rails6 seems to have broken our attempts under testing to use different ActiveJob queue adapters for different tests. 

Very confused about what's going on, not sure if it's broken that for everyone, everyone using rspec, just us, or what. Haven't yet figured out how to fix it. 

https://github.com/rails/rails/issues/37270

jrochkind:
Spending some time with https://guides.rubyonrails.org/upgrading_ruby_on_rails.html to potentially manually  upgrade certain other in-app things.

jrochkind:
One test is failing because of a bug in rspec-rails. 

A fix has been merged into rspec-rails master, but is not yet released. https://github.com/rspec/rspec-rails/pull/2242

I would like to give it a few weeks to see if we get an rspec-rails release; the only other option we know of for moving forward is making a temporary local fork of rspec-rails and depending on that instead of an official rspec-rails release. That is possible, but would be kind of unfortunate and require us to remember to switch back to real rspec-rails in the future. (Trying to depend on rspec-rails git master instead of a release -- while ordinarily an option with git-hosted rubygems, doesn't work with rspec-rails for idiosyncratic reasons).

jrochkind:
Oops, need to do a bit of cleanup with regard to new zeitwerk autoloader too, especially wrt STI. 

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#require-dependency

May want to put some of that in kithe, or not.

jrochkind:
After deploy of rails6, people will probably have to log in again -- rails 5 cookies will end up invalidated due to new security measures in Rails6.

Issue #394, opened 2019-10-02
Email from kithe jobs server?


eddierubeiz:
Hey Dan! We had discussed this briefly last week, but would it be possible to rig up the Kithe jobs server to send mail?

Currently, on `scihist_digicoll-jobs1-staging`, if you run:
```
$ cd /opt/scihist_digicoll/current/ && bin/rails runner "ActionMailer::Base.mail(from: 'no-reply@sciencehistory.org', to: 'digital-tech@sciencehistory.org', subject: 'Test message', content_type: 'text/html', body: 'Test message').deliver_now"
```
you get the error `sh: 1: /usr/sbin/sendmail: not found`

Fixity failures are being reported via Honeybadger, but it would be nice to have this in addition (see https://github.com/sciencehistory/scihist_digicoll/pull/387/files ).

Thanks for looking into it!

jrochkind:
@sanfordd you said in slack you had done something about this, but I'm confused cause I thought our brainstormed solution involved a single SMTP server and would Rails app config switched to use it. Please outline on this ticket what your plan is and where it's at, thanks!

sanfordd:
Documenting this here,

What we'll be doing is opening the email server to the jobs server(s) so we can pass along SMTP requests as per @jrochkind's recommendation in Slack.

This requires that our email server be open in a very limited manner, we don't want to start being a source of spam. We'll be limiting access with both the firewall rules and postfix itself will limit access to allowed servers.

sanfordd:
We've got what should be an open setup now. 
The local_env file has a variable `email_sever` which points to the right location. We're using the AWS public DNS, which EC2 instances will properly resolve to the private IP address of the server in question. 
Since it uses the elastic IP we can easily move to new servers without any need to push out local_env changes with each new email server build.

jrochkind:
`email_server` is SMTP server data?  Can you give an example of the value, is it just a host, host and port, etc? 

The next step will be making the app _use_ that config, if it is set.

sanfordd:
It's currently just a host,
`ec2-[Public-IP-Address-here].compute-1.amazonaws.com`
Where [Public-IP-Address-here] is the public IP with dashes instead of dots, i.e. 1-2-3-4
Pinging or using that causes EC2 instances to _not_ use the public address, but resolve the internal IP address (172.XX.YY.ZZ)
No port, though that's easy to add if needed.

jrochkind:
[...]

PR #515, opened 2019-11-27
Send email from jobs

eddierubeiz:
These settings should allow the Jobs server to send email.
Ref #394

### DIFF
--- a/app/lib/scihist_digicoll/env.rb
+++ b/app/lib/scihist_digicoll/env.rb
@@ -266,6 +266,7 @@ module ScihistDigicoll
     define_key :no_reply_email_address
     define_key :digital_tech_email_address
 
+    define_key :smtp_host
 
   end
 end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -74,7 +74,7 @@ Rails.application.configure do
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.smtp_settings = {
-    address: ScihistDigicoll::Env.smtp_host,
+    address: ScihistDigicoll::Env.lookup!(:smtp_host),
     domain: 'sciencehistory.org',
     ssl: false,
     enable_starttls_auto: false
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -71,7 +71,6 @@ Rails.application.configure do
   }
 
   config.action_mailer.perform_caching = false
-  config.action_mailer.raise_delivery_errors = true
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.smtp_settings = {
     address: ScihistDigicoll::Env.lookup!(:smtp_host),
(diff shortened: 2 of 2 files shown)


jrochkind:
Aha, I see the config got a bit complicated with starttls and all, thanks for figuring it out!

Do you think we should wrap the config in a conditional, to only do the new thing when `ScihistDigicoll::Env.lookup(:smtp_host)` is present, otherwise leave the old thing?

i'm not sure what will happen under dev or test if you set the smtp host config to nil. 

Actually, with the code exactly as it is, I know what will happen: `ScihistDigicoll::Env.lookup!(:smtp_host)`, the `!` on the end of `lookup` is the form of the method that will *raise* an exception if no value is present. This is convenient for "fail fast with clear error instead of confusing error down the line" for values we insist on being present -- but in this case, we don't expect it to be present in dev or test, so I think it may end up raising an error in dev or test on boot? (Can you even start the app in dev or run tests locally without an error with this code the way it is?)

So minimally, you should turn that into `lookup` without the `!` -- but maybe also make the new config conditional only if the value is present? Or is that too dangerous cause if it's accidentally not present, emails will just silently fail again?

i'm not totally sure what we should do -- what do we want to happen in dev and test?  Any thoughts?

jrochkind:
> We actually don't want to raise mail delivery errors in production.

I'm not sure if we do or not; maybe we do, to find out about email delivery errors and get a honeybadger notification, instead of having them just silently fail?  Thoughts?

eddierubeiz:
You know what? Let's have them fail loudly if they're misconfigured in prod. I would absolutely want to know right away if e.g. a password reset email wasn't going to get sent.

jrochkind:
okay you’ve changed it to only configure if `ScihistDigicoll::Env.lookup(:smtp_host)` is present — that handles test/dev, where we want the app to be able to start up without local_env :smtp_host defined. (In general, we don't want a local_env.yml to be required in dev/test, the app should start up and run with default configuration for dev/test without it at all). 

But as far as in production… maybe it should raise right away in production? which it now no longer does?

There are two notions of "production" that we could check.  

One is the Rails "environment", which you can check with `Rails.env.production?`. That sets a bunch of Rails configuration defaults for "production mode". We run with it in both staging and true "production", but also very occasionally I run in Rails production environment on dev, when I need to reproduce something only with that config. So if we checked that and raised, then if I was trying to do that on dev, I'd need to set that value (We already have some values like that, which have already been annoying to me :) )

Alternately, we have our own `ScihistDigicoll::Env.lookup(:service_level)` value, which can be `production`, `staging`, or nil. (and ScihistDigicoll::Env [enforces that it must be only one of those three values](https://github.com/sciencehistory/scihist_digicoll/blob/master/app/lib/scihist_digicoll/env.rb#L48)). So we could instead say if it's non-nil, then we'll raise right away if no smtp_host is given.  in dev and test it is nil, it is only production or staging on our deploy environments.

Issue #391, opened 2019-10-01
Post-launch: refactor email notification after fixity check problems.


eddierubeiz:
In https://github.com/sciencehistory/scihist_digicoll/pull/387 there's a discussion of rewriting the fixity check code to use an ActionMailer sub-class and and template file, instead of putting the html inline in a string (as the code was in Sufia).

We decided to postpone this work until after launch.

PR #522, opened 2019-12-02
Better fixity emails, using a proper ApplicationMailer object 

eddierubeiz:
Per discussion in https://github.com/sciencehistory/scihist_digicoll/pull/387#issuecomment-535629994 , we had decided to do this after launch, and it seemed a good idea to refactor this now, since we were already working on https://github.com/sciencehistory/scihist_digicoll/pull/515 . 

Ref #391


app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  default from: ScihistDigicoll::Env.lookup!(:no_reply_email_address)
```
jrochkind: Nice catch for consistency, but =this will raise on loading the class if no Env `no_reply_email_address` is provided. And our class doesn't provide a default in dev/test. So. 

(I'm actually surprised and a bit alarmed that tests pass with this, not sure if I don't understand something, or if tests just don't cause this class to be loaded). 

Maybe change to `lookup` without the `!`, which will just return nil instead of raising. Or, backwards compatibly, `(ScihistDigicoll::Env.lookup(:no_reply_email_address) || "from@example.com")

app/mailers/fixity_failure_mailer.rb
```diff
@@ -0,0 +1,24 @@
+#class FixityFailureMailer < ApplicationMailer
+
+class FixityFailureMailer < ApplicationMailer
+
+  def fixity_failure_email
+    @fixity_check = params[:fixity_check]
```
jrochkind: It's been a while since I wrote a mailer -- I'm a bit disturbed by the way the `params` are not documented in a method signature, is this really the standard way to do it? I'm going to double check in the [Rails Action Mailer guide](https://guides.rubyonrails.org/action_mailer_basics.html)...

That does seem to be what the examples there do, okay then. 🤷‍♂   What if we raise if the arg isn't supplied though, to avoid more confusing debugging if it's accidentally missing?

    @fixity_check = params[:fixity_check] || raise ArgumentError.new("Required params[:fixity_check] missing")

app/mailers/application_mailer.rb
eddierubeiz: Good call. Will use the latter.

app/mailers/fixity_failure_mailer.rb
jrochkind: @eddierubeiz You told me this was ready for review, but you don't seem to have responded to or addressed this comment? Please use the comments on a PR as a checklist of things to consider before asking for another review -- if you don't think one is necessary or is a bad idea, you can say that in the comment thread instead of just leaving it there hanging.

eddierubeiz: I am checking for the presence of params[:fixity_check] and raising an error if it's not present. Just not on the same line (the way you had it resulted in a syntax error)

Issue #372, opened 2019-09-18
Required fields to publish a record


yonyitz:
Please require the following fields before publishing a record (other than title and external ID, which are already required to save a record):

Date
Format
Genre
Department
Rights

Thanks!

jrochkind:
Part of this would be also preparing a report of things already in the db that are invalid according to these rules. 

We might want to fix them before deploying the code to require published items to have those things, to avoid odd interactions.

jrochkind:
There are a LOT of published items missing date. (I did do a couple spot checks confirming missing date in sufia too, does not appear to be a migration error). 

It's possible many/all of them should be marked "Undated"? Which is a possible thing to do, but these aren't marked Undated, they have no dates at all. We could write a simple script to change them all to Undated if we wanted. 

A few others missing genre, rights, or others. 

Here is the list. 

https://gist.github.com/jrochkind/9ab75d74b0b7042f7d0a956075611e5f

@HKativa @yonyitz would it make sense to fix these manually, at the same time we are putting code in to prevent you from publishing items without these in the future?

yonyitz:
Thanks for this info, Jonathan.  The metadata guidelines state that undated can be used.  Not sure whether it has been in the past @HKativa .  I'm fine with fixing manually depending on what @HKativa  recommends!

jrochkind:
we're going to hold off on the software change until the existing bad data is reviewed

jrochkind:
Actually I might start dev'ing it now, even if we don't deploy it. 

The UI I am thinking of is -- if you hit the 'publish' button, and there are missing required fields, it will drop you into a Work Edit form, with a "Missing fields required to publish" message, and those fields highlighted. 

If an item is already published, and you're editing it, the fields will show up and validate same as existing universally required fields (title and external id). 

Hopefully that makes sense, please let me know if it does not! @yonyitz @HKativa

jrochkind:
So as above, my original idea was if you try to publish a work that lacks required fields, it drops you to the work edit field, with an error message. 

But working on it, I realize now that when you publish a work it ALSO publishes all that work's children. If the work has work children that lack required fields to be published... the UI is not simple, best we can do is maybe let you know that? 

Or should it publish the children it _can_ and just leave some children un-published? I think that probably is too much of a risk of not noticing you haven't done what you intended to do and leaving some children unintentionally unpublished. 

[...]

PR #525, opened 2019-12-02
required fields to publish a record

jrochkind:
Ref #372


jrochkind:
So our default `work` factory for tests created a "published" work, but didn't include the now required fields for published works. So a bunch of tests failed on validation errors. 

I can:

A) Change the default `:work` factory to include the required-for-publication fields. 

B) Change the default `:work` factory to be a non-published work, add a `:published` trait for a published work, and change any existing tests that just do `create(:work)` to instead do `create(:work, :published)` if they need a published work. I'd find out which tests these are by simply changing the default work factory to be non-published, and looking at test failures. 

@eddierubeiz Any sense of whether A or B would be preferable? I'm not really sure.

eddierubeiz:
Hm. I would argue B is preferable. It's a little more work but we won't regret having that choice in the future. And I think the factory should yield unpublished works by default.

jrochkind:
OK, I guess we can merge and deploy this one now. 

But there's no rush, and it's some kinda weird code, so do please take a look at the code in a non-hurried fashion. :)

eddierubeiz:
Let's plan on merging / deploying this afternoon.

Issue #440, opened 2019-10-28
Post launch: remove Fedora check code from the codebase.


eddierubeiz:
In particular, we should remove `faraday` from the gemfile.

PR #527, opened 2019-12-02
Removing the fedora_check directory.

eddierubeiz:
Ref #440 
Looks like we are using Faraday in several other contexts, so we can't remove it from the Gemfile.


jrochkind:
> Looks like we are using Faraday in several other contexts, so we can't remove it from the Gemfile.

Can you expand on that? When I grep for Faraday in the project, I don't see it anywhere else, what are you seeing?

eddierubeiz:
You're right. It's nowhere else in the project once I remove the fedora checker code. It's also *not* in the `Gemfile` -- only in `Gemfile.lock`. I thought I had *added* it to the `Gemfile` as part of the fedora checker code, but that's apparently not the case. Not sure how to deal with Gemfile.lock in this case; Faraday is in there several times (e.g. as a dependency for `qa`). Suggestions welcome.

jrochkind:
So, I think you DID add it to the `Gemfile` as as part of fedora checker code -- a git history/blame would answer that. Oh, I see, it's not in Gemfile at all!

OK! So we don't need to remove it! 

So, it's in Gemfile.lock because it's an _indirect_ dependency -- some of our dependencies in our Gemfile themselves depend upon Faraday (or something that does). That's all fine. So it's in Gemfile.lock. If those dependencie stop depending on faraday, it'll disappear from Gemfile.lock.   (You can figure out exactly what dependency graph leads to faraday by examining the Gemfile.lock manually, not sure if there's a command to spit it out for you). 

When you were using it directly, you _should_ have added it to `Gemfile`, instead of relying on it being an indirect dependency -- so if it stopped being an indirect dependency, it'd still be available, and so you could say eg "We use Faraday 3.x, and the 4.x release might not be compatible, don't automatically upgrade it ever." And then it'd be appropriate to remove it when we removed our own use of it. 

But since we didn't add it in the first place... indeed nothing to be done with it here!

PR #532, opened 2019-12-03
Add ./bin/cap stub

jrochkind:
Now you can run `./bin/cap` which is basically equivalent but shorter to `bundle exec cap` -- it'll run the version of capistrano specified in your Gemfile.lock, even if you have different/newer versions installed locally.

Since the Capistrano deploy.rb currently locks to a _specific_ version of capistrano and complains and refuses to run if you are using a different one -- I was getting errors when I wound up with a newer capistrano on my system, and tried just running `cap whatever deploy`. but having to type `bundle exec` all the time is annoying, `./bin` is a little bit less annoying. So let's try this.

This ./bin/cap binstub was created with `bundle binstubs capistrano`


eddierubeiz:
Noted. Sounds good to me.

Issue #349, opened 2019-09-11
File extension missing from certain assets on original download.


eddierubeiz:
- Visit either of these two assets:
    - https://staging.kithe.sciencehistory.org/admin/asset_files/d791sg519
    - https://staging.kithe.sciencehistory.org/admin/asset_files/9593tv46x
- Click "Download Original"
Expected behavior: file is saved with a ".tif" or ".tiff" extension
Actual behavior: file name is the (very long) title of the asset.

On my laptop this means the downloaded file is not recognized as an image.

eddierubeiz:
Not sure if this is generalized; found out about it in the middle of doing something unrelated; further research needed.

jrochkind:
Ah, the download link on the ADMIN page hasn't been touched in a while, so that probably explains it. I'll take a look.

jrochkind:
Ah, the download link on the ADMIN page. Hasn't been touched in a while, was one of the first things written when the app hardly exists, not surprised it's buggy, although I'm surprised and curious about the _nature_ of the bug, I'm going to take a look. 

Good job finding.

jrochkind:
Ah, right, it's doing just:

    @asset.file.url(response_content_disposition: ContentDisposition.format(disposition: :attachment, filename: @asset.title))

Yep, that is setting the content-disposition filename to the title!

We could go through the DownloadsController like ordinary. 

Or should the admin download do something special that gives you the _direct_ download from S3 without even going through downloads controller? Should it even give you the filename as actually stored in S3, which is going to be something weird like `09983434.tiff`? Probably not.

jrochkind:
For now, use same logic as on public UI for download names from staff UI.

PR #534, opened 2019-12-03
Asset download filename fix

eddierubeiz:
Use `DownloadFilenameHelper.filename_for_asset` to calculate the proper download filename, and pass that to the `response_content_disposition` in the URL.
This means that e.g. `/admin/asset_files/0g354f48h` will result in a compact downloaded filename with the correct file extension on it instead of a rambling six-word title.
`Orig. Filename` metadata is still available for reference on the asset view page, by the way.

Ref #349

### DIFF
--- a/app/views/admin/assets/show.html.erb
+++ b/app/views/admin/assets/show.html.erb
@@ -24,7 +24,7 @@
 
     <% if @asset.file %>
       <%= link_to "Download Original",
-            @asset.file.url(response_content_disposition: ContentDisposition.format(disposition: :attachment, filename: @asset.title)),
+            download_path(@asset),
             class: "btn btn-outline-primary btn-lg mt-4" %>
       <% end %>
   </div>
--- a/app/views/admin/assets/show.html.erb
+++ b/app/views/admin/assets/show.html.erb
@@ -24,7 +24,7 @@
 
     <% if @asset.file %>
       <%= link_to "Download Original",
-            download_path(@asset),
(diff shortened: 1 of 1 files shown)


app/views/admin/assets/show.html.erb
```diff
@@ -24,7 +24,7 @@
 
     <% if @asset.file %>
       <%= link_to "Download Original",
-            @asset.file.url(response_content_disposition: ContentDisposition.format(disposition: :attachment, filename: @asset.title)),
+            download_path(@asset),
```
jrochkind: Hmm, this changes things to through the downloads controller now. So the URL is pointing to an app URL, which will then redirect you to the S3 location. 

While convenient, I was thinking that the admin download link ought to be _directly_ to s3 like it was before. To avoid any bugs or problems that might be in the downloads controller, and to be a little bit more 'raw' access for admins. 

Also because while presently we want the same content-disposition filename, that's sort of just an "accident", I don't think they should actually be tied together, it should be easy to change our mind about admin download content-disposition filename separate from public. 

So, I'm thinking the same code that was there before -- don't switch to pointing to our local downloads controller -- just where before it said `filename: @asset.title`, switch to `filename: x`, where `x` is whatever logic the downloads controller uses to figure out filename. Just re-use that filename-determining logic, instead of actually pointing to an in-app URL. 

Does that seem reasonable?

eddierubeiz: Absolutely reasonable. I hesitated about it, but I see your point Will do.

eddierubeiz: Fixed.

PR #538, opened 2019-12-04
Reset password instructions subject -- remove EXPERIMENTAL

eddierubeiz:
Not important but it bothered me so I'm fixing it.


jrochkind:
good catch thank you!

Issue #518, opened 2019-12-02
Add S3 backend file name to admin view


sanfordd:
Related to ticket #349, it would be helpful for future troubleshooting if the admin view provides the S3/backend file identification about the asset for being able to quickly check the item(s) directly in the backend, currently S3.

jrochkind:
To avoid confusion let's output the bucket name (can be found from shrine configuration), as well as the path on that bucket.

PR #541, opened 2019-12-04
Adding a link to the file in the s3 interface.

eddierubeiz:
I broke Jonathan's recipe for determining a file's URL *within* the s3 interface (basically a link to the bucket with a canned search to the file itself.) out of the `fixity_check` object, where it had lived before, and moved it into a service object.
The link is now available for all assets.
Does not display if the file doesn't exist, or if (like me) you're storing files locally.
Ref #518

### DIFF
--- a/app/models/fixity_check.rb
+++ b/app/models/fixity_check.rb
@@ -20,50 +20,7 @@ class FixityCheck < ApplicationRecord
     passed
   end
 
-  # A link to the checked_uri in the AWS S3 Console. Kind of hacky and fragile, will break
-  # if AWS changes their console URLs, or if our assumptions weren't quite right in other
-  # ways, but it's so clever and useful when it works!
   def checked_uri_in_s3_console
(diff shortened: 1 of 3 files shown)


jrochkind:
So this is still just filename, not complete path? Wait for another PR for complete path?

To get complete path, you'll have to make sure you're checking to see if the shrine S3 storage has a `prefix` set, and if so use it as a... prefix. 

Also, once we hae complete path, If we wanted to, we could make this one line on the screen using the pseudo-URL format I've seen used for S3 urls, like: `s3://bucketname/path/to/thing`

eddierubeiz:
I'm actually going to use this same PR for the complete path to the file in the console. Stay tuned please!

jrochkind:
Looks good. In an ideal world we might have a unit test or two for the new S3ConsoleURI object, but I'm honestly not really sure if it's worth it anyway. merging.

Issue #537, opened 2019-12-03
Add digital@sciencehistory.org to the recipients of failed fixity check notifications


eddierubeiz:

eddierubeiz:
`digital_tech_email_address` still needs to be added to `local_env.py` in Ansible; will do in early 2020.

sanfordd:
@eddierubeiz It's been added to ansible and should be in staging shortly for testing.

eddierubeiz:
A recent email went out with (among others) the following headers:
```
Date: Thu, 09 Jan 2020 11:19:52 -0500
From: no-reply@sciencehistory.org
To: digital-tech@sciencehistory.org, 
 digital@sciencehistory.org
Message-ID: <5e1752a823456_3e052ab0fe0e59642181c@scihist-digicoll-jobs1-production.mail>
Subject: FIXITY CHECK FAILURE: https://digital.sciencehistory.org,
 "b1057296x1_042.tif" (asset cz30pt621)
Mime-Version: 1.0
Return-Path: no-reply@sciencehistory.org
```
digital-tech@ received the email, but digital@ did not.

eddierubeiz:
Hypothesis: removing the carriage return in the To: field will result in both email messages receiving the mail.

eddierubeiz:
This is a duplicate of #576. Closing.

PR #544, opened 2019-12-05
Send fixity failure emails to digital@sciencehistory.org as well.

eddierubeiz:
If a `digital_email_address` is present in the local_env, send fixity failure emails to that email as well.
Otherwise, behave as before.
Ref #537


jrochkind:
@eddierubeiz Has that value already been added to local_env.yml, or does it still need to be?

eddierubeiz:
It still needs to be -- just added #546  to track that.

Issue #551, opened 2019-12-09
Facebook share icon on Work page broken


jrochkind:
I think browsers are blocking it now, although they didn't when we implemented. 

It's implemented as a `<a href="javascript:(code)">` link... maybe browsers don't allow `javascript:` anymore?

In chrome, address bar says `about:blank#blocked` after you click. 

In Firefox, it opens a new tab with the `javascript:...` code in the location bar, tab content empty.

jrochkind:
The link also has a `target="_blank"` on it which seems superfluous -- the inline JS code is already telling it to open a new window -- and also removing that seems to let chrome and FF unblock the link again. Don't entirely underestand what's going on, but if it works. 

The reason we're using `window.open` in the first place instead of just `target=_blank` is because a tab/window opened with `window.open`, the Facebook page that opens in there, when it tries to execute JS to close the tab/window... the browser lets it. Is what my comments in source say.

PR #552, opened 2019-12-10
remove superfluous target=_blank

jrochkind:
We are doing a javascript window.open anyway (for convoluted purposes involving letting Facebook JS close the window itself), so the target blank was really superfluous. And for some reason we don't entirely understand, triggers browser blocking, causing #551, which this seems to resolve.


eddierubeiz:
Thanks for figuring this out.

PR #555, opened 2019-12-10
Refactor DownloadDropdownDisplay a bit

jrochkind:
Ran into it when trying to use in redesign of audio work show page


eddierubeiz:
Thank you for un-hacking this and especially for adding a test.

PR #553, opened 2019-12-10
upgrade the yarn dependency serialize_javascript

jrochkind:
To resolve vulnerability warning flagged by github for older versions of serialize_javascript. https://github.com/advisories/GHSA-h9rv-jmmf-4pgx

We edited the package.json to requirement for newer version of `serialize_javascript` in the `resolutions` block, then ran `yarn install` to generate new yarn.lock.

PR #558, opened 2019-12-11
Revert "upgrade the yarn dependency serialize_javascript"

jrochkind:
This reverts commit 8e18a8cc01447b0cd70569babb995840dd928c72 and PR #553

533 did *not* work!

Now my local app is in the condition where it refuses to start, cause yarn complains:

```
yarn check v1.19.1
success Folder in sync.
Done in 0.11s.
yarn check v1.19.1
error "@rails/webpacker#compression-webpack-plugin#serialize-javascript" is wrong version: expected "^1.4.0", got "2.1.2"
error "@rails/webpacker#webpack#terser-webpack-plugin#serialize-javascript" is wrong version: expected "^1.7.0", got "2.1.2"
error Found 2 errors.
```

Not sure how we upgrade the vulnerable dependency, but apparently it wasn't that. yarn is driving me crazy!

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,7 @@
   "resolutions": {
     "lodash": ">= 4.17.13",
     "lodash.template": ">= 4.5.0",
-    "js-yaml": ">= 3.13.1",
-    "serialize-javascript": ">= 2.1.1"
+    "js-yaml": ">= 3.13.1"
   },
   "browserslist": [
     "defaults"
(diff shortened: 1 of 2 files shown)


jrochkind:
@eddierubeiz Merging this on my own, cause without it master seems to be broken for dev for me (apparently not for deploy to production, which I honestly can't explain why!)

PR #562, opened 2019-12-17
Add license

jrochkind:
Copied from
https://github.com/sciencehistory/chf-sufia/blob/master/LICENSE.txt

Assume we want same license in new repo.


jrochkind:
Approved by Chuck Hornstein (@CHF-IT-Staff ) in Slack.

Issue #549, opened 2019-12-09
R&R queue


eddierubeiz:
This ticket covers preliminary work on expanding the Digital Collections application to provide R&R queue functionality, currently provided through a Google form and spreadsheet.

eddierubeiz:
**Facts:**
- 3/4 of R&R requests for on-demand digitization result in the ingestion and cataloging of item(s) on the Digital Collections site.
- Currently, for those items, metadata needs to be re-entered into the digitization queue.
- There's significant overlap between the metadata stored on the current R&R spreadsheet with that in the digitization queue.
- When an item requested is marked "closed" in the R&R spreadsheet, it has the equivalent status of "Post-Production completed" in the digitization queue.

eddierubeiz:
**Lining up the metadata**

An R&R form would generate the following metadata:

- Curator or point of contact
- Collecting area (select one of the digitization queues)
- Patron name
- Patron email
- Deadline for digitization
- Ingest into digital collections (options: _yes_, _no_, _uncertain about copyright_)
- Scope (Pages to digitize for R&R only)
- Additional pages for Digital Collections
- Status (see below)
- Bibliographic metadata (OPAC link; External ID; Title; Title; Box/Folder/ Copyright status or holder)

eddierubeiz:
**Statuses for R&R requests**

- Awaiting digitization on cart
- Imaging in process
- Post-production completed
- Files sent to patron
- Closed as an R&R request

Notes:

- The following statuses (currently in the R&R spreadsheet) would not be needed in the new implementation: Batch Metadata Completed; Individual metadata completed; Hold; Re-pull object; Imaging completed; in post production.

eddierubeiz:
**When** an item's status is set to "Closed as an R&R request":
**If** its "Ingest into digital collections" field is set to **yes**, 
**Then** its status is equivalent to "Post-Production Completed" in the regular digitization queue.

It can be considered:
- removed from the R&R queue
- added to the regular digitization queue, ready for description and ingestion.

jrochkind:
Huh, the desire is that something that started out in the R&R queue (with it's different statuses and form fields) -- could sometimes _move_ to one of the other queues (with the original  statuses and form fields)?

That makes sense, but definitely makes the implementation a bit more complicated.

eddierubeiz:
Open questions:
Should closed R&R requests all be visible in one place?
[...]

PR #567, opened 2020-01-06
Rights and reproductions

eddierubeiz:
This is a reworked version, using a separate model (and DB table) of the work in branch `r_and_r_db_mods` (which is in github as a deleted branch.)
Ref #549


app/controllers/admin/r_and_r_items_controller.rb
```diff
@@ -0,0 +1,140 @@
+class Admin::RAndRItemsController < AdminController
```
jrochkind: Add a comment saying this is based off the other controller it's based off of

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
...
+        record.errors[attribute] << (options[:message] || "must start with b")
+      end
+    end
+  end
+  validates :bib_number, length: { is: 8 }, allow_blank: true, starts_with_b: true
+
+  # collecting areas could have been normalized as a separate table, but
+  # not really needed, we'll just leave it as a controlled string.
+  COLLECTING_AREAS = %w{archives photographs rare_books modern_library museum_objects museum_fine_art}
+  validates :collecting_area, inclusion: { in: COLLECTING_AREAS }
```
jrochkind: use some constant that already exists for collection areas in Admin::DigitizationQueueItem::COLLECTING_AREAS

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  STATUSES = %w{
+    awaiting_dig_on_cart imaging_in_process post_production_completed
+    files_sent_to_patron closed_r_and_r_request
```
jrochkind: should the status just be "closed" now that it's unique to an R&R table?

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
...
+   }
+
+  CURATORS = %w{ ashley hillary jim patrick molly other }
```
jrochkind: maybe just a text entry instead of a fixed list with 'other'?

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
...
+
+
+  before_validation do
+    if self.will_save_change_to_status? || (!self.persisted? && self.status_changed_at.blank?)
+      self.status_changed_at = Time.now
+    end
+  end
+
+  # Is this ready to make a DigitizationQueueItem out of?
+  def ready_to_move_to_digitization_queue
```
jrochkind: method name should end in question mark

app/views/admin/r_and_r_items/show.html.erb
```diff
@@ -0,0 +1,152 @@
...
+</div>
+
+
+<% if @admin_r_and_r_item.ready_to_move_to_digitization_queue %>
+  <div class="row mb-2">
+    <div class="col">
+      <%= link_to "Send to Digitization Queue", new_admin_digitization_queue_items_path(r_and_r_item: @admin_r_and_r_item.id, collecting_area: @admin_r_and_r_item.collecting_area), class: "btn btn-primary" %>
+    </div>
+  </div>
+<% end %>
```
jrochkind: 1. Should we try to enforce `ready_to_move_to_digitization_queue?`, or just let the user do it when appropriate

2. If we are enforcing it, guarding the display in the view is not enough -- the controller also has to refuse to do it if it's not allowed

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
...
+    return false if self.digitization_queue_item
+    return false unless self.is_destined_for_ingest
+    return false if self.copyright_research_still_needed
+    return false unless self.ready_to_move_to_digitization_queue_based_on_status
+    return true
+  end
+
+  def ready_to_move_to_digitization_queue_based_on_status
```
jrochkind: method name should end in question mark

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+    digitization_queue_item.title = self.title
+    # self.scope refers to the R&R request scope.
+    digitization_queue_item.scope = self.additional_pages_to_ingest
+
+  end
+
+  def digitization_queue_item
+    Admin::DigitizationQueueItem.find_by(r_and_r_item_id:self.id)
+  end
```
jrochkind: you shoudln't need to make a method like this manually, it should be a Rails association. 

Since you already have `belongs_to :digitization_queue_item, optional: true`, you don't need to define this method, the method already exists based on that `belongs_to`, you are overriding it here to be less efficient, just use the method Rails already gave you.

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def remove_reference_from_dq_table
+    my_dq_item =  self.digitization_queue_item
+    return unless my_dq_item
+    my_dq_item.r_and_r_item_id = nil
+    my_dq_item.save!
+  end
```
jrochkind: You don't need this method, or to manually call it in a before_destroy hook. 

Instead, when define `belongs_to :digitization_queue_item, optional: true`, you should say:

`belongs_to :digitization_queue_item, optional: true, dependent: :nullify` -- if it's not already doing this for you? I thought it did it automatically. 

But at any rate, look into how to get ActiveRecord `belongs_to` to do this for you -- instead of trying to re-invent it.

app/views/presenters/_digitization_queue_item_status_form.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= simple_form_for(view.model, data: {remote: true, auto_submit: true}) do |f| %>
   <div class="input-group input-group-sm flex-nowrap">
     <%= f.collection_select :status,
-          Admin::DigitizationQueueItem::STATUSES.collect {|s| [s.humanize, s]}, :second, :first, {},
+          view.model.class::STATUSES.collect {|s| [s.humanize, s]}, :second, :first, {},
```
jrochkind: `view.model.class::STATUSES` is a bit weird ruby syntax, I"m a bit surprised it works! Although it is clever. 

Another way to do this, would be just make a `def available_statuses` instance method on each model, and then you can just call `view.model.available_statuses`. 

Both ways, rely on both classes implementing a certain convention. But this one seems a bit clearer to me.

app/views/admin/digitization_queue_items/_form.html.erb
```diff
@@ -1,7 +1,12 @@
 <%= simple_form_for(@admin_digitization_queue_item) do |f| %>
+
+  <% if !@admin_digitization_queue_item.persisted? && @admin_digitization_queue_item.r_and_r_item_id %>
+    <%= hidden_field_tag "admin_digitization_queue_item[r_and_r_item_id]", @admin_digitization_queue_item.r_and_r_item_id  %>
+  <% end %>
```
jrochkind: can you put in a comment explaining why this is needed?

app/views/admin/r_and_r_items/_form.html.erb_bak
```diff
@@ -0,0 +1,93 @@
+<%= simple_form_for(@admin_r_and_r_item) do |f| %>
```
jrochkind: delete this file you didn't mean to commit it

app/views/admin/r_and_r_items/new.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>New R &amp; R Item</h1>
+<%= render 'form', admin_r_and_r_item: @admin_r_and_r_item,  :url => admin_create_r_and_r_item_path, :method => :post %>
```
jrochkind: Why do you need to pass in `:url => admin_create_r_and_r_item_path, :method => :post `?

method post should be the default anyway. `simple_form_for` should come up with the right url anyway. 

It's possible these args aren't being used at all and you can just remove them. If removing them does break things -- can we use conventional routes to not break things still without these args?

app/models/admin/r_and_r_item.rb
```diff
@@ -0,0 +1,100 @@
+class Admin::RAndRItem < ApplicationRecord
+  belongs_to :digitization_queue_item, optional: true
```
jrochkind: This allows multiple digitalization queue items for one R&R item. 

If we only want one-to-one, we need to use the rarely used Rails feature `has_one`. 

That means the foreign key could be on RAndRItem OR DigitizationQueueItem -- I'm not sure which is best but you can consider which you want to put it on.

jrochkind: Regardless of what the model says though --- you need to check your UI, is the "make digitzation queue item" button still available AFTER you've already done it once. Should it be?

config/routes.rb
```diff
@@ -178,16 +178,30 @@ def self.matches?(request)
...
+    # PATCH /admin/r_and_r_items/1/update
+    patch "/r_and_r_item/:id", to: "r_and_r_items#update", as: "update_r_and_r_item"
+    # GET /admin/r_and_r_items/new
+    get "/r_and_r_items/new", to: "r_and_r_items#new",  as: "new_r_and_r_item"
+    # POST  /admin/r_and_r_items/
+    post "/r_and_r_items", to: "r_and_r_items#create", as: "create_r_and_r_item"
+    # DELETE /admin/r_and_r_items/1
+    delete "/r_and_r_item/:id", to: "r_and_r_items#destroy", as: "delete_r_and_r_item"
+
+    # END R&R
```
jrochkind: Instead of listing each route individually, it might be better to use the Rails routing `resources` method for standard routes. Compare to what's being done for `resources :digitization_queue_items`. That might help with the issues with having to override simple_form_for for non-standard routing.

app/controllers/admin/r_and_r_items_controller.rb
eddierubeiz: Done.

app/models/admin/r_and_r_item.rb
eddierubeiz: Done. I could also wrap this in a method - not sure exactly if this works.

app/models/admin/r_and_r_item.rb
eddierubeiz: I will leave this as if for now, and prompt them when I demo.

app/views/admin/digitization_queue_items/_form.html.erb
eddierubeiz: Done.

app/views/admin/r_and_r_items/show.html.erb
eddierubeiz: Because I share your fear of one-to-one relationships, I've made some simple changes, and the relationship is many-to-many. You can now create more than one DQ item from an R&R item.

app/models/admin/r_and_r_item.rb
eddierubeiz: Stay tuned... gonna make an adjustment.

eddierubeiz: Never mind -- the way it is is fine. Was confusing this with STATUSES, which is actually used in a presenter; slightly different scenario.

app/views/presenters/_digitization_queue_item_status_form.html.erb
eddierubeiz: Good call. Done.

app/models/admin/r_and_r_item.rb
eddierubeiz: Yes, done!

app/models/admin/r_and_r_item.rb
eddierubeiz: Done.

app/models/admin/r_and_r_item.rb
eddierubeiz: Done!

app/models/admin/r_and_r_item.rb
eddierubeiz: Done!

app/views/admin/r_and_r_items/_form.html.erb_bak
eddierubeiz: Done.

app/views/admin/r_and_r_items/new.html.erb
eddierubeiz: Done! They were indeed not needed.

app/models/admin/r_and_r_item.rb
eddierubeiz: I've decided to go with a one-to-many relationship, per our earlier discussion. The button remains available after a new dq item is created.

config/routes.rb
eddierubeiz: Done.

app/models/admin/r_and_r_item.rb
eddierubeiz: Done!

PR #571, opened 2020-01-08
Routine to copy a work from staging to dev

jrochkind:
cap staging copy_work[$work_friendlier_id]

Note that to test, this branch needs to be deployed to staging, so the rake task is there on staging for the cap task to call on staging.


jrochkind:
* It copies postgres metadata
* it copies originals (by issuing an S3 "copy" command from bucket to bucket, via shrine doing that automatically)
* it _locally recreates_ all derivatives. 

Each step is progressively slower. It can take quite a while to copy a large work. This is a first try at it, there are things we can try to do to change how the task approaches things to improve performance. 

It does not currently copy collection membership. 

The code is kind of hacky, has to include several workarounds -- including turning off postgres referential integrity checking for the import, which is kind of ugly, but it's one way of getting it to work.

jrochkind:
OK, this now has concurrency, so it will issue the S3 "COPY" operations in parallel, instead of waiting for S3 to confirm them each one by one. 

In my tests, I can copy ramelli down to staging in about 5 minutes. Not bad!

The concurrency *might* be disastrous if you are using local file system storage; not sure. It might also be okay (just prob a lot slower than copying to S3). Or it could destroy your disk trying to save 50 files at once to it!

I didn't go as much concurrency as I could, cause concurrency is scary, there are some limits (50 parallel, and it won't let the queue of things waiting for a parallel worker to grow unbounded before stopping to wait for it to empty). 

@eddierubeiz This is ready for review and testing and merge. Also no hurry on it, you can wait until you are done or at a nice break with R&R (since this should prob be deployed to staging to test fully). But no hurry at all, R&R (and stakeholder convenience testing of it) can be first priority.

app/models/copy_staging/restore_work.rb
```diff
@@ -0,0 +1,168 @@
...
+            model = model_class.new(attributes)
+            model.save!
+
+            if model_class <= Kithe::Asset
+              restore_asset_file(model)
+            elsif model_class <= Kithe::Derivative
+              restore_derivative_file(model)
+            end
+
+            # let's keep from getting too far in front of the thread pool --
```
eddierubeiz: Clever.

eddierubeiz:
The s3-to-s3 copy is ... extremely fast.

As for copying down to my local dev setup, it took about 2 seconds per page for a sample 200-page book, whereof each page contained roughly a 60MB original and all its derivatives.

Issue #580, opened 2020-01-16
[scihist_digicoll/production] ActionView::Template::Error: undefined method `start_with?' for nil:NilClass


jrochkind:
## Backtrace

    line 70 of [PROJECT_ROOT]/app/decorators/audio_work_show_decorator.rb: link_to_non_audio_member
    line 68 of [PROJECT_ROOT]/app/views/works/show_with_audio.html.erb: block in _app_views_works_show_with_audio_html_erb___58343582107713764_47082047641260
    line 64 of [PROJECT_ROOT]/app/views/works/show_with_audio.html.erb: each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/59430164)

jrochkind:
Looks like an actual bug from honeybadger, we should fix it. perhaps happens when something hasn't been fully ingested yet and doesn't have a content_type attached, so may be temporary and not fully reproducible from the data on the HB error, but needs to be investigated.

PR #581, opened 2020-01-17
On audio pages, make PDFs downloadable; all other non-audio files should not be linked.

eddierubeiz:
Instead of testing for an image, let's just explicitly check to see if the file is a PDF when deciding whether to create a link or not should fix ref #580 .  If it's not (or we don't have a content_type), don't link to it.


jrochkind:
To avoid the null reference exception, we don't need to change the logic, we can just change it to be safe for null references. 

I think we should leave the logic alone -- for instance, what if it's an audio file?  That's handled differently in your change vs original, I'm not sure if it should be.  

I think checking if it's an image is the correct logic, and we should not change the logic, instead just do it safely for null reference to fix the bug we are presented with. 

`member.content_type && member.content_type.start_with?("image/")`

or using the newer ruby "safe navigation operator":

`member.content_type&.start_with?("image/")`

What do you think?

PR #586, opened 2020-01-21
update capistrano to latest

jrochkind:
$ bundle update capistrano

Just cause why not, and its an inconvenience where I have to use `bundle exec cap` for everything if I have a more recent cap installed otherwise.


jrochkind:
@eddierubeiz ping?

jrochkind:
oops, had a conflict. 

Redone on master. 

@eddierubeiz let's get this merged before it has a conflict again. :)

eddierubeiz:
Thanks for fixing the conflict. WIll merge as soon as it passes.

Issue #584, opened 2020-01-21
Some UI (including viewer) may not work well on mobile


jrochkind:
When using the site on my Android phone, I noticed there seemed to be some problems, especially with viewer. Things were not fitting on page correctly, the multi-page scroll area at bottom was off screen and hard ot use etc. 

We can investigate and fix viewer specifically, and also do a survey for any other problems. 

We can use an actual phone a developer might have; Chrome's screen-size simulation; or browserstack.com that we now have a free account to. 

There are some ways to get browserstack access to your dev copy via a proxy for easier debugging. https://ngrok.com/ is one of them.

jrochkind:
Some things noticed in quick review in my actual Android phone:

* On single item page, there are visible "•" bullets next to metadata that is not expected. 
* viewer does seem to have problems
   * on viewer for single-page, works okay, but bottom toolbar is entirely missing
   * on viewer for a multi-page item, toolbar IS there, although background colors are not filled in properly
      * But additionally, the actual horizontal line of pages on bottom has the android navigation bar on top of it, covering up part of it, making it kind of unusable. 

Not immediateley seeing any other problems. 

Will investigate using Chrome dev tools for simulating small screen, as well as browserstack.com.

jrochkind:
No need for ngrok -- browserstack has a preference "resolve all URLs through my network: Enabling this feature makes testing websites slower" -- which lets me connect to via browserstack to my dev copy on `localhost` -- amazing!

PR #588, opened 2020-01-21
btn-group necessary in download-dropdown-display after all

jrochkind:
Tried to remove it in 6ea353eb. But that messed up viewer display, not just in small screens but in all screens.

Screenshot 2020-01-21 15 41 06

Hopefully putting it back won't break popper positioning as suggested in the commit log for 6ea353eb -- i think we're good.

Ref #584 although it turns out it wasn't just small screens!


eddierubeiz:
Thanks for identifying and fixing.

PR #590, opened 2020-01-21
Android viewer height correction

jrochkind:
Wasn't correct due to "100vh" not working as expected on Android (and possibly other browsers).
https://stackoverflow.com/questions/52848856/100vh-height-when-address-bar-is-shown-chrome-mobile

This hack seems to fix it, although it's a hack on top of a hack... tested in a variety of browsers (thanks browserstack.com), and seems to be good on all of em.

Ref #584


eddierubeiz:
Looks fine to me. Thanks for testing in Browserstack.

Issue #302, opened 2019-08-20
separate ActiveJob worker queues, to keep large on-demand derivatives in their own queue


jrochkind:
@sanfordd says in #292 

> keeping the on-demand queue distinct from the ingest queue also has the benefit of avoiding interruptions. While any one ingest job is fast, if I throw a book in just before someone starts a on-demand file it can get delayed.

Would actually keep large ingests from interfering with on-demand derivatives, as well as many on-demand derivative requests from interfering with ingests. 

"Ingests" an "on-demand derivatives" are the main categories of our BG jobs here, but making a separate queue for on-demand-derivatives, it'd really be "on-demand-derivatives" and "everything else". 

Requires having the separate groups of workers processing the separate queues, as well as making sure jobs get assigned to correct queues when enqueued.

sanfordd:
Adding here, from #292 that we don't want more than 2 on-demand workers, to avoid potential crashes from pdf generation.

sanfordd:
Breaking out the workers into distinct queues is fairly simple (https://github.com/nevans/resque-pool#yaml-file-config)

We'll need to tell our existing tasks what queues to use though once we go from a single pool
"*": 10
to breaking the pools apart into groups like
staff: 10
public: 2

I'm not sure if our tasks or logic support it, but if we can distinguish pdf on-demand requests easily from others we could further divide pools. PDF generation is the only one that's highly RAM intensive on large works, ZIP creation is fairly light and doesn't need to be quite as limited. If they're harder to distinguish then just staff/public seems like a fine divide for now until we have more public facing worker tasks.

jrochkind:
Ideally it would be nice to say some set of workers handle queue X, and other set of workers handle "all queues EXCEPT for X". 

Sadly, I don't believe that is possibe in resque/resque-pool. I thought I saw an unmerged PR to resque-pool that would have provided it, but now I can't find it. 

We can look back at how we handled this in chf_sufia. 

I am not certain if it's easy to separate out PDF from ZIP, I think they are currently both the same ActiveJob, they'd probably have to be split into separate ones to put em in separate queues. 

Rather than "staff" and "public", I'd probably just leave it "default" and "public-on-demand-derivs" or something. There are all sorts of things that can end up in bg jobs, including things that may not be strictly divided into staff or public; right now, I *think* everything is in the single default queue. The redis admin can show us what queues any actually existed jobs have been put in currently.

sanfordd:
[...]

PR #594, opened 2020-01-22
update resque-pool and resque

jrochkind:
Resque will now be 2.0. Done by changing Gemfile to allow resque 2.x, and then:

    bundle update resque-pool resqu3e

So other dependencies of resque/resque_pool are updated too.

Let's got on the most recent version of resque/resque-pool, including resque 2.0, before enhancing the setup of our resque workers per #302


jrochkind:
Tested on staging with an ingest, seemed to work fine. 

@eddierubeiz I'm going to go ahead and merge this in your absence, so I have it in master as a starting place.

PR #595, opened 2020-01-22
Separate ActiveJob worker pools

jrochkind:
Instead of having 10 workers set up, we set up two pools of 10 and 2 workers each. Right now we have one jobs server, but if we had multiple this would be true on each of them.

The 2-worker pool is working ONLY the `on_demand_derivatives` queue -- which are on-demand derivatives job is now set to use.

The 1o-worker pool is working first priority `mailers` queue, then the `default` queue which includes all other work we know about -- mostly ingests. At the moment, we aren't actually sending any email using mailers in the bg, so the mailers queue won't have any work on it. But preemptively setting it up in case we do so it will keep working. Rails will use the `mailers` queue for any emails sent "later", and our workers will preferentially handle those (fairly quick) tasks when there are some waiting, before handling any other work such as (much slower) ingest-related work.

After deploying this, we should keep an eye on the resque admin, to make sure there isn't any work piling up in OTHER queues that we DON'T have workers working. Previously our workers were working "any" queue, but that is not possible with resque configuration options available and the desired two-separate-pool setup.

Ref #302

### DIFF
--- a/app/jobs/on_demand_derivative_creator_job.rb
+++ b/app/jobs/on_demand_derivative_creator_job.rb
@@ -1,5 +1,7 @@
 # placeholder
 class OnDemandDerivativeCreatorJob < ApplicationJob
+  queue_as :on_demand_derivatives
+
   def perform(work, derivative_type)
     OnDemandDerivativeCreator.new(work, derivative_type: derivative_type).attach_derivative!
   end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -62,7 +62,12 @@ Rails.application.configure do
 
   # Use a real queuing backend for Active Job (and separate queues per environment)
   config.active_job.queue_adapter     = :resque
-  config.active_job.queue_name_prefix = "scihist_digicoll_#{Rails.env}"
+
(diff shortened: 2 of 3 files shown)


jrochkind:
Also set `config.active_job.queue_name_prefix = nil`

jrochkind:
Deployed to staging and tested -- everything seemed to go well. 12 total workers. Did an ingest, it's bg jobs were handled, by appropriate workers. Did a on-demand PDF generation, it's bg job was handled, by one of the 2 appropriate workers as well.

sanfordd:
I threw a set of large items from the collection as simultaneous PDF requests 
Museum Wormianum
Le diverse et artificiose machine del capitano Agostino Ramelli
Neft︠i︡ana︠i︡a promyshlennostʹ v s︠i︡evero-amerikanskom shtat︠i︡e Pensilʹvanīi i na Kavkaze
and some smaller ones
The Mineral Kingdom
Musaeum Hermeticum Reformatum et Amplificatum 

We stayed around a max of 6 GB of RAM used and did not generate any OOM errors.

Issue #572, opened 2020-01-08
PA Digital reports problems with OAI-PMH


jrochkind:
> We are planning to fully reprocess all contributors in the new aggregator, and Science History is scheduled to be reprocessed in February so we can explore more on this soon. In the meantime, we did a quick test run and found an issue with the resumption tokens. Here is the error message we got:

    <?xml version="1.0" encoding="UTF-8"?><OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
    <responseDate>2020-01-08T15:44:00Z</responseDate><request>https://digital.sciencehistory.org/oai</request>
    <error code="badResumptionToken">The value of the resumptionToken argument is invalid or expired.</error>
    </OAI-PMH>

> I'm copying Leanne Finnigan here who can help with questions you may have about this. If this issue can be resolved, we'll continue to do a full reprocessing in February and reach out with any additional issues.

jrochkind:
Sometime in February is when next PA Digital harvest is.

jrochkind:
Tried to reproduce, I believe I am reproducing the problem. 

First, do a request for:

    https://digital.sciencehistory.org/oai?verb=ListRecords&metadataPrefix=oai_dc

The response includes a 'resumptionToken':

 ```
<resumptionToken>
oai_dc.f(2016-05-24T16:57:17Z).u(2020-01-14T20:21:48Z):08612p06f
</resumptionToken>
```

Now we try to do another request with that resumptionToken (properly URI encoded) to get the next page. 

    https://digital.sciencehistory.org/oai?verb=ListRecords&oai_dc.f%282016-05-24T16%3A57%3A17Z%29.u%282020-01-14T20%3A21%3A48Z%29%3A08612p06f

Indeed we get an error message from our server:

```
<error code="badArgument">
The request includes illegal arguments, is missing required arguments, includes a repeated argument, or values for arguments have an illegal syntax.
</error>
```

This is frustrating, it must be a bug of some kind in the [oai](https://github.com/code4lib/ruby-oai) gem. 

[...]

PR #601, opened 2020-01-27
use thumb_large_2X for oai pmh thumbnail

jrochkind:
PDFs didn't have a download_large. But should have thumb. It's only a bit smaller than download_large. Ref #572

Added tests in this PR, but they didn't actually go red to catch the bug. Oh well, good tests anyway, good enough.

### DIFF
--- a/app/serializers/work_oai_dc_serialization.rb
+++ b/app/serializers/work_oai_dc_serialization.rb
@@ -143,9 +143,9 @@ class WorkOaiDcSerialization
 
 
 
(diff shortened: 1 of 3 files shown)


app/serializers/work_oai_dc_serialization.rb
```diff
@@ -218,8 +218,11 @@ def publishers
   # We are hand-building the URL instead of using Rails route helper to further improve
```
eddierubeiz: Noted!

PR #583, opened 2020-01-21
[WIP] Temporarily remove transactions & batching in staging.

eddierubeiz:
Please do not merge.

Temporarily removing transactions and batching in staging, to help diagnose the bug described in #498  .

Issue #498, opened 2019-11-18
Fixity checks happening more often than anticipated


jrochkind:
Our intent with the automatic nightly fixity checking code, is to set it up so it fixity checks 1/7th of all assets every night, in such a way that after 7 days they've all been checked -- every asset gets checked once every 7 days. 

However, the fixity checks seem to be happening _more_ frequently. 

Our [fixity health report page](https://digital.sciencehistory.org/admin/fixity_report) gives us the "Fully-ingested Asset with oldest most-recent fixity check" -- we expect that to be 7 days old, but looking at it now on 18 Nov 2019, it says 17 Nov 2019, only two days ago.  (It's at least possible the query to look up this statistic is wrong). 

As another example, for asset [ft848s025](https://digital.sciencehistory.org/admin/asset_files/ft848s025), it had a fixity check failure on 16 Nov 2019. Eddie re-ran a fixity check that same day, which then passed. (Mystery why it temporary failed, that's #464). Then the automatic fixity check on 18 Nov -- choose to fixity check it again, even though it had a fresh check from only 2 days ago. Not sure why. 

Logic for choosing what assets to fixity check nightly might be wrong -- although it's got a SQL `limit`in there, so not sure how it could be fixity checking more things per run. 

Alternately, perhaps the task for "fixity check 1/7th of things" is being run _more than once per night_, perhaps on more than one machine? 

We might need more logging to figure out what's going on, logging as to exactly what's being checked when by whom we don't have now. But keep in mind that if the fixity check is being run on a *host* we don't expect, if we were just logging to local file system, we'd never *see* those logs on a host we're not thinking of looking at. So might need some kind of loggable visibility in a manner other than just writing to local file system on host where activity occurs. (Emails or slack messages?)

sanfordd:
I'm switching tracks, but I want to make a couple quick notes for myself or the next person to check this for what I've been able to rule out so far.

- There's definitely only one cron role server in production or staging, so it's not another cron server anywhere.
  - The other cron jobs don't seem to be running multiple times. I took a look at sitemap on S3 to confirm we're not seeing 2 made in the same day. So that rules out a non-cron tagged server running tasks.
- Manually running the fixity check correctly handles item counts
- Our current log level isn't dumping any details about the work being done by the fixity task, we might want to bump staging up to debug?
- My stopping point was checking if somehow the CYCLE_LENGTH env variable was getting lost or set to 0 somehow when it was run as a cron-like job.

[...]

PR #605, opened 2020-01-28
AssetsNeedingFixityChecks avoid find_each

jrochkind:
Accomplish the same thing more manually -- fetching in batches of 1000, but still yielding in a single stream.

We think find_each race condition may have been responsible for #498, and this may solve it.

@eddierubeiz 's debugging in #583 was crucial for figuring out what was going wrong and coming up with possible solution. Possibly we're still being too clever -- but I just didn't want to give up on avoiding fetching al objects or IDs into memory, to have an implementation that should be able to scale to orders of magnitude more assets no problem.

### DIFF
--- a/app/lib/scihist_digicoll/assets_needing_fixity_checks.rb
+++ b/app/lib/scihist_digicoll/assets_needing_fixity_checks.rb
@@ -19,6 +19,7 @@ module ScihistDigicoll
   # Note: if you pass in 0 as the cycle length, you just get all the assets.
   #
   class AssetsNeedingFixityChecks
+    BATCH_FETCH_SIZE = 1000
     DEFAULT_PERIOD_IN_DAYS  = 7 # in days
     attr_reader :cycle_length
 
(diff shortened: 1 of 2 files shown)


jrochkind:
i did do a test of this in a manually-created crontab with a really high `CYCLE_LENGTH` (5000) -- it seemed to work, and only check the expected (in this case) 5 assets. 

I will leave this branch on staging overnight, so we can see what it does with the overnight cron. 

I think I have an understanding of how the previous code could lead to checking all the assets. I *don't* understand why it would be only reproducible in a cronjob! But I guess race conditions can be mysterious. 

Basically, Rails `find_each` tries to execute your query with a batch size limit (by default 1000).  then, to get the rest, it executes the *same query*, _starting from where it left off_, with another batch limit. Until it's done. "starting from where it left off" means sorting by primary key (in our case uuid), and starting `>` last encountered PK. 

So find_each is imcompatible with a custom sort, but we tried to use an inner sub-query to be really clever, and still sort by "most in need of a fixity check" -- which was confusing to implement in the first place because it required a join and `group by`. 

But the `find_each` docs warn:

> NOTE: By its nature, batch processing is subject to race conditions if other processes are modifying the database.

In particular, if they are modifying the database in such a way that the query under `find_each` returns a different value on each iteration. And indeed, adding new fixity check records will change the list of "X records most in need of a fixity check cause they have stalest lack check".  So every time `find_each` went back to get the "next batch of records most in need of fixity check", the overall list of "records most in need of fixity check" would change to be a NEW list, which hadn't been exhausted yet... and it could basically never be done... I'm actually kind of surprised it wasn't infinite looping.... which would have been even harder to diagnose!

I can't explain why this would only happen in a cronjob though. I might have the diagnosis not quite right. 

But `find_each` is clearly dangerous in these sorts of conditions, and, I dunno, this way seems to work?

jrochkind:
I might have broken `CYCLE_LENGTH=0` -- that might have actually been broken before too?

jrochkind:
This did seem to fix it on staging:

```
I, [2020-01-29T02:30:03.254812 #10513]  INFO -- : checking asset fixity (task ID 735) for 3744 of 26214 assets
I, [2020-01-29T03:55:22.784151 #10513]  INFO -- : Finished checking asset fixity for 3744 of 26214 assets. The task (ID 735 ) took
 5119.515423673 seconds
```

The new logic leaves me a bit less confident that it will check the *right* resources. 

The intended logic is:  

* Let's say we wanted to check 3000 of the most-stale assets tonight
* We check the 1000 most stale assets
* We again check the 1000 most stale assets (differnet list, cause the first 1000 are no longer most stale)
* We again check the 1000 most stale assets. 

Get it?  This should work -- assuming it's not buggy. But this is kind of confusing stuff. (I still wonder if there's some totally simpler way to do this, that would still be performant -- would probably involve storing more task-state in the database, making new db table; simpler in some ways, more moving parts in others). 

I think we won't really know for sure unless we leave it deployed for a week or two, and check distribution of latest fixity check dates to make sure it meets expectations. 

I think we can probably go ahead and deploy to production and then just check on it -- fixity check every week is kind of more than we really need anyhow, it's no big deal if they aren't being done as expected, so long as we catch it in a few weeks. 

But we should check with stakeholders at a weekly check-in to confirm before deploying.

eddierubeiz:
We do have `assets_needing_fixity_checks_spec.rb` file. The clever new check you added makes it more likely it'll do what it's supposed to now.

PR #615, opened 2020-02-03
[WIP] Comments for r and r items

eddierubeiz:
Allow comments for R & R items. This will closely follow the pattern of the existing comments on digitization queue items. A new table will not be necessary; we will just allow comments to be attached to *either* R&R items *or* dq items.

### DIFF
--- a/app/controllers/admin/r_and_r_items_controller.rb
+++ b/app/controllers/admin/r_and_r_items_controller.rb
@@ -1,6 +1,6 @@
 # This controller is based on Admin::DigitizationQueueItemsController.rb
 class Admin::RAndRItemsController < AdminController
-  before_action :set_admin_r_and_r_item, only: [:show, :edit, :update, :destroy]
+  before_action :set_admin_r_and_r_item, only: [:show, :edit, :update, :destroy, :add_comment, :delete_comment]
 
   # GET /admin/r_and_r_items
   # GET /admin/r_and_r_items.json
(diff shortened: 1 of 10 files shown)


jrochkind:
> A new table will not be necessary; we will just allow comments to be attached to either R&R items or dq items.

I believe that would require using Rails "polymorphic assocation" feature. it's a bit of an advanced feature, but I think mostly works. It requires adding a "model name" column of some kind next to the fk column. I haven't done this in a while/ever, we'd have to look up the details. 

A separate comments table might be more straightforward, once we've decided on a separate table for the R&R things as a whole

jrochkind:
Just did a quick skim, but looks fine. 

As discussed, we know that the dual alternate fk/associations on the comment model/table:

     belongs_to :digitization_queue_item, optional:true
     belongs_to :r_and_r_item, optional:true

Is not ideal and is a weird non-normalized schema, to have these two paralell fks and expect one and only one of them to be set.  But still seemed to be the least bad option at this time, we'll try it out and see. 

Validations intend to insist that one and only one of these must be set:

      # A comment must be EITHER on an r_and_r_item,
      # XOR on a digitization_queue_item
      # (but can't and shouldn't be on both).
      validates :digitization_queue_item_id, presence: true, unless: :r_and_r_item_id
      validates :r_and_r_item_id, presence: true, unless: :digitization_queue_item_id

Issue #603, opened 2020-01-28
Display "location" in list view on digitization queue


HKativa:
Please add "location" to the fields displayed on the list view of the digitization queue for the following collecting areas:

-  Modern library
- Rare books

jrochkind:
Is it okay if we add it to the list view for all collecting areas? We don't yet have collecting-area-specific logic in Digitization Queue like that, and I'd like to avoid doing it unless we really need it.

HKativa:
Sure, that's not a problem. It's really essential for those two collecting areas and often will be blank in the case of the other collecting areas anyway.

jrochkind:
@HKativa Does this need attention on the R&R queue too, or just on the existing "Digitization Queue" pages?

If I don't hear otherwise, I'll assume just Digitization Queue pages. And as discussed, will put it on all collecting areas, since you said that was acceptable.

HKativa:
@jrochkind we would like to add "location" to the list view on the R&R queue as well. That's on my list of tweaks/feedback for @eddierubeiz.

HKativa:
Also, as discussed at the tech check-in on 2/3/20202, please break out the IDs (bib number, accession number, and object ID(s)) into separate columns, a la the R&R queue.

HKativa:
@jrochkind I'm not seeing "location" on the list view for the various collecting areas in the digitization queue. (I do see "location" on the list view on the R&R queue). I do see that the various IDs were separated out - thanks!

jrochkind:
@HKativa Odd, indeed it isn't, i recall doing it but maybe it slipped through the cracks, nice catch thanks, I'll look into it.

eddierubeiz:
@HKativa  -- deploying this feature to production now.

PR #618, opened 2020-02-03
break out the IDs into separate columns on dig queue index list

jrochkind:
ref #603 https://github.com/sciencehistory/scihist_digicoll/issues/603#issuecomment-581512956

Tried to size the columns such that there would be enough space for each, based on trying to find sample data for each type of ID from production.

No fancy getting rid of columns on small screen, you'll just get horizontal scroll when too small.

### DIFF
--- a/app/views/admin/digitization_queue_items/index.html.erb
+++ b/app/views/admin/digitization_queue_items/index.html.erb
@@ -48,12 +48,14 @@
 
 
 <div class="table-responsive">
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Looks good. It occurs to me horizontal scroll is perfectly fine on this particular screen:
a) patrons never see this screen, and
b) the admin users are using wide monitors anyway.

PR #620, opened 2020-02-04
R&R adjustments

eddierubeiz:
Adding some more tweaks based on feedback from Hillary and Nicole:

- Add "location" to the list view 

- Drop-down menu for "Curator" field on R&R edit form: Remove "Other." R&R requests always involve a named curator who is responsible for managing a specific collection.

- On the show page : add status menu/save feature

- When an R&R item is sent to the Digitization Queue: set status to "post production completed?"

- Removing too-clever functionality to hide or disable the button to send items to the DQ. It wasn't helping that much and conflicted with the new dropdown menu above.

### DIFF
--- a/app/views/admin/r_and_r_items/index.html.erb
+++ b/app/views/admin/r_and_r_items/index.html.erb
@@ -74,7 +74,7 @@
   <table class="table">
     <thead>
       <tr>
-        <th>Title and notes</th>
+        <th scope="col" class="w-25 p-3">Title and notes</th>
         <th scope="col" class="d-none d-sm-table-cell" >Status</th>
         <th scope="col" class="d-none d-sm-table-cell" >Deadline</th>
         <th scope="col" class="d-none d-sm-table-cell" >Collecting area</th>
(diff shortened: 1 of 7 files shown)


app/views/admin/r_and_r_items/index.html.erb
```diff
@@ -81,6 +81,7 @@
         <th scope="col" class="d-none d-md-table-cell" >Bib number</th>
         <th scope="col" class="d-none d-md-table-cell" >Accession number</th>
         <th scope="col" class="d-none d-md-table-cell" >Object ID</th>
+        <th scope="col" class="d-none d-md-table-cell" >Location </th>
```
jrochkind: There's room in the table for "Location", you tested this?

On Dig Queue Item, I put location under "title" instead, because I thought location can get quite long. 

But I guess R&R and Dig Queue list view just continue to diverge, I guess that's just what happens when they aren't DRY. 

In general, have you tested the CSS/html for this table with what happens when values get lengthy?  Make sure it handles them okay without totally falling apart?

eddierubeiz: Good call. I limited the width for the title and notes field, which can get long. We can always adjust further once we have some real values in the table, of course. But the metadata columns fit just fine. They get hidden on narrow screen widths of course, but this part of the admin won't be used on narrow screen widths, in practice.

eddierubeiz:
Also made "Curator" a free-text field.

Issue #624, opened 2020-02-06
Changing status and saving feature not working


yonyitz:
When changing the status of a digitization queue item and clicking the save icon, status does not change. The status should update to status selected after clicking save, and this is not functioning properly.  I attempted to change/save the status of an item on the [modern library queue page](https://digital.sciencehistory.org/admin/digitization_queue_items/modern_library) as well as the page of the [queue item](https://digital.sciencehistory.org/admin/digitization_queue_items/45).

yonyitz:
I keep trying to select "closed" and it keeps reverting back to "batch metadata completed" which is the old status.

jrochkind:
I think I did reproduce on staging, it does seem to be just plain broken. 

I think maybe that menu is doing nothing at all, no JS is executing at all for it. Not sure why haven't debugged more than that.

eddierubeiz:
The code to fix issue https://github.com/sciencehistory/scihist_digicoll/issues/604 changed the rules we use to determine if a DQ item is "valid". Under the new rules, seven items were invalid -- they were just missing a "Location".

https://digital.sciencehistory.org/admin/digitization_queue_items/38
https://digital.sciencehistory.org/admin/digitization_queue_items/39
https://digital.sciencehistory.org/admin/digitization_queue_items/40
https://digital.sciencehistory.org/admin/digitization_queue_items/41
https://digital.sciencehistory.org/admin/digitization_queue_items/43
https://digital.sciencehistory.org/admin/digitization_queue_items/44
https://digital.sciencehistory.org/admin/digitization_queue_items/45

eddierubeiz:
Meanwhile, when you update a DQ item via the dropdown menu, if the item is not deemed "valid" by the underlying code, the item is left unchanged, and no error message is shown to the user.

jrochkind:
Nice job figuring that out. 

So we'llwant a better error condition for an invalid one -- it shouldn't just completely fail silently, it should at least give you some kind of error message.  For invalid or ANY failure to save. 

And of course we'll have to make those valid (which can be easily done just by entering "not entered" or something). 

Also shows the danger of this kind of validation in Rails, that it's universal -- this is why some people don't like model-level validations!

eddierubeiz:
Yup. See PR 625 for a rough draft of how we could do that. (Albeit I'm not a huge fan of rails-ujs based on the docs I saw today.)

PR #625, opened 2020-02-07
Error handling for Digitization Queue status menu

eddierubeiz:
By requesting that the update method in the controller return json (instead of JS) we fix a bug that was causing the front end to always think the update had succeeded, even when it hadn't.

I also solved another smaller problem (only apparent once the main bug was fixed) that was incorrectly determining when to run the ajax listener code.

Ref #624

### DIFF
--- a/app/javascript/src/js/admin/queue_status_submit.js
+++ b/app/javascript/src/js/admin/queue_status_submit.js
@@ -19,7 +19,7 @@ import domready from 'domready';
 
 domready(function() {
   function notOurForm(form) {
-    return !(form && form["data-auto-submit"] == "true");
+    return !(form && form.dataset['dqStatusForm'] === 'true');
   }
 
 
--- a/app/javascript/src/js/admin/queue_status_submit.js
+++ b/app/javascript/src/js/admin/queue_status_submit.js
@@ -15,6 +15,8 @@
 // Our actual HTML uses the poorly documented rails-ujs "data-disable-with" feature to
 // have a progress spinner and disable form when in proress.
 
+var eddie_event;
+
 import domready from 'domready';
 
 domready(function() {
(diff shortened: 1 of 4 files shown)


app/javascript/src/js/admin/queue_status_submit.js
```diff
@@ -19,7 +19,7 @@ import domready from 'domready';
 
 domready(function() {
   function notOurForm(form) {
-    return !(form && form["data-auto-submit"] == "true");
```
eddierubeiz: form["data-auto-submit"] wasn't a good expression for getting that data attribute. I just created a new one just for the purposes of identifying the form.

jrochkind: I don't understand what made it not a good expression. Can you say more?

jrochkind:
I don't understand how/why this solves the problem. It's very mysterious. 

I think we already are using rails-ujs here? If so, I think figuring out the right way to do error handling with rails-ujs is the way to go, rather than a not-well-understood hacky workaround. 

I can look into error handling with rails-ujs auto-submit, or you can?

app/javascript/src/js/admin/queue_status_submit.js
```diff
@@ -49,9 +49,11 @@ domready(function() {
     //debugger;
     // And warn the user, with info for developer in console
     var detail = event.detail;
-    var data = detail[0], status = detail[1], xhr = detail[2];
-    console.error("DigitizationQueueItem AJAX status change failed\n\n" + status + " " + data);
-    alert("Uh oh, digitization Queue Item status change failed!");
+    var data = JSON.stringify(detail[0], null, 2);
```
jrochkind: It's coming accross as a String that needs to be parsed into json instead of a JS object already? i'm surprised, but I believe you!

eddierubeiz: Oh no, that's not me parsing it. I just wanted to pretty-print the JS object.

eddierubeiz: https://codingwithspike.wordpress.com/2016/11/08/pretty-printing-javascript-objects/

jrochkind: Oh i see, it's not parsing, it's actually the reverse, serializing to string. ok!

PR #628, opened 2020-02-11
Encrypt patron data

eddierubeiz:
For now, we're just encrypting patron_name and patron_email.
A 64-bit master key is present in `local_env.yml`, so this can be safely deployed whenever.


jrochkind:
Looks great, thanks @eddierubeiz !

jrochkind:
Please mention/link to lockbox in "Interesting gem dependencies" section in README, as discussed. And mention what for it's used. Also make sure a README reader can access instructions on how to rotate keys, here, or in lockbox PR or in the README itself or other docs or whatever you think someone like us would need.

Issue #631, opened 2020-02-18
Digitization queue status menu throws error


eddierubeiz:
At e.g. https://digital.sciencehistory.org/admin/works/kasznnp
Attempt to change the status
You will see a JS error alert.
However, the item's status *is* changed, as is apparent if you reload the page.

jrochkind:
Is this HB error the same thing? Or different? https://app.honeybadger.io/projects/58989/faults/60952006

eddierubeiz:
Yes, the HB error is the same.

eddierubeiz:
This is fixed in production as of Wed. afternoon.

PR #632, opened 2020-02-18
[WIP] Fix for DQ and R&R controllers' update method when json is requested.

eddierubeiz:
This is a fix for Ref #631 
I believe https://stackoverflow.com/questions/20006951/rails-looking-for-template-for-json-requests is a good description of the problem here.
Calling update method of dq and r&r controllers now returns json, instead of a string, if you ask for json.
It's not clear to me why this bug was not visible in production before; further research may be necessary.
@jrochkind  points out we may want some front-end tests for this finicky feature.

### DIFF
--- a/app/controllers/admin/digitization_queue_items_controller.rb
+++ b/app/controllers/admin/digitization_queue_items_controller.rb
@@ -47,8 +47,9 @@ class Admin::DigitizationQueueItemsController < AdminController
   def update
     respond_to do |format|
       if update_with_action_comments(@admin_digitization_queue_item, admin_digitization_queue_item_params)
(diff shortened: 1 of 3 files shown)


jrochkind:
Do you know if we already have any tests of the auto-submit JS status menu? Capybara I guess?

If not, any thoughts as if it's feasible to add one? If so, more of em?

I don't like how we keep accidentally breaking this (or just not fully fixing it?) when trying to fix another bug involving it, or add features to it. And our breakages don't get caught by tests. 

If we could add a basic "smoke test" that would fail before this fix but pass after it, I think we'd be ahead for trying to avoid more bugs in the future.

eddierubeiz:
`spec/system/digitization_queue_spec.rb` can easily be amended to include a smoke test for this, without much overhead or slowing down the tests. I'm going to include that in this PR.

PR #634, opened 2020-02-24
Modelling for OralHistoryContent

jrochkind:
A sort of 'sidecar' on a Work with oral-history-specific content and metadata, mostly related to OHMS. See the docs at the top of the oral_history_content.rb class, it should be self-documenting. 

This is where the combined audio derivatives will be attached. Also the OHMS XML file exported from OHMS.


eddierubeiz:
This all looks good. I'm merging it!

Issue #560, opened 2019-12-12
Audio files (and other JS) won't play in MS Edge/IE11


jrochkind:
I could have sworn we checked/verified this, but currently it looks like our audio player won't work in MS Edge (successor to Internet Explorer), or Internet Explorer itself. 

I tested in a VirtualBox virtual machine. It's possible the failure is an artifact of the VirtualBox, and it would work on other windows machines? We should maybe ask someone with a Windows machine to test. 

On MS Edge:

![Screenshot 2019-12-12 15 39 04](https://user-images.githubusercontent.com/149304/70747243-92949e00-1cf5-11ea-92de-b4f73f77d0b3.png)

On IE 11, it looks like it might work, but doesn't actually play. (This may be unrelated JS error on IE7, not sure if we still need to support IE?).

jrochkind:
We're providing an html5 `<audio>` tag with an audio/mpeg mp3 file, and an audio/webm file. 

eg

```
<audio controls="" controlslist="nodownload" data-role="audio-elem">
                <source src="https://scihist-digicoll-production-derivatives.s3.amazonaws.com/f9bdb82f-51cb-47fb-8ec5-972af4ec905d/small_mp3/5caf1582aaf452f2e753dddeebf889a9.mp3" type="audio/mpeg">
                <source src="https://scihist-digicoll-production-derivatives.s3.amazonaws.com/f9bdb82f-51cb-47fb-8ec5-972af4ec905d/webm/cc5eaf3278d625db7daf0b115cc6ac2c.webm" type="audio/webm">
            </audio>
```

I'm confused by Edge not seeming to be able to use the mp3.  In fact some information I find online suggests that Edge should even be able to use the webm. I wonder if this is an artifact of my VirtualBox VM.

jrochkind:
Ah, indeed may be a false alarm. It seemed to play fine testing in MS Edge with a free trial on browserstack.com . (Free trial didn't let me test in IE11). 

I think the VirtualBox VM for MS Edge was just broken with regard to audio support. :(

sanfordd:
@jrochkind I've tested on two PCs, a Windows 7 PC and a Windows 10 PC.
The Windows 7 PC under IE 11 could not play or download audio files.
The Windows 10 PC worked fine under Edge, but could not play or download audio files under IE 11.

sanfordd:
Adding some details here,
[...]

PR #636, opened 2020-02-26
ohms features: transcript and index

jrochkind:
Doesn't actually work yet, becuase we don't have combined audio derivative yet, so audio file that is loaded in player and which we seek to for sync links -- won't be right. But can be used to demo with single-audio-file works like duarte. 

Provides an 'oral history' tab in work admin (only for works with genre oral histories) that lets you upload an OHMS XML. (And has some placeholders for where we'll trigger generating a combined audio deriv). 

Then, on oral history audio show page, if OHMS data is present, will show transcript, or index, or both. With sync links. Styling of index is just tentative. 

If no OHMS data is present, audio show pages should be basically unchanged. So, if all is well, we should be able to deploy this now. 

One difference is rights-and-social section swapped to be above download file list, instead of below. Works better for when we have OHMS tabs.


jrochkind:
At some point in the future when it won't conflict with ongoing work in parallel -- we should rename the "show audio" stuff to "oral_history_with_audio" -- it's really OH specific, and getting more so. And change the condition deciding when to use it to make sure genre == oral histories.

jrochkind:
Hmm, potentially has problems with IE11 on browserstack. 

Bootstrap tabs don't seem to work. Audio player doesn't seem to work. Need to investigate more, and compare to what's already deployed.

jrochkind:
OK, IE11 problem not unique to this branch. 

Audio does not play on master in IE11 either. I don't think the problem is necessarily specific to audio -- I think all JS is broken on IE11. It looks like it might be Blacklight's fault -- perhaps Blacklight does not support IE11 anymore?

sanfordd:
This relates back to #560, not sure if we ever came to a conclusion on our desire to support IE anymore.

jrochkind:
Good call @sanfordd . This branch does seem to work fine on MS Edge.

sanfordd:
Thankfully Edge moving to a Chromium based engine should mean anything that works on Chrome works on Edge.

jrochkind:
This is ready to be _reviewed_ now @eddierubeiz , but  before we _merge_ I wanna check with stakeholders that:

* "index" (ToC) view looks okay -- prefer accordion style? 
* This will change existing audio OH's to have the "Tabbed" UI with Description and Downloads in separate tabs instead of two columns -- is that okay to deploy?

eddierubeiz:
Noted. Will review this evening.

jrochkind:
The OHMS transcript really needed a higher line-height to be legible, so this includes some kind of hacky CSS to allow some areas to have a higher line-height than usual, as well as higher line-height for not only OHMS areas, but work/collection detail page full description areas too. (Stakeholders agreed that was an improvement there too). 

I looked into changing it more places and/or making the longer line-height the default for our serif body font -- but it was turning into more work than I wanted to do right now to make sure the design was still ship-shape changing it in more places.

jrochkind:
This is now ready for review/merge.

eddierubeiz:
No comments. Great work. Merging now!

Issue #626, opened 2020-02-10
Updates to supporting text on metadata form


HKativa:
Nicole is drafting updates to supporting text for fields on metadata form (changing "CHF" to "Science History Institute," expanding on "TBD" fields, etc.). Once the updated text is ready, we'll pass the edits along to Eddie and Jonathan.

eddierubeiz:
OK. These are very easy changes, for what it's worth.

yonyitz:
Please find supporting text updates in spreadsheet below:

P:\Othmer Library\Digital Collections - Internal Access\documentation\supporting_text.xlsm

eddierubeiz:
This can probably go live Monday morning, by the way.

PR #638, opened 2020-02-27
Updated copy for work form.

eddierubeiz:
Adding new copy per Nicole & Hillary.
Ref #626 .


jrochkind:
Thanks @eddierubeiz !

PR #639, opened 2020-02-28
Combined audio derivatives for oral histories

eddierubeiz:
Featuring lots of ffmpeg trickery.
Re: #644


app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,163 @@
...
+# given oral history work.
+#
+# some_work = Work.find_by_friendlier_id(friendlier_id)
+# CombinedAudioDerivativeCreator.new(some_work).generate
+#
+#
+# Sample output:
+#     {
+#      :combined_audio_mp3_data=>
+#          "/var/folders/n7/t4zt2w751bj_6pbnf95sgpnrmz8lz8/T/output20200228-61637-ec6kjz.mp3",
```
jrochkind: Please return a ruby TempFile, not a file path.

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,163 @@
...
+#      :combined_audio_webm_data=>
+#          "/var/folders/n7/t4zt2w751bj_6pbnf95sgpnrmz8lz8/T/output20200228-61637-1btjscz.webm",
+#      :fingerprint=>"4998d366f5edb6222db1181c7b153e21",
+#      :component_metadata=>
+#        {:durations=>
+#          [
+#            "00:30:30.29",
+#            "00:29:13.32",
+#            "00:33:03.26",
+#            "00:24:28.94"
```
jrochkind: So, these free floating times makes me a bit worried that we'd get confused about which one goes with which file -- but I guess the fingerprint (unless it's buggy) ensures that these go with a certain list of files in a certain order. 

Still, can we make this something that is actually pairs of ids representing files to times, instead of just times? That will make it easier to access in the access patterns we're going to need too -- we're going to have a file, and want to look up it's timecode in order to add a link on that file to "jump to" a certain timecode. 

That can probably be returned as a hash -- hashes are ordered in ruby, ordered by insertion order. Or it could be an array of two-element arrays `[[id, time], [id, time]]` (which makes it easier to keep in order, and can be easily turned into a hash). 

The ids can be actual UUID pks, or friendlier_ids. Friendlier_ids are a lot easier to deal with, but also are hypothetically mutable, at which point they'd invalidate the data. I wonder if it actually makes sense just to use the UUIDs, they shoudln't change and are more universally available internal to the app. 

**Wait though**, also, "durations" aren't really what we want -- we want START TIMES. So actually calculate them for us, they are what we want. (ie, start time of 3rd one is duration1+duration2).  And then call it :start_times instead of :durations.

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,163 @@
...
+      member.stored? && member.file.mime_type.start_with?('audio')
+    end
+  end
+
+  # If this checksum changes, you need to regenerate the audio
+  def fingerprint
+    @calculated_checksum ||= begin
+      parts = [work.title, work.friendlier_id] +
+        audio_members.map { |a| a.file.metadata['sha512'][0..10] }
+      Digest::MD5.hexdigest(parts.join)
```
jrochkind: Why are you including only the first 10 chars of the sha512?  While that *might* be safe, it runs some risk, and I don't see any downside of including the whole thing. 

Additionally, when we make `durations`  or `start_times` a pair of id=>time, we could include whatever id we choose there in the fingerprint too, to ensure that if that changes, the fingerprint shows up as stale. That might make it fine to use the easier friendlier_id instead of uuid pk.

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,163 @@
...
+
+  def audio_member_files
+    audio_members.map { |asset| asset.file }
+  end
+
+  # Filters out non-audio items
+  def audio_members
+    work.members.order(:position, :id).select do |member|
+      member.stored? && member.file.mime_type.start_with?('audio')
+    end
```
jrochkind: Cache/memoize this with `@audio_members ||=`, since you refer to it multiple times. Without that, as it is, every time you refer to it, it might go to the db again to fetch it again. Which is a performance problem, but also could be a semantics problem if it's *changed* in the db on subsequent fetch, could lead to unexpected and hard to debug behavior. 

Am I sure this will go to the DB? No, I'm not sure, `work.members` should cache/memoize itself, but I _think_ adding `order` on the end will make it go to the DB. 

But rather than be sure (or be sure if it's called more than once) -- just memoize, and then you know you're safe, even if code changes after you were 'sure'.

spec/services/work_combined_audio_creator_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
```
jrochkind: Add a couple tests for fingerprint, make sure that if the members change in a way that should change fingerprint (add or remove member, change position), it does change fingerprint.

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,163 @@
...
+    result = []
+    audio_member_files.each do |original_file|
+      log "Downloading #{original_file.metadata['filename']}"
+      new_temp_file = Tempfile.new(['temp_', original_file.metadata['filename'].downcase], :encoding => 'binary')
+      original_file.open(rewindable:false) do |input_audio_io|
+        new_temp_file.write input_audio_io.read until input_audio_io.eof?
+      end
+      result << new_temp_file
+    end
+    result
```
jrochkind: As this method is called in the constructor it means just constructing one of these objects will trigger download of all components. That seems like a bad pattern to me, the "expensive" thing shouldn't happen in the constructor, it should only happen when you call some "action" method. I think that will be more intuitively expected behavior. 

While you could make this method lazy, I think it may make sense to keep the list of actual Asset models that will be used internally, and then only download them into files later. This will make it easier to get metadata from them as needed (like to use as keys above in duration!), to have them all there. We may also have duration already stored in metadata as an asset, so you don't need to use a shell out to ffmpeg to look it up?  (But if we don't, that's cool).

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,163 @@
...
+#      :component_metadata=>
+#        {:durations=>
+#          [
+#            "00:30:30.29",
+#            "00:29:13.32",
+#            "00:33:03.26",
+#            "00:24:28.94"
+#          ]
+#        }
+#     }
```
jrochkind: Returning as a hash is kind of messy, it's easy for the consumer to typo a key and miss the thing. 

How about making a very simple response object to return instead? Here's one easy way to make a nested response class:

```ruby
Response = Struct.new(:webm_file, :mp3_file, :fingerprint, keyword_init: true)

# then you can do
Response.new(webm_file: whatever)
# and also
r = Response.new
r.webm_file = whatever
r.webm_file
# etc
```

app/services/combined_audio_derivative_creator.rb
jrochkind: Also, you can just do `asset.sha512` as a shortcut, don't need to do `asset.file.metadata['sha512']`. The shortcut seems less error prone too. (Should we raise if any of the sha512's are missing to avoid a subsequent harder to debug problem?)

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Done.

app/services/combined_audio_derivative_creator.rb
eddierubeiz: OK  -- returning :start_times -- just a running total of durations, minus the last duration which I suppose we don't care about. Keys are UUIDs (not friendlier IDs). Good call.

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Yup!

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Removed the method from the constructor, and memoized the `components` method (not the `audio_member_files` method, since that is only called once, from the `components` method).

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Good call. Followed your advice.

spec/services/work_combined_audio_creator_spec.rb
eddierubeiz: Done. Took forever but I don't regret it.

app/services/combined_audio_derivative_creator.rb
eddierubeiz: `asset.sha512` -- good call. I'm also deferring to you re: using the entire hash, although 40 bits of info per asset seem plenty for our purposes here. (On the other hand, truncating makes things more complicated and bug prone.)

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,170 @@
+require 'tempfile'
+
+
+Response = Struct.new(:webm_file, :mp3_file, :fingerprint, :start_times, keyword_init: true)
```
jrochkind: Nest this inside of `CombinedAudioDerivativeCreator`, by putting it under the `class CombinedAudioDerivativeCreator`. 

As you've done it, it's a top-level constant `Response`, which could conflict with something else trying to make a top-level `Response`. Instead, it'll be `CombinedAudioDerivativeCreator::Response`, avoiding name conflicts and making it clear where it comes from.

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,170 @@
...
+      end
+      result
+    end
+  end
+
+  # A list of arrays; the first item in each is the UUID of each audio member,
+  # while the second is the *starting point* of that audio w/r/t the combined audio.
+  def calculate_start_times()
+    durations = components.map do |f|
+      duration_of_audio_file_in_seconds(f.path)
```
jrochkind: This is going to do a lot of extra shell-outs to `ffmpeg`. When you ask for start time for file 2, it'll shell out to `ffmpeg` for 1. When you ask for 3, it'll shell out for 1 and 2 (again). Etc. 

Because the info is never cached, every time you ask for it it shells out. Is there another way to do this to avoid that? Or does it not really matter?

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,170 @@
...
+#
+# Sample output:
+# <struct Response
+#  webm_file=<File:/path/to/file.webm>,
+#  mp3_file=<File:/path/to/file.mp3>,
+#  fingerprint="e0248c40015d6dc90b0d02937950b5d7",
+#  start_times=
+#   [["ddbd1a8d-c2eb-47b3-85a4-11b4ffb41719", 0],
+#    ["df773502-56d7-4756-a58c-b1f479910e97", 1.593469]]
+#  >
```
jrochkind: Might show up when printing as `<File:`, but please make it clear in docs, somehow, that what is returned is a Tempfile. Also make sure there's a test ensuring this.

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,170 @@
...
+
+  def generate
+    output_files = {}
+
+    # Create the two output files:
+    ['mp3', 'webm'].each do |format|
+      output_files[format] = output_file(format)
+      ffmpeg_args = args_for_ffmpeg(output_files[format].path)
+      log "Creating #{format} using: #{ffmpeg_args.join(" ")}"
+      @cmd.run(*ffmpeg_args, binmode: true)
```
jrochkind: Does TTY::Command not already log itself?  If it doesn't it might be becuase we TOLD it not to! `TTY::Command.new(printer: :null)`? Instead of telling TTY::Command not to log, and then logging separately, can we just tell it to log instead? Only if there's a way to tell it to log at our desired level I guess?

Alternately, we could just not log it in case of successful completion -- the places you copied where I was doing `printer: null` are places whereI decided I didn't want it clogging up the logs (although in those places it was much more frequent things).

app/services/combined_audio_derivative_creator.rb
```diff
@@ -0,0 +1,170 @@
...
+      work.members.order(:position, :id).select do |member|
+        member.stored? && member.file.mime_type.start_with?('audio')
+      end
+    end
+  end
+
+  # If this checksum changes, you need to regenerate the audio
+  def fingerprint
+    @fingerprint ||= begin
+      digests = audio_members.map(&:sha512).compact
```
jrochkind: Since start_times includes the UUID pk, should we make sure the UUID pk hasn't changed, by including it in the fingerprint too? Maybe not necessarily, because we don't expect these to ever change, although they ARE technically mutable. But maybe doens't matter.

spec/services/work_combined_audio_creator_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    expect(combined_audio_info.start_times).to match([
+      ["ddbd1a8d-c2eb-47b3-85a4-11b4ffb41719", 0],
+      ["df773502-56d7-4756-a58c-b1f479910e97", 1.593469]
+    ])
+    cmd = TTY::Command.new(printer: :null)
+    stats_command = ['ffprobe', '-v', 'error', '-show_format', '-show_streams' ]
+    mp3_details   = cmd.run(*stats_command + [combined_audio_info.mp3_file.path] ).out.split("\n")
+    webm_details  = cmd.run(*stats_command + [combined_audio_info.webm_file.path]).out.split("\n")
+    expect(mp3_details).to  include('format_name=mp3')
+    expect(webm_details).to include('format_name=matroska,webm')
```
jrochkind: I don't understand why you are you doing your own TTY::Command shell out in the test and testing that. Is this a mistake? It would appear to be testing code that is only located in this test, instead of testing the code that is in CombinedAudioDerivativeCreator . Please explain, if it isn't a mistake. 

Also, nothing in this test seems to be testing that the `Response` object returned actually includes the Tempfile for the webm and the mp3. Please ensure these are there, that they are ruby Tempfiles, and maybe that the file is actually non-0, or if you want to be extra that the file really looks like a webm/mp3 (maybe using the same file analysis tool that kithe uses, or shelling out to `file`, or whatever).

spec/services/work_combined_audio_creator_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(mp3_details.any?  {|x| x.include? 'duration=4.7'}).to be true
+    expect(webm_details.any? {|x| x.include? 'duration=4.7'}).to be true
+    # The fingerprint depends on the title, friendlier_id, and audio_file_sha512.
+    # As long as those are fixed, you can count on this fingerprint.
+    expect(combined_audio_info.fingerprint).to eq '8402b003dd07775a83b1a000b3319fd9'
```
jrochkind: Prob just necessar to check that `fingerprint` is present?, rather than that it's a specific thing, we don't really know what it's supposed to be, don't need the test to break if we change algorithm, but not really a big deal either way.

spec/services/work_combined_audio_creator_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+describe "Combined Audio" do
+  # We need to set the friendlier_id explicitly, because it's part of the fingerprint recipe.
+  let!(:work) { FactoryBot.create(:work, title: "top", friendlier_id: '1ciyp1i')}
+
+  let!(:audio_asset_1)  { create(:asset, :inline_promoted_file,
+      position: 1,
+      parent_id: work.id,
+      id: 'ddbd1a8d-c2eb-47b3-85a4-11b4ffb41719',
+      file: File.open((Rails.root + "spec/test_support/audio/ice_cubes.mp3"))
```
jrochkind: Would not have expected you had to set the `id` manually here, but you did?

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Eh, no harm including UUID pk as well.

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Good call!

app/services/combined_audio_derivative_creator.rb
eddierubeiz: We cleared this up on Slack; resolving. (TLDR: no, there are no extra shell outs.)

spec/services/work_combined_audio_creator_spec.rb
eddierubeiz: So I'm using ffprobe here to determine that the files output by the service are in fact audio, that their duration is correct, and that they are indeed mp3 or webm. (I *could* use file, but ffprobe gives me duration in particular.)

eddierubeiz: Definitely should be checking that `combined_audio_info.mp3_file.class == Tempfile`

spec/services/work_combined_audio_creator_spec.rb
eddierubeiz: Only if I needed to check that the same id resulted in the same fingerprint. Since I'm not checking that anymore per your comment above ... no need to set the `id`!

spec/services/work_combined_audio_creator_spec.rb
jrochkind: OH NOW I SEE

spec/services/work_combined_audio_creator_spec.rb
eddierubeiz: Cool!

spec/services/work_combined_audio_creator_spec.rb
eddierubeiz: Removed it!

spec/services/work_combined_audio_creator_spec.rb
eddierubeiz: OK -- resolving this!

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Turned on logging for cmd, removed all logs from the code!

PR #647, opened 2020-03-04
ability to upload and attach an OHMS XML file

jrochkind:
JUST that, extracted from #636. So we can merge this first, slightly reducing the size of the other -- and also giving @eddierubeiz the place to put the admin UI for generating combined deriv and getting URL for it etc.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -84,8 +84,16 @@ class Admin::WorksController < AdminController
       return
     end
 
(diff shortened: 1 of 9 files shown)


jrochkind:
OK, per @eddierubeiz 's advice, added checking to make sure it's a valid XML file (against OHMS XSD even), before allowing it to be saved. 

Also added a couple controller tests for adding OHMS XML.

eddierubeiz:
Thanks for indulging me. I do think it'll save us drama in the long run.

Issue #648, opened 2020-03-05
combined audio derivative: admin UI 


jrochkind:
We need UI to create the combined derivative.  There's some placeholder stuff on teh screen here:

https://digital.sciencehistory.org/admin/works/wh246t17v#nav-oral-histories

It should tell you if combined derivative(s) are already created. (Prob verifying we have both of the formats expected). 

It should tell if you if the 'fingerprint' is fresh, maybe putting a big alert in red if it's not. 

It should give you a button (re)create, which kicks off a bg job to create the combined derivs and record them and their metadata in the db. 

* Bg job should remember to delete the Tempfile when it's done with it -- while you can get let ruby GC take care of it, more efficient to do it if you know you're done with it. See https://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Explicit+close

Anything else you can think of?

eddierubeiz:
I have a basic version of this working. There isn't much of a way for a regular user to find out if the bg job fails for any reason. However, once we mass-create the combined audio derivatives, requesting one to be generated or regenerated will only happen once in a while, so this might not be a huge problem.

jrochkind:
> There isn't much of a way for a regular user to find out if the bg job fails for any reason.

That's true of all our derivatives, really, so may not be a requirement for first pass. 

Although I could imagine adding a column to the oral_history_content table where error information is recorded, if it happens. But rescuing errors to reprot them could interfere in odd ways with things we might do in the future to eg auto retry failed jobs etc.

jrochkind:
> Are you sure? This could take several minutes.

This made me wonder if it was going to occupy my browser for several minutes preventing me from doing other stuff, which of course is not true. 

Under what circumstances would you imagine a staff user confronted with this warning deciding that no, after all, they didn't really want to do it? (because it would take several minutes?).

eddierubeiz:
Excellent point. Will remove the warning.

jrochkind:
Looking at stuff in the staff UI as I sit at Code4Lib. 

Right now, both webm and mp3 are presented, as links on the word. 

I think maybe the user doesn't need webm at all, but I think they need MP3 in a more clear and usable way, such as:

> Use this URL as the OHMS Media URL:
>       http://......

(Make sure you generate that url with `url(public:true)`, since that guarantees a non-expiring S3 URL without a `?` in it, which OHMS insists upon to work!)

[...]

PR #650, opened 2020-03-06
Background job to create audio files with basic UI

eddierubeiz:
Includes a rake task as well to run the background job.
Includes tests.
Ref #648

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -109,16 +109,10 @@ class Admin::WorksController < AdminController
       redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), flash: { error: "This oral history doesn't have any audio files." }
       return
     end
-
-    current = @work.oral_history_content!.combined_audio_fingerprint
-    if current != CombinedAudioDerivativeCreator.new(@work).fingerprint
-      # As things currently stand, the job can fail silently if e.g. the
-      # originals can't be downloaded from AWS for any reason.
-      CreateCombinedAudioDerivativesJob.perform_later(@work)
-      notice = "The combined audio derivative job has been launched."
-    else
-      notice = "We already have up-to-date combined audio for this item."
-    end
+    # As things currently stand, the job can fail silently if e.g. the
+    # originals can't be downloaded from AWS for any reason.
+    CreateCombinedAudioDerivativesJob.perform_later(@work)
+    notice = "The combined audio derivative job has been launched."
     redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), notice: notice
   end
 
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -97,13 +97,16 @@ class Admin::WorksController < AdminController
   end
 
 
-  # Create_combined_audio_derivatives, if warranted.
+  # Create_combined_audio_derivatives in the background, if warranted.
   # PATCH/PUT /admin/works/ab2323ac/create_combined_audio_derivatives
+  # Unfortunately, if the job fails for any reason, the user will not be notified.
   def create_combined_audio_derivatives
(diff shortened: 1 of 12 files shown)


app/controllers/admin/works_controller.rb
```diff
@@ -96,6 +96,33 @@ def submit_ohms_xml
...
+
+    current = @work.oral_history_content!.combined_audio_fingerprint
+    if current != CombinedAudioDerivativeCreator.new(@work).fingerprint
+      # As things currently stand, the job can fail silently if e.g. the
+      # originals can't be downloaded from AWS for any reason.
+      CreateCombinedAudioDerivativesJob.perform_later(@work)
+      notice = "The combined audio derivative job has been launched."
+    else
+      notice = "We already have up-to-date combined audio for this item."
+    end
```
jrochkind: I would not include the code to prevent running if the fingerprint is already up to date. Maybe it's corrupt or was done with the wrong settings and needs to be re-done -- there's no reason to forbid doing it again, and it's extra code, and the most bug-free code is the code you don't include. 

Not saying you need to change the UI at present, but let's take out the guard here that refuses to create if the fingerprint is up to date -- if the action is triggered, fire off the bg job.

app/controllers/admin/works_controller.rb
```diff
@@ -381,4 +408,36 @@ def cancel_url
...
+      oh_content.combined_audio_webm&.url(public:true)
+    end
+    helper_method :combined_webm_audio
+
+    def combined_audio_fingerprint
+      return nil unless @work.genre.present?
+      return nil unless @work.genre.include?('Oral histories')
+      @work.oral_history_content!.combined_audio_fingerprint
+    end
+    helper_method :combined_audio_fingerprint
```
jrochkind: These are an awful lot of methods to include in the controller for a very focused/targetted functional area. 

Let's figure out how to extract them into a presenter or ViewModel or something? If these are all for the UI component for the button to trigger creation and show status, maybe it would be a good place for a ViewModel for that purpose, which can include the helper methods?

Let me know if you're not sure what I mean or want to discuss the best way to do it.

eddierubeiz: Yeah, a presenter is a fine idea. Not a huge change either. I can take care of that.

app/controllers/admin/works_controller.rb
eddierubeiz: Yeah, that makes sense. Will remove.

app/controllers/admin/works_controller.rb
```diff
@@ -381,4 +402,36 @@ def cancel_url
...
```
jrochkind: You moved these to a ViewModel, but you didn't remove them from here yet...

eddierubeiz: Yeah -- oops. I also never actually changed the controller and view to refer to the new methods. My bad; fixed.

app/views/admin/works/_combined_audio_derivatives.html.erb
```diff
@@ -0,0 +1,53 @@
+<div class="card bg-light mb-3">
+  <h2 class="card-header h3">Combined Audio Derivative Files</h2>
+  <div class="card-body">
+    <% if view.work_audio_members.count == 0 %>
+      This oral history doesn't have any audio files associated with it.
+    <% else %>
+      <% if view.combined_audio_fingerprint.present? %>
+        <% if CombinedAudioDerivativeCreator.new(model).fingerprint != view.combined_audio_fingerprint %>
```
jrochkind: You *could* extract this logic into the ViewModel, once you have a ViewModel it's best to keep logic in the html template as simple as possible, mostly just calling methods in the viewmodel for anything non-trivial. 

But I'm not gonna block the PR over it at this point!

eddierubeiz: Good call; done.

Issue #649, opened 2020-03-05
Modify audio player front end to use the new combined audio derivatives


jrochkind:
To follow up #648

First:

* Create "combined audio derivatives" for all existing OH audio works
  * This could be very time-consuming to do for all OH works, we'll have to see and see if we need to do special thigns to speed it up, like a special jobs server working only on that in parallel or something. 

Then we need to move the UI over to USE them. 

* audio player should load the combined derivative
  * If it's not available, a "Sorry not available" message of some kind, maybe even "Administrator: You need to create the combined derivative", maybe guarded behind a "if logged in user is admin". 
* Clicks on individual download tracks should advance to that portion of the combined derivative, as recorded in the metadata on the oral history content, that was created with the combined audio derivative. 
  * No need for the 'now playing filename" line currently in the player, that gets swapped every time you click on a different link -- just eliminate that whole area. Click on a download filename is just a 'seek' operation. 
  * No need for highlighting the now playing track in the list either I don't think, the JS gets significantly simpler, just a seek!  (Maybe the time to seek to is put in a data attribute on HTML generation? actually maybe can use the same JS that is doing seeks for transcript/index! Oh, except that's not merged yet, so I dunno)

After THAT is done and merged and released, 

* Remove code that creates existing single-file webm derviatives, and delete all existing webm single-file derivatives. Cause we're not using them anymore, they were only used for the player. 
* The single-file mp3 derivatives are currently being offered as downloads -- need to talk to stakeholders about if we should retain these as a download option, or get rid of them too.

eddierubeiz:
Sample code to generate all combined audio derivatives:
```
require 'ruby-progressbar'
progress_bar = ProgressBar.create(total: Work.where("json_attributes -> 'genre' ?  'Oral histories'").count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")

Work.where("json_attributes -> 'genre' ?  'Oral histories'").find_each(batch_size: 10) do |w|
  progress_bar.log("INFO: Starting #{w.title}")
  CreateCombinedAudioDerivativesJob.perform_now(w)
  progress_bar.log("INFO: Done with #{w.title}")
  progress_bar.increment
end
```

eddierubeiz:
Creating all the combined audio derivatives in production should take roughly 36 hours.
In staging, roughly 24 hours after kickoff, most of them are done:
```
irb(main):013:0> Work.where("json_attributes->'genre'?'Oral histories'").count
=> 228
[...]

PR #659, opened 2020-03-17
Oral history front  end: use new combined audio derivs.

eddierubeiz:
Ref #649
Includes basic tests.


app/decorators/audio_work_show_decorator.rb
```diff
@@ -83,4 +83,38 @@ def has_ohms_transcript?
...
+  end
+
+  # The start time or audio offset for a particular audio asset,
+  # relative to the entire oral history interview (the work as a whole).
+  # We're rounding to the nearest tenth of a second; easier to read.
+  # Returns -1 if there is no start time for this asset, or for any assets.
+  def start_time_for(audio_asset)
+    return -1 if start_times == {}
+    start_times[audio_asset.id]&.round(1) || -1
+  end
```
jrochkind: how about returning `nil` instead of `-1`?   `nil` is a pretty standard value for "no value available".

jrochkind: Also you don't need line 116, I don't think. Line 117 handles that case already. And if returning `nil` instead of `-1`, you probably don't need the `|| -1`, since it'll be returning nil already if not in the array!

jrochkind: oh, I see this maybe makes the JS more confusing if it's not -1?  I dunno if it's really painful that's fine, but maybe it just translates to JS `undefined` or whatever, which is still a better value for "no value available" than -1. But not a big deal.

app/javascript/src/js/playlist.js
```diff
@@ -43,23 +29,10 @@ $( document ).ready(function() {
...
-		$(this.audioElement).find("source[type='audio/webm']").attr("src", $(track).data('webmUrl'));
-
-		// Tell HTML audio to load new stuff
-		// See: https://stackoverflow.com/questions/9421505/switch-audio-source-with-jquery-and-html5-audio-tag
-		this.audioElement.pause();
-		this.audioElement.load();
+		var startTime = $(track).data('startTime');
+		if (startTime != -1) {
+			this.audioElement.currentTime = startTime;
+		}
```
jrochkind: I don't think you need any JS at all, located in master in simple_ohms_player.js is already code that will catch any link with a `data-ohms-timestamp-s="#{time_in_seconds}"` attribute, and advance the player to that timecode. So you can just make use of that, with such an attribute, and no custom JS at all?

Maybe if the start time is missing (which would be an error condition), simply don't add that attribute to the link at all. Now no need for special -1 handling either?

app/decorators/audio_work_show_decorator.rb
eddierubeiz: Yup, absolutely right; nil works great.

app/javascript/src/js/playlist.js
eddierubeiz: Got rid of the entire file!

PR #663, opened 2020-03-23
[WIP] Rake task to store audio asset file durations.

### LINKED ISSUE #661: Consider displaying either start times or durations alongside the file listings in the oral history audio front end.
Show start time next to each interview segment, with a label.
Related to #652

eddierubeiz:
This is only a band-aid; we will not need it in the long run:
Download audio asset files, calculate their duration, and add to file metadata.
Does not fix #661, but it makes it a lot easier to do.
Ultimately, the right way to do this is as part of file characterization, on ingest, via Shrine 2.0.

### DIFF
--- a/lib/tasks/calculate_audio_asset_durations.rake
+++ b/lib/tasks/calculate_audio_asset_durations.rake
@@ -1,0 +1,40 @@
+namespace :scihist do
+  desc """
+    Goes through all the oral histories and generates combined audio derivatives for
+    those missing them:
+
+    bin/bundle exec rake scihist:calculate_audio_asset_durations
+  """
+
+  task :calculate_audio_asset_durations => :environment do
+    begin
+      cmd = TTY::Command.new
+      duration_command = ['ffprobe', '-v', 'error',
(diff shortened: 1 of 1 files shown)


jrochkind:
I don't want to do this until we have a way to automatically do it for new ones. 

Then we could use a task like this to get the old ones that were previous to our way to automatically do it for new ones. 

Without a way to automatically do it for new ones, I don't see any point to doing this. Like, we'd set it up in a cronjob to run hourly or something? I don't think it's worth it, it's just more moving parts that can go wrong, and confusing to someone who sees durations and doesn't realize they aren't automatically added.

jrochkind:
There are certainly ways to wire it up to go automatically without waiting for the kithe refactor that will make it easier, if we wanted to do it sooner rather than later. I am not sure it's necessary.

jrochkind:
Not sure if this is currently relevant?

eddierubeiz:
Let's close this for now. I still think it's a good idea to have these numbers on hand, but this code is not something we need at the moment.

PR #666, opened 2020-03-23
ohms search

jrochkind:


app/javascript/src/js/audio/ohms_search.js
```diff
@@ -0,0 +1,463 @@
...
+  var indexResults = Search.searchIndex(query);
+  Search.currentIndexResults = new Search.SearchResults(
+    $("*[data-ohms-search-results]").get(0),
+    indexResults,
+    "index"
+  );
+  $("*[data-ohms-hitcount='index']").html('<span class="badge badge-pill badge-danger">' + indexResults.length + '</span>');
+
+  Search.currentResults().draw();
+  Search.currentResults().scrollToCurrentResult();
```
eddierubeiz, line 287: The scrollToCurrentResult line should be wrapped in an `if(Search.currentResults().results.length) {`

jrochkind: Thanks! Did you reproduce an error condition without this, verified something going wrong?

eddierubeiz: Yeah - just if there are zero results there's an error somewhere in scrollToCurrentResult. You could also return at the beginning of scrollToCurrentResult - whatever works.

jrochkind: awesome, thanks for catching this, I will try to reproduce before fixing, just so I understand what I'm fixing.

jrochkind: Reproduced!  

Normally it defaults to starting with "Table of Contents". If there are no hits in ToC, but htere is a hit in Transcript.... should it switch you to Transcript instead to show you the hit? Or just keep showing you no hits in ToC?  

Not sure, not sure it matters or we'll pick the right UX anyway, so I guess I'll just do the simplest thing, which is probably show you the no hits in ToC.

eddierubeiz:
Impressive work!

PR #668, opened 2020-03-23
Fix for bug in combined audio decorator

eddierubeiz:
Two related bugs which cancelled each other out. Fixing them both at the same time seemed like a good idea.


eddierubeiz:
Just noting that the diff for the template is much easier to understand if you click the gear icon and ask for a split diff. I just moved the contents of the `if` tag into the corresponding `else` tag, and vice versa.

Issue #664, opened 2020-03-23
eliminate individual original file webm derivatives


jrochkind:
We don't use them any more at all, not behind the scenes and not in the UI. 

* Stop them being created automatically on ingest
* Delete existing ones to stop using file spacce

jrochkind:
Also the individual file "optimized MP3" -- we decided to get rid of it! 

The individual file MP3 also needs to be removed from "downloads" list.

PR #670, opened 2020-03-25
Remove audio asset derivatives

eddierubeiz:
Ref #664 .
Audio asset derivatives, now replaced by work-level audio derivatives:
 - Are no longer generated on ingest
 - Aren't available for download
 - Aren't referred to in any of the tests.

This PR includes a simple Rake task to get rid of all the now-obsolete derivatives.

### DIFF
--- a/lib/tasks/remove_audio_asset_derivatives.rake
+++ b/lib/tasks/remove_audio_asset_derivatives.rake
@@ -1,7 +1,7 @@
 namespace :scihist do
   desc """
-    Goes through all the derivatives and if they are audio, removes them.
-
+    Goes through all derivatives of assets where key is either 'webm' or 'small_mp3', and removes them.
+    NOTE: Remove this file once the task has been run in production; it won't be needed after that.
     bundle exec rake scihist:remove_audio_asset_derivatives
   """
 
--- a/lib/tasks/remove_audio_asset_derivatives.rake
+++ b/lib/tasks/remove_audio_asset_derivatives.rake
@@ -6,7 +6,7 @@ namespace :scihist do
   """
 
   task :remove_audio_asset_derivatives => :environment do
-    progress_bar = ProgressBar.create(total: Work.where("json_attributes -> 'genre' ?  'Oral histories'").count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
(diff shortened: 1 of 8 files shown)


jrochkind:
> They're still actually in our S3 storage, though; removing them is not covered in this PR.

I _think_ still referenced by a Kithe::Derivative model too?  We should make sure we know where they are referenced -- and how to get rid of them -- before merging this PR, to make sure we aren't accidentally removing the info we'd need to easily remove them from S3. 

I _think_ it's good though.

eddierubeiz:
Yeah, I don't think Kithe knows anything about them. Will double-check!

jrochkind:
Something clearly knows about them, or how would we get a URL to them?

We need to make sure we know about them in order to find them and delete them from S3. 

I _think_ the application knows about them via the `work.derivatives` relationship, the objects at the end of that relationship are `Kithe::Derivative` models, which can have enough info to get a URL to one, and also to delete it from S3. The `Kithe::Derivative` models themselves will need to be deleted -- just deleting the model instance may automatically remove from s3 too.

eddierubeiz:
Oh, I see what you mean. Yup, the s3 files are wrapped by a `Kithe::Derivative` model. Absolutely, we'll have to delete audio `Kithe::Derivative`s .

eddierubeiz:
It's fine to merge this and *then* run the rake task, by the way. Each `Kithe::Derivative` AR keeps tabs on its corresponding S3 file via Shrine.

lib/tasks/remove_audio_asset_derivatives.rake
```diff
@@ -0,0 +1,20 @@
+namespace :scihist do
+  desc """
+    Goes through all the derivatives and if they are audio, removes them.
```
jrochkind: To be accurate, if they are marked 'webm' or 'small_mp3', it doesn't check to see if their content-type is audio, which it could. It doens't need to just saying the description is inaccurate. 

Should we also note in comments "can/should be removed once it's run in production and no longer needed."

eddierubeiz: Good catch. I wrote the comment before the code :) Will amend.

Issue #665, opened 2020-03-23
add user-facing download option for combined MP3 derivative


jrochkind:
Not sure what to call it or how to present it, needs some thought. But we do want to give the user the option to download this derivative we are already creating to power the player.

PR #672, opened 2020-03-30
Offer downloads for combined audio derivatives

eddierubeiz:
Ref #665


eddierubeiz:
@jrochkind adds:

Maybe put an icon indication “download” next to the link, so it’s more consistent with the individual links above?

Have to think about the case (like Duarte) where there’s only one file; we still prob want to offer the ‘optimized’ mp3 (cause original might be a FLAC too). Just have to make sure the heading is not weird/wrong i that case; maybe by special case for one file, or a generic heading that works for both. Maybe incorporate the term “optimized MP3" like we did before, somewhere?

Also can you make the whitespace consistent for what are the “three” headings now? Instead of different whitespace before PDF than before “all 7 segments” — I think the increased whitespace before “PDF” is probably good.

eddierubeiz:
For single-segment interviews: I'll just re-add the derivative to the existing download dropdown menu as "Optimized MP3." I think it's less confusing that way.

eddierubeiz:
Also I want to display the size of the item to download.

app/presenters/download_options/audio_download_options.rb
```diff
@@ -7,9 +7,34 @@ def initialize(asset)
...
 
+    # How many of this asset's parent's members are audio assets.
+    def parent_audio_member_count
+      @audio_members ||= begin
+        members = asset.parent.members.with_representative_derivatives
+        members = members.where(published: true) if current_user.nil?
+        members.count { |m| m.leaf_representative&.content_type&.start_with?("audio/") }
+      end
+    end
+
```
jrochkind, line 18: I believe this has a problem with "n+1 queries".  This method will do a fetch to the DB. Since there is an AudioDownloadOoptions object instantiated for *each* file in the list, it will be executed once for each file in the list -- even though each query, since based on the parent, will be the same query, it'll be executed multiple redundant times. 

(Correct me if I'm wrong, haven't verified interacting with code, just what it looks like to me). 

Its also not as efficient a query as it could be -- its' fetching ALL memberes, then in-memory filtering to get the count of audio ones. It could, with more difficulty, use postgres jsonb operators to have the DB do the count and get back only the count value. However, I think that wouldn't be good enough, it still wouldn't be good enough, we don't want the n+1. 

This is a problem I've run into before with the "ViewModel" archiecture -- when you need to calculate/fetch something "expensive" that will be the same for each model in a list, but isn't already easily available pre-calculated, you end up with an "n+1" problem. I haven't found any solutions I love yet. 

One is to have an _argument_ passed into the ViewModel for `parent_audio_member_count`. The thing calling it needs to calculate it (once), and pass it in to each AudioDownloadOptions it instantiating. But this kind of complicates and confuses the method signature for instantiating, making it more confusing to re-use the view model in different contexts, which is supposed to be part of the advantage of a view model. And you can end up having to pass this pre-caluclated value through a chain of calls to get from a common ancestor to where it's needed, which is even uglier. 

But since the "X" in "All X Segments as a single file" is already being calculated, maybe available right at the point AudioDownloadOptions is instantiated, perhaps it's feasible here. 

Another solution I"ve played with is passing in a draper-decorated model instead of the raw model. The draper decoration can have a method to calculate/fetch this "expensive" thing and memoize it. you need to make sure each AudioDownloadOptions gets the *same* draper-decorated model, so the memoized value is really reused. 

I think it might also be worth considering if we really need this variant UI for one vs more than one file. While I agree it is a *better* UI you've come up with... is it enough better to be worth this pain? Could we instead come up with a UI where basically the same things are on the page for both one and more-than-one file, so we don't need to know the count inside AudioDownloadOptions at all?

eddierubeiz: We talked about this on Slack, but to summarize the consensus, we went for a simple UI that obviates need to know the count inside a ViewModel which is *itself* inside another ViewModel (!).

eddierubeiz: Using a much simpler solution, on reflection. Full discussion is on Slack. Summary: "Could we instead come up with a UI where basically the same things are on the page for both one and more-than-one file, so we don't need to know the count inside AudioDownloadOptions at all?" We can and we should.

jrochkind:
Tried it out in a browser. Looking good!

Our normal downloads set content-disposition headers to force the browser to actually _download_ (with an appropriate filename). 

The combined download is not doing that, so the browser is like "I can an MP3 in a browser window", which I think is not what we want. And if you *do* choose "save", it gives you a name "combined.mp3", also cause of lack of content-disposition header. 

That made me notice for our existing downloads, we do not generally provide a direct to S3 URL, we provide an app URL that redirects to them. (I think that's to try to keep people from bookmarking the possibly temporary S3 one). We could do that same thing here, or we could just keep doing the S3 directly, but construct the signed S3 URL to specify content-disposition headers, I think that'd be fine too.

eddierubeiz:
Good call re: content-disposition for the links. Should have caught that  - sorry!

app/decorators/audio_work_show_decorator.rb
```diff
@@ -88,6 +90,22 @@ def combined_mp3_audio
     model&.oral_history_content&.combined_audio_mp3&.url(public:true)
   end
 
+  def combined_mp3_audio_download
+    base = "Combined audio for #{model.title}".
+      gsub(/([[:space:]]|[[:punct:]])+/, '_').
+      downcase
+    filename = DownloadFilenameHelper.filename_with_suffix(base, content_type: 'audio/mpeg')
```
jrochkind, line 97: Hmm, you are writing literal "Combined audio for", but then immediately downcasing and replacing spaces with underscore. I understand why you need to do do the gsub/downcase for the `model.title` because that's dynamic, but for the literal, you could just start out with "combined_audio_for_", since that's what you want to end up with. 

But I'm not sure that's really what we want to end up with. We'll end up wtih download names like "combined_audio_for_title_of_work.mp3".  Is that really what we want?  

I think `title_of_work.mp3` would probably be quite suffiicent. Although we also might want to limit the number of words in case of title of work is really long. 

Let's actually just look at how the title is constructed for the analogous case of our whole-work PDF/zip downloads....

https://github.com/sciencehistory/scihist_digicoll/blob/master/app/models/on_demand_derivative.rb#L77-L84

First three words of title, plus friendlier id. (Not sure what's going on in there with `deriv_type`). Can we do something similar here, using similar code? For consistency, and because that seems like a pretty good way to construct the filename too. There's even a helper there for extracting first three words that you can re-use. 

I see you decided not to do a controller to redirect, putting the temporary S3 URL directly in the HTML. This is probably fine. I can't recall why i did the redirect situation in the case of the PDF/ZIP derivatives... one thing I can think of is, using a redirect URL makes the page containing the link long-term cacheable by an HTTP cache, for longer than the S3 URL is good for. I'm not sure if this really matters or not though.

eddierubeiz: First 3 words of title + friendlier_id - done.
I really don't mind that much about page being cached, but I figure if it becomes a problem we can come back to this!

jrochkind:
Cool, now there's a conflict, want to resolve it when you have a chance @eddierubeiz ?

PR #676, opened 2020-03-31
Update blacklight range limit from 7.1.0 to 7.5.0

jrochkind:
With:

    $ bundle update --conservative blacklight_range_limit

When we tried updating to previous > 7.1.0 versions, there were bugs and problems preventing us. But in 7.5.0 they seem to be resolved.

There are some UI changes here that I'm not a huge fan of, but we determined in Slack we don't hate them enough to try to local override customize, it's what blacklight_range_limit has chosen.

On the plus side, there is now a "view larger" link to look at the histogram in a larger modal, which is convenient.


eddierubeiz:
Thanks for testing!

Issue #678, opened 2020-03-31
update blacklight range limit to 7.5.0


jrochkind:
Oddly doing so in #676 seems to have caused a failure not caught by our tests but reported by honeybadger. Reverted in #677. 

```
ActionView::Template::Error: No route matches {:action=>"range_limit_panel", :collection_id=>"6w924c45h", :controller=>"collection_show", :f=>{"creator_facet"=>["Mazurs, Edward G."], "department_facet"=>["Archives"], ..
```

https://app.honeybadger.io/projects/58989/faults/62413696

and

https://app.honeybadger.io/projects/58989/faults/62413752

Diagnose, hopefully fix, plus make a test to catch it, since none of our tests caught it!

PR #679, opened 2020-03-31
improve collection show page test

jrochkind:
It wasn't actually succesfully testing that collection members were displayed on page -- they weren't actually being properly solr-indexed in test, and were not being displayed.

Also give them different dates, to exersize blacklight_range_limit a bit.

This improved test would have failed when updating blacklight_range_limit, so would have caught problem in #678, that was uncaught and let us deploy with a bug.


jrochkind:
Good call with improving the other test @eddierubeiz , thanks!

PR #680, opened 2020-04-01
update blacklight_range_limit from 7.1.0 to 7.6.0

jrochkind:
Second try. See #676 and #678.

This time we include custom routes in collection and special topic controllers, for the new routes added by new version of blacklight_range_limit, to support 'view larger' functionality.


jrochkind:
@eddierubeiz blacklight_range_limit 7.6.0 is about to be released, so actually let's hold off, and upgrade to that, make sure it still works.

jrochkind:
OK, all the way up to blacklight_range_limit 7.6.0 and ready for review/merge. 

(Hopefully we're not making a mistake updating to just-released version... but I felt weird not, since this PR is about updating to stay up to date!)

eddierubeiz:
I'm ready to merge this. We can push on Monday of course.

Issue #673, opened 2020-03-30
Don't allow a user to publish an oral history until / unless there is an up to date combined audio deriv.


eddierubeiz:
First of several adjustments we'll probably need to make now that we're using the combined audio derivatives in production.

jrochkind:
I think the other half is "don't let staff edit anything that would invalidate a combined derivative, for a published audio oral history."

Those two things might take care of it all nicely?

eddierubeiz:
That is a very good idea.

eddierubeiz:
Very good.

jrochkind:
The implementation is going to be a little bit tricky to get right and reliable though, I'm afraid.

PR #683, opened 2020-04-01
Try to prevent publishing oral histories w/ out of date combined audio derivatives

eddierubeiz:
If an oral history is published, you can't reorder or rename its audio members.
If an oral history is NOT published, you can't publish it unless its c-a-d's are up to date.

Ref #673


eddierubeiz:
Note: for now, I'm not adding any validations on the model -- just adding an additional check in the controller before publishing. That way changes to metadata can happen separately from generating c-a-d's.

jrochkind:
Can't reorder or rename -- how about adding or deleting or changing 'public' status on em? Those would invalidate the combined deriv too, right? Not sure if that catches everything; this is a bit tricky.

eddierubeiz:
I prevented deleting, publishing and unpublishing, but not adding. Good catch.
Note: I'm relying on the templates for these prohibitions so far. Not sure we need to add extra code in the controllers for extra safety. Leaning towards not doing so, as this is not public-facing functionality anyway.

jrochkind:
Hm. Yeah I'm less worried about an "attacker", but more worried about our own bugs. How hard is it to add it in the controller, do you think?

(I agree that model validation is probably a best avoided tar pit).

eddierubeiz:
I could certainly add checks to the controller; we can discuss.

eddierubeiz:
Adding new segments to unpublished OHs is now also prohibited btw.

app/controllers/admin/works_controller.rb
```diff
@@ -392,4 +398,22 @@ def cancel_url
       admin_works_path
     end
     helper_method :cancel_url
+
+    # If this work is an oral history,
+    # return true if the user needs to (re)calculate combined audio derivatives.
+    def need_combined_audio_derivatives?
+      return false
```
jrochkind, line 405: This seems to be a mistake, you immediately return false without checking any conditions... which means this method will always return false. Which makes me suspect any tests that are present here must not be suffiiciently reliable, if they pass despite this broken method?

I may be wrong?

jrochkind:
Maybe should have a test? Also maybe a test that would have failed due to the bug I think I saw?

I am not totally following how the view-level restirctions on editing works. I may need a walk-through. The trickiness in following the code to be confident it's doing what's expected makes me think that we do want controller-level restrictions too if it's not infeasible (since it seems pretty easy to mess up), and also that we need some tests. (Ideally maybe including a test that would have failed on the bug I caught, unless I was wrong about that bug!)

app/controllers/admin/works_controller.rb
eddierubeiz: Ohhh no, you're not wrong. I'm wrong. Will fix. And add tests.

app/controllers/admin/assets_controller.rb
```diff
@@ -54,6 +64,11 @@ def fixity_report
 
   def display_attach_form
     @parent = Work.find_by_friendlier_id!(params[:parent_id])
+    if @parent.genre && @parent.genre.include?('Oral histories') && @parent.published?
```
jrochkind, line 67: Should this re-use the logic from `published_oral_history_asset?`like the others, or is this different?

jrochkind: oh nevermind, I see that this one is based on @parent, which is different. It could be DRY'd but I'm not gonna worry about it.

app/controllers/admin/works_controller.rb
```diff
@@ -392,4 +422,48 @@ def cancel_url
...
+      existing_fingerprint = @work.oral_history_content!.combined_audio_fingerprint
+      (existing_fingerprint != CombinedAudioDerivativeCreator.new(@work).fingerprint)
+    end
+
+    def restrict_audio_member_edit?(member)
+      return false if member.work?
+      return false unless @work.published?
+      member.content_type && member.content_type.start_with?('audio')
+    end
+    helper_method :restrict_audio_member_edit?
```
jrochkind, line 448: I think we talked about moving all of these helper methods out of the controller to the decorator before maybe?

eddierubeiz: Sorry, only just now getting back to this question. We're not using decorators or presenters to wrap any of this admin-side functionality at the moment. (`app/views/admin/works/_member_list_table.html.erb` is just a garden-variety partial.)

eddierubeiz: I'd certainly consider doing a refactor, though, now we're up to three tabs; something to consider.

jrochkind: We have a sort of standard top-level Draper decorator, I think. 

As opposed to the more custom thing we call a "ViewModel" that acts as a "component", a more standard Draper decorator just "wraps" a model, usually to provide view-specific or context-specific logic -- this does get a bit confusing.  In retrospect, we maybe should have kept these two "kinds" in two different directories, if we had to have two kinds a tall. 

But what I'm thinking about is AudioWorkShowDeocrator in ./app/decorators/audio_work_show_decorator.rb.  That already has "helper" method kind of methods, keeping them out of the controller. Hypothetically, even the partials views can access these helpers, as long the as the decorated model is being passed down as an argument (instead of an undecorated model). I think sometimes we call the argument `decorator`, but sometimes just `model`, but it might really be a decorated model. There might be Reasons (tm) that it won't work for these, but I think it's worth at least a little bit of investigation maybe.

eddierubeiz: OK (per separate discussion in Slack)`AudioWorkShowDecorator` is not being used in the admin pages.

PR #688, opened 2020-04-03
Bug fix for broken audio file download links

### LINKED ISSUE #687: Audio file download links broken
In oral history audio, on the downloads tab, if you open up the "download" menu, and choose a thing from it -- either the Rights link or the Download link -- it doesn't work. 

Instead it does the same thing as if you had clicked on the title -- it loads the segment timecode in the scrubber. 

I can see how this is happening, since you are "inside" the link, it's catching it somehow. 

I'm pretty sure this worked fine before we refactored to single-combined-audio in scrubber. Not sure what the difference was, or if it's worth checking out that old code to compare. 

Not totally sure the best way to fix it. 

Note one reason the "a" is extending the length of the row is because we wanted a click anywhere (but the download menu) to be like a click on the title, we wanted a bigger target for clicks, the whole row, cause that's nice.

eddierubeiz:
Only add the data-ohms-timestamp-s to links that are actually supposed to play audio.

This should fix #687.

### DIFF
--- a/app/views/works/show_with_audio.html.erb
+++ b/app/views/works/show_with_audio.html.erb
@@ -77,13 +77,18 @@
                 <div  class="track-listing" data-role="track-listing"
                       data-title="<%= track.title %>"
                       data-member-id="<%= track.id %>"
-                      data-original-url="<%= main_app.download_path(track.id)    %>"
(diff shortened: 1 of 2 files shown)


jrochkind:
Lovely. I tested it out in dev, and you indeed still can click on anywhere in the row that isn't a 'download' link to load the segment in the player. Great!

PR #690, opened 2020-04-03
Show start times for each segment

eddierubeiz:
Show the start times before each link in the download section. Relies on @jrochkind's method to format the times.
Ref #661


jrochkind:
So we decided just start time was sufficicent. Need to be formatted as `mm::ss` instead of decimal minutes.  I think I already wrote a helper to do this somewhere in some helper file. 

Let's think about ways to typographically put this on the page. 

1. What do you think about putting the start time on a separate line underneath, smaller grey, similar to other places we've done smaller grey second lines for extra info (like for type info underneath combined deriv and PDF)?  Not sure if it needs a label "start time" or not... it might, maybe just "Starts at hh:mm", like a sentance instead of a label/colon. 

3. Alternately, look at table of contents for a guide, it just puts the `hh:mm` start time on the same line as the link, but at the beginning of the line, separated by a wide space. 

Not sure which would work better, if either.

eddierubeiz:
Yeah. I think 2 is better since we're only showing 1 number. It should be obvious from context.

jrochkind:
Let's show to Lee on staging and get approval before merging, otherwise approved!

Issue #689, opened 2020-04-03
OHMS Front End: TOC Segment Scrolling


lsberry16:
If you expand a TOC segment and then expand the next segment without collapsing the previous one, the display scrolls all the way to the bottom of the newly-expanded segment (bad), I guess because it's auto-collapsing the previous one (good). I noticed this specifically in the Renfrew TOC:
https://staging-digital.sciencehistory.org/works/1n79h560c#ohToc

but I don't know that it's limited to that.

jrochkind:
I'm having trouble reproducing this. There is a little bit of "jumping around" that happens, but when it all settles down, for me, the section that you just expanded is properly alligned with the top of it right below the navbar. 

The "jumping around" is unfortunate, but I couldn't figure out how to fix it, using the accordion tools we are using. I also think the weird "jumping around" happens too, at least to some extent, in the standard OHMS viewer and the Brooklyn Historical Society too.

eddierubeiz:
I interpreted this as just "WHEN you change which accordion flap is shown, IF the newly visible flap is taller than the screen, THEN please show the top of that flap, INSTEAD OF the bottom."

jrochkind:
@eddierubeiz Thanks, that helps.  Now that you put it like that, I can reproduce! Sorry, I missed the part about the newly visible section being taller than the screen, good catch. 

Which surprises me, I thought this would already do the right thing. Hmm. I can hopefully fix it.

jrochkind:
Aha, I think I actually already have this fixed in the "search" branch that isn't deployed yet. Was part of my confusion!

I should have fixed it in a separate branch that could have been merged/deployed independently -- will work on extracting it out into it's own thing, so we can get it merged, since it was identified by @lsberry16  as high priority.

PR #693, opened 2020-04-06
fix accordion open off-screen in OHMS ToC

jrochkind:
This had already been done/fixed in the ohms-search branch, but we need to "back port" it to be in master while the unfortunately long-running ohms-search is still under construction. We'll have to deal with merge conflict over in ohms-search branch.

This is a little bit janky, it first opens the accordion, then repositions it to be at the right place. But this seems to be the best we can do with the bootstrap accordion stuff. I did notice similar "open then reposition" even in ohms standard viewer and other ohms-like viewers, it seems to be a common problem -- our fixed navbar makes it even worse. This is the best we can do, and hopefully good enough, we should ask stakeholders to verify it resolves problem reported in #689

### DIFF
--- a/app/javascript/src/js/simple_ohms_player.js
+++ b/app/javascript/src/js/simple_ohms_player.js
@@ -12,3 +12,23 @@ $(document).on("click", "*[data-ohms-timestamp-s]", function(event) {
   html5Audio.currentTime = seconds;
   html5Audio.play();
 });
+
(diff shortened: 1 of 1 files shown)


jrochkind:
@lberry confirms that this seems to resolve sufficiently, please review and merge at your convenience @eddierubeiz

PR #701, opened 2020-04-14
Lock style generation to csl chicago-note-bibliography-16th-edition

jrochkind:
Instead of leaving it `chicago-note-bibliography`, meaning latest version of the chicago style in the csl-styles gem.

We developed our func for the 16th edition styles, and when csl-styles updated to a 17th ed one, it changed some things, that interacted oddly with some of our workarounds. Rather than deal with it, we'll just lock to the style version we developed under for now, so our app behavior remains consistent.

This is an alternative to locking csl-styles gem itself, now we can allow updates to the gem, but keep using the locked stylesheet.

See https://github.com/inukshuk/csl-styles/issues/5

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -168,9 +168,4 @@ gem 'aws-sdk-core'
 
 # we use for data structures for citation models, and for generating citations
  gem "citeproc-ruby", '~> 1.0'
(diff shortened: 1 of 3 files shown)


eddierubeiz:
This makes sense. Merging.

PR #702, opened 2020-04-14
Update to webpacker 5.x

jrochkind:
Did this by:

1. Change Gemfile to ask for webpacker 5.x
2. `bundle update` (to change Gemfile.lock)
3. `yarn upgrade @rails/webpacker --latest` (changes package.json and yarn.lock)

Webpacker 5 should have no real backwards incompats that affect us, so this should be smooth.

See: https://prathamesh.tech/2020/03/25/webpacker-5-0-released/


jrochkind:
Aha, we'd actually need to update our Node version to update to webpacker 5. 

    Webpacker requires Node.js ">=10.13.0" and you are using v8.17.0

Will create another ticket.

Issue #706, opened 2020-04-16
Log all downloads of combined audio derivatives in Google Analytics.


eddierubeiz:
Downloads of the new combined audio derivatives are not yet being tracked in GA.

PR #713, opened 2020-04-20
Log downloads of combined audio derivatives in Google Analytics

eddierubeiz:
Ref #706 
I'm using as the `eventAction`:
```data-analytics-action="download_combined_audio_derivatives"```
which I assume does not have to be registered with the GA interface before it Just Works (tm).

### DIFF
--- a/app/views/works/_show_with_audio_downloads.html.erb
+++ b/app/views/works/_show_with_audio_downloads.html.erb
@@ -51,7 +51,11 @@
   </h2>
   <div class="combined-audio-download-links">
     <div class="combined-audio-download-icon-container">
-      <a href="<%= decorator.combined_mp3_audio_download %>" download>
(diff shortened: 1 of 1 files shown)


eddierubeiz:
See also 
https://developers.google.com/analytics/devguides/collection/analyticsjs/events

jrochkind:
So it should match the GA metadata from our existing download links, we should use consistent metadata so it's put in the report consistently. Can you confirm you checked that? 

You can also confirm it's actually being caught by our GA reporting code that should be looking for those data- attributes, by putting a debugger breakpoint in the debugger code or whatever. Please do confirm it's working instead of assuming.

eddierubeiz:
Yeah, the code is indeed getting submitted by `custom_google_analytics_events.js`. The actual value I'm using for `data-analytics-action` is new, but that's a good thing since we're talking about a new type of event.

PR #714, opened 2020-04-20
Stop using tzinfo gem

jrochkind:
It is only needed on Windows or JRuby, which we don't use. Having it in the Gemfile results in the confusing error message:

> The dependency tzinfo-data (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mswin32, x64-mingw32, java. To add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mswin32 x64-mingw32 java`

Every time you do a bundler thing.

Trying to follow it's `--add-platform` directions just seems to result in other problems. I don't think we need it, let's just comment it out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -141,7 +141,8 @@ group :test do
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Noted.

PR #716, opened 2020-04-21
Handle either shrine 2.x or shrine 3.x file not found exception classes

jrochkind:
Shrine 3 helpfully standardizes the exception raised from storage when a file is not found. But that is a backwards compat.

In order to safely and easily get ready for shrine 3, we're making some changes such that tests will pass and the app will work on either shrine 2 OR shrine 3, make our app ambidextrous, with some tiny easily comprehended non-dangerous changes.

This is one of them.

### DIFF
--- a/app/services/combined_audio_derivative_creator.rb
+++ b/app/services/combined_audio_derivative_creator.rb
@@ -73,15 +73,11 @@ class CombinedAudioDerivativeCreator
     @components ||= begin
       result = []
       audio_member_files.each do |original_file|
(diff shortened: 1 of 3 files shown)


jrochkind:
@eddierubeiz This is ready again.

eddierubeiz:
Thanks for making the `app/services/combined_audio_derivative_creator.rb` fail loudly on failing to download a file from S3 -- will save us a world of pain, I imagine.

Issue #708, opened 2020-04-17
UI to remove uploaded XML file from oral history record


lsberry16:
I uploaded Alyea's XML file to the Digital Collection site and then realized the italics in a segment title didn't display properly (separate issue). I was going to take it back down off the site while I fixed the XML and then reupload, but I don't think there's a way to take it back down once I've added it. (Maybe I missed it, though.) We may need to be able to do that easily in the future.

jrochkind:
Yes, there should be a way to remove XML files, I agree. 

For now I will go this manually on the back-end for Alyea record cf95jc49c ...  removed!

Issue #686, opened 2020-04-02
Allow access to OHMS xml files once they're uploaded into the Digital Collections.


eddierubeiz:
It doesn't make sense for users to *only* be able to upload XML, and then never be able to see it again.
Could be a download button (suggested by @jrochkind )
And / or possibly a "show raw xml" ?

eddierubeiz:
@jrochkind argues there's probably no need for "show".

eddierubeiz:
This could arguably be addressed in the same PR as the related #708 .

PR #718, opened 2020-04-21
Download or remove OHMS xml file

eddierubeiz:
Adding two buttons on the OHMS tab of the work admin page.
Created a partial, as the template was getting on the large side.
Ref #708 
Ref #686 
Still needs tests.


jrochkind:
I don't think this really needs tests, honestly.

eddierubeiz:
I can see what you're saying. In that case...

app/controllers/admin/works_controller.rb
```diff
@@ -96,6 +97,18 @@ def submit_ohms_xml
...
+  def remove_ohms_xml
+    @work.oral_history_content!.update!(ohms_xml_text: nil)
+    redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), notice: "OHMS XML file removed."
+  end
+
+  # GET /admin/works/ab2323ac/download_ohms_xml
+  def download_ohms_xml
+    send_data @work.oral_history_content!.ohms_xml_text,
+      :type => 'text/xml; charset=UTF-8;',
+      :disposition => "attachment; filename=#{@work.oral_history_content!.ohms_xml.accession}.xml"
```
jrochkind, line 110: Rather than construct the header manually, we have a gem we use in other places, that helps avoid errors. Like special encoding needed if there's unicode in the filename, which there shouldn't be in this case, but just DRY up to use the utility is safer...

```ruby
:disposition => ContentDisposition.format(disposition: "attachment", filename: "#{@work.oral_history_content!.ohms_xml.accession}.xml")
```

Already have this gem available and used in our project. https://github.com/shrinerb/content_disposition

PR #720, opened 2020-04-22
Proper logic for adding directly to 'storage' in shrine 2 or 3, with condition check

jrochkind:
Part of attempt to get ready for shrine 3 with small changes that leave the app working in both shrine 2 and shrine 3.

This does pass tests (and is necessary to pass tests) when run directly against the in-progress branch of kithe I have that uses shrine 3. But that isn't demo'd by CI, and I think it's too much trouble to try to get CI running like that, too much is still up in the air. So it's a bit 'dangerous' that the branch of code for shrine 3 added here is not actually used by any regular tests; but I still think this is an overall helpful approach.


eddierubeiz:
No problem with this --  as long as we get rid of it after we upgrade Shrine!

Issue #721, opened 2020-04-23
Zip downloads trigger a NameError


eddierubeiz:
The  rubyzip object isn't being invoked correctly.

https://app.honeybadger.io/projects/58989/faults/63143233

/app/services/work_zip_creator.rb:36:in `create'
34     derivative_files = []
35 
36     Zip::File.open(tmp_zipfile.path, Zip::File::CREATE) do |zipfile|
37       # Add attribution as file and zip comment text
38       zipfile.comment = comment_text

eddierubeiz:
Readily reproducible in staging, but not in dev.

eddierubeiz:
In the staging console, requiring `'zip'` before invoking `Zip::File` fixes the problem.

eddierubeiz:
Fixed in production after deploying #722

PR #722, opened 2020-04-23
Require zip (fixes rubyzip NameError)

eddierubeiz:
We now need to explicitly require the rubyzip gem before invoking it.
#721


jrochkind:
If require zip is necessary, another option would be requiring it in a central place like config/application.rb, such that it is required on app boot and then just there. 

But this is also fine for now. Thanks for taking care of it!

PR #724, opened 2020-04-23
model.reload is neccessary to pass in shrine3, doesn't hurt in shrine 2

jrochkind:
We're doing some weird hacking to prepare a test object, in shrine3 if you manually change data in the underlying column, the shrine file_attacher doesn't pick it up unless you do something to make it do so, in this case model.reload works and is good enough.


eddierubeiz:
Ah. Glad the test was written the way it was then :)

PR #726, opened 2020-04-27
update kind-of npm dependency

jrochkind:
Github warned us about a security vuln to kind-of CVE-2019-20149, Vulnerable versions: = 6.0.2

We have LOTS of versions of kind-of in our yarn.lock, but only 6.x are relevant here.

Manually edited `yarn.lock` to delete the block for ` kind-of@^6.0.0, kind-of@^6.0.2:`, that had been using 6.0.2. ran `yarn-install`, it replaced it with 6.0.3 in yarn.lock, the diff you see here.


eddierubeiz:
See also #725, resolved using the same procedure.

PR #732, opened 2020-04-30
Footnotes

eddierubeiz:
Loosely based on
http://hiphoff.com/creating-hover-over-footnotes-with-bootstrap/#return
Includes ARIA labels.

- Some validation of footnotes markup , in #743


jrochkind:
I see the commit message “deal properly with references to empty or missing footnotes” — a better commit message would explain at least briefly what “properly” is. Its not apparent from the code. What do you now consider the “proper” thing to happen with these?

Also you ABSOLUTELY need a test for this. Like please add a test immediately that has the bad data and tests how it outputs. It is dangerous not to add this right away when you add the fix, we will forget, or have trouble finding the bad data or get confused about what it is.

This is a classic example of calling for adding a test that fails before fixing, and then passes after fixing — TDD.

eddierubeiz:
Note to self:
Need to test for the case where one line contains several footnotes, too!

app/models/oral_history_content/ohms_xml.rb
```diff
@@ -43,24 +43,43 @@ def index_points
...
         #      [[footnotes]]
         #
-        #     [[note]]William E. Hanford (to E.I. DuPont de Nemours &amp; Co.), &quot;Polyamides,&quot; U.S.
-        #     Patent 2,281,576, issued 5 May 1942.[[/note]]
+        #         [[note]]William E. Hanford [...] issued 5 May 1942.[[/note]]
+        #         [[note]] [ another footnote ...] [[/note]]
+        #         ( and so no)
         #
         #      [[/footnotes]]
         #
```
jrochkind, line 85: remove commented out stuff

app/presenters/ohms_transcript_display.rb
```diff
@@ -43,6 +43,27 @@ def display
...
+    # upload time and rejecting ones that are invalid.
+    # We can thus assume the footnotes are valid at display time;
+    # if the array of footnotes doesn't contain a footnote for `number`,
+    # we are *not* going to throw an error.
+    raw_footnote = model.footnote_array[number.to_i - 1]  || ''
+    if raw_footnote == ''
+      Rails.logger.warn("WARNING: Reference to empty or missing footnote #{number} for OHMS transcript #{model.accession}")
+    end
+    render "works/ohms_footnote_reference",
+      footnote_text: raw_footnote.gsub('"', '&quot;'),
```
jrochkind, line 64: I'm suspicious of this gsub -- if we want to escape for HTML, we should use a proper existing HTML escaping routine, not try to re-create one that only escapes quotes.

jrochkind: should be a test around this -- create a test with that input has HTML special characters that aren't meant to be HTML tags, test teh output is it what you expect?

app/presenters/ohms_transcript_display.rb
```diff
@@ -43,6 +43,27 @@ def display
     content_tag("div", safe_join(paragraph_html_arr), class: "ohms-transcript-container")
   end
 
+  def test_render
+    render "works/ohms_footnote_reference"
+  end
```
jrochkind, line 48: is this method supposed to be called `test_render`?

app/views/works/show_with_audio.html.erb
```diff
@@ -105,6 +105,10 @@
         <% if decorator.has_ohms_transcript? %>
           <div class="tab-pane long-text-line-height" id="ohTranscript" role="tabpanel" aria-labelledby="ohTranscriptTab">
             <%= OhmsTranscriptDisplay.new(decorator.oral_history_content.ohms_xml).display %>
+            <% if decorator.oral_history_content.ohms_xml.footnote_array.length > 0 %>
+              <%= render "ohms_footnotes_section",
+                footnote_array: decorator.oral_history_content.ohms_xml.footnote_array  %>
+            <% end %>
```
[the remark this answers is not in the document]
jrochkind: When we call `OhmsTranscriptDisplay.new(decorator.oral_history_content.ohms_xml).display`, it should render HTML that already has the "ohms_footnotes_section" in it. 

This is just part of OhmsTranscriptDisplay, we'd never want to do one without the other.

jrochkind:
Remove the "blinky" on reference link

app/presenters/ohms_transcript_display.rb
eddierubeiz: That was a typo left over from testing. Got rid of it.

app/presenters/ohms_transcript_display.rb
eddierubeiz: Done!

app/views/works/show_with_audio.html.erb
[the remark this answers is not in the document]
eddierubeiz: Good call; done.

jrochkind:
@eddierubeiz Sorry, so some more. 

How you fixed that rspec render error with `render "works/ohms_footnotes_section"`... I think reveals an actual problem. 

It's not something special about rspec that made it unable to find it. It's becuase you were relying on the current lookup path for the current controller. 

When we move to ViewModels, we try to make ViewModels totally independent and de-coupled from any controller, so they can be called without assuming a particular controller context.  Which is what the rspec tests were doing, so that's why it was failing. 

While your fix does that fix that, it points out an inconsistency in our conventions... usually for ViewModels, we put the templates to be rendered inside `app/views/presenters`, and then render them like `render "/presenters/some_template_name"`. 

Note the beginning `/` in `/presenters`, which tells Rails _not_ to do any relative lookup based on the current controller context, but to just insist on top level `/presenters/some_template_name`. 

Can you move these templates around and do this here, to be consistent, for the same reasons?

Also, this may be the first ViewModel that has *two* templates it uses. One *could* say it should be extracted out into more than one ViewModel, to have one-to-one view model to template, and when we move to `view_component` as in #742 it might require that, but I don't think we need to insist on it now.

app/models/oral_history_content/ohms_xml.rb
eddierubeiz: Done!

PR #741, opened 2020-05-05
WIP fix WorkZipCreator returning Tempfile with empty content

jrochkind:
Fails on problem causing #740

Somehow even though the Zip existed at the path the Tempfile was set to, the Tempfile itself returns empty string when `read` for content. 

Somehow that results in the `Aws::S3::Errors::XAmzContentSHA256Mismatch` when asking S3 SDK to upload it. Not sure why. But clear what to fix.

### DIFF
--- a/app/services/work_zip_creator.rb
+++ b/app/services/work_zip_creator.rb
@@ -69,6 +69,9 @@ class WorkZipCreator
       end
     end
 
+    # tell the Tempfile to (re)open so it has a file handle open that can see what ruby-zip wrote
+    tmp_zipfile.open
+
     return tmp_zipfile
   ensure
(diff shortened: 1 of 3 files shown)


jrochkind:
In the end, just had to tell `temp_zipfile.open` to reopen it so it could see the content written by the Ruby::Zip process. 

This has probably always been a bug, but one that somehow shrine 2 accidentally forgave, but shrine 3 does not. 

That was a lot of hours to figure out a simple fix!

jrochkind:
confirmed fix on staging

eddierubeiz:
OK! This is just bizarre, but OK. Merging.

Issue #743, opened 2020-05-06
Robust validation on footnotes in OHMS xml


eddierubeiz:
Footnotes are stored in OHMS following an easy-to-understand syntax (`[[footnote]]` tags need to be closed; each footnote reference should refer to exactly one footnote; footnotes are numbered with consecutive positive integers; and so on).

With dozens of footnotes spread out over hundreds of pages in (soon) dozens of oral histories, small errors will inevitably creep in, leading to bugs that are unpredictable and often tedious to diagnose.

To ensure we won't waste a lot of time in the future hunting down the *one* footnote typo that's breaking a particular oral history, it is worth:

- listing the rules that describe the syntax;
- making sure they are enforced when users upload OHMS XML;
- in certain cases, making sure our display code is somewhat lenient, so it behaves reasonably when a particular rule is broken;
- adding tests to describe how our code responds to each of these rules.

jrochkind:
@eddierubeiz Please link this to the original related issue by mentioning issue number here?

eddierubeiz:
#732 is the general "Footnotes" issue; don't close it before we deal with validation.

PR #746, opened 2020-05-07
More robust validation for ohms XML

eddierubeiz:
Ref #743  .
The current validation for OHMS XML checks for malformed XML.
This PR adds checks preventing the upload of:
- Footnotes with no references to them;
- References to nonexistent footnotes.
It also lists the problems to the user on the front end at upload time rather than just noting "Could not accept invalid OHMS XML", as there are more things that can go wrong.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -103,7 +103,7 @@ class Admin::WorksController < AdminController
     else
       Rails.logger.debug("Could not accept invalid OHMS XML for work #{@work.friendlier_id}:\n  #{xml.slice(0, 60).gsub(/[\n\r]/, '')}...\n\n  #{validator.errors.join("\n  ")}")
       redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), flash: {
-        error: "OHMS XML file was invalid and could not be accepted: #{validator.errors.join(';<br/>')}"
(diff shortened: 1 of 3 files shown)


jrochkind:
Just to see what would happen,I tried deleting a close tag `[[/footnote]]`, I deleted the one after footnote 1. 

The error message is 

> OHMS XML file was invalid and could not be accepted: Missing reference(s) to footnote(s): 1

Is that clear enough, or would it be better to have an error message that understood it meant a footnote tag hadn't been closed, there was mismatch between opening and closing tags?

Similarly, missing the closing `[[/footnotes]]` tag is an error staff have actually made before. I tried deleting that to see what happened. It turns into:

> OHMS XML file was invalid and could not be accepted: Reference to missing footnote 1;<br/>Reference to missing footnote 2;<br/>Reference to missing footnote 3;<br/>Reference to missing footnote 4;

1. I think it might be better to catch this and say "unclosed [[footnotes]]" tag, or at least "missing [[footnotes]] section entirely", 

2. We can see that HTML is not allowed in these top of screen error messages, the `<br>` is showing up escaped.

eddierubeiz:
It is a common enough error. I wrote some code to detect missing closing [[/notes]] and [[/footnotes]] and gives you a hint about where to look.

jrochkind:
Tested manually. This is great, love it! Nice work. Merging.

Issue #729, opened 2020-04-27
test current OHMS behavior around 'piled up' timecodes


jrochkind:
See for example the Smythe transcript, search for `smyth personal rec`. 

Due to "dead air", there are multiple timecodes all at the same point in the "transcript" (since no transcript corresponds to them).  (In addition to dead air, this can happen when we 'skip' a portion of the recorded audio transcript). 

the standard OHMS viewer handles this pretty poorly. 

Our custom viewer handles it better -- but it wasn't on purpose, it's kind of an accident! It looks like our custom viewer may be ignoring all but the first timecode in a "line". (Or all but the last?)

Now that we want this intentional behavior, we should write a test verifying it, so it stays true. 

(Confirm with @lberry that "only use first timecode" is preferable to "only use last").

jrochkind:
Lee says when a string of timecodes in "dead space" exist, the FIRST one should be displayed not the ast.

PR #750, opened 2020-05-12
Handle "pileups" of timestamps better

eddierubeiz:
Currently, the code ignores multiple timestamps on the same line, except for the last one. This new PR will do almost exactly the same thing: per a discussion with Lee on Slack, we actually want to ignore **all consecutive** timestamps. This will result in a slightly different display when there are long gaps in the audio.

Ref #750

~~The new test ensures that if there's a long interval in the OHMS transcript with no timestamps associated with it, only the timestamps before and after the interval will be printed out.
This test does not change the behavior of the transcript display; instead, it **describes** the current behavior on the assumption that what we currently have is what we want. This is illustrated below with some screenshots, as it's confusing to discuss without examples.~~

Note that in the `https://ohms.uky.edu/sync.php`, the empty chunk is associated with a number of timestamps with no text associated with them ( I believe this is what we've been referring to as the "pileup".)

~~Note also that these timestamps are *not* represented in the XML as ingested into the Digital Collections.~~ <-- This is incorrect, I'm happy to report.

### DIFF
--- a/app/models/oral_history_content/ohms_xml.rb
+++ b/app/models/oral_history_content/ohms_xml.rb
@@ -102,12 +102,13 @@ class OralHistoryContent
     # It looks like: 1:|13(3)|19(14)|27(9)
     #
     # We believe that means:
-    # * `1:` -- 1 minute granularity, so each element is separated by one minute.
-    # * "13(3)" -- 13 line, 3rd word is 1s timecode (as it's first element and 1s granularity)
-    # * "19(14")  -- 19th line 14th word is 2s timecode
+    # * `1:`      -- 1 minute granularity: each timecode is separated by 60 seconds.
+    # * "13(3)"   -- Timecode 1: minute 1 ends at line 13, word 3.
+    # * "19(14)"  -- Timecode 2: minute 2 ends at line 19, word 14.
+    # * "27(9)"   -- Timecode 3: minute 3 ends at line 27, word 9.
     # * Etc.
     #
-    # OHMS seems to actually ignore the word position in placing marker, we may too.
+    # OHMS seems to actually ignore the word position in placing marker; we may too.
     def parse_sync!
       sync = parsed.at_xpath("//ohms:sync", ohms: OHMS_NS).text
       return {} unless sync.present?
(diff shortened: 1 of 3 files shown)


jrochkind:
So what's currently being tested is current already existing behavior? Do you have a sense of *why* that happened, what code led to that?  I wasn't expecting it, so I'm curious! 

> only the timestamps before and after the interval will be printed out.

Hmm, my understanding of what was spec'd was that if there was a long interval, like let's say 10 timecodes in a row, we wanted only the FIRST of those timecodes to be output. 

As opposed to 1st and 10th, or -1st and 11th. 

But this stuff is confusing to talk about; and it's possible what it's doing is good enough and/or better anyhow. Maybe show/tell it with @lberry1 to confirm?

jrochkind:
@eddierubeiz Try to remember to link PR's to Issues, in this case #729, thanks!

eddierubeiz:
Here's an image of the corresponding OHMS sync screen.
You can also go to https://ohms.uky.edu/sync.php?id=89689 and search for the word "monocle".
timecodes

eddierubeiz:
The 25 minutes after 4:23:00 and before 4:28:00 do *not* have timecodes in the OHMS xml as uploaded into the Digital Collections.

eddierubeiz:
Meanwhile, here's the output on our site.
You can also just search for the word "monocle" a https://digital.sciencehistory.org/works/4j03d097t#ohTranscript

our_ui

eddierubeiz:
My understanding is that the current behavior *is* considered correct by Lee, and this test describes the current functioning of the system. @lberry, please let me know if I'm wrong!

eddierubeiz:
(Correct: what's currently being tested is current already existing behavior. See the issue for a confirmation that this PR is not supposed to change the actual transcript display behavior.)

jrochkind:
I think I considered desirable behavior to be for 04:24:00 to show up in the output, and the 04:48:00 *not* to show up. 

In this example, it doesn't seem necessary. But in other examples I've seen (I can find one again), without that first timecode showing up, and with the last timecode essentially showing up in the *place* the silence begins -- it has seemed very confusing to me as a reader understanding where the silence begins or ends. 

But if the OHMS XML given to us no longer contains enough information to make it like that, then it's a non-starter anyway. 

We can try to find an example of where it's confusing maybe. It's one of the one's Lee showed us before, on that first day, that I recall.

eddierubeiz:
I'm closing this as an out of date duplicate of #752.
(TL;DR: we are in fact going to change the behavior of the timecode display.)

PR #752, opened 2020-05-16
WIP Ignore consecutive OHMS timecodes.

eddierubeiz:
Ref. #750 .
Related to #729 .
Changes the code to ignore all *consecutive* timecodes, per our discussion on Slack.

We could get rid of a bunch of this new code if we changed the rule slightly so that any line with more than one timecode will have *all* its timecodes ignored. This is actually almost the same thing from the point of view of whoever is syncing the interview. But for now I'm following the instructions to the letter.

### DIFF
--- a/app/models/oral_history_content/ohms_xml.rb
+++ b/app/models/oral_history_content/ohms_xml.rb
@@ -119,24 +119,30 @@ class OralHistoryContent
         return {} unless sync.present?
 
         mbt, stamps = sync.split(":")
-        minutes_between_timecodes = mbt.to_i
 
         result = {}
 
         stamps.split("|").enum_for(:each_with_index).each do |stamp, index|
-          next if stamp.blank?
-          stamp =~ /(\d+)\((\d+)\)/
-          line_num, word_num = $1, $2
-          next unless line_num.present? && word_num.present?
-
-          (result[line_num.to_i] ||= []) << {
-            word_number: word_num.to_i,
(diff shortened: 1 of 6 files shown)


app/models/oral_history_content/ohms_xml.rb
```diff
@@ -94,36 +94,90 @@ def initialize(xml_node)
...
 
-    # A hash where key is an OHMS line number. Value is a Hash containing
-    # :word_number and :seconds .
-    #
-    # We parse the somewhat mystical OHMS <sync> element to get it.
+    # A hash where key is an OHMS line number. Value is a hash containing
+    # word_number and seconds.
+    def parse_sync!
+      remove_consecutive_timecodes(raw_timecodes)
+    end
```
jrochkind, line 101: Not a big deal, but any reason to have this just be a method that calls another method, instead of only one method? But there might be a reason!

jrochkind:
@eddierubeiz Looks great, nice job! Code-review-approved. Not merging since it's still marked Draft, not sure if you're ready.

app/models/oral_history_content/ohms_xml.rb
eddierubeiz: Good call.

jrochkind: I mean there might actually be a reason, it might be fine. Maybe I was just reaching for something to comment on I wouldn't worry about it actually.

Issue #734, opened 2020-04-30
"Create combined audio derivatives" button should always be available


eddierubeiz:
Regardless of the status of the combined derivs for a work.
(Per our discussion on Slack.)

Issue #711, opened 2020-04-20
Work admin OHMS page: list number the of segments incorporated in the (already-created) combined audio derivatives.


eddierubeiz:
Just to provide an additional cue to the user.
Small change but I believe this would be helpful.

Issue #704, opened 2020-04-16
Hide "combined audio derivatives" button if there are zero published segments in an oral history


eddierubeiz:
Instead, show copy to the effect that:
* only published segments are incorporated into the combined audio derivatives, per our policy
* thus we have no segments from which to create said derivatives
* thus we're not showing the button.
This is related to #674 but is much simpler to implement.

eddierubeiz:
This actually conflicts with #734 but I'll at least show an error message.

jrochkind:
Not totally sure it does; I see a difference between refusing to show the button becasue the software thinks "this is not needed, you already have a derivative", and refusing to show the button because the software thinks "I am literally UNABLE to create the derivative, because there are no source files". 

But also not totally sure I see your point too. Do what you think is best for both UI and code simplicity/clarity!

PR #753, opened 2020-05-16
WIP Adjustments to the UI around the combined audio derivs admin interface

eddierubeiz:
Ref #734 
Ref #704 
Ref #711 

- Use `work_published_audio_members.count` rather than `work_audio_members_count` in figuring out whether to display the "redo combined audio derivs button".
- Offer the user a chance to recreate them even if they are considered "up to date".
- Also, list the number of segments incorporated into the already-created combined audio derivs.

TODO
- [x] Actually hide the button if there are no published segments (per @jrochkind 's comment on 704.)
- [x] Tests will need to be adjusted.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -126,6 +126,13 @@ class Admin::WorksController < AdminController
   # PATCH/PUT /admin/works/ab2323ac/create_combined_audio_derivatives
   # Unfortunately, if the job fails for any reason, the user will not be notified.
   def create_combined_audio_derivatives
+    if CombinedAudioDerivativeCreator.new(@work).audio_members.empty?
+      redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), flash: {
+        error: "Combined audio derivatives cannot be created, because this oral history does not have any published audio segments."
+      }
+      return
+    end
+
     CreateCombinedAudioDerivativesJob.perform_later(@work)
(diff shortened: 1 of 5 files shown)


app/controllers/admin/works_controller.rb
```diff
@@ -126,6 +126,13 @@ def download_ohms_xml
   # PATCH/PUT /admin/works/ab2323ac/create_combined_audio_derivatives
   # Unfortunately, if the job fails for any reason, the user will not be notified.
   def create_combined_audio_derivatives
+    if CombinedAudioDerivativeCreator.new(@work).audio_members.empty?
```
jrochkind, line 129: Small thing, but to be a little bit more "law of demeter" encapsulated, let's make a method for this on `CombinedAudioDerivativeCreator`.  Like `if CombinedAudioDerivativeCreator.new(@work).no_input_available?` or `~ CombinedAudioDerivativeCreator.new(@work).available_members?`. 

Then you can use the same method *internal* to `CombinedAudioDerivativeCreator` to be sure that we're using the same logic here to decide to error redirect that the `CombinedAudioDerivativeCreator` is using to decide if it can do it's thing.

jrochkind:
looks good

app/controllers/admin/works_controller.rb
eddierubeiz: Good call.

PR #757, opened 2020-05-18
update to kithe 2.0.0.alpha2 with derivative migration rake task

jrochkind:


jrochkind:
(Ran tests locally all passed, merged early)

PR #765, opened 2020-05-26
Fix WorkZipCreator for new style derivatives

jrochkind:
In 3f12510b00ec we accidentally changed a `||` to a `&&`, breaking logic. Change it back. 

Tests didn't catch it, because it didn't break jpeg originals, but did break tiff originals. And our tests generally use jpeg originals. Which is kind of a mistake, since it's not realistic, our actual original sare almost never jpg and usually tiff. But fixing the test data will be left for another PR. Not worth trying to fix the tests here IMO. Problem solved, fixing production post derivative migration

### DIFF
--- a/app/services/work_zip_creator.rb
+++ b/app/services/work_zip_creator.rb
@@ -107,7 +107,7 @@ class WorkZipCreator
                             where(published: true).
                             order(:position).
                             select do |m|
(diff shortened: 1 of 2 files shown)


jrochkind:
@eddierubeiz I'm going to go ahead and merge after it passes!

Issue #694, opened 2020-04-07
Improve cap copy_data task to copy oral history content


jrochkind:
The OralHistoryContent model, and the combined derivatives.

jrochkind:
@HKativa I'd like to propose bumping this up in priority, perhaps for Eddie to work on after he's gotten through footnotes. It's lack is making it a bit harder for us to debug problems.

PR #768, opened 2020-05-27
include oral_history_content sidecar in copy_data serialization/restore

jrochkind:
Turns out this was all it took. Ref #694


eddierubeiz:
Thank you so much for doing this!

Issue #674, opened 2020-03-30
Combined audio derivatives: keeping users posted about the status of these background jobs.


eddierubeiz:
The combined audio derivatives take a while to complete (often well over 10 minutes.)
We already have them running in a background job. What are some ways we could let users know what state the job is in? Is it:
- Not started yet?
- Started, not complete yet?
- Started, but failed?
Could we provide the user with an estimate of how long the job might take?

eddierubeiz:
@jrochkind  suggests:
"we could add a ‘status’ column to OralHistoryContent.
Set it to “queued” when queuing a job; “in progress” when the job starts; and “complete” or “error” when it’s done or errored.
Then we could show that status on the admin page."

eddierubeiz:
Other idea: keep users posted on start, success or failure via email.

eddierubeiz:
Or just encourage users to check the job queue ?

jrochkind:
I don't expect the job queue is a usable UI for... really anyone, definitely not anyone that's not a developer. Not even totally sure it's possible to identify from a large queue which job corresponds to what thing. 

I don't think emails are much easier and are I think a very substandard UI.

jrochkind:
We could look into/discuss using a 'generic' ActiveJob solution for tracking job status.

PR #772, opened 2020-05-28
Show status of audio derivative creation on front-end

eddierubeiz:
Ref #674 .


app/models/oral_history_content.rb
```diff
@@ -64,6 +66,12 @@ def has_ohms_index?
     @has_ohms_index ||= ohms_xml&.index_points&.present?
   end
 
+  def set_combined_audio_derivatives_creation_status(status_string)
+    self.combined_audio_derivatives_creation_status = status_string
+    self.combined_audio_derivatives_creation_status_changed_at = DateTime.now
+    self.save!
+  end
```
jrochkind, line 73: Let's not use any arbitrary string for this, including things like `"Error: #{e}"`. 

Let's use a controlled vocabulary that can only be a limited set of things. I would imagine `queued`, `in_progress`, `success`, and `failure`.  (**Not sure if you are currently recording/tracking "queued", but we should** Should be able to override a method in the ActiveJob to automatically set to queued when the job is queued....).  I think these are the right states: just "complete" isn't clear if it means success or failure, those are two different states. 

That would allow us to do things like easily search on all failed ones, etc. 

One way to do that is to use the ActiveRecord "enum" feature. https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html  By default that will store integers in the database (and want an integer column), but translate them to words in code. You can also have it store actual strings in a string column, although undocumented, with:

    enum combined_audio_derivatives_creation_status: { queued: "queued", in_progress: "in_progress" }

Then it automatically gives you methods like `sidecar.queued?` and stuff like that. 

If you think I'm being overly persnickety let me know; but I just don't like trying to figure out if something is an error or not by looking if the message starts with "Error:", the semantics should be separated from the user-displayed text like this. We might want to put an error in red, or search for all errors, etc.  It's possible there are other easier ways to do this than ActiveRecord `enum`, but it is a feature meant for it. 

It does mean if you wanted to store error info, you'd need a separate column. To begin wtih, we could just not store error info -- it should get a honeybadger report already, I think?

Next, let's get rid of the special `set_*` method, and just let the caller use ordinary `sidecar. combined_audio_derivatives_creation_status = whatever`. How do we get that to automatically set the "changed_at"? The easiest way is we can actually override the method behind that, turns out when you do `something.something = `, that's actually a method!

```ruby
def combined_audio_derivatives_creation_status=(value)
  super
  self..combined_audio_derivatives_creation_status_changed_at = DateTime.now
```

There would be cleverer ways to do that with ActiveRecord callbacks that would work even if the caller used a different method to set the status -- but your original implementation required the caller to use the "right" method anyway, so this is a good improvement. (And should have tests that when you set, it updates the `changed_at`). 

Lastly, `combined_audio_derivatives_creation_status` and `combined_audio_derivatives_creation_status_changed_at` are, phew, really long, right? I guess we were calling the basic thing `combined_audio_derivatives_creation_status` for clarity, so maybe it should match, but what do you think about just `derivatives_status` and `derivatives_status_changed_at`, there's only one derivative in the model that it could apply to!  But not sure about this, maybe the long name is good?

The next step could be making the record use activerecord callbacks to *automatically* update the "changed_at" whenever it changes, so the user

app/presenters/work_combined_audio_derivatives.rb
```diff
@@ -45,4 +45,22 @@ def derivatives_up_to_date?
...
+    @combined_audio_creation_status ||= begin
+      status = combined_audio_creation_status
+      return nil if status.nil?
+      "#{status} as of #{combined_audio_creation_date}"
+    end
+  end
+
+  def combined_audio_creation_date
+    model&.oral_history_content&.combined_audio_derivatives_creation_status_changed_at
+  end
```
jrochkind, line 59: this is not necessarily a "creation" date, right? It could be the date it was set to "in progress" or the date it was marked as an error...

app/models/oral_history_content.rb
eddierubeiz: I did almost exactly this -- of course I thought of it AFTER posting the initial PR which you reviewed. The latest version of the code is in fact adding a controller vocab. But will add queued. :)

app/presenters/work_combined_audio_derivatives.rb
eddierubeiz: Good point!

app/models/oral_history_content.rb
```diff
@@ -32,6 +32,15 @@ class OralHistoryContent < ApplicationRecord
...
   include CombinedAudioUploader::Attachment.new(:combined_audio_webm, store: :combined_audio_derivatives)
 
+  job_status_options = {
+    queued: 'queued',
+    started: 'started',
+    failed: 'failed',
+    done: 'done'
+  }
+
+  enum combined_audio_derivatives_job_status: job_status_options
```
jrochkind, line 42: No reason to have this as a separate local variable first, that's never re-used. Just put the hash inline, I think it's more readable inline, not making the reader look in two places and put it together in their head.

app/models/oral_history_content.rb
```diff
@@ -32,6 +32,15 @@ class OralHistoryContent < ApplicationRecord
   include CombinedAudioUploader::Attachment.new(:combined_audio_mp3, store: :combined_audio_derivatives)
   include CombinedAudioUploader::Attachment.new(:combined_audio_webm, store: :combined_audio_derivatives)
 
+  job_status_options = {
+    queued: 'queued',
+    started: 'started',
+    failed: 'failed',
+    done: 'done'
```
jrochkind, line 39: "done" is ambiguous, cause failed is a form of done too. Can we please use "succeeded", to go with "failed", as originally proposed?

app/jobs/create_combined_audio_derivatives_job.rb
```diff
@@ -2,14 +2,18 @@ class CreateCombinedAudioDerivativesJob < ApplicationJob
...
+    sidecar.combined_audio_derivatives_job_status = 'started'
+    sidecar.save!
     # Generate the derivatives:
     deriv_info = deriv_creator.generate
     if deriv_info.errors
       Rails.logger.error("Unable to create combined audio derivatives for work #{work.friendlier_id}:")
       Rails.logger.error(deriv_info.errors)
+      sidecar.combined_audio_derivatives_job_status = 'failed'
+      sidecar.save!
       return
```
jrochkind, line 15: I know this wasn't being done before either -- but what if we actually raise an exception on errors here, so that honeybadger will report it. Then we can set the status to `failed` in an ActiveJob `rescue_from`, like we've done before -- which will also catch other unexpected errors... is that good or bad?

app/views/admin/works/_combined_audio_derivatives.html.erb
```diff
@@ -8,10 +8,34 @@
...
+              <i class="fa fa fa-thumbs-down pr-1" aria-hidden="true"></i>
+              The combined audio derivatives were not recreated. The job failed <%=view.time_since_job_status_change %>.
+            </p>
+          <% end %>
+          <% if model&.oral_history_content&.done?  %>
+            <p class="alert-success p-3">
+              <i class="fa fa fa-thumbs-up pr-1" aria-hidden="true"></i>
+                The combined audio derivatives, created <%=view.time_since_job_status_change %> from <%= view.work_available_members_count %> published audio segments, are up to date. Their fingerprint is <code class="border"><%= view.combined_audio_fingerprint%></code>.
+            </p>
+          <% end %>
```
jrochkind, line 38: This is way too much logic/code for this staff facing feature. The more branches you have, the more opportunity for bugs and such, the more to maintain. 

I think it should just be:  "Combined audio derivative creation status: #{status}, since #{formatted status changed at time}". Maybe with a red color for error, that's about it. 

Not sure if it should be there at all for `done` (to be changed to `succeeded`) -- if it is, it means it will basically be there forever? Not sure we need it for `done` at all, or alternately only if `done` is in the last X hours (24? I dunno). But I think for "done" it might be okay just that... you see the derivative is there now, that's all the feedback you need?

app/views/admin/works/_combined_audio_derivatives.html.erb
```diff
@@ -38,16 +62,48 @@
...
+          <% if model&.oral_history_content&.started? %>
+            <p class="alert-info p-3">
+            A job to recreate them started <%=view.time_since_job_status_change %>.
+            </p>
+          <% end %>
+          <% if model&.oral_history_content&.failed? %>
+            <p class="alert-danger p-3">
+            A job to recreate them failed <%=view.time_since_job_status_change %>.
+            </p>
+          <% end %>
```
jrochkind, line 79: This is just repeating the status info that was already earlier in this file? Is there a reason we need it in two places?

app/views/admin/works/_combined_audio_derivatives.html.erb
```diff
@@ -38,16 +62,48 @@
...
+              They are currently being created;
+              the job started <%=view.time_since_job_status_change %>.
+            </p>
+          <% end %>
+          <% if model&.oral_history_content&.done? %>
+            <p class="alert-danger p-3">
+              Someone tried to create them,
+              but the job failed <%=view.time_since_job_status_change %>.
+            </p>
+          <% end %>
```
jrochkind, line 106: Wait a THIRD time repeating the same status info? What am I missing?

jrochkind:
oh hey, since we merged structure.sql => schema.rb into master…. your PR that makes schema changes needs to make sure it has master merged in, and the new schema.rb committed (and the new structure.sql NOT accidentally put back). Haven’t checked any of these things myself yet, but I noticed something fishy with a schema.rb.

app/models/oral_history_content.rb
eddierubeiz: No problem.

app/jobs/create_combined_audio_derivatives_job.rb
eddierubeiz: Done!

app/views/admin/works/_combined_audio_derivatives.html.erb
eddierubeiz: Yup -- drastically simplified. Good call.

app/views/admin/works/_combined_audio_derivatives.html.erb
eddierubeiz: Gone now!

app/views/admin/works/_combined_audio_derivatives.html.erb
eddierubeiz: Fixed!

app/jobs/create_combined_audio_derivatives_job.rb
```diff
@@ -1,24 +1,32 @@
...
+    @sidecar.combined_audio_component_metadata = { start_times: deriv_info.start_times }
+    @sidecar.combined_audio_derivatives_job_status = 'succeeded'
+    @sidecar.save!
+  end
 
-    sidecar.save!
+  rescue_from(StandardError) do |exception|
+    @sidecar.combined_audio_derivatives_job_status = 'failed'
+    @sidecar.save!
+    raise
```
jrochkind, line 30: This needs to say `raise exception` to re-raise the same exception you caught, instead of a generic one.

db/migrate/20200527142734_remove_kithe_derivatives_table.rb
```diff
@@ -1,5 +1,6 @@
 class RemoveKitheDerivativesTable < ActiveRecord::Migration[6.0]
   def change
-    drop_table :kithe_derivatives
+    drop_table :kithe_derivatives do |t|
+    end
```
jrochkind, line 4: how'd this end up getting changed in this PR? The change is basically no-op I think, I don't think it'll hurt, also isn't necessary, just not sure how it crept in here

app/jobs/create_combined_audio_derivatives_job.rb
eddierubeiz: Checked -- you're right, and my source was wrong.

db/migrate/20200527142734_remove_kithe_derivatives_table.rb
eddierubeiz: I added it in because your migration, which was chronologically *after* my migration, had been preventing me from reversing my migration. No longer needed.

Issue #684, opened 2020-04-02
OHMS front end: segment-level links to other works


eddierubeiz:
A segment can contain as part of its metadata a link to another work. We should consider displaying these in a manner comparable to e.g. https://ohms.uky.edu/preview/?id=89262, click on "Du Pont and Tetraethyl Lead".

Lee Berry:
_I further improved Plunkett this morning in another way that currently only works in the OHMS viewer.
I added hyperlinks to the oral history interviews in the digital collection with people Plunkett mentions in the appropriate TOC segments--you can see them in the segment that starts at 57:53 and the one that starts at 1:47._

jrochkind:
Ah, thanks for tracking that down @eddierubeiz 

It's just metadata on a segment; should be pretty straightforward to enhance our viewer to include these, if prioritized above other things, could be something you could do @eddierubeiz . 

I wouldn't necessarily lay it out on the page just like that, I'd probably use the "Related URL" heading (same one we use on work metadata), just a single heading, either styled to match our existing segment metadata ("inline" rather than a value per line) would be easiest to avoid having to do CSS/HTML refactoring, but if not good enough HTML/CSS refactoring could be done.

jrochkind:
Also let's just make sure to remind Lee (I forgot and can't find her github account) that we already have work-level Related Links functionality that can be used to link one OH to another (unconnected to OHMS), and confirm that additional segment-level metadata really is desired/required. 

The fewer features we have, the fewer bugs and fewer maintenance hours on them.

jrochkind:
Was selected/prioritized by @HKativa / @lsberry16 

While standard OHMS viewer has a list of links each one of which says "Hyperlink: [url]", I think it probably works better with a single heading "References" and/or "Links", followed by list of links. 

Compare to our work-level references on whole bib, maybe use same language and similar format for consistency. (Although in this case we have link text, not just URLs, which is great).

PR #774, opened 2020-05-28
ohms index segment hyperlinks

jrochkind:
Ref #684


jrochkind:
Ready for code review @eddierubeiz , although no rush, thanks!

PR #780, opened 2020-06-03
search fulltext with solr highlights

jrochkind:
There's a "feature flag" in here at present, so even if merged to master/production, changes to search relevancy and highlighting are by default not active on production. 

Ref #785


jrochkind:
Not ready to deploy this yet, but we can merge it, as it's hidden behind a feature flag. Then we can keep doing other stuff necessary to be ready to deploy, from #785

PR #782, opened 2020-06-04
Import code for full-text indexing of oral histories

eddierubeiz:
Assumes the presence of files at `/tmp/ohms_transcript_files/`, but you can override it using ` FILES_LOCATION`.

### DIFF
--- a/lib/tasks/add_transcripts_to_oral_histories.rake
+++ b/lib/tasks/add_transcripts_to_oral_histories.rake
@@ -1,0 +1,52 @@
+namespace :scihist do
+  desc """
+    Goes through all the oral histories and adds transcripts from a file on the disk, for those missing them:
+
+    bundle exec rake scihist:create_full_length_audio_derivatives
+  """
+
+  task :add_transcripts_to_oral_histories => :environment do
+    files_location = '/tmp/ohms_transcript_files/'
+    progress_bar = ProgressBar.create(total: Work.where("json_attributes -> 'genre' ?  'Oral histories'").count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
+    Work.where("json_attributes -> 'genre' ?  'Oral histories'").find_each(batch_size: 10) do |w|
+      accession_num =  w.external_id.find { |id| id.category == "interview" }&.value
+      unless accession_num
+        progress_bar.log("ERROR: #{w.title}: no accession number.")
+        progress_bar.increment
+        next
+      end
+      filename = "#{files_location}#{accession_num}.txt"
+      unless File.file?(filename)
+        progress_bar.log("ERROR: #{w.title}: couldn't find a file on disk for #{accession_num}.")
+        progress_bar.increment
+        next
+      end
+
+      begin
+        full_text = File.read(filename)
+      rescue Errno::ENOENT => e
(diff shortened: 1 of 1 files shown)


jrochkind:
Change title of this I think? It's actually about importing searchable transcripts i think? We have many other PR's about searchable transcripts in general! thanks!

jrochkind:
> Assumes the presence of files at /tmp/ohms_transcript_files/.

Would it make sense to make this an optional argument, for convenience? It can still use that as a default. May not matter, if we run this as rarely as expected, but should be easy to do?

lib/tasks/add_transcripts_to_oral_histories.rake
```diff
@@ -0,0 +1,57 @@
...
+    bundle exec rake scihist:add_transcripts_to_oral_histories
+  """
+
+  task :add_transcripts_to_oral_histories => :environment do
+    files_location = '/tmp/ohms_transcript_files/'
+    progress_bar = ProgressBar.create(total: Work.where("json_attributes -> 'genre' ?  'Oral histories'").count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
+    Work.where("json_attributes -> 'genre' ?  'Oral histories'").find_each(batch_size: 10) do |w|
+
+      if w.oral_history_content!&.searchable_transcript_source.present?
+        progress_bar.log("INFO: #{w.title} already has a searchable transcript.")
```
jrochkind, line 14: If you're running this more than once, you're going to get a LOT of these on second run -- is it going to make it hard to see any actual errors you want to know about, crowding them out of log and making it easy to miss them?

lib/tasks/add_transcripts_to_oral_histories.rake
```diff
@@ -0,0 +1,57 @@
...
+  desc """
+    Goes through all the oral histories and adds transcripts from a file on the disk, for those missing them:
+
+    bundle exec rake scihist:add_transcripts_to_oral_histories
+  """
+
+  task :add_transcripts_to_oral_histories => :environment do
+    files_location = '/tmp/ohms_transcript_files/'
+    progress_bar = ProgressBar.create(total: Work.where("json_attributes -> 'genre' ?  'Oral histories'").count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
+    Work.where("json_attributes -> 'genre' ?  'Oral histories'").find_each(batch_size: 10) do |w|
```
jrochkind, line 11: Nice job figuring out how to use find_each here, to accomodate cases where we have very large numbers -- it's just the right way to do "iterate through everything". 

I don't think you need the `batch_size` argument, you can just leave it the Rails default. (Which yes, means that for our actual number of records, it won't be doing much batching at all -- that's fine, Rails knows what it can handle, generally, unless we have reason to know/believe otherwise).

lib/tasks/add_transcripts_to_oral_histories.rake
eddierubeiz: Correct!

lib/tasks/add_transcripts_to_oral_histories.rake
eddierubeiz: Removed!

PR #787, opened 2020-06-10
Solr index ohms ToC index terms as searchable fulltext

jrochkind:


jrochkind:
Let's go ahead and merge this in whenever, so stakeholders can see it on staging too. 

(Overall feature is still guarded by a feature flag and won't be deployed production)

Issue #785, opened 2020-06-09
Allow full text of oral histories to be downloaded; uploaded; deleted.


eddierubeiz:
**Downloads**:
- The UI will be on the oral history admin page (the same one where you can upload OHMS transcripts.).
- Download file name will be the accession  number plus some suffix, e.g. `_full_text.txt`.

**Uploads**:
- You can upload a file with an arbitrary name, as long as it ends in `txt`.
- We will not store the name of uploaded files.
- Minimal validation on the files, but we should probably make sure they have `Content-Type: text/something; charset=UTF-8`
- If a transcript already exist, we will prompt the user to make sure s/he wants to replace it with the contents of the uploaded file.

**Deletes**:
- Just a delete button with a confirmation.

jrochkind:
Nice, thanks @eddierubeiz . I'd say try to get an "MVP" ready FIRST, in a PR, being even more "conservative" with anything that may not be strictly necessary. 

Like set the content-type limit on the HTML file input, sure, but don't bother with back-end validation. (This is staff!).  Don't bother with prompt to make sure you want to replace an existing one (I don't think our OHMS XML has that). 

we can potentially add those things even before merging the PR (or not, I'm not sure how much time/code complexity they are worth), but get the basic MVP functionality first, cool?

PR #788, opened 2020-06-10
Upload, download, or remove full-text transcripts

eddierubeiz:
Allows the user to upload, download, or remove full-text transcripts.
Ref #785

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -3,7 +3,6 @@
 # Started with generated code from Rails 5.2 scaffold.
 #
 # We'll probably handle `show` in a different controller, for now no show.
-require 'mimemagic'
 class Admin::WorksController < AdminController
   before_action :set_work,
     only: [:show, :edit, :update, :destroy, :reorder_members,
@@ -136,7 +135,7 @@ class Admin::WorksController < AdminController
       @work.oral_history_content!.update!(searchable_transcript_source: transcript)
       redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), notice: "Full text has been updated."
     else
-      Rails.logger.debug("Could not accept this file as a full-text transcript.  MimeMagic calls it a #{MimeMagic.by_magic(transcript).to_s}")
+      Rails.logger.debug("Could not accept this file as a full-text transcript for work #{@work.friendlier_id}")
       redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), flash: {
         error: "Could not accept this file: it's not a text file."
       }
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -8,7 +8,8 @@ class Admin::WorksController < AdminController
     only: [:show, :edit, :update, :destroy, :reorder_members,
            :reorder_members_form, :demote_to_asset, :publish, :unpublish,
            :submit_ohms_xml, :download_ohms_xml,
-           :remove_ohms_xml, :create_combined_audio_derivatives]
+           :remove_ohms_xml, :submit_searchable_transcript_source, :download_searchable_transcript_source,
+           :remove_searchable_transcript_source, :create_combined_audio_derivatives]
 
   # GET /admin/works
   # GET /admin/works.json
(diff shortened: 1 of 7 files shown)


app/controllers/admin/works_controller.rb
```diff
@@ -119,7 +121,48 @@ def download_ohms_xml
...
+      redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), flash: { error: "No file received" }
+      return
+    end
+    transcript = params[:searchable_transcript_source].read
+    # Very basic check for now. Not stripping HMTL tags,
+    # even though they really don't belong in these transcripts.
+    #transcript_checks = [true]
+    transcript_checks = [
+      (params[:searchable_transcript_source].content_type.start_with?('text/')),
+      ((MimeMagic.by_magic(transcript).nil?)||(MimeMagic.by_magic(transcript).to_s.start_with? 'text/')),
```
jrochkind, line 138: this is going overboard I think, just trust the content_type, no need to bring mime_magic into it. Which should simplify this code a lot. Let's keep it simple, less code to maintain, less code that can have bugs! no need for `transcript_checks`, just check the content type.

app/controllers/admin/works_controller.rb
```diff
@@ -119,7 +121,48 @@ def download_ohms_xml
...
+      transcript.valid_encoding?
+    ]
+    if transcript_checks.all?
```
jrochkind, line 141: I don't understand what `all?` with no block does -- is it just the same thing as `present?`?  But maybe we can abolish `transcript_checks` entirely.

app/views/admin/works/_full_text.html.erb
```diff
@@ -0,0 +1,29 @@
+<div class="card bg-light mb-3">
```
jrochkind, line 1: `works/_full_text` is not descriptive enough for what this is. It's really`oral_history_transcript.html.erb` maybe?

app/views/admin/works/_full_text.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <h2 class="card-header h3">Full text
+  </h2>
+  <div class="card-body">
+    <div class="text-muted small mb-2">The full text file is the transcript we use for full-text indexing.</div>
```
jrochkind, line 5: Le'ts not call it "the full text file", that doesn't mean a lot. Can we call it "Searchable Transcript Text"?  

Can we also specify something like "This will be used for searchable transcripts only if there is NOT an OHMS XML transcript. If there is an OHMS XML transcript, this file will not be used.", for clarity?

app/controllers/admin/works_controller.rb
eddierubeiz: No problem -- removed.

app/views/admin/works/_full_text.html.erb
eddierubeiz: Done.

app/views/admin/works/_full_text.html.erb
eddierubeiz: Yup. Added verbiage to that effect. Good call.

app/controllers/admin/works_controller.rb
eddierubeiz: Just checking that all items in an array are true. No longer relevant though.

app/controllers/admin/works_controller.rb
```diff
@@ -119,7 +121,42 @@ def download_ohms_xml
...
+      return
+    end
+    transcript = params[:searchable_transcript_source].read
+    # Very basic check for now. Not stripping HMTL tags,
+    # even though they really don't belong in these transcripts.
+    if params[:searchable_transcript_source].content_type.start_with?('text/')
+      @work.oral_history_content!.update!(searchable_transcript_source: transcript)
+      redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), notice: "Full text has been updated."
+    else
+      Rails.logger.debug("Could not accept this file as a full-text transcript.  MimeMagic calls it a #{MimeMagic.by_magic(transcript).to_s}")
```
jrochkind, line 139: Are you intentionally using mimemagic in the debug message, even though no longer using it in the actual check? Do we need a debug message at all, is the user error displayed enough? (In prod, we don't generally output debug messages anwyay).

app/controllers/admin/works_controller.rb
```diff
@@ -3,12 +3,14 @@
 # Started with generated code from Rails 5.2 scaffold.
 #
 # We'll probably handle `show` in a different controller, for now no show.
+require 'mimemagic'
```
jrochkind, line 6: Even if we are using mimemagic... I'm surprised we need to explicitly require it. I think the correct "fix" might be adding mimemagic to our Gemfile (if we are using it directly, we should express it as a dependency directly) -- and then I think bundler would take care of it requiring it on boot and we wouldn't need to. 

If we stop using it in this file entirely, that would also eliminate the need for the "require".

app/controllers/admin/works_controller.rb
```diff
@@ -119,7 +121,42 @@ def download_ohms_xml
     send_data @work.oral_history_content!.ohms_xml_text,
       :type => 'text/xml; charset=UTF-8;',
       :disposition => ContentDisposition.format(disposition: "attachment", filename: "#{@work.oral_history_content!.ohms_xml.accession}.xml")
-      #:disposition => "attachment; filename=#{@work.oral_history_content!.ohms_xml.accession}.xml"
+  end
+
+  # PATCH/PUT /admin/works/ab2323ac/submit_ohms_xml
```
jrochkind, line 126: We COULD try to combine these actions with the existing ones for OHMS XML file -- since it is kind fo similar, need to add/remove/download "a file", we could try to it into a generic set of actions for adding/removing/downloading files, that could be re-used. 

BUT I'm honestly not sure if that is over-engineering or not; I actually think this is probalby just fine, and we should just go ahead with it rather than spending more time on this, but wanted to mention it just to share knowledge/thoughts with "the team" (Ie you and me).  But I think probably no need and this is good!

app/controllers/admin/works_controller.rb
eddierubeiz: Nah - that was just a stray mention that I neglected to delete. Good call!

app/controllers/admin/works_controller.rb
eddierubeiz: Fair enough. I'm usually the one who always insists on super-DRYing everything but in this case -- yeah, this is fine.

app/controllers/admin/works_controller.rb
eddierubeiz: Got rid of it!

Issue #712, opened 2020-04-20
More flexibility when (un)publishing a work and its assets


eddierubeiz:
Currently, publishing a work publishes all its assets, and un-publishing a work un-publishes all its assets.

This poses a problem when it comes to recreating combined audio derivatives, which 
a) can't be recreated when the work is published
b) only incorporate published assets.

Possible fixes:
* Easier way to publish all assets in a work without publishing the work itself.
* Prompt user, when un-publishing a work, whether to un-publish all assets as well.
* Allow the user to publish/un-publish assets directly from the asset list table, instead of only allowing it on the asset edit page.

eddierubeiz:
OK -- talked to Nicole this morning and I would like to propose a simple fix.
**IF** a work is an oral history
**AND** it is unpublished
**AND** it has a bunch of audio segments. Let's say 12 of them.
**CURRENTLY** users do not have access to click the blue button that says "Generate combined audio derivatives".
**INSTEAD**, show the button.
**WHEN IT IS CLICKED**, prompt the user with the two following options:
**OPTION 1**: "Publish all 12 audio segments, and create audio derivatives from them. (The work will remain unpublished.)"
**OPTION 2**: "Cancel"

eddierubeiz:
My *belief* is that the change above will solve a large number of the problems needed to solve #767.  My *hope* is it will solve all of them.

jrochkind:
This is a clever solution. 

What about instead:

* A feature to "publish all audio members without publishing the work" *separately* from "combined audio derivatives"?  Or in addition?  Is this need really connected specifically to creating a combined audio derivative?  Or is it really a general feature to "publish all members" Right now, there is no convenient way to do that without publishing the work, but sounds like we have a need?  Maybe it would be useful generally?


* Separately, I'm not sure what feedback you get currently when the combined audio derivatives button doesn't show up -- is it just missing invisible with no reason to know why it's not there? If part of the point of this feature is to make taht less confusing -- would an on-screen explanation work better? Like a greyed out/inactive button, with the explanation "You can't create because there are no published audio members". 

* I am just trying to brainstorm to see if things can be kept *simpler* and more *general purpose*. Worried about continuing to add more complicated logic to deal with previous logic we added; can we simplify instead?  In general purpose ways? A generic feature "make all members published" (without publishing parent) seems potentially simpler than adding a UI choice form, with two logic paths, triggered by certain state, etc. (Although not clear how it applies to multi-level children. :( )

[...]

PR #789, opened 2020-06-12
Allow user to publish all unpublished audio members in an oral history before creating combined audio derivatives.

eddierubeiz:
Ref #712 .

This modifies works_controller#create_combined_derivatives by adding a new option, `params['publish_all_audio_members_first']`.

If `publish_all_audio_members_first` is 'true', the controller publishes all unpublished audio members of the item, and then creates the job as it would normally. _It does *not* publish the work; this is meant to be done later, after the user has tested the work along with its combined audio derivatives._

If `publish_all_audio_members_first` is not 'true', the method behaves as before.

On the front end:
We now show a button that triggers the method (with publish_all_audio_members_first = true )  in the (very common) case in which there are no published audio members. (Previously, we didn't show the button, and required the user to publish all the audio items one by one.)

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -125,17 +125,36 @@ class Admin::WorksController < AdminController
   # Create_combined_audio_derivatives in the background, if warranted.
   # PATCH/PUT /admin/works/ab2323ac/create_combined_audio_derivatives
   def create_combined_audio_derivatives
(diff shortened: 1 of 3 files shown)


eddierubeiz:
We reached consensus via Slack and a demo with Nicole and Lee that this feature is OK for now. We can revisit in a few weeks if it doesn't behave as it should in production.

Issue #786, opened 2020-06-09
upgrade solr to latest 8.5.x


jrochkind:
We are currently on Solr 8.0. 

I happened to notice Solr 8.5 (released recently March 2020) adds some configuration options related to "results highlighting snippets" features we are using with transcript searching, that might be nice to play around with. 

At this point, those extra features are *not* known to be essential to our efforts, so this is not an urgent thing.  It's possible our desire for those options will increase after more evaluation of our highlighting feature. 

But for now, just having noticed there were some solr releases out past what we have, if we have some time it might be nice to get up to date. 

I think Solr minor releases *should* not require any index rebuidling, the on-disk indexes should be perfectly forward-compatible. But of course we'd have to test to be sure.  It would probably require at least some minimal downtime regardless to switch Solr's; not sure the most efficient/reliable/least downtime way to do it.

sanfordd:
Given how fast the reindexing process is, I think that worrying about carrying the index over may not be worth the effort.
The "simplest" thing to do would be to build out a new solr server, deploy the code, swap the servers out, and then reindex the system. We'd be down for couple minutes during the server swap and reindex.

The slower but ideally no downtime option would be to rebuild the solr server, take a backup of the old server's index and import it to the new server, then swap it in for the old server. This assumes there won't be any index rebuilding needed. We would need to do it when people aren't adding new material so the index doesn't change during that process.

jrochkind:
Some minutes of scheduled downtime seems fine, if that's easiest.

sanfordd:
It saves me 10-15 minutes of work with making, copying, and loading the backup to the new server. We don't keep a regular backup around anymore since we improved the indexing times.

sanfordd:
Swapped staging to 8.5.2 we can test it tomorrow before any production changes.

PR #791, opened 2020-06-16
update solr_wrapper solr version from 8.0.0 to 8.5.2

jrochkind:
Ref #786

This does not effect production machines, but should effect solr in dev and test, including on travis. 

On your dev machine, you may have to manually force solr upgrade, after merging this, this OUGHT to work, I think:

    $ ./bin/rake solr:stop solr:clean solr:start
    $ RAILS_ENV=test ./bin/rake solr:stop solr:clean


eddierubeiz:
Noted - will do clean solr in dev and test on my dev box shortly,

Issue #738, opened 2020-05-04
Oral history transcripts searchable in digital collections search


jrochkind:
Current To do: 

- [x] Do not deploy this until we have fixed HTML showing up in imported searchable transcripts. 
   * not only do we have to strip HTML tags, but possibly replace `&nbps;` with something reasonable. 
   * #797
- [x] Do not deploy until we've fixed too large full text snippet excerpt, for instance search for "christianity", look at Nico M. Nibbering hit. #794 
- [x]  deal with weird characters (up arrow?) from some of the imported fulltext, around back-of-the-book index terms
   * looks like the weird char is `0x0C`, or ascii "form feed" -- at least in our pg database, not sure if it gets turned into something else when we index it into Solr, from which it's ultimately displayed. Not sure why it shows up as an arrow on some browser/OS. Our best move is probably to strip control characters on input, or at least control characters except for LF and CR?
  * #797
- [x] reindex transcripts in *production* with fixes in #797

https://chemheritage.atlassian.net/wiki/spaces/HDC/pages/846397441/Priorities+for+the+Next+Stage+of+Development+Work

* Includes OHMS-enabled and non-OHMS-enabled -- so we'll have to add another input field for transcript plain text. 
* Search results should give you the 'hit in context' like microsite does. 
  * This involves a new feature in search results, although it only applies to OH content. We'll have to figure out how it should look on the page
  * The implemention will use the Solr [highlighting](https://lucene.apache.org/solr/guide/8_2/highlighting.html) feature. I haven't used that one before, it might take some work to figure out. We will probably want to put these results in their own Solr field, so we can use taht field for highlighting
  * This is tricky becuase our display right now is using only the ActiveRecord postgres model for results. We'll need ot make sure the relevant template is getting the solr response too, and can use it for solr highlighting return results. 
* We'll have to pay attention to how this effects relevancy results in our general search, to have some items with "full text" in the Solr search index, and other items with only metadata
   * There is a danger it will make OH items show up "too well", always being on top of search results; or "not well enough", hardly ever getting full text OH matches. Have to figure out the right balance
   * This can be hard to figure out when we don't *have* the transcripts in there yet to test. So we may want to figure out a way to *automate* import of existing full text from the microsite, so we can actually test it all in staging ASAP (as well as save all that manual labor)
   * We may want to ask other peer institutions if anyone has handled mixed fulltext/metadata-only content in Solr/blacklight, and any tips they have or lessons they've learned 
[...]

PR #794, opened 2020-06-16
Tweak solr highlighting parameters

jrochkind:
To control the length of the snippets better, we switch to WORD delimiting instead of default SENTENCE. With SENTENCE, if there was a partiuclarly long "sentence" (or not really a sentence, but a section of, say, the back of the book index with a lot of text with no periods!), then it would return a single "sentence" and be far too long. With WORD, it's less likely there will be a single word that's far too long. (Although still possible, and sort of a lack of feature in Solr that it doesn't let you cap it).

Until Solr 8.5, the "WORD" delimiting in highlighter didn't really work well, it would give you a snippet with the highlighted match always at the beginning! But 8.5 fixes it to put the highlighted match in the middle of the snippet, as you'd want.

We also tweaked highlighting settings to provide *3* highlights if avaialble instead of just two; and tweaked the fragSize to be a bit smaller; both of these get a bit closer to what microsite appears to be doing.

Ref #738

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -129,13 +129,18 @@ class CatalogController < ApplicationController
         "hl.method" => "unified",
         "hl.fl" => "searchable_fulltext",
         "hl.usePhraseHighlighter" => "true",
-        "hl.snippets" => 2,
+        "hl.snippets" => 3,
         "hl.encoder" => "html",
         # Biggest current transcript seems to be OH0624 with 817,139 chars.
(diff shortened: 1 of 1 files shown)


jrochkind:
Fixes too long snippet, example `Christianity` in Nibbering. 

Before (it actually goes on longer than this, screenshot was cut off):

68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3565386234393665396465643531343263633034643631372f30613430363464302d356532372d343933612d383532652d666661353431666231616239

After:

Screen Shot 2020-06-16 at 12 26 07 PM

PR #796, opened 2020-06-17
don't try to copy integer pk of OralHistoryContent when copy_data betweem hosts

jrochkind:
Can lead to collisions, there's no need to preserve that pk


eddierubeiz:
Makes sense.

Issue #807, opened 2020-07-01
Full-text transcript mismatch(es)


eddierubeiz:
Regrettably, in at least one case, the code I wrote to associate full-text transcripts to oral histories attached the wrong file to the wrong oral history.

Paul Weisz https://digital.sciencehistory.org/admin/works/cj82k863f#nav-oral-histories

is the only example we have for now.

eddierubeiz:
Compare:

Paul B. Weisz
https://oh.sciencehistory.org/oral-histories/weisz-paul-b
Interview no.: Oral History 0141

Pouné Saberi
https://oh.sciencehistory.org/oral-histories/saberi-poune
Interview no.: Research Interview 0141

eddierubeiz:
I've removed the transcripts from the following mismatches:
```
    0138
    Roland
    https://digital.sciencehistory.org/admin/works/w37637891#nav-oral-histories

    0139
    Hanford
    https://digital.sciencehistory.org/admin/works/g445cf063#nav-oral-histories

    0141
    Weisz
    https://digital.sciencehistory.org/admin/works/cj82k863f#nav-oral-histories

    0142
    McKennon
    https://digital.sciencehistory.org/admin/works/1v53jz09q#nav-oral-histories

    0147
    Biondi
    https://digital.sciencehistory.org/admin/works/pn89d750s#nav-oral-histories

    0148
    Labovsky
    https://digital.sciencehistory.org/admin/works/pg15bf97z#nav-oral-histories
```

jrochkind:
We forget for sure if we resolved this, @lsberry16 will double check.

lsberry16:
I reviewed all six and successfully added the text files for the ones that were missing them. This ticket can be closed, as far as I'm concerned. :)

jrochkind:
Thanks for reviewing and completing!

PR #809, opened 2020-07-06
additional features around upload of plain text transcript

jrochkind:
include catch catchable encoding errors. 

Ref #808

Blocking #807.


jrochkind:
Lee tested and signed off.

Issue #811, opened 2020-07-13
[scihist_digicoll/production] NoMethodError: undefined method `strip' for #<ActionController::Parameters:0x...>


jrochkind:
## Backtrace

    line 15 of [PROJECT_ROOT]/app/views/catalog/index.html.erb: _app_views_catalog_index_html_erb__2392805746771157780_47099852032960
    line 24 of [PROJECT_ROOT]/app/controllers/application_controller.rb: block in batch_kithe_indexable
    line 23 of [PROJECT_ROOT]/app/controllers/application_controller.rb: batch_kithe_indexable

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/63082701)

jrochkind:
woah, this is actually facebook doing something weird. 

A normal BL facets url might look like `?f[subject][]=value.`

If you try to share that on Facebook as a link, facebook will INSIST on changing it to `?f[subject][0]=value` — which will NOT work in Blacklight.

Meaning the facebook preview scraper will immediately make this request to your app (and generate an error), but also that the FB post someone was trying to do is actually broken, any humans clicking on it will also get an error. The search URL is not actually share-able on FB.

So this is actually a "problem", in that our search URLs with facet limits are not shareable on Facebook! Not sure why Facebook is doing this though.

jrochkind:
More info. This has been noted by other Blacklight useres, Blacklight previously tried to fix it, but apparently BL's fix has stopped working. 

https://github.com/projectblacklight/blacklight/issues/1784

I may try to contribute fix upstream again.

jrochkind:
Having trouble making a test that reproduces, very odd!

PR #812, opened 2020-07-13
Redirect corrupted facet params

jrochkind:
When sharing a search results URL on facebook, facebook will erroneously alter the facet limit params. From eg `?f[facet_name][]=value` to `?f[facet_name][0]=value`

This will result in a Blacklight error. When the facebook crawler crawls it for a preview, as well as when any user clicks on it. We try to catch these, fix them to what Blacklight wants, and REDIRECT to the correct one again with an HTTP 301 Moved Permanently. (Better than two URLs representing the same search, and it looks like Facebook crawler will follow redirect).

Ref #811

Discussion in Blacklight tracker indicates this is a kind of regression, Blacklight already tried to work around facebook mangling of URLs earlier: https://github.com/projectblacklight/blacklight/issues/1784

There's a pending Blacklight PR to fix this "in place", instead of doing a redirect: https://github.com/projectblacklight/blacklight/pull/2313

We'd like to fix this without waiting for a Blacklight release (we are several minor releases behind, and BL upgrades may not be breakage-free), and this approach of a redirect gives us a way to do that without monkey-patching. We also maybe actually like doing the 301 Moved Permanently Redirect to the correct canonical URL for the search?

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -3,6 +3,7 @@
 require 'kithe/blacklight_tools/bulk_loading_search_service'
 
 class CatalogController < ApplicationController
+  before_action :redirect_hash_facet_params, only: :index
   before_action :redirect_legacy_query_urls, only: :index
 
   include BlacklightRangeLimit::ControllerOverride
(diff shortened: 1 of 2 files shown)


jrochkind:
More context:

A query string `key[]=val1&key[]=val2` is converted by Rails into `key = [val1, val2]`

But facebook turns that query string into `key[0]=val1&key[1]=val2` which is converted by Rails into `key = { 0: val1, 1: val2}`. So we just need to convert it back.

The f param is the only place we know of where this pattern is used, so it’s all we’re fixing. It's the only place Blacklight itself tries to account for too. 

The standard `f` params for facets look like:  `&f[facet_name][]=val` which turns into `f: { facet_name: [val] } }`

eddierubeiz:
@jrochkind spells out the operation further:

A query string `key[]=val1&key[]=val2` is converted by Rails into `key = [val1, val2]`.
But Facebook turns that query string into `key[0]=val1&key[1]=val2` which is converted by Rails into `key = { 0: val1, 1: val2}`. So we just need to convert it back.
The f param is the only place we know of where this pattern is used, so it’s all we’re fixing.

PR #813, opened 2020-07-14
Handle mis-ordered date range limit params

jrochkind:
If you have a query with a date range limit, where the "start" is AFTER the "end", blacklight_range_limit will raise a `BlacklightRangeLimit::InvalidRange`. This results in an HTTP 406 "Not Acceptable" being delivered to the browser. This could be considered acceptable behavior. But it also results in an uncaught exception  being reported by Honeybadger.

This happens from time to time in our logs. It looks like it probably is real users, just entering dates out of order. While a 406 response might be okay... why not just treat it as if they were entered in the "correct" order instead, probably giving the user what they wanted, and avoiding the honeybadger uncaught error report?

That is what this commit does. Code adapted from something similar we were doing in earlier chf_sufia app.


eddierubeiz:
Awesome.

PR #816, opened 2020-07-16
DownloadOptions use our homegrown bytes_to_human_string

jrochkind:
We put the size of derivatives next to a link to download in the download menu, and we format it human-readably like "104 MB" instead of just a raw number of bytes like `104567233`

We were using ActiveSupport::NumericHelper.number_to_human_size, but that turned out to be a real performance problem,  in a way that really really matters when we might do it 6 derivatives * 600 children == 3600 times on a page. (Whether we should re-architect the app to NOT do that is another story). 

The Rails built-in method is super slow perhaps because it does so many checks for i18n localization (to get the right units? How do you say GB in Chinese? I don't think we are concerned), perhaps because it has even fancier handling of edge cases that we don't really need. 

Turns out a very short DIY implementation (cribbed from StackOverflow, modified a bit) is orders of magnitude faster. This can speed up delivery of Ramelli by like 40%!

It was kind of a hacky nightmare figuring this out, it's hard to explain how I figured it out, at least right now in text. 

There is definitely more optimization still to be done, but this is a big jump!

### DIFF
--- a/app/lib/scihist_digicoll/util.rb
+++ b/app/lib/scihist_digicoll/util.rb
@@ -33,5 +33,40 @@ module ScihistDigicoll
 
       mime_obj.symbol.to_s.upcase
     end
+
+    # Quite similar to Rails' ActiveSupport::NumberHelper.number_to_human_size
+    # But it turns out that Rails `number_to_human_size` is pretty expensive, expensive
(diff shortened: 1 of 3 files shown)


eddierubeiz:
Great. Thanks for the test in particular!

PR #819, opened 2020-07-22
update lodash yarn dependency

jrochkind:
For security warning.

It's a bit tricky to figure out how to update "indirect" dependencies (not specifically mentioned in package.json) with yarn.

Here, we first REMOVED the lodash "resolution" in package.json, which was an earlier attempt to deal with updating indirect dependencies, which wasn't really a good solution.

Then we followed instructions at https://medium.com/@ayushya/upgrading-javascript-packages-deep-dependencies-using-yarn-8b5983d5fb6b

* Hand-edit the yarn.lock to remove the relevant lodash resolution
* re-run `yarn install`

Bingo, now we have lodash 4.17.19 which is what we wanted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "webpack-dev-server": "^3.9.0"
   },
   "resolutions": {
-    "lodash": ">= 4.17.13",
     "lodash.template": ">= 4.5.0",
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Noted. Merging.

PR #837, opened 2020-08-04
tell travis to cache more stuff including yarn install

jrochkind:
yarn install is taking like 120s on travis, telling it to cache yarn might speed up our builds?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ cache:
     - /home/travis/.rvm/
     - ./tmp/solr_dist
     - vendor/bundle
+    - node_modules
 before_install:
   - gem install bundler
   - nvm install node
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ services:
   - postgresql
 cache:
   bundler: true
+  yarn: true
   directories:
     - /home/travis/.rvm/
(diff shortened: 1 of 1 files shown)


jrochkind:
It's hard for me to figure out if this is actually working, but doesn't hurt.

jrochkind:
Adding cache dir `vendor/bundle` explicitly to travis.yml seemed to work to speed up bundle install?

Travis may not be operating quite as documented. We're just trying to get it to cache the yarn and bundle installs, cause those are like half of our total travis build time!

eddierubeiz:
I guess I'll merge -- we can remove later if we get the sense it's not helping. Hard to tell w/ travis.

jrochkind:
Hmm, a recent PR after this had it's travis finish in 4 minutes (had been 8 or even 9 routinely before), with less than 1 second spent on `bundle install` (instead of 120+), and 20 seconds on yarn install (instead of 120+). 

https://travis-ci.com/github/sciencehistory/scihist_digicoll/builds/178612787

So this seems to have worked to significantly speed up travis!  It seems like travis maybe wasn't even caching bundler before this... which it should have been with `cache: bundler: true`....  but something was maybe going wrong or is mis-documented or we didn't understand? But we seem to have improved things, hooray.

Issue #830, opened 2020-08-03
identify request-form-gated oral histories on back end


jrochkind:
For "access control" purposes, these request-form-gated oral history assets are probably still marked 'private'. 

So how does the software know to display the proper page with request form and grant access after request? They have to be identified on back-end somehow. 

Need to figure this out. Trying to avoid making more complex the basic published/private access control distinction in digital collections generally, for this feature only needed for Oral Histories specifically.

PR #839, opened 2020-08-05
An alternate work show page, meant for OH's with gated access

jrochkind:
work in progress.

At first I thought we'd use this for OH's without audio generally, but there are a couple tricks. Makes sense to get it in place for it's motivating use case first, then come back. 

Going to put this aside for a second while I do other things necessary to actually put it in place for it's motivating use case. Like #830


jrochkind:
We might want to look at download filenames for OH PDFs too.

jrochkind:
needs tests

app/views/works/work_file_list_show.html.erb
```diff
@@ -0,0 +1,93 @@
...
+
+
+          <p class="text-brand-serif">Available upon request
+            <%= decorator.multiple_files? ? "are" : "is" %>
+            <%= decorator.available_by_request_summary %>.
+          </p>
+          <p class="text-brand-serif">After submitting a brief form, you will receive immediate access to
+            <%= decorator.multiple_files? ? "these files" : "this file" %>.
+            If you have any questions about transcripts, recordings, or usage permissions, please
+            <%= link_to "contact the Center for Oral Histories", "https://www.sciencehistory.org/center-for-oral-history/contact" %>.
```
eddierubeiz, line 37: Should read "Center for Oral History".

jrochkind: thanks, good catch!

PR #840, opened 2020-08-06
Staff UI for marking Oral history assets as "available by request" even when private

jrochkind:
Ref #830

Assets have an `oh_available_by_request` flag individually (using attr_json). 

But the main UI to set it is a list of assets on the work edit page. This is also where the switch for "with" and "without" approval/human intervention will go, once we support with-human-intervention. Or should we plan for this now, to get the metadata flag in there?

Limited controller test. No test of the front-end checkboxes right now. I feel like calling it a day, but let me know if you think a presenter test of the staff front-end checkboxes is needed.

### DIFF
--- a/app/views/admin/assets/show.html.erb
+++ b/app/views/admin/assets/show.html.erb
@@ -30,7 +30,17 @@
   </div>
 
   <div class="col-sm-6">
+
+      <% if !@asset.published? && work_is_oral_history? && @asset.oh_available_by_request %>
+        <div class="row">
+          <div class="col-sm-12 h3">
+            <span class="badge badge-warning">Un-published, but marked available by request</span>
+          </div>
+        </div>
+      <% end %>
+
     <dl class="row">
+
       <dt class="col-sm-4">Created</dt>
       <dd class="col-sm-8"><%= l @asset.created_at, format: :admin %></dd>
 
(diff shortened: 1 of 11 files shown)


jrochkind:
Actually, let's put in a toggle for on/off too -- to eventually become off/without-intermediation/with-human-approval

That'll make it easier to use as the trigger for showing the alternate view with request button. having to actually see if any assets are marked "available by request" would be tricky (cause would require fetching all children), way easier to have a toggle at Work level, so we'll do that.

jrochkind:
@eddierubeiz remember when you said you'd review staff UI for this before I showed it to staff, for a sanity check?

It's ready when you are! You can deploy this branch to staging, or look in dev? (It does have a migration from #842 in it)

app/views/admin/assets/show.html.erb
```diff
@@ -30,7 +30,22 @@
   </div>
 
   <div class="col-sm-6">
+
+      <% if !@asset.published? && work_is_oral_history? && @asset.oh_available_by_request %>
+        <div class="row">
+          <div class="col-sm-12">
+            <span class="h3">
+              <span class="badge badge-warning">Un-published, but marked available by request</span>
```
eddierubeiz, line 38: Great. Glad this is read-only.

Issue #760, opened 2020-05-19
Move emails from self-hosted server to AWS SES


sanfordd:
Since we're already getting SES setup for other projects, moving our collections emails to SES will remove the need to manage a local mailserver.

jrochkind:
How to configure in Rails? Looks like best practices have changed over the few years since SES was released, so hard to google and not get outdated info. Maybe this gem?

https://github.com/drewblas/aws-ses

jrochkind:
Possibly you can use SES with a standard SMTP interface? Then it wouldn't require any special gems on Rails end, just the SMTP host and other config info?

https://hixonrails.com/ruby-on-rails-tutorials/ruby-on-rails-action-mailer-configuration/#amazon-ses-account-setup

sanfordd:
You can use STMP with SES
https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-smtp.html

jrochkind:
We're probably going to need some emails for OH work, so tagging this with "oral histories" category, and we may have to get to it sooner rather than later. 

We think Dan may have already created an SES resource for us. It's unclear if he got it cleared for "production" use yet -- I think Amazon makes you manually email them to get a non-toy-sized email limit for production use. 

Then we just have to configure the app to use it, probably via connection details supplied in local_env.yml. 

Maybe @eddierubeiz would be interested in carrying this over the finish line.

PR #841, opened 2020-08-07
Use Amazon Simple Email Service for outgoing email

eddierubeiz:
Ref #760 

Documentation is at https://chemheritage.atlassian.net/wiki/spaces/HDCSD/pages/1046151169/Amazon+Simple+Email+Server

Do not deploy before merging the Ansible staging branch to production ( the PR is at https://bitbucket.org/ChemicalHeritageFoundation/ansible-inventory/pull-requests/111/ )and ensuring the new variables are available in production. (That said, if you do, nothing bad will happen. But we want to do things in the correct order).

### DIFF
--- a/app/lib/scihist_digicoll/env.rb
+++ b/app/lib/scihist_digicoll/env.rb
@@ -294,8 +294,8 @@ module ScihistDigicoll
     define_key :digital_email_address
 
     # These are the credentials used for accessing Amazon's simple email server:
-    define_key :ses_smtp_username
-    define_key :ses_smtp_password
+    define_key :smtp_username
+    define_key :smtp_password
     define_key :smtp_host
 
     ##
--- a/app/lib/scihist_digicoll/env.rb
+++ b/app/lib/scihist_digicoll/env.rb
@@ -275,11 +275,7 @@ module ScihistDigicoll
 
     define_key :honeybadger_api_key
 
-    # Used as the from (no_reply) and to (digital_tech, digital) email addresses
-    # when sending notices about failed fixity check.
-    define_key :no_reply_email_address
-    define_key :digital_tech_email_address
-    define_key :digital_email_address
+
 
     # Used for sending mail, and if no queue is specified:
     define_key :regular_job_worker_count, default: 0
@@ -288,6 +284,18 @@ module ScihistDigicoll
     # Used (infrequently) by additional job servers whose only purpose is to handle special tasks:
     define_key :special_job_worker_count, default: 0
 
+
+    # OUTGOING EMAIL:
+
+    # From address:
+    define_key :no_reply_email_address
+    # To addresses (these are email lists maintained by IT.)
+    define_key :digital_tech_email_address
+    define_key :digital_email_address
+
+    # These are the credentials used for accessing Amazon's simple email server:
+    define_key :ses_smtp_username
+    define_key :ses_smtp_password
     define_key :smtp_host
 
     ##
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -93,10 +93,10 @@ Rails.application.configure do
 
     config.action_mailer.delivery_method = :smtp
     config.action_mailer.smtp_settings = {
-      :address => "email-smtp.us-east-1.amazonaws.com",
+      :address => ScihistDigicoll::Env.lookup(:smtp_host),
       :port => 587,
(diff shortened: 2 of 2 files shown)


app/lib/scihist_digicoll/env.rb
```diff
@@ -288,6 +284,18 @@ def self.shrine_dzi_storage
...
+    # From address:
+    define_key :no_reply_email_address
+    # To addresses (these are email lists maintained by IT.)
+    define_key :digital_tech_email_address
+    define_key :digital_email_address
+
+    # These are the credentials used for accessing Amazon's simple email server:
+    define_key :ses_smtp_username
+    define_key :ses_smtp_password
     define_key :smtp_host
```
jrochkind, line 299: These values would apply even if it wasn't `ses`, right? Change variable names to just `smtp_username` and `smtp_password`, what do you think?

config/environments/production.rb
```diff
@@ -90,12 +90,15 @@
...
       raise RuntimeError, "Please specify smtp_host in local_env.py so we can send emails."
     end
+
     config.action_mailer.delivery_method = :smtp
     config.action_mailer.smtp_settings = {
-      address: ScihistDigicoll::Env.lookup(:smtp_host),
-      domain: 'sciencehistory.org',
-      ssl: false,
-      enable_starttls_auto: false
+      :address => "email-smtp.us-east-1.amazonaws.com",
```
jrochkind, line 96: Don't hard-code this address in, use it from `ScihistDigicoll::Env`, there is even an `stmp_host` key already there. 

oops sorry just noticed this was marked "wip", I"m reviewing prematurely sorry!

app/lib/scihist_digicoll/env.rb
eddierubeiz: Absolutely, duh!

config/environments/production.rb
eddierubeiz: Yup -- but I have fixed this. I also got rid of the vestigial `email_server_ip_dns` setting in both Ansible and scihist_digicoll.

jrochkind:
@eddierubeiz 

1. Have you looked into the "getting SES approved for production traffic" situation? Please let us know here on PR ticket that you've confirmed this. 

2. Please confirm on PR ticket here that you've verified email being sent via SES using this code has been received. (I think it should work in dev even, if the Env is all set? Might be fun to test that out and verify). 

3. In dev or test WITHOUT the variables being set, what happens, and does it differ from what happened before this PR? Please investigate and let us know on this PR. 

Thanks!

jrochkind:
Also can you link here to ansible PR url, so I can find it, and to make it easier to verify it's been merged? I'm not used to dealing with that ansible bitbucket, so have trouble finding things there. Thanks!

eddierubeiz:
`1) Have you looked into the "getting SES approved for production traffic" situation? Please let us know here on PR  ticket that you've confirmed this, what it entailed, link to relevant SES docs etc.`

No, we are *not* yet approved for production traffic; we're only allowed to send up to 200 messages a day, at a rate of 1 per second. You can monitor our progress against this quota at https://console.aws.amazon.com/ses/home?region=us-east-1#dashboard. (I've already sent out a few test messages today, as you can see.)

Docs are at https://docs.aws.amazon.com/ses/latest/DeveloperGuide/request-production-access.html for getting us to production -- we just have to fill out a prose description of how we plan to use our new email powers for good, basically.

I can take care of this before we merge / deploy this PR, as promised.

eddierubeiz:
`2. Please confirm on PR ticket here that you've verified email being sent via SES using this code has been received. (I think it should work in dev even, if the Env is all set? Might be fun to test that out and verify).`

Confirmed -- it works fine in staging. (You could test it in dev too, of course, but testing it in staging has the added advantage of being able to double-check it works correctly with the changes I've made to Ansible.)

eddierubeiz:
`3. In dev or test WITHOUT the variables being set, what happens, and does it differ from what happened before this PR? Please investigate and let us know on this PR.`

In dev or test, if you don't have any mail variables set, ActionMailer::Base.mail( stuff ).deliver_now will fail silently, as it did before.

(Note that we are just changing the value of smtp_host and adding two new credentials to allow connections to SES: as you noted the other day, we're already using SMTP, so it's just a question of changing *which* server we're connecting to and making sure we're all set up.)

Issue #838, opened 2020-08-05
staff UX for publishing should not automatically (un)publish all children


jrochkind:
Right now the publish/unpublish buttons will automatically publish/unpublish all children, because (we thought) that matched desired workflow universally. 

But at least with oral histories, it does not. 

We need that not to be automatic. In sufia, after changing visibility status of a work, you got a confirmation form asking you if you wanted to apply that to all children. That would be one possibilty. 

Or instead, we could move from a button in the navbar, to a control somewhere with more room, that has a checkbox or radio button that lets you select "apply to all children" yourself. Possibly also a control to publish/unpublish all children without changing status of parent. 

(Does this need to be 'transitive', applying through an entire tree of multi-level children? Probably? Was it before in our app? Is it in sufia/hyrax?)

jrochkind:
This change doesn't actually look like it's in production to me, even though #844 looks merged, and master was deployed to production. 

Not sure what's going on, we need to figure it out before closing this. 

I thought when clicking on 'publish' or 'unpublish' on a work, you should get a popup select menu asking if you want it to apply to children or not? That does not seem to be happening to me. 

It's tricky to experiment with in production, since you don't want to actually publish or unpublish anything. The one test record we have in production at w37636967 may be helpful.

eddierubeiz:
Logged into production; did not see the feature; redeployed master to production; now the feature is visible. Is it possible you thought you had deployed master to production but somehow you didn't?

jrochkind:
I guess it is possible! Thanks! I didn't mean for you to have to interrupt other work to do that it wasn't urgent, but I guess we can close this now!

PR #844, opened 2020-08-13
Adjustments to publishing / unpublishing UI.

eddierubeiz:
Ref #838 .
Nicole has approved the UI for this.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -192,11 +192,12 @@ class Admin::WorksController < AdminController
   # recursive CTE so it'll be efficient-ish.
   def publish
     authorize! :publish, @work
-
     @work.class.transaction do
       @work.update!(published: true)
-      @work.all_descendent_members.find_each do |member|
-        member.update!(published: true)
+      if params[:cascade] == "true"
+        @work.all_descendent_members.find_each do |member|
+          member.update!(published: true)
+        end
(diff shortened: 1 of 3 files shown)


app/controllers/admin/works_controller.rb
```diff
@@ -192,11 +192,12 @@ def create_combined_audio_derivatives
   # recursive CTE so it'll be efficient-ish.
   def publish
     authorize! :publish, @work
-
     @work.class.transaction do
       @work.update!(published: true)
-      @work.all_descendent_members.find_each do |member|
-        member.update!(published: true)
+      unless params[:cascade] == "false"
+        @work.all_descendent_members.find_each do |member|
```
jrochkind, line 198: I'm still wondering about leaving the "default", when no `params[:cascade]` is present at all -- to be cascading. Since cascading is the more 'dangerous' operation, it seems like maybe if no `params[:cascade]` is present at all -- or `params[:cascade]` is set to some nonsense like `adf `--  it should NOT cascade.  

I think `params[:cascade] == 'true'` is also a more usual pattern, that matches how we do boolean params everywhere else, instead of `params[:cascade] != 'false'`

eddierubeiz: Yeah -- that's how I wrote it initially! I couldn't make up my mind... so thanks for weighing in.

PR #848, opened 2020-08-18
Remove the .show-title layout div from TitleAndDates ViewModel component

jrochkind:
Each use of TitleAndDates should wrap it in it's own .show-title. The .show-title selector is used for page layout. This lets us be more flexible on an upcoming Oral History layout (#839) where we want to put some additional content in the .show-title section of the screen, that isn't from TitleAndDates.

This actually seems to make it consistent with how these elements were factored on collection show pages! Kind of accidental that htey didn't match, until now.

I am coming to the general principle that selectors targetted for layout should never be in viewcomponent, they should always be in the parent. Then it's more flexible to make different decisons about what content goes in what layout slot.

I am also realizing that our CSS around show page layouts is kind of a spagetti mess! But so it goes, not gonna fix that now.

### DIFF
--- a/app/views/works/_title_and_dates.html.erb
+++ b/app/views/works/_title_and_dates.html.erb
@@ -1,4 +1,4 @@
-<header class="show-title">
+<header>
   <% if view.genre.present? %>
     <div class="show-genre">
       <%= view.display_genres %>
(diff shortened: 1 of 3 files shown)


eddierubeiz:
OK, this makes sense.

PR #856, opened 2020-09-14
store restricted derivatives in prefix in originals bucket

jrochkind:
We still need to discuss which to do permanently, but this gets things deployable again.


jrochkind:
Screen Shot 2020-09-14 at 11 24 43 AM

Issue #832, opened 2020-08-03
Figure out how to and implement the app to properly and securely enforce restrictions on access to files on S3


jrochkind:
We could go over to all "signed" S3 URLs, with the app gatekeeping whether a user can get a signed url. but the problems with this are potentially:

* could interfere with google indexing of our images and other assets, if their URLs are different on every page load. 
  * Right now, we include S3 non-changing "public" URLs in our google sitemap, which we think is what has led to succesful indexing of lots of our images in google images.  These are also the URLs we actually use on pages to display thumbs. (in multiple resolutions per thumb). 
  * For originals, we have an app URL that *redirects* to a *signed* (different on every access) S3 URL, with a HTTP 302. (Temporary redirect).  Will this still allow google to index them? Well, Google *is* still indexing 40 of our OH PDF's that are delivered like this, so it's not an absolute barrier. Could it be related to why Google is not indexing the other ~200 though? Hard to say. 
  * The redirect approach might be too slow/expensive for thumb src in pages anyway, when we have hundreds of thumb urls on a page. We don't really want those hundreds of extra requests to our app per page load. 
* could be a performance problem for pages with hundreds of asset URLs?

## Plan

Our originals are already treated as non-public, we will make sure originals bucket has public access blocked, the app is already set up to work with that. 

Derivatives are the problem, due to performance and efficiency problems with delivering all the thumbnails. 

After [much analysis](https://bibwild.wordpress.com/2020/08/27/access-patterns-for-s3-hosted-resources-with-access-control/), the least bad thing to do is: Leave existing derivatives how they are on a public S3 bucket, but provide an additional flag on Assets for "secure derivative storage type". If an asset has that flag set, derivatives will be stored in a different location, in a public-access-blocked S3 bucket.  

Such assets for now will generally *not* have thumbnails that can be delivered. As our use case is restricted OH content (PDF and MP3) that won't be shown to the public with thumbnails, and doens't really have useful thumbs to show staff either. So we'll work with what we got. 

## Steps

* bucket — in S3, ansible, Rails Env setting. 

* [x] attribute on asset that can toggle between `public` and `restricted` storage type #855
* [x] Configure shrine so at time of ingest, look at flag to decide which storage to put it in. By also configuring a new Shrine `storage` called `restricted_kithe_derivatives`, with support in ScihistDigicoll::Env like existing storages.  #855
[...]

PR #857, opened 2020-09-14
Facilities to move derivatives between public and restricted storage type

jrochkind:
Ref #832

We catch in ActiveRecord callback if derivative storage type has changed, and execute a background job to move derivatives between storage locations if so. 

* Bg job also makes sure to delete any DZI files for `restricted` storage; we have no way to serve DZI's but from public storage.  It does *not* create DZI files if you move a file from `restricted` to `public`. Images aren't really our use case at all anyway. 

* Not currently handling removing things from derivatives backup bucket, leave that for another ticket. (Is a checkbox in master Issue, to help us not  forget)

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -39,6 +39,17 @@ class Asset < Kithe::Asset
   after_promotion DziFiles::ActiveRecordCallbacks, if: ->(asset) { asset.content_type&.start_with?("image/") }
   after_commit DziFiles::ActiveRecordCallbacks, only: [:update, :destroy]
 
+
+  # Ensure that recorded storage locations for all derivatives matches
+  # current #derivative_storage_type setting.  Returns false only if
+  # there are derivatives that exist, in wrong location.
+  def derivatives_in_correct_storage_location?
+    file_derivatives.blank? ||
+      file_derivatives.values.collect(&:storage_key).all?(
+        self.class::DERIVATIVE_STORAGE_TYPE_LOCATIONS.fetch(self.derivative_storage_type)
+      )
+  end
+
   # What is total number of derivatives referenced in our DB?
   #
   # Since they are now referenced as keys inside a JSON hash in Asset, it's a bit
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -18,6 +18,11 @@ class Asset < Kithe::Asset
   # bucket. Intended for non-published Oral History assets, eg "free access no internet release"
   attr_json :derivative_storage_type, :string, default: "public"
   validates :derivative_storage_type, inclusion: { in: ["public", "restricted"] }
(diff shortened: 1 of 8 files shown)


app/models/asset.rb
```diff
@@ -50,4 +70,12 @@ def self.all_derivative_count
       "select count(*) from (SELECT id, jsonb_object_keys(file_data -> 'derivatives') FROM kithe_models WHERE kithe_model_type = 2) AS asset_derivative_keys;"
     ).first["count"]
   end
+
+  # If derivative_storage_type changed in last save, fire off bg job
+  # to move derivatives to correct place.
+  def ensure_correct_derivatives_storage_after_change
```
eddierubeiz, line 76: I like this declarative approach.

jrochkind:
@eddierubeiz So I used a callback in the model to automatically move derivatives whenever the attribute is changed. By launching a bg job. 

I wonder if that's too clever, and I should remove it from the model, and just have the controller do it upon form submission from the form on the admin asset show page that lets you change `derivative_storage_type`?  And I wonder if the form on on the admin asset show page should have a way to "copy derivatives to correct location" without changing `derivative_storage_type`, in cases where they may be mis-matched?

eddierubeiz:
_So I used a callback in the model to automatically move derivatives whenever the attribute is changed. By launching a bg job.

I wonder if that's too clever, and I should remove it from the model, and just have the controller do it upon form submission from the form on the admin asset show page that lets you change derivative_storage_type?_

I like having this on the model. I'm thinking of a case in which we might want to invoke the "move derivatives" code on a large number of models from a rake task.

_And I wonder if the form on on the admin asset show page should have a way to "copy derivatives to correct location" without changing derivative_storage_type, in cases where they may be mis-matched?_

I'm not opposed to this idea, but it might be unnecessary. I'd consider not including it in this PR; rather, you could create a someday/maybe issue that we can delete if we find the feature isn't needed in practice.

jrochkind:
Cool thanks for feedback and merging. 

Some people don't like this callback approach because it makes it infeasible to *turn it off* if you want to turn it off -- like maybe you want to change the attribute on hundreds of  Assets *without* triggering hundreds of bg jobs. As a model callback makes it pretty hard to disable when using standard ActiveRecord functions. 

Or alternately, if you wind up changing the value in the db *without* using standard AR functions, it wont' trigger the callbacks, but you might forget you need to move things over yourself anyway because of being used to not thinking about it as "automatic". 

But we could pick through details on this forever, appreciate the feedback, and the decision to just go forward with it as is for now!

jrochkind:
@eddierubeiz Also when quoting someone in a github comment using markdown, you can use `>` markdown for a quote, instead of trying to use `_` for italics. 

> This is a quote on a line that begins with `>`

A bit more readable.

PR #859, opened 2020-09-14
Avoid n+1 query in home page recent items

jrochkind:
I noticed an "n+1 query" problem in my logs in dev when loading home page. Each of the recent items needs it's `representative_asset` to display a thumb. They were being fetched in a separate SQL query for each, since they weren't being pre-loaded. Easily fixed at the point of DB fetching by adding the ActiveRecord 'includes'.


eddierubeiz:
Thanks for finding and fixing!

PR #861, opened 2020-09-15
Don't create DZI if derivative_storage_type != public

jrochkind:
Ref #832

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -31,7 +31,7 @@ class Asset < Kithe::Asset
     @dzi_file ||= DziFiles.new(self)
   end
 
-  after_promotion DziFiles::ActiveRecordCallbacks, if: ->(asset) { asset.content_type&.start_with?("image/") }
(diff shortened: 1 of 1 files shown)


eddierubeiz:
By `== public` I assume you mean `!= public` :) ?

jrochkind:
@eddierubeiz thanks for careful read, but I *think* it's right. 

     after_promotion DziFiles::ActiveRecordCallbacks, if: ->(asset) { asset.content_type&.start_with?("image/") && asset.derivative_storage_type == "public" }

After promotion, run `DziFiles::ActiveRecordCallbacks` only if the asset's content type starts with `image/` and the asset has a public derivative storage type. 

Otherwise, don't run the callbacks, which create DZI files. 

I didn't include any specs for this change, which is a bit annoying, but it's a bit annoying to write specs for this stuff. I just tried changing to `!= 'public` and running existing tests...  Indeed it made some existing tests fail to do that, existing tests that made sure assets with default value of public derivatives had DZI created! Including a failure on:  `DziFiles Asset life-cycle automatic actions asset creation queues dzi creation`. So that's good news. 

There *isn't* an additional test here to make sure DZI *isn't* created for an asset which had initial value restricted derivatives. Despite being annoying, I guess you could tell me there should be.

Issue #860, opened 2020-09-14
[scihist_digicoll/production] TypeError: String does not have #dig method


jrochkind:
## Backtrace

    line 456 of [PROJECT_ROOT]/app/controllers/catalog_controller.rb: swap_range_limit_params_if_needed
    line 24 of [PROJECT_ROOT]/app/controllers/application_controller.rb: block in batch_kithe_indexable
    line 23 of [PROJECT_ROOT]/app/controllers/application_controller.rb: batch_kithe_indexable

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/67713855)

PR #862, opened 2020-09-15
catch query params bad for Blacklight

jrochkind:
Ref #517 and #860

https://app.honeybadger.io/projects/58989/faults/57865973
https://app.honeybadger.io/projects/58989/faults/67713855


eddierubeiz:
Great idea. Awesome.

PR #867, opened 2020-09-22
basic derivative storage type auditor

jrochkind:


jrochkind:
This doesn't have an in-app "dashboard" report yet, but does have a nightly task that notifies via HB and email. It can be merged, with future work in another PR.

app/lib/scihist_digicoll/env.rb
```diff
@@ -315,10 +315,10 @@ def self.shrine_dzi_storage
     # OUTGOING EMAIL:
 
     # From address:
-    define_key :no_reply_email_address
+    define_key :no_reply_email_address, default: "no-reply@sciencehistory.org"
     # To addresses (these are email lists maintained by IT.)
-    define_key :digital_tech_email_address
-    define_key :digital_email_address
+    define_key :digital_tech_email_address, default: "digital-tech@sciencehistory.org"
+    define_key :digital_email_address, default: "digital@sciencehistory.org"
```
eddierubeiz, line 321: Good call re: the default. Thanks.

jrochkind:
I think I left the defaults off initially to *require* the local_env to supply it, to make sure it didn't get out of sync or whatever. 

But in this case it made testing hard, since the test wanted to test something that required these values to be there. I guess I could have provided a fake testing default only in test, but I was lazy and tired.

app/lib/scihist_digicoll/env.rb
jrochkind: I think I left the defaults off initially to require the local_env to supply it, to make sure it didn't get out of sync or whatever.

But in this case it made testing hard, since the test wanted to test something that required these values to be there. I guess I could have provided a fake testing default only in test, but I was lazy and tired, good enough.

Issue #866, opened 2020-09-21
Analysis of access control issues with S3 backup buckets


jrochkind:
Pulling out from #832 this specific issue for more analysis. 

_If you change derivatives storage from public to restricted, we need to make sure any copies from derivative *backup* bucket get properly deleted too. Also DZI backup bucket. Also how does this apply to version history_

Some questions involved:

* If an object is deleted from the derivative bucket or dzi, does that deletion get automatically replicated to backup bucket already?

* Even if so, it might still be there in version history. Are the *versions* publically available, so they need to be purged too?

jrochkind:
Note we don't have backup buckets (derivatives, dzi, originals) for staging, only for production.

jrochkind:
OK, here's some analysis. Did this with a test object on production, since staging doesn't use backup buckets. 

1. Created an asset on public derivatives. 

2. Verified that derivatives *were* automatically replicated to backup bucket `scihist-digicoll-production-derivatives-backup` (although had to wait for a couple minutes for them all to replicate)

3. Using staff UI, changed derivatives storage to "restricted"

4. Verified that derivatives were gone from derivatives bucket `scihist-digicoll-production-derivatives`

5. They were NOT deleted from the backup bucket. Do our replication rules not replicate deletes? Or did I just not wait long enough for replication to happen?

6. Just to see what happened, *manually* deleted a derivative from backup bucket. 

7. The older version in version history **IS** still there and **IS** publicly accessible if you know the URL!
  * The URL is even more obscure. In addition to a fairly random URL, you'd ALSO have to have the fairly random `versionID` that goes on the end of the URL.  Eg: `https://scihist-digicoll-production-derivatives-backup.s3-us-west-2.amazonaws.com/0014db87-8949-47f6-bdf2-5b6541eeeb5b/download_small/375450757aca863d65156b1441e35c53.jpeg?versionId=SexTgv_pQAtUS8QGBIlTY4eKmANAYwHT`. But anyone that does have this URL can access the thing, even though it was deleted, the version is still there at that URL. 
  * Why does our derivatives backup bucket keep version history for 30 days? (It only keeps old versions for 30 days). Not sure, not sure we need it. But we are not looking to changing policy decisions like this in a hurry as part of the OH feature implementation, especially since without Dan we don't necessarily have all the context. 

[...]

PR #868, opened 2020-09-23
remove all versions and backups when moving derivatives to restricted

jrochkind:
Ref #832 and #866 

Needs ansible PR with new local_env.yml keys just in production, s3_derivatives_backup_bucket; s3_dzi_backup_bucket; s3_backup_bucket_region

Needs to be manually tested in
- [x] staging (everything works when moving derivative storage type; derivatives, dzis, and versions of both are gone in buckets)
- [x] production (above plus all versions are gone in _backup_ buckets too)

This doesn't have a lot of tests, I just can't see a good way to make a worthwhile test that doesn't assume too much.

### DIFF
--- a/app/jobs/ensure_correct_derivatives_storage_job.rb
+++ b/app/jobs/ensure_correct_derivatives_storage_job.rb
@@ -6,12 +6,8 @@
 class EnsureCorrectDerivativesStorageJob < ApplicationJob
 
   def perform(asset)
+    byebug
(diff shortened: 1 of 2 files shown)


jrochkind:
This seems to be making tests fail in weird hard to reproduce ways, both on travis and locally. Which is weird, not sure why.

eddierubeiz (approved):

Issue #843, opened 2020-08-10
No Google Analytics events for in browser view/listen on Oral History download page


jrochkind:
If you choose a "download" link on Oral History Downloads tab, it gets recorded in a Google Analytics custom action. 

![Screen Shot 2020-08-10 at 3 22 45 PM](https://user-images.githubusercontent.com/149304/89821982-77e12600-db1d-11ea-9100-c3ed9b79c4ec.png)

But if you click to view the PDF in-browser:

![Screen Shot 2020-08-10 at 3 24 04 PM](https://user-images.githubusercontent.com/149304/89822085-9e9f5c80-db1d-11ea-87fa-e47bae6eca4f.png)

No Google Analytics event is generated. Is this a problem? @HKativa 

(note, there are also no Google Analytics events being generated for in-browser audio player listens, that's a separate issue that can be created).

HKativa:
Hmm, we're tracking "view" for all other works so we should do the same for oral history PDFs. With regards to the in-browser audio player listens, it seems worthwhile to track these as well, but I'd also like to hear what @lsberry16 thinks as the stakeholder.

jrochkind:
Makes sense, but please just remember to ask us to prioritize this at a Monday meeting, for us to work on it.

jrochkind:
Makes sense, I don't think anyone would not want that!

The issue is more priorities. Should we work on this *now*, instead of focusing on limited access oral histories and in general on trying to replace OH microsite features, or should we just leave this in the backlog for later?

Just let us know when you'd like us to work on it is all.

jrochkind:
Also just do a review/audit of all relevant links on our new OH "audio"/OHMS page, to make sure they are google analytics tracked.

PR #869, opened 2020-09-28
Better Google Analytics reporting for oral histories.

eddierubeiz:
Ref #843 
The main task in 843 is done and tested.
Also adding a GA action for in-browser audio player listens as well.

### DIFF
--- a/spec/system/oh_audio_front_end_spec.rb
+++ b/spec/system/oh_audio_front_end_spec.rb
@@ -29,7 +29,7 @@ describe "Audio front end", type: :system, js: true do
     audio_assets.select {|a| a.published }
   }
 
-  let!(:regular_assets) {
+  let!(:image_assets) {
     (5..8).to_a.map do |i|
       create(:asset_with_faked_file,
         title: "Regular file #{i}",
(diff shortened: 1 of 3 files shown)


eddierubeiz:
@lsberry16 and @HKativa, a note for yous:

When the user clicks on a track on the OH player front-end, this code would notify Google Analytics that the user has performed action `play_oral_history_audio_segment` action on an *unspecified* audio segment of a particular oral history work.

**Pro**: This makes it easy to see which oral histories are being listened to in the browser.
**Con**: It makes it impossible to compare the popularity of individual *segments*  within a particular oral history.

I'm assuming this is fine; please let me know otherwise.

eddierubeiz:
Lee confirmed on slack: "I can’t imagine wanting to compare the popularity of segments. I think we’ll be great with just knowing which interviews are getting played." Please review whenever convenient, @jrochkind!

spec/system/oh_audio_front_end_spec.rb
```diff
@@ -59,6 +67,24 @@
 
       click_on "Downloads"
 
+
+      # In oh_audio_work_show_decorator.rb we specify that only PDFs should be linked.
```
eddierubeiz, line 71: Added an extra test here for untested functionality in `oh_audio_work_show_decorator.rb` -- out of scope, I know, but seemed like a good idea at the time.

PR #872, opened 2020-10-04
Oral history audit reporting

eddierubeiz:
No actual report yet; just storing the results of the audit in the DB and testing the fact that they are stored.

TODO:
- [x] Adjust nomenclature as needed so it's consistent and makes sense. (Log? Report?)
- [x] Move the model to its proper spot -- for now it's at the root level at /app/models which is obviously not good
- [x] Delete reports older than a month
- [x] Create a route / controller / template for the admin page to show the report
- [ ] Probably no need to actually test the admin page.


jrochkind:
Is there any reason to store more than ONE record, the most recent one? Is there any reason anyone would want to see a HISTORY of report values , instead of just the current status?

If not, can the code be made simpler, the front-end only needs to display one the current one, the deletion code can just delete all but one.

eddierubeiz:
Don't feel that strongly about it either way, but changed the code to only store one report. We can always go back if we feel the need to store more than one. The code is of course simpler without iterating over n reports.

app/services/asset_derivative_storage_type_auditor.rb
```diff
@@ -31,6 +31,9 @@
 class AssetDerivativeStorageTypeAuditor
   attr_reader :incorrect_storage_locations, :incorrectly_published
 
+  # We're storing a short list of sample friendlier_ids of assets with
+  # problems. How many?
+  PROBLEM_SAMPLE_SIZE = 10
```
jrochkind, line 36: Should we make sure to mention to stakeholders when presenting this that if there's a failure, the report will only tell about at most 10 failed ones? Then I guess they'll have to ask developers for help finding if there are more? I have no opinion on whether this is sufficient, we should just clear it with stakeholders.

eddierubeiz: Oh, I highlighted the recipe in the copy on the form: ```This report confirms that files that should not be accessible to the public are indeed locked down, and vice-versa. To save space, we only list the total number of assets with problems, along with a short sample of actual links to some of those assets. Reports are run nightly, but we only keep the latest one.```

app/services/asset_derivative_storage_type_auditor.rb
```diff
@@ -79,13 +98,47 @@ def notify_if_failed
...
+  end
+
+  def delete_stale_reports
+    sql = """
+      SELECT old.id FROM asset_derivative_storage_type_reports old
+      WHERE old.id NOT IN (
+        SELECT id FROM asset_derivative_storage_type_reports new
+        ORDER BY created_at DESC
+        LIMIT 1 )
+    """
```
jrochkind, line 136: You could probably do this with ActiveRecord instead of raw SQL, which might be more reliable with regard to future code changes and avoiding other kinds of bugs. But this is probably good enough for now, why not. 

Although wait we already know what report we created. How about simply:  `AssetDerivativeStorageTypeReports.where.not(id: report.id).delete_all` (or `destroy_all` if you wanted to trigger Rails callbacks, which shoudln't probably matter here).

app/services/asset_derivative_storage_type_auditor.rb
eddierubeiz: And the UI makes it clear that it's not the full list. When there are problems to display, I list the ones we have stored, but followed by an ellipsis, and then the total number of items in parenthesis.

app/services/asset_derivative_storage_type_auditor.rb
eddierubeiz: You are so, so right about this. Thanks :)

app/services/asset_derivative_storage_type_auditor.rb
jrochkind: OK let's just make sure to bring it up when presenting for review/approval too, since it's easy to miss on a quick skim. They may ask "why'd you do it that way".

PR #879, opened 2020-10-06
if these required bucket config env are missing, fail earlier with clearer error message

jrochkind:


eddierubeiz:
Great.

eddierubeiz:
I merged master into this branch first, to resolve a conflict from the FasterS3Url branch.

PR #882, opened 2020-10-06
fail earlier on missing AWS keys when configuring buckets

jrochkind:
Debugging puma problems, was a failure here that wasn't as obvious to debug as it could have been. This way it will actually prevent deploy with a raise on deploy


eddierubeiz:
Weird. I thought I had merged this already, but OK.

jrochkind:
> Weird. I thought I had merged this already, but OK.

there was a similar one, but I realized it didn't go far enough! This adds to it.

PR #884, opened 2020-10-07
specific config for puma in heroku

jrochkind:
from https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server

These files shouldn't effect existing non-heroku environments, we just have a Procfile telling heroku we want it to start our web dynos with puma, with a specific config file we've also added for this purpose.

Originally we tried telling puma to default to 2 workers with 5 threads each -- that was overwhelming heroku. This one for now says 1 worker with 5 threads. Can be adjusted without code change by heroku config/env change. We are investigating performance/resource characteristics still; things are looking a bit problematic.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,3 @@
+# This specifies what dyno types (worker VMs) are set up for heroku
+
+web: bundle exec puma -C config/heroku_puma.rb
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Noted.

PR #885, opened 2020-10-08
More careful abolute path to load .ruby-version

jrochkind:
Our Gemfile is dynamically loading the local .ruby-version file and doing a bit of trivial parsing of it's contents to supply a `ruby` decleration in Gemfile.

To load .ruby-version, it was supplying a relative path assuming that the current working directory was app root. This seems a safe assumption, but for some reason it was failing in a weird edge case involving passenger on heroku. Without debugging exactly what was going on -- it doesn't hurt to be more specific, using the ruby special variable `__dir__` (directory of current source file) to construct an absolute path to `.ruby-version` in the same directory. Less sloppy, more careful, resolves our weird problem with passenger on heroku.


eddierubeiz:
Sounds good.

Issue #828, opened 2020-08-03
Oral history request form, and logging


jrochkind:
for "free access without internet release". 

Should have:

* name (required)
* email (required)
* institution (optional)
* intended use (required)

Will get logged to back-end. All fields should be kept "encrypted" using same technique we used for R&R. https://github.com/sciencehistory/scihist_digicoll/issues/621  https://bibwild.wordpress.com/2020/03/05/encrypting-patron-data-in-rails-why-and-how/

Staff should be able to download log in spreadsheet form, giving date ranges of requests to include. 

Staff should be able to see "recent" delivered requests in browser. 

Note: our app does not currently support different feature access for different staff, anyone with a staff account *will* be able to see these. 

What happens once you fill out the form will be a different ticket.

eddierubeiz:
Thanks for gathering these requirements, @jrochkind.

Issue #829, opened 2020-08-03
Delivering request-form-gated oral history assets


jrochkind:
See #828  re request form. 

After request form is filled out for these "free access no internet release" histories, the asset files (PDF as well as audio) can be delivered to user without staff intervention. 

How will we deliver?

1.  email with attachments (audio files start getting big)
2.  email including time-limited non-guessable private URLs
3.  just getting those time-limited non-guessable private URLs in browser after filling out form
4. Combination of 2 and 3, get it in the browser, but also get it in email in case you close your browser but haven't gotten em yet? 

We are leaning toward 3 or maybe 4.

jrochkind:
Thought of another thing: What address should the email come from?

I think it's time to move on from our `no-reply` email with our emails that actually go to users. Why not give them an email address they can actually reply to if they need help?

`digital@sciencehistory.org` might be the obvious one -- it already appears on https://digital.sciencehistory.org/contact, so just save them the step of having to look it up, and let them reply. 

Or if OH team would rather an email addr that goes right to them for these OH emails specifically, their choice.

eddierubeiz:
Sending it from digital@ for now -- easy to change.

PR #886, opened 2020-10-12
Oral history request form

eddierubeiz:
Simple public-facing form to request access to restricted oral histories; ref #828.
Patron PII is encrypted using Lockbox.

**TODO:**
- [x] Get approval for the legal verbiage, design, etc.
- [x] Actually connect the OH front-end to this form.
- [x] Tests

**NOT IN THE SCOPE OF THIS PR:**
- Staff should be able to download a log of requests, in the form of a spreadsheet, giving date ranges of requests to include.
- Delivery of the actual links to the patron.
- All admin functionality.


jrochkind:
Hi @eddierubeiz 

I think you really want to ref #828 rather than #829 for this one? #828 is the OH requset form; #829 is actual delivery of assetes. Let's start with #828, yes? 

I also am thinking it would be best if we separate "filling out and recording the form" and "staff view of forms that have been filled" out into two separate PRs. So, please, for this PR:

So:

* Can you put staff view of requests in a *different* PR please, leave it out of this initial one?

* Contrarily, can you include "Actually connect the OH front-end to this form" in this PR?  

#828 specifies the fields that should be on the form; that the form submission should be recorded in a database table; and that the database table should use our lockbox encryption techniques. 

I'd like to see a PR with just those things; then a separate PR with the staff interface for viewing the list of recorded requests; then thirdly, another PR (or more than one, we haven't fully spec'd it and need to discuss it more) that is about actually delivering assets to patrons.

eddierubeiz:
Noted. In that case this PR is almost done, except for tests.

app/controllers/oral_history_access_requests_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      redirect_to work_path(@work.friendlier_id), notice: 'Your request has been logged.'
+    else
+     render :new
+    end
+  end
+
+private
+  def cls
+    Admin::OralHistoryAccessRequest
+  end
```
jrochkind, line 26: What's the point of putting this in a method? Why not just say `Admin::OralHistoryAccessRequest.new` above, instead of the layer of indirection? Actual question, you can have an answer! Curious where you saw this pattern or whatever.

app/controllers/oral_history_access_requests_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  def new
+    @oral_history_access_request = cls.new
+  end
+
+  # POST "/request_oral_history_access"
+  def create
+    @oral_history_access_request = cls.new(oral_history_access_request_params)
+    @oral_history_access_request.work = @work
+    if @oral_history_access_request.save
+      redirect_to work_path(@work.friendlier_id), notice: 'Your request has been logged.'
```
jrochkind, line 17: For now, since this functionality isn't actually done yet, let's actually leave it in a state where nobody can accidentally think it's working if we accidentally turn it on in production. Instead of the redirect with a message that tells them it's been logged, how about:

    render plain: "This functionality is not activated yet, and is only present for testing!"

app/controllers/oral_history_access_requests_controller.rb
eddierubeiz: It was just sheer irritation at having such a bulky identifier for the class. Maybe this wasn't such a great idea :)

app/controllers/oral_history_access_requests_controller.rb
eddierubeiz: Sure, I can see that. Will do.

app/controllers/oral_history_access_requests_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      :patron_institution, :intended_use, :status, :notes)
+  end
+
+  def set_work
+    if params['admin_oral_history_access_request'].present?
+      @work_friendlier_id = params['admin_oral_history_access_request'].delete('work_friendlier_id')
+    else
+      @work_friendlier_id = params['work_friendlier_id']
+    end
+    @work = Work.find_by_friendlier_id(@work_friendlier_id)
```
jrochkind, line 40: This condition is because the ID comes in two different ways depending on whether it's the #new or #create action? 

Instead of one `before_action` with a condition, let's just inline this into the two actions? For instance, the first line of new could just be:

    @work = Work.find_by_friendlier_id(params['work_friendlier_id'])

In fact, also, use the verison with a `!` to get an exception raised if no work could be found: `Work.find_by_friendlier_id!(params['work_friendlier_id']`

Then you'd have to re-add the work for raising if the work is not marked requestable. But the net change will be FEWER lines code, trying to "DRY" with a before_action actually resulted in more code and more complicated code! Simplify it, have each action do what it needs.

eddierubeiz: Good-- I wrestled with this for a bit and this is a better approach.

app/models/admin/oral_history_access_request.rb
```diff
@@ -0,0 +1,11 @@
+class Admin::OralHistoryAccessRequest < ApplicationRecord
+  encrypts :patron_name, :patron_email, :patron_institution, :intended_use
+  belongs_to :work
+  validate :work_exists
+  def work_exists
+    Work.exists?(self.work_id);
+  end
```
jrochkind, line 7: I can't find this `work_exists` method used anywhere in the code?  

I don't think it's needed.   If the work doesn't exist, `request.work` will be nil. So you can always just check `request.work.present?` whenever you want to check if it exists, no need for this method. (If the method WAS going to remain, it should be end in a `?`, convention for boolean returning methods!

app/views/oral_history_access_requests/_form.html.erb
```diff
@@ -0,0 +1,64 @@
...
+        <%= f.input :intended_use,
+          label: "Intended use", hint: "So that we can better understand our audience, please tell us how you plan to use the oral history you have requested." %>
+      </div>
+    </div>
+
+    <%= f.hidden_field :work_friendlier_id, value: @work.friendlier_id %>
+
+  <div class="form-actions mb-3">
+    <%= link_to "Cancel", work_path(@work.friendlier_id), class: "btn btn-outline-primary" %>
+    <%= f.button :submit, value: "#{f.object.new_record? ? 'Grant access' : 'Edit'}" %>
```
jrochkind, line 61: "Grant Access" isn't the right term. Maybe just "Submit Request"? And I don't think this form will ever be used with an existing record in `edit` mode? We could leave that boilerplate so it will magically work if it ever is... or just simplify to what we're really using.

app/models/admin/oral_history_access_request.rb
eddierubeiz: I'm using the method for validation: `validate :work_exists`, but this is also enforced at the DB level. Will change the method name -- yes, question mark for boolean return values. I was just skeeved by the idea of saving requests for works that didn't exist, I gues.

app/views/works/work_file_list_show.html.erb
```diff
@@ -23,7 +23,13 @@
...
         <div class="pr-2 long-text-line-height">
         <div class="pb-sm-3 pl-sm-3 text-nowrap float-sm-right">
-          <%= link_to "Request Access", "javascript: alert('under construction, not yet implemented')", class: "btn btn-emphasis mb-2" %>
+          <%= link_to "Request Access",
+            request_oral_history_access_form_path(@work.friendlier_id),
+            class: "btn btn-emphasis mb-2",
+            analytics_category: 'Work',
+            analytics_action: "show_restricted_oral_history_request_form",
+            analytics_label: @work.friendlier_id
+          %>
```
jrochkind, line 32: Let's confirm that stakeholders want this tracked in google analytics before tracking it. 

Do we want to track when the user just *displays the form* without submitting it? Which is what's being done here? Or instead, track when the user submits the form? Or track both actually, I guess would let us see if people ask for the form but don't submit it?  Or maybe we don't actually care about any of this and don't need to track anything, we don't need to make extra work for ourselves if our stakeholders don't want/need it.

app/views/oral_history_access_requests/_form.html.erb
eddierubeiz: Correct -- the patron is not granting him/herself access- that's kinda wrong. Thanks for catching.

app/views/works/work_file_list_show.html.erb
eddierubeiz: Fair. I can simply remove. We are tracking actual submitted requests after all so this is probably overkill.

app/models/admin/oral_history_access_request.rb
jrochkind: Hmm, I see. I'm not sure I've ever seen this done before in Rails. i guess I see what you're doing, but it should be extremley unusual for this to happen, the app wouldn't normally allow it, unless I guess a work was deleted right as someone were filling out a request? And then the db will enforce it anyway, so....

I'm just thinking the less code, the less code to maintain/bugs/etc. But if you feel strongly, you could leave the validation I guess.

app/views/works/work_file_list_show.html.erb
jrochkind: Oh, I guess actual submission of form is tracked in the db, no need to track it in GA, right!

I'm not sure we need to track every time someone asks for the form to be displayed (and may or may not submit it) though, unless this is something stakeholders brought up?

app/controllers/oral_history_access_requests_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+private
+  def oral_history_access_request_params
+    params.require(:admin_oral_history_access_request).permit(
+      :work_friendlier_id, :patron_name, :patron_email,
+      :patron_institution, :intended_use, :status, :notes)
+  end
+
+  def set_work
```
jrochkind, line 34: How about `set_work(friendlier_id)` instead of relying on instance variable `@friendlier_id` as state? 

Or even how about `@work = load_work(params['work_friendlier_id'])` for even more explicitness? Now `load_work` neither relies on nor sets mysterious instance state, it's just a function with inputs and outputs all visible. 

I know Rails really likes instance variables, and they are the standard way of communicating with views. But they are also rightly widely critisized for making the coupling between code really mysterious. When `@work_friendlier_id` is not being used in the view, there is no reason not to make this code much more explicit and standard good design.

config/initializers/filter_parameter_logging.rb
```diff
@@ -5,3 +5,6 @@
 
 Rails.application.config.filter_parameters += [:patron_email]
 Rails.application.config.filter_parameters += [:patron_name]
+Rails.application.config.filter_parameters += [:patron_institution]
+Rails.application.config.filter_parameters += [:intended_use]
```
jrochkind, line 9: nice catch I never would have remembered this!

app/controllers/oral_history_access_requests_controller.rb
eddierubeiz: That works!

config/initializers/filter_parameter_logging.rb
eddierubeiz: The magic of grep

PR #890, opened 2020-10-14
Oral history request: admin page(s)

eddierubeiz:
This is the logging part of Ref #828:
**TODO:**
- [x] Staff should be able to download log in spreadsheet form, giving date ranges of requests to include.
- [x] Staff should be able to see "recent" delivered requests in browser.
- [x] Tests

**NOT PART OF THIS PR:**
Delivery of the actual OH content to the patron.


jrochkind:
Set up a work in staging that is faked to have restricted-by-request media. https://staging-digital.sciencehistory.org/works/02870x01s

(Note Front Matter and Index PDF is from another OH, for demonstration purposes only)

spec/controllers/admin/oral_history_access_requests_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(response.headers["Content-Disposition"]).to match(/attachment; filename=.*oral_history_access_requests.*csv/)
+      expect(response.media_type).to eq "text/csv"
+      response_lines = response.body.lines
+      expect(response_lines.count).to eq 11
+      expect(response_lines[0]).to eq  "Date,Work,Name of patron,Email,Institution,Intended use\n"
+      expect(response_lines[8]).to match  /Oral history interview with William John Bailey/
+      expect(response_lines[8]).to match  /Patron 8/
+      expect(response_lines[8]).to match  /patron@institution_8.com/
+      expect(response_lines[8]).to match  /Institution 8/
+      expect(response_lines[8]).to match  /I will write 8 books/
```
jrochkind, line 34: let's actually "parse" the CSV returned. Instead of just testing that "Patron 8" is in the line SOMEWHERE, let's actually use ruby code to parse CSV, make sure it parses, and the various fields are in the right places. 

Also I just added a column for Work URL, so that can be included.

jrochkind:
@MDiMeo  @lsberry16  The text on the form says:

> You are receiving a copy of this oral history and <strong>you are not permitted to reproduce it</strong> in any way without the permission of the Science History Institute.

But we're also giving them a "rights statement" that says "NON-COMMERCIAL USE PERMITTED", and links to a description saying: "no permission is required from the rights-holder(s) for non-commercial uses. For other uses you need to obtain permission from the rights-holder(s)."

I feel like we're giving them confusing mixed messages -- this stuff is confusing enough already. 

Thoughts?

(As a larger issue, I've thought for some time that for things where the Institute is the rights holder, we ought to be giving them a CC license, not just a rights statement, but that's a different larger issue and I could be wrong).

jrochkind:
Code looks good @eddierubeiz, nice job. Made some UI tweaks, including to form that was part of previous UI. 

Made one comment about tests, which you are free to take or leave. Other than that, code is reviewed, but I think maybe it makes sense to demo and get approval from stakeholders before merging and closing issue? Can you take on that?

jrochkind (approved):

app/controllers/admin/oral_history_access_requests_controller.rb
```diff
@@ -0,0 +1,59 @@
+# STAFF FACING
+# Public facing actions are in app/controllers/oral_history_access_requests_controller.rb
+require 'csv'
+
+class Admin::OralHistoryAccessRequestsController < AdminController
+  def index
+    @oral_history_access_requests = Admin::OralHistoryAccessRequest.
+    where('created_at > ?', 3.months.ago).order(created_at: :desc).to_a
```
jrochkind, line 8: Would it make more sense to show most recent N requests instead of 3 months? Say most recent 50 or something? `.order(created_at: :desc).limit(50)`?  Not sure.

eddierubeiz: It occurred to me. Will ask Lee.

eddierubeiz:
Re: verbiage: this is above my pay grade, but I would suggest something like:

- Your receipt of an electronic copy of an oral history indicates your agreement to abide by U.S. copyright law **and the oral history's rights statement** with regard to the reproduction and dissemination of the oral history.
- Get rid of the part where it says you can't disseminate it; that's superseded by the above.
- Get rid of the "You are receiving a copy of this oral history" part -- that's superfluous.

jrochkind:
Alternately it's possible the rights statement isn't right for these "free access no internet release" documents, and it could be changed to just "in copyright". 

I'm not opposed to providing additional guidance than just the phrase-and-link of our current rights statement; I am just concerned if doesn't seem to match!

lsberry16:
Thanks, Eddie and Jonathan! Let me check in with Dave about the wording and get back to you.

Sent from my iPad


Lee Berry   (she, her, hers)
Curator of Oral Histories
Center for Oral History
t. +1.215.873.8251

Science History Institute

On Oct 20, 2020, at 11:51 AM, Jonathan Rochkind <notifications@github.com> wrote:

﻿

Alternately it's possible the rights statement isn't right for these "free access no internet release" documents, and it could be changed to just "in copyright".

I'm not opposed to providing additional guidance than just the phrase-and-link of our current rights statement; I am just concerned if doesn't seem to match!

spec/controllers/admin/oral_history_access_requests_spec.rb
eddierubeiz: Thanks for adding the work URL (duh!). CSV is parsed now.

eddierubeiz:
Include the Oral History number as well in both the table in admin and the downloaded spreadsheet.
The Oral History number is listed in the work metadata under the label "External ID".

eddierubeiz:
Confirmation message should read "Links to the files are being sent" rather than "Files are being sent"

app/controllers/admin/oral_history_access_requests_controller.rb
eddierubeiz: No, past 3 months is fine, said Lee.

jrochkind: I'm worried past 3 months would be TOO MANY lines, too big a page, performance problems. Since it's unbounded as to actual number, it just depends on how many requests there have been. Is why I would have made it a fixed number. But we can just deal with that if it becomes a problem, I guess.

Issue #895, opened 2020-10-15
fix n+1 problem in current member display page calling #member_content_types


jrochkind:
This was always hacky. Current code for work show page is triggering an n+1 fetch for all leaf representatives for #member_content_types, console log looks like:

```
Calling Work#member_content_types without pre-loading members => leaf_representative may be dangerous to performance
  CACHE Kithe::Model Load (0.0ms)  SELECT "kithe_models".* FROM "kithe_models" WHERE "kithe_models"."id" = $1 LIMIT $2  [["id", "bf0d487e-fb6a-4375-b977-25f89a85e584"], ["LIMIT", 1]]
  ↳ app/models/work.rb:157:in `block in member_content_types'
  CACHE Kithe::Model Load (0.0ms)  SELECT "kithe_models".* FROM "kithe_models" WHERE "kithe_models"."id" = $1 LIMIT $2  [["id", "ba213170-6783-4955-87fc-b22661a6f72c"], ["LIMIT", 1]]
  ↳ app/models/work.rb:157:in `block in member_content_types'
  CACHE Kithe::Model Load (0.0ms)  SELECT "kithe_models".* FROM "kithe_models" WHERE "kithe_models"."id" = $1 LIMIT $2  [["id", "ba23681a-bb11-4d64-9dd5-55673919d517"], ["LIMIT", 1]]
  ↳ app/models/work.rb:157:in `block in member_content_types'
  CACHE Kithe::Model Load (0.0ms)  SELECT "kithe_models".* FROM "kithe_models" WHERE "kithe_models"."id" = $1 LIMIT $2  [["id", "15037510-f2ad-4c16-92cc-5e764d3e4a18"], ["LIMIT", 1]]
  ↳ app/models/work.rb:157:in `block in member_content_types'

[…]
```

We need to change to something less clever, that makes it harder to mess this up. Logs have probably been printing this for a while with nobody noticing.

PR #896, opened 2020-10-19
Avoid expensive/extra loading of #members association on standard Work show page

jrochkind:
When dealing with works that can have hundreds of memberes, loading all #members can be expensive. Worse is doing it more than once, or doing it in a way that causes an "n+1 query" problem, with an extra SQL query per member. We had two main problems/categories of problem discovered here:

1. To actually load the members for display, WorkShowDecorator#member_list_for_display constructs a query with SQL WHERE limits and ORDER BY clause, that does NOT end up memoized/cached in #members. So any subsequent calls to `work#members` will end up loading all members AGAIN a second time. It was expensive enough the first time, we don't want to do this. In most cases we want to replace it by a targetted SQL query getting the aggregate info we wanted -- since it's a work#show page, we can do one query per main work displayed -- there's only one main work displayed. Doing a few queries to get a few different pieces of info is okay, and better than loading #members to do it in-memory. 

2. One thing we want in a couple places is the mime types of all members. This is tricky to get! The previous implementation of work#member_content_types loaded #members and then looked at them in-memory to calculate. In some cases, if members hadn't been pre-loaded properly, this would result in an n+1 query problem. The previous implementation recognized that, and printed a warning to logs -- but we never noticed that, it was deployed to production with the warning!  The new implementation won't just warn, it will raise refusing to do it. BUT ALSO provides a different mode, where instead of looking at #members in-memory, it actually goes to database with another query. Either could be disastrously expensive depending on circumstances; the caller needs to decide which mode they want.  (Ref #895)

This is all a bit hacky, and I do worry it's easy to have a performance regression where it gets slow again when we write new code. But it's just part of how it goes I think. We also put a check in a controller spec that ensures work#members actually was NOT loaded! Can protect against some things, but not everything. 

Some attempt at measuring what effect this has on performance coming in a comment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -145,6 +145,8 @@ group :test do
   gem "db-query-matchers", "< 2.0"
   # Used for Cap deployment to auto-lookup hosts from EC2 tags.
   gem 'aws-sdk-ec2', '>=1.74'
+
+  gem 'rails-controller-testing'
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -423,6 +423,10 @@ GEM
       bundler (>= 1.3.0)
       railties (= 6.0.3.3)
       sprockets-rails (>= 2.0.0)
+    rails-controller-testing (1.0.5)
+      actionpack (>= 5.0.1.rc1)
+      actionview (>= 5.0.1.rc1)
+      activesupport (>= 5.0.1.rc1)
     rails-dom-testing (2.0.3)
       activesupport (>= 4.2.0)
       nokogiri (>= 1.6)
@@ -673,6 +677,7 @@ DEPENDENCIES
   puma (~> 4.3)
   qa (~> 5.2)
   rails (~> 6.0.0)
+  rails-controller-testing
   rails_autolink (~> 1.0)
   ransack (~> 2.1)
   resque (~> 2.0)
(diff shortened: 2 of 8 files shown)


jrochkind:
There are a variety of things/ways we can measure.  

* `time curl url >/dev/null` how long does it take curl to get it, dumping everything it gets into the trash
* Chrome inspector window time, both "time to first byte"  (TTFB) and total time
* number of object allocations in rails logs (effects RAM use, and also tends to effect performance) (just looked at Rails logs on staging vs production when I had this new code deployed on staging). 

These are all for [Ramelli](https://staging-digital.sciencehistory.org/works/4b29b614k) -- a very large/expensive page, that maybe can't exist this large anyway.  I'm just doing a few requests by hand and seeing what they look like, not being done very scientifically, so could be subject to just incidental variance. 

|   | Original | After this PR | 
| ------------- | ------------- | ------------- |
| EC2 curl /dev/null   | 2.1-2.8s  | 1.6-1.8s |
| EC2 Chrome TTFB  |  2.0s  | 1.4s | 
| EC2 Chrome total time | 3-3.5s | 1.5-2s |
| Object allocations | 1,724,882 | 1,276,613 |
| Heroku standard-1x curl /dev/null | 4.5-5.8s | 2.3-3.0s (one outlier at 3.6s) |

These measurements weren't done entirely scientifically, but these do seem to be significant improvements.

PR #901, opened 2020-10-21
Deliver oral history via email

eddierubeiz:
Ref #829
Just focusing on the email for now.
Per our discussion at today's meeting, I'm advising we make the *design* of "professional"-looking outgoing emails a separate issue; hence ref #904  .
Making sure the verbiage is correct is in issue #912 .

TODO:
- [x] Tests
- [x] There's no need for a a special page with the download URLS listed there; the UI in this PR is fine, per Lee.


jrochkind:
Start out with a confirmation page that just says "they've been emailed to you", then we can ask stakeholders if they want asset links direclty there too. 

One thing for verbiage is making sure we let them know the links will expire after 7 days (also have to check with stakeholderes that 7 days is what they want; we can go less but not more using S3 presigned links), and they should download to their own computer if they want to keep the files longer.

jrochkind:
@eddierubeiz  

> One week minus 10 seconds ought to work.” 

— were you getting an error at exactly one week? That might be a bug in the FasterS3URL gem I wrote. Can you duplicate and file a ticket on that gem if so?

The error I saw made it look like you were trying TWO weeks, which would be too long. One week plus one second would be too long. But one week exactly (in seconds) ought to be okay.

i think one week in seconds is 60 * 60 * 24 * 7 == 604,800. When using Rails, you can also say 1.week  or 1.week.to_i, which is… 604800, yup.

jrochkind:
> Order the files correctly and make the URL attachment rather than inline

I would have gone inline, so users could look at it in the browser if they wanted, and choose 'save as' from browser if htey wanted to save. 

We should probably ask stakeholders. 

You can do a header that allows inline but still sets filename for when the user chooses 'save as' if you want to.

eddierubeiz:
> @eddierubeiz
> 
> > One week minus 10 seconds ought to work.”
> 
> — were you getting an error at exactly one week? That might be a bug in the FasterS3URL gem I wrote. Can you duplicate and file a ticket on that gem if so?
> 
> The error I saw made it look like you were trying TWO weeks, which would be too long. One week plus one second would be too long. But one week exactly (in seconds) ought to be okay.
> 
> i think one week in seconds is 60 * 60 * 24 * 7 == 604,800. When using Rails, you can also say 1.week or 1.week.to_i, which is… 604800, yup.

I started with two weeks because I wasn't really paying attention. Then I changed it to exactly `1.week.to_i` but I *still* got the error. The current code does not trigger the error.

Happy to file a ticket on fasterS3URL; stay tuned.

lsberry16:
B, maybe? I can be convinced either way.

Sent from my iPad


Lee Berry   (she, her, hers)
Curator of Oral Histories
Center for Oral History
t. +1.215.873.8251

Science History Institute

On Oct 22, 2020, at 10:31 AM, Eddie Rubeiz <notifications@github.com> wrote:

﻿

@lsberry16<https://github.com/lsberry16>, a penny for your thoughts: should clicking the links emailed to the patron

 *   a) download the files directly to the patron's computer or
 *   b) bring them up in the patron's browser, at which point the patron can choose "Save" in order to keep them?
TIA!

eddierubeiz:
I'm popping the from address infrastructure into Ref #903 so we can deal with it separately.

app/views/oral_history_delivery_mailer/oral_history_delivery_email.html.erb
```diff
@@ -0,0 +1,11 @@
+<p>Dear <%= mailer.patron_name %>,</p>
+<p>Thanks for your recent visit to the <%= link_to "Science History Institute’s digital collections website", mailer.hostname %>. On <%=mailer.created_at.strftime("%m/%d/%Y")%>, you requested access to the files for <%= mailer.work_link %>. The files are ready for you to download.</p>
+<h2>Files for <%= mailer.work_link %>:</h2>
+<ul>
+  <% mailer.asset_links.each do |link| %>
+    <li><%= link_to link[:text], link[:url]%></li>
```
jrochkind, line 6: We could give them the file type (PDF, MP3, FLAC), and the file size here too. Would that be nice of us? Or is this really just a placeholder until we do more review, and we don't care about polishing it more now?

config/environments/production.rb
```diff
@@ -78,6 +78,9 @@
   config.action_mailer.perform_caching = false
   config.action_mailer.raise_delivery_errors = true
 
+  # For images in emails:
+  config.action_mailer.asset_host = ScihistDigicoll::Env.app_url_base_parsed.to_s
```
jrochkind, line 82: nice.

app/mailers/oral_history_delivery_mailer.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def to_address
+    request.patron_email
+  end
+
+  # Note: any value greater than 604800 will raise an exception.
+  # See https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Presigner.html
+  def how_long_urls_will_be_valid
+    1.week.to_i - 10
+  end
```
jrochkind, line 21: How about calling this method `asset_expiration_time` instead? `how_long_` rubs me funny as a name. 

And I'm suspicious of that `-10`?  If it raises an error without the `- 10`, that might be a bug in my FasterS3Urls gem, or something else, that I want to get to the bottom of.

app/mailers/oral_history_delivery_mailer.rb
```diff
@@ -0,0 +1,70 @@
...
+  def patron_name
+    request.patron_name
+  end
+
+  def work
+    request.work
+  end
+
+  def assets
+    work.members.order(:position).select {|x| x.is_a? Asset}
```
jrochkind, line 37: we shouldn't be sending all assets, but only those marked for restricted delivery by request. All assets might include for instance the public "front matter" PDF.  i forget where in the data the mark is for "available by request", but it's somewhere or other. 

There should be a test that verifies this too. 

Nice job catching `order(:position)`, I actually came here to confirm that.

app/mailers/oral_history_delivery_mailer.rb
```diff
@@ -0,0 +1,70 @@
...
+  def subject
+    "Science History Institute: files from #{work.title}"
+  end
+
+  def hostname
+    ScihistDigicoll::Env.lookup!(:app_url_base)
+  end
+
+  def work_url
+    @work_url ||= "#{hostname}#{Rails.application.routes.url_helpers.work_path(work.friendlier_id)}"
```
jrochkind, line 64: Are you sure you have to do `Rails.application.routes.url_helpers`?  i think just `work_path` is supposed to work in ActionMailer. 

And instead of concatenating the host manually you can do:

    work_path(work.friendlier_id, host: hostname)

But even better, we can configure the default hostname globally for ActionMailer, and then just do `work_path(work.friendlier_id)`. https://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views

app/mailers/oral_history_delivery_mailer.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def work_link
+    "<a href=\"#{work_url}\">#{work.title}</a>".html_safe
+  end
```
jrochkind, line 69: Just put this in the .erb, instead of having a `work_link` method in the Mailer. Then you can use `link_to work.title, work.url`, no need to write out the HTML and then `html_safe` it. 

You should actually never call `html_safe` like this, it's not safe. What if work.title had a `<` or `>` mark in it? you'd be generating invalid HTML, and then falsely claiming it was html safe. It is only safe to call `html_safe`, if you've actually escaped things yourself, for instance you could do `ERB::Util.html_escape(work.title)`. but if you just use Rails helpers like `link_to`, it takes care of it. Which will be convenient if you put the code in the .html.erb file.

app/mailers/oral_history_delivery_mailer.rb
```diff
@@ -0,0 +1,70 @@
...
+        public: false,
+        expires_in: how_long_urls_will_be_valid,
+        response_content_type: asset.content_type,
+        response_content_disposition: ContentDisposition.format(
+          disposition: "inline",
+          filename: DownloadFilenameHelper.filename_for_asset(asset)
+        )
+      )
+      { text: asset.title, url: url }
+    end
```
jrochkind, line 52: generating a list of hashes that then we need to coordinate what to expect in the hash is weird. How about you just have a method `download_url(asset)`, and then the actual .html.erb can just iterate through the assets itself, call `asset.ttitle` itself, and call `mailer.download_url(asset)` when needed.

app/views/oral_history_delivery_mailer/oral_history_delivery_email.html.erb
eddierubeiz: It certainly did occur to me. It's a quick change; I think I'll just add it in.

app/mailers/oral_history_delivery_mailer.rb
eddierubeiz: Noted and fixed!

app/mailers/oral_history_delivery_mailer.rb
eddierubeiz: No you're not supposed to be able to use *_path helpers inside of an email (hence this clumsy workaround), but you can use `work_url`. I had to add a bit of config to `config/environments/test.rb` so it would work in the tests but the default hostname is already in the other two config environments, so we're good to go.

app/mailers/oral_history_delivery_mailer.rb
eddierubeiz: Good call. Fixed!

app/mailers/oral_history_delivery_mailer.rb
eddierubeiz: Fixed that. Makes sense.

app/mailers/oral_history_delivery_mailer.rb
eddierubeiz: Fixed method name.

 I do *not* believe your `FasterS3Urls` has a bug.
I did some basic testing in the staging console that indicated this should work.
https://github.com/sciencehistory/scihist_digicoll/pull/901#issuecomment-714528948

eddierubeiz: Got rid of the - 10. Works fine.

PR #903, opened 2020-10-23
New default "from" email address

eddierubeiz:
Suggested by jrochkind; I'm popping this feature out into its own PR so we can deal with it separately.

This PR has any and all email addresses specified in env.rb, and one of them (the one we want, `digital_collections_email_address`) is set as the default in the application_mailer.rb file.

_Furthermore_, even if we were to *not* set the default in env.rb, `application_mailer.rb` would still default to "from@example.com", so I feel we can simplify the three (soon to be 4) outgoing mailers' code. They no longer specify the "from" value at all, and thus default to what's set in `application_mailer.rb`.

### DIFF
--- a/app/lib/scihist_digicoll/env.rb
+++ b/app/lib/scihist_digicoll/env.rb
@@ -383,6 +383,8 @@ module ScihistDigicoll
 
     # From address:
     define_key :no_reply_email_address, default: "no-reply@sciencehistory.org"
+    define_key :digital_collections_email_address, default: "digital@sciencehistory.org"
+
     # To addresses (these are email lists maintained by IT.)
     define_key :digital_tech_email_address, default: "digital-tech@sciencehistory.org"
     define_key :digital_email_address, default: "digital@sciencehistory.org"
--- a/app/lib/scihist_digicoll/env.rb
+++ b/app/lib/scihist_digicoll/env.rb
@@ -383,7 +383,6 @@ module ScihistDigicoll
 
     # From address:
     define_key :no_reply_email_address, default: "no-reply@sciencehistory.org"
-    define_key :digital_collections_email_address, default: "digital@sciencehistory.org"
 
     # To addresses (these are email lists maintained by IT.)
     define_key :digital_tech_email_address, default: "digital-tech@sciencehistory.org"
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,10 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
   default from: ScihistDigicoll::Env.lookup!(:digital_email_address)
   layout 'mailer'
-
-
-  def hostname
-    ScihistDigicoll::Env.lookup!(:app_url_base)
-  end
-
 end
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: (ScihistDigicoll::Env.lookup(:no_reply_email_address) || "from@example.com")
+  default from: (ScihistDigicoll::Env.lookup(:digital_collections_email_address) || "from@example.com")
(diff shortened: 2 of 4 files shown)


app/lib/scihist_digicoll/env.rb
```diff
@@ -383,6 +383,8 @@ def self.dzi_backup_bucket
 
     # From address:
     define_key :no_reply_email_address, default: "no-reply@sciencehistory.org"
+    define_key :digital_collections_email_address, default: "digital@sciencehistory.org"
```
jrochkind, line 386: Hi, is there an intended difference in meaning between this key and `:digital_email_address` below, pointing to same email address? Or did you just not notice that one?

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: (ScihistDigicoll::Env.lookup(:no_reply_email_address) || "from@example.com")
+  default from: (ScihistDigicoll::Env.lookup(:digital_collections_email_address) || "from@example.com")
```
jrochkind, line 2: As long as we're tweaking this stuff, lets get rid of the default `|| from@example.com` (we never want to send from there in production, if we accidentally did we might get penalized on spam lists and such?), and maybe change to `lookup!` which will raise  if no value is supplied.  Raise on app boot, since that's when this config code is executed.

jrochkind:
Good idea to separate in it's own PR, and nice work.

app/mailers/application_mailer.rb
eddierubeiz: OK, great.

app/lib/scihist_digicoll/env.rb
eddierubeiz: Didn't notice it! Thanks for catching this.

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,10 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: (ScihistDigicoll::Env.lookup(:no_reply_email_address) || "from@example.com")
+  default from: ScihistDigicoll::Env.lookup!(:digital_email_address)
   layout 'mailer'
+
+
+  def hostname
+    ScihistDigicoll::Env.lookup!(:app_url_base)
+  end
```
jrochkind, line 8: I'm confused, do we actually NEED this in more than one mailer? Why'd you add it to the new DerivativeStorageTypeAuditMailer in the first place? (To then immediately remove it). 

It's a bit weird to need this thing like this; the fixity check mailer did, for it's own baroque reasons, which we don't need to change now. But I want to make sure the new one really does before adding it there or raising this to the superclass to be used by everything and kind of encouraging doing things this way.

jrochkind: Or you only use it for `Digital Collections Warning: #{hostname}`.  Yeah, maybe just use the :service_level Env key?  (which will be like "staging" or "production" usually). Or, really, `ScihistDigicoll::Env.lookup!(:app_url_base)` is simple enough that I don't think it needs to be encapsulate din a new method?

`ScihistDigicoll::Env.lookup!(:app_url_base)` is ALSO not actually going to return a "hostname", it's going to return something like `https://hostname`.  Calling it hostname is a bit misleading too. Don't necessarily need to fix a misleading name that already existed in code, but I get more cautious when we're boosting htis method to be in the superclass ApplicationMailer, and without any comments too. I don't think it really needs to be there.

eddierubeiz: Eh, I just saw two identical methods in two subclasses and automatically thought -- eh, let's move it to the superclass. Don't feel strongly about it; moved the method back down; it's not really part of the scope of this PR and it's not necessary, as you say.

jrochkind: Cool, I'm not sure either one really needs it, but it's not this PR that's adding it, so I guess it's not this PR's problem! Thanks.

Issue #656, opened 2020-03-17
Add lograge to application


sanfordd:
Add the lograge gem: https://github.com/roidrage/lograge
This is to reduce the noise of our logs to something more manageable for technical staff to review logs.

jrochkind:
also configure to include user-agent in action logs

jrochkind:
An alternative to lograge is semantic_logger:

https://github.com/rocketjob/rails_semantic_logger

eddierubeiz:
[Semantic logger docs]( https://logger.rocketjob.io/) looks awesome.

jrochkind:
* lograge
  * benefits:
     * can optionally log in a JSON format, which additional log aggregating tools increasingly can understand as structured data 
     * looks pretty customizable
     * Rails-focused
     * seems popular/maintained. Other third party log aggregator services sometimes mention/suggest lograge. 
  * possible disadvantages:
      * in default situation, it doesn't actually even log the complete URL requested, only the path component not query component! That seems weird to me. Also may leave out other things we need. May need substantial config to work for us. 
       * they seem to not want you to log errors, unclear how errors show up in logs, may be hard to associate error in logs with corresponding request

* semantic_logger
  * benefits:
     * also very customizable
  * possible disadvantages:
     * not Rails-specific, which is good if you need that (we don't), but possibly makes it more complicated/confusing to set up for Rails specifically, since it's not Rails-focused. 
     * seems to be less popular
     * also may need custom setup to make sure it's logging everything we want
     * like lograge, anything that collapses each request to one line may make it hard to correlate errors to request they correspond with -- usually you have a logged 'start of request' and 'end of request' which the error in between but with only one logged line, you need to associate error (or any other warn/error log messages) with a subsequent logged request line I guess?

jrochkind:
We've done some other things to reduce logging somewhat, although not as much as lograge. See #906. For now, we don't necessarily want to do lograge, let's close this.

PR #906, opened 2020-10-28
Reduce view render logging we consider excessive, in production

jrochkind:
As I try to monitor the app in heroku, I am finding current logs pretty hard to deal with due to excessive verbosity. A more drastic change to logging could be taken ala #656, but without wanting to take time to figure that out and configure it, for now we can just turn off view/render logging (which is also part of what lograge or semantic_logs would do), which will get us a lot of the way there to less awful logs, or at least we can see how far it gets us.


eddierubeiz:
I'm totally fine with this. Merging.

PR #909, opened 2020-11-02
Suppress actionview logging *without* triggering exceptions!

jrochkind:
New try to #906

Much hackier than the elegance of the previous solution, but this one doesn't break the app. Found this hacky solution at 
https://stackoverflow.com/questions/12984984/how-to-prevent-rails-from-action-view-logging-in-production/61893582


jrochkind:
I'm going to go ahead and merge this one, just to get some sane logging to continue with my heroku investigations.

PR #910, opened 2020-11-02
WIP Resque on heroku

eddierubeiz:
Some changes I made to the code to get resque working on Heroku, based on 

https://devcenter.heroku.com/articles/ruby-resque-pool.

This isn't yet guaranteed not to break our existing setup, nor to be the minimum number of changes to make it possible to run background jobs on Heroku.

Closing this branch -- its content is broken into 3 smaller ones.


Rakefile
```diff
@@ -4,3 +4,15 @@
...
+
+# this task will get called before resque:pool:setup
+# and preload the rails environment in the pool manager
+task "resque:pool:setup" do
+  # close any sockets or files in pool manager
+  ActiveRecord::Base.connection.disconnect!
+  Resque::Pool.after_prefork do
+    ActiveRecord::Base.establish_connection
+  end
+end
```
jrochkind, line 18: I don't totally understand this -- can you document where you got this from? Can we be confident this won't cause any problems on our existing infrastructure, where things seem to be working fine without it?  I wonder why we didn't need it before.

config/resque-pool.yml
```diff
@@ -1,3 +1,4 @@
+'*': <%= WORKER_CONCURRENCY %>
```
jrochkind, line 1: Hm, we are changing our configuration for what workers we run?  Previously we had our configuration below, under `production`, we had various workers running on various queues. 

Now we also have this "*" key for all environments, saying to run default 5 workers that work on ALL queues. But we may ALSO be running those additional workers under `*`?

I don't think this is quite what we want to do. 

1. We should keep all our config under the `production` key
2. This change would change how many workers are running in our current production and staging environments if merged, we don't want to do that. We need a PR that allows us to keep things runnig as they are on current environment, but also if necessary allows us to set configuration to run them differently on heroku. 
2. Ideally, we could keep running the same configuration we have -- if you want to try different counts of workers, note you can already do that with no code change, just set heroku config vars `REGULAR_JOB_WORKER_COUNT`, `ON_DEMAND_JOB_WORKER_COUNT`, and `SPECIAL_JOB_WORKER_COUNT`.  Because our `ScihistDigicoll::Env` already uses the ENV variables that heroku config sets. You can set any of these to 0 (the last one already defaults to 0). 
3. If you DO need, at least for initial experimentation, to have a setup where you have 5 workers working all queues on heroku, instead of the setup we do have -- add a new line for `*` _under_ `production` area, which uses a *new( `ScihistDigicoll::Env` key you create, perhaps `ALL_QUEUE_JOB_WORKER_COUNT`. We can make sure it's set to 0 on our current infrastructure, but 5 on heroku, and you can set all the others to 0 on heroku. 

So, we need a code change that via config lets us keep our worker configuration the same on our existing infrastructure, but allows us to experiment with different things on heroku. But if we aren't going to end up with the same config on heroku eventually, we need to think through the ramifications.

Rakefile
eddierubeiz: The rationale given at https://devcenter.heroku.com/articles/ruby-resque-pool is:

"If you are using any connections in your application, such as connections to a database you need to be sure to disconnect and reconnect after the resque pool worker forks."

In other words, the rationale is to disconnect and reconnect to the database as part of the setup of the resque pool.

Just out of curiosity I'll try commenting it out and see if I get an error in Heroku, though.

config/resque-pool.yml
eddierubeiz: Yup -- will fix this. (This was me just copying and pasting without regard to the existing setup -- we definitely won't merge this as is.)

Rakefile
jrochkind: I think it's possible that is outdated info from heroku... I *think* it should have applied to our existing setup too the same, if it applied at all. But it is a bit confusing. resque is only barely supported software, which doesn't help. 

I'm not too worried about putting this in if we want to -- so long as we test it on staging first, since it will apply to our existing environment, AND leave a comment with the heroku docs URL that told us to do it, so someone can check later ot understand why, and see if heroku recs have changed at a later date.

eddierubeiz: Looks like equivalent code is already present in `lib/tasks/resque.rake`; if I understand correctly, there's no need for this after all.

jrochkind: Aha, good catch! Yep, we ahd already done what heroku suggested, just in a different file, ok!

config/resque-pool.yml
eddierubeiz: Absolutely no need for this line; option 3 above works perfectly.

jrochkind: Cool. I think it's possible you won't be able to fit more than TWO workers total in a standard-2x, once they get busy... but that's what we'll find out from testing! I can show you various ways to look at RAM usage on heroku that I have discovered, to figure out how much RAM my stuff is using, if I'm running out, etc.

jrochkind:
still says WIP so i've been ignoring it, but maybe it's actually ready for review/merge? Please rebase/merge master, so it'll run the new CI.

jrochkind:
@eddierubeiz you seem to be doing additional work on this branch that is not to me related to "Resque on heroku". Eg "wrapping work indexing in a boolean env variable". 

Is that intentional?

Would it make sense to split this into more smaller PR's? Like if there are a few changes needed to make resque work at all on heroku, that's different than additional features we are adding?

Or maybe not, it all goes together/is interrelated, or is small enough even when it's all together? Just asking the question, you can answer it how you like; I was seeing things I wasn't expecting, but I may have been misunderstanding what was going on.

jrochkind:
Especially when it seems to be failing tests though... it might make sense to get resque *working* on heroku in a PR that passes tests?

I also recommend not leaving tests failing for too long in the PR, it tends to work a lot smootherif you can keep tests passing with most commits.

PR #913, opened 2020-11-10
Switch from travis to github actions for CI

jrochkind:


jrochkind:
I'm going and marking the Travis build as NOT required for a PR merge, because we won't be running them anymore. 

I will try to mark the Github Actions CI build as required for a merge; I think it should work, although there is some weirdness in how this works in GH Actions that may require us to give up on that, but for now we'll try it. 

@eddierubeiz I think I'm ready to actually merge this! There isn't too much to review, just seeing where the file is that defines the GH Action build (instead of .travis.yml), and some sense of what it looks like (took me a couple days of research to figure out how I wanted to construct it). But do you want to take a look?

jrochkind:
Hm, I edited settings to `Travis CI - Pull Request` is not required for merge anymore, but it may not update here unless we make another commit. 

I think we can just go ahead and merge pull request overriding the warning that the thing we don't want to require anyway hasn't happened.

PR #918, opened 2020-11-21
Allow config variable to bypass automatic SOLR indexing on callbacks

eddierubeiz:
For testing Heroku background jobs even without SOLR present.


jrochkind:
This is a simple way to do it, nice. 

Config variable is `solr_indexing` in local_env.yml, or `SOLR_INDEXING` ENV on heroku. If set to "false", the app won't write to solr index on updates of works/collections at all, which means it won't cause an error if the solr is missing. 

I'm just going to add a comment above the config variable in `env.rb`

PR #919, opened 2020-11-21
Add an Aptfile specifying packages for apt to install on heroku build.

eddierubeiz:
To add the apt buildpack that will actually read the file and install the packages, you need to run:

    heroku buildpacks:add --index 1 heroku-community/apt

More docs are at:
https://elements.heroku.com/buildpacks/heroku/heroku-buildpack-apt

Also need to add this buildpack to get ffmpeg (installing it via apt on heroku does not work)

    heroku buildpacks:add -i 1 https://github.com/heroku/heroku-buildpack-activestorage-preview

https://github.com/heroku/heroku-buildpack-activestorage-preview


jrochkind:
I was expecting ffmpeg, libvips, and mediainfo. 

I was not expecting ghostscript (which is a PDF library); have you tried without it? I bet it's pulled in automatically as a libvips dependency if needed. 

I was also expecting imagemagic, which is not here -- maybe that ends up being also pulled in automatically as a libvips depdendency?  Since I think we also sometimes use it directly, I think I like adding imagemagick as an explicit dependency maybe, I dunno. 

This is still marked draft, intentional?

jrochkind:
Just for the heck of it and cause I wanted to see it with my own eyes, I pushed this branch to heroku scihist-digicol app:

    git push heroku apt_buildpack:master

Then I opened up a bash terminal:

    $ heroku run bash

(Huh, that does take a while to start. I wonder if the extra apt-gets we're asking it to do slow it down? Well, we need em either way). 

And then I have a bash prompt on heroku, and verified that the command line utilities expected are there:

```
~ $ mediainfo --version
MediaInfo Command line,
MediaInfoLib - v17.12

~ $ vips --version
vips-8.4.5-Wed Jan 18 12:36:44 UTC 2017

~ $ ffmpeg --version
ffmpeg: error while loading shared libraries: libvpx.so.5: cannot open shared object file: No such file or directory
```

Oops, ffmpeg not actually working. 

Also imagemagick isn't actually there; let's try adding imagemagick (I think we might end up using it for PDFs even if we aren't now), taking out ghostscript (I think we don't need it), and we gotta troubleshoot ffmpeg. 

You wanna work on that Eddie?

jrochkind:
(We could try filing a ticket for heroku on it, see what happens)

jrochkind:
Googling, it looks like other people have trouble getting ffmpeg working on heroku... one answer is using this buildpack instead of asking for it to be apt-installed:

https://github.com/heroku/heroku-buildpack-activestorage-preview

They created that buildpack for the ActiveStorage (a part of Rails) "preview" feature, which also needs ffmpeg, but the important thing is it gives you ffmpeg.

PR #925, opened 2020-11-30
rails auto-scale heroku add-on needs a gem

jrochkind:


jrochkind:
Should be safe to have this in the app even when it gets deployed to current infrastructure. 

>> 1. What does the rails_autoscale gem do in the app in cases where it is not deployed to heroku at all?
>
> It looks for the RAILS_AUTOSCALE_URL env var, which is only present when the add-on is installed. If not present, it does nothing.

Although if we didn't play to use heroku and rails autoscale we woudln't want to leave it in Gemfile forever. But for now, let's merge it so our app on heroku can use autoscaling.

jrochkind:
hold off on this for a second, i want to explore hirefire.com as an alternative.

jrochkind:
we're probably going to use hirefire.io instead, not going to use this gem.

PR #927, opened 2020-12-01
build PDFs from download_medium instead of download_large

jrochkind:
download_large is 2880px wide, download_medium is 1200px wide. Initial thought was to support high-res printing from PDF, but this leads to unworkably large PDF (ramelli is 1.8GB). It also may be contributing to excess RAM consumption involved in creating them. Let's try with download_medium instead.

After review, does reduce RAM use significantly on heroku, but not enough to fit ramelli. But still may be better for users anyway, stakeholders are considering.


jrochkind:
* 50-image work, qf85nc451. 99MB PDF to  20MB PDF

* 100-image work, fx719n43f. 171MB PDF to 35MB PDF

* 325-image work, 1831ck38c. 386MB PDF to 92MB PDF

* Ramelli, 694 items. 1.8GB PDF(!), to  325MB PDF

The smaller-image PDFs may be basically indistuishable for our desired use cases. If you zoom in far, you can see the loss of resolution, but that's not the intended use case. If you print to a standard inkjet/laser can you? Would need to test.

jrochkind:
@MDiMeo approved this -- she said she couldn't tell the difference printing on office printer. 

One ~500-page PDF we tested became ~200MB instead of ~1G. Who wants a 1G PDF? I think we were using too large images (and may still be, but it's better!). 

I'm going to go ahead and merge @eddierubeiz , since there really isn't much to review.

PR #930, opened 2020-12-03
Rewrite WorkPdfCreate to use less RAM

jrochkind:
By creating multiple PDF files with 30 images max, then joining them together with the unix command line pdfunite, which does so fairly efficiently

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
         - name: Install apt dependencies
           run: |
             sudo apt-get update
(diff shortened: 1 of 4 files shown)


jrochkind:
Has been tested on staging, and works -- pdfunite is there. 

Should not be merged/deployed on prod until pdfunite is installed there. 

I'm still working on seeing if I can get PDF Info Properties metadata set on the pdfunite'd file.

jrochkind:
We decided we could give up setting metadata on generated PDF, so this is ready to merge. 

Make sure production servers have `pdfunite` before merging.  As long as they do, this works fine on existing servers as well as heroku.

PR #937, opened 2020-12-14
Tools for syncing our local solr config files to a solr cloud instance

jrochkind:
Such as via SearchStax, although it will work for any solr cloud instance.

The SolrConfigsetUpdater class has some general purpose tools for working with "config sets" (what Solr Cloud calls a config directory packaged for use with a remote solr cloud instance) via API in a solr cloud instance. It has some good documentation. 

There are a couple new rake tasks that use SolrConfigsetUpdater to do common tasks we need. 

To set up a brand new collection in an empty solr, using our on-disk solr config, there's `scihist:solr_cloud:create_collection`. You might run it on a heroku worker, which should already have config to talk to the desired solr instance:  

    heroku run rake  scihist:solr_cloud:create_collection

Or you could run it locally if you set SOLR_CLOUD via ENV or local_env.yml, it will of course be using whatever solr config is on disk where you are running it. 

    SOLR_URL=https://user:pass@solrhost.searchstax.com/scihist_digicoll ./bin/rake scihist:solr_cloud:create_collection

There is also a task `sync_configset` that will sync the current on-disk solr config to an already existing solr cloud instance. In our ansible-managed EC2 infrastructure, we do something like this on every capistrano deploy, so this is a way to do it with the searchstax infrastructure. 

    heroku run rake scihist:solr_cloud:sync_configset
    SOLR_URL=https://user:pass@solrhost.searchstax.com/scihist_digicoll ./bin/rake scihist:solr_cloud:sync_configset

And in fact we edit the Procfile to use the heroku ["release phase"](https://devcenter.heroku.com/articles/release-phase) to have a solr config sync (as well as a db migrate!) run on every heroku release; our existing capistrano for ansible-managed EC2 release was doing both those things on release, so now we perpetuate that. 

Note also wiki at https://chemheritage.atlassian.net/wiki/spaces/HDC/pages/1372192769/SearchStax+Solr

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,6 +72,9 @@ gem 'font-awesome-rails', '~> 4.7'
 # intend to merge to master like this.
 gem 'kithe', ">= 2.0.0.beta", "< 3", git: "https://github.com/sciencehistory/kithe.git"
 
+# temporarily using master traject
+gem "traject", github: "traject/traject"
+
 gem 'simple_form', "~> 5.0"
 gem "cocoon"
 
--- a/Gemfile
+++ b/Gemfile
@@ -71,9 +71,7 @@ gem 'font-awesome-rails', '~> 4.7'
 # temporary kithe indexing branch, for scihist_digicoll indexing branch, do not
 # intend to merge to master like this.
 gem 'kithe', ">= 2.0.0.beta", "< 3", git: "https://github.com/sciencehistory/kithe.git"
-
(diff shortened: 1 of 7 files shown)


jrochkind:
Please review wiki at https://chemheritage.atlassian.net/wiki/spaces/HDC/pages/1372192769/SearchStax+Solr as well as docs in this PR in general as part of review, thanks!

Issue #934, opened 2020-12-09
Database backup from Heroku to S3.


eddierubeiz:
Currently our postgres database is backed up by a cronjob `scihist-digicoll-backup.sh` running on the production database server, to an S3 bucket; namely, `s3://chf-hydra-backup/PGSql/digcol_backup.sql`.

This S3 bucket is "versioned" and keeps 30 days worth of past versions.

The DB server is going away, and so are the cronjobs on that server. What's the simplest way to regularly copy the Heroku DB to a particular location on S3?

For now, let's have it go to: `s3://chf-hydra-backup/PGSql/test_heroku_backup.sql`

jrochkind:
It would need to be running *somewhere*. I guess one possibility is continuing to run an EC2 for that purpose, but that's not very attractive. 

Perhaps a heroku scheduled job? 

We also need to look at what backup facilities are built into heroku. They will make certain backups for us. Is that enough? Is there any facility already built into heroku to ask for postgres backups to an S3 location of your choice? Maybe not, but maybe there are some facilities there to make this easier?

https://devcenter.heroku.com/articles/heroku-postgres-data-safety-and-continuous-protection

https://devcenter.heroku.com/articles/heroku-postgres-backups

Hmm, wait

> Both pg_dump and pg_restore running on Heroku are available through Heroku Postgres. During a scheduled run of PGBackups or manual run of heroku pg:backups:capture, PGBackups will launch a dedicated dyno to take a dump of your database and upload it to S3. 

Not totally sure what this means, but it sounds promissing?

jrochkind:
@jrochkind and @eddierubeiz will demo doing the cronjob on github actions to compare and contrast.

PR #938, opened 2020-12-14
Upload a database dump s3

eddierubeiz:
Ref #934 .

This creates a new file on S3, opens a write stream to it, and directly dumps the database into the write stream; the dump is not written to a temp file.

The dump is a logical (not physical) backup of the database (i.e.: not coupled to a particular database version; slower to restore; more portable;  consistent with pre-Heroku backup workflow; encoded as a UTF-8 file rather than in binary.)

Note that Heroku also offers physical backups which are fast and useful for short-term DB maneuvers. This workflow is meant to supplement, not replace, those backups.

We create a write stream into a new s3 object, then `pg_dump` the database directly into the write stream. No temp file is saved to disk.


jrochkind:
Looking good!

* Is `'PGSql/digcol_backup.dump` the current location of the thing? 

* I am not totally sure what you mean by "Consider uploading an ASCII SQL file rather than a binary .dump file." -- what was our previous backup routine doing? If different from what this is doing, is there any way to get heroku to do the thing we used to do be doing? If not, what are the implications of doing a binary dump instead?

* If we want to use different AWS credentials for backup, so that we don't have the AWS credentials normally used by our app with access to backup location, which is I think what we want, then you need to use something other than  `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. Since those standard variable names are already in use for our 'main' credentials. There should be some way to pass other values in to the AWS SDK though.

lib/tasks/copy_database_to_s3.rake
```diff
@@ -0,0 +1,24 @@
...
+    Copy the latest copy of the database to a location on s3.
+
+    AWS_ACCESS_KEY_ID=joe AWS_SECRET_ACCESS_KEY=schmo bundle exec rake scihist:copy_database_to_s3
+    Note: this is a Postgres binary .dump file, not ASCII SQL. You can convert it to regular sql like this: pg_restore -f mydatabase.sql latest.dump
+  """
+
+  task :copy_database_to_s3 => :environment do
+    cmd = TTY::Command.new(output: Rails.logger)
+    get_database_url_args = ['heroku', 'pg:backups:url']
+    result =  cmd.run!(*get_database_url_args, binmode: true, only_output_on_error: true)
```
jrochkind, line 15: I don't think there's any benefit to abstract this as `get_database_url_args`, it is easier to read if you just put them right in there:

   cmd.run!("heroku", "pg:backups:url"....

jrochkind:
Also I think I'm still confused about what we were talking about. I thought you were saying that you could provide a presigned S3 URL as an arg to `heroku pg:backups:url` or some similar command, so we wouldn't need to use down/put, we could just tell heroku where we wanted it to write backups too. But no?

If no, then this does look like the way to do it, and indeed this way doesn't use presigned URLs anymore, because there is no purpose for it, cool.

lib/tasks/copy_database_to_s3.rake
jrochkind: I also don't think there's any reason this needs `binmode`?

eddierubeiz:
Yeah, I realized after we talked that there is in fact no way to get Heroku to upload its own db backup to an arbitrary presigned S3 URL; I misinterpreted the docs for another function and realized belatedly that I'd been barking up the wrong tree. There's indeed no need for a presigned url!

jrochkind:
> Yeah, I realized after we talked that there is in fact no way to get Heroku to upload 

Cool, the good news is that you still figured out a very simple way to do what's needed, great!

jrochkind:
> Check that there is a DB to download before proceeding; fail gracefully if there isn't.

I don't think there's any need for that. We probably want it to fail hard with an error if that happens, it's not expected. 

Ideally that error would get caught/reported by honeybadger. So that's actually a checkbox to add, verify that honeybadger is going to catch errors reported by heroku scheduler regularly scheduled rake tasks, or figure out how to get them to be.

Issue #942, opened 2021-01-05
"Profile incompatible with image" problem with thumbnail creation for certain TIFF files


eddierubeiz:
`vipsthumbnail` is reporting a "profile incompatible with image problem" and failing to generate a thumbnail for the TIFF at https://digital.sciencehistory.org/works/yj7tsjw .

The image looks normal enough. Output from `file` is:
`TIFF image data, little-endian, direntries=20, height=4512, bps=8, compression=none, PhotometricIntepretation=BlackIsZero, orientation=upper-left, width=5347`

Honeybadger report: https://app.honeybadger.io/projects/58989/faults/73028093

Note: we'll leave this issue open until a permanent solution is in place.

eddierubeiz:
Easy to reproduce in staging: ```vipsthumbnail test.tiff --eprofile /opt/scihist_digicoll/shared/bundle/ruby/2.6.0/bundler/gems/kithe-b4e4e18b61fa/lib/vendor/icc/sRGB2014.icc```

eddierubeiz:
@HKativa, this could be a colorspace problem. The files we're talking about appear to be grayscale, and it looks like our thumbnail creation recipe assumes sRGB. Could it be that that our scanning workflow normally results in sRGB files, but somehow these particular ones ended up being saved in grayscale instead?

HKativa:
@eddierubeiz Thanks for looking into this! The files are definitely grayscale - the originals are b&w negatives, which were digitized on the Epson scanner, not the overhead camera. You're right that the normal workflow results in sRGB files, but we do have occasion to save certain materials in grayscale, even if it's rare. What solution would you recommend - changing the creation recipe or the profile for the files themselves?

eddierubeiz:
The former, absolutely.

HKativa:
Excellent, sounds good to me. I wonder if the recipe was re-written at some point because I've definitely ingested digitized negatives in the past, albeit back in the Hydra/Samvera days.

eddierubeiz:
Yeah, we've messed around with thumbnail creation quite a bit -- balancing speed and memory use vs. nice-looking thumbnails.

[...]

PR #943, opened 2021-01-06
Temporary remove icc profile, so we can ingest grayscale files.

eddierubeiz:
Files newly ingested with this code will need to have their derivatives regenerated once we have a permanent fix.
Ref #942


jrochkind:
Hmm, I think there was a reason we used the ICC profile though. I am nervous that we are no longer doing it. 

> Files newly ingested with this code will need to have their derivatives regenerated once we have a permanent fix.

Is there an issue or anything to prioritize or remind us to make a "permanent fix"?

Would we have any way at all of identifying which "files were ingested with this code" once we did so?

The fix might be fine permanently; if it is not, however, It doesn't seem to me we are currently in a position to actually reverse it/recover from it.

eddierubeiz:
Good call. Created issue above describing my understanding of the situation.

eddierubeiz:
Note: https://libvips.github.io/libvips/API/current/Using-vipsthumbnail.md.html makes me wonder if we might be able to fix this by using `--eprofile srgb`.

jrochkind:
Issue is #946

PR #944, opened 2021-01-08
Add date range validation to range_limit

eddierubeiz:
Ref #945 

In regular operation, calls to `CatalogController#range_limit` are first processed by `CatalogController#index`, which checks the dates and fixes them, if necessary, before making a second request to `CatalogController#range_limit`.

However, if you know the URL, it's still possible to make a request directly to `CatalogController#range_limit` with bad dates. We have a very enthusiastic generator of such requests living at IP `13.66.139.128`.  `HTTP_USER_AGENT` is always `Mozilla/5.0 (compatible; bingbot/2.0; +http://www.bing.com/bingbot.htm)`. This is resulting in over a hundred 500 errors a day, logged dutifully by Honeybadger : see https://app.honeybadger.io/projects/58989/faults/56829566/ .

This adds a second validation step in `range_limit` and throws a 406 error instead of a 500. This means the parameters are checked twice, so it's not a particularly elegant solution.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -496,7 +496,7 @@ class CatalogController < ApplicationController
   def screen_params_for_range_limit
     if (params['range_end'].nil?) ||
       (params['range_start'].nil?) ||
-      (params['range_start'].to_i < params['range_end'].to_i)
+      (params['range_start'].to_i > params['range_end'].to_i)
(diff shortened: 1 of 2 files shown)


jrochkind:
Is there an existing GH issue that should be referenced here? (So we can go move it to review/closed when merged)

If not, no worries, not saying we need to make one necessarily.

PR #949, opened 2021-01-11
Rails 6.1

jrochkind:


jrochkind:
The one weird thing was that stuff with manual `require` of things in `lib`, that I had to move to a `before_configuration` block. It is weird and confusing!

We can talk about it more, but basically....

All of our code in "app" is auto-loaded by Rails. This is standard for Rails, and things just magically are available without having to write `require` statements as you would have to in "normal ruby". 

But using auto-loaded code (ie, code in `app`) in the *configuration/initialization* phase of rails startup is deprecated, and stops working in some cases in Rails 6.1. Including a case we ran into.  Why? Because auto-loaded code is also automatically *reloaded on changes* in development mode, and this is just kind of a contradiction with using it in the boot process (or can be), and Rails is like, no, don't do that. 

But we really want to use ScihistDigicoll::Env in the boot process, that's kind of what it's for! So we had to move it to `./lib` so it's not auto-loaded... but now how do we make it available?  Well, we should just be able to `require "scihist_digicoll/env"`, since `lib` is supposed to be on the load path.. but I had the dangest time getting that to work, I think the load path wasn't set up right yet at the points I was trying to put the `require` in.  

To have it work and be required early enough for everywhere in configuration we were trying to use it.... somehow I came up with using that `before_configuration` block, I can't even say exactly how I arrived at that. :(

jrochkind:
OK, I think everything is working with app upgraded to Rails 6.1!

I'm still nervous about deploying it. I don't know if we should do more testing on staging. But... all tests pass, it deploys to staging, I did some basic trivial manual tests... probably no reason to waste anyone's time with manual testing. Maybe just be intentional about when we deploy this, and warn people?

jrochkind:
Planning on deploying this Monday probably, going to leave it in 'draft' until then, but if you feel like reviewing and sticking an 'approval' on it or asking any questions, feel free @eddierubeiz .

PR #952, opened 2021-01-14
Re-adding icc profiles pursuant to vips upgrade

eddierubeiz:
This reverses the changes made in https://github.com/sciencehistory/scihist_digicoll/pull/943/. We now have VIPS 8.10.5 in staging and production, so the "bandaid" fix in 943 is no longer necessary.


jrochkind:
Nice, thanks. 

There probably would have been a way to get git to reverse the changes made before, by mentioning specific commits (possibly including a merge commit) and telling git to reverse them. That could help reduce chance of human error in doing the revert.  (Or maybe you did that and I just don't realize it?)

But that's just something for future reference. This is small and looks good on my review, merging. 

Do we care about regenerating any thumbs created in the interim? My gut is, nah, let's not worry about it. Not sure if we should check with stakeholders. Again my gut is, they aren't going to want to be bothered. We could try to remember to bring it up at next weekly check-in just in case.

Issue #875, opened 2020-10-06
investigate terraform with heroku


jrochkind:
For proof of concept, we are configuring all things in heroku manually. This includes:

* list of config variables (including some that are secret)
* add-ons and plans selected for add-ons
* buildpacks

If we want to switch to 'infrastructure as code' for documentation/history/reproducibility benefits, it looks like terraform with heroku is the way to go. 

https://devcenter.heroku.com/articles/using-terraform-with-heroku

https://medium.com/rackbrains/manage-heroku-infrastructure-with-terraform-4a167b850300

We could also use terraform to set up our S3 buckets, abandoning ansible.

jrochkind:
Managing secrets via terraform seems to be kind of a pain and require special consideration. https://blog.gruntwork.io/a-comprehensive-guide-to-managing-secrets-in-your-terraform-code-1d586955ace1

Would be worth asking around in the samvera and code4lib communities if anyone is using terraform and how they are managing secrets.

jrochkind:
Hey I didn't realize this feature also existed in heroku, another way to do a bit of infrastructure-as-code with heroku, using a less powerful/less complicated one-file format that might be specific to heroku. 

https://devcenter.heroku.com/articles/app-json-schema

jrochkind:
The app.json heroku thing seems to be *not* suitable for this. It's intended as a "quick start template" to make a new heroku app, but there's no way to use it to update an existing heroku app, so it unfortunately doesn't work well trying to evolve it as the app setup evolves. 

Another heroku option is switching to [heroku's docker support](https://devcenter.heroku.com/categories/deploying-with-docker). This is a different way to deploy heroku apps than we've been using. If I understand it right, it would (like terraform) mean giving up using the heroku manual UI's that we like so much. 

It would require us to learn more about docker, to invest a bit more in skill-building to use. 

What we'd get out of it, is having a more docker-centric setup that is similar to other non-heroku trendy options too, it would probably make it easier for us to migrate to them -- possibly using some of the specific docker-based setup, but also just having the docker skills.  

It also might make it easier to install things like specific versions of vips -- it does make it possible to have different things installed on eg `web` vs `worker` dynos, I think docker-based-deploy is the only way to do that on heroku. 

This is a good thing to know that exists -- but I don't think we want to commit to investing in it before a heroku migration. It may be something we want to investigate later though.

jrochkind:
[...]

PR #956, opened 2021-01-15
Remove vips from aptfile; we're using machinio buildpack instead

eddierubeiz:
Removing vips from the Aptfile and instead adding the machinio vips buildpack by running:
`heroku buildpacks:set https://github.com/machinio/heroku-buildpack-vips
`
results in vips 8.10.2 getting installed on Heroku.
This is not the latest version (we currently have 8.10.5 installed in production on s3 boxes), but it does fix the grayscale thumbnail problem described in ref #942.

### DIFF
--- a/Aptfile
+++ b/Aptfile
@@ -1,6 +1,13 @@
-libvips-tools
+# Removing libvips-tools
+# in favor of https://github.com/machinio/heroku-buildpack-vips
+# which gives us vips 8.10.2 .
+# We may want to switch back to libvips-tools
+# in the future and remove the buildpack.
+# libvips-tools
+
 mediainfo
 imagemagick
+
 # poppler-utls was probably already getting installed as a dependency
 # of above anyway, but we need `pdfunite` command line util from it, so we'll
 # list it explicitly.


eddierubeiz:
Note: Once the standard-issue version of apt-installed `vips` provided with the latest version of Ubuntu on Heroku reaches 8.10, we will want to reverse this PR by removing this buildpack and adding `libvips-tools` back into the `Aptfile`.

jrochkind:
nice okay that works. 

Yeah, the lack of "infrastructure as code" for the heroku setup is concerning me a little bit -- like the fact that we added this buildpack isn't checked into our git repo, or documented anywhere other than here. 

I'm not sure the right solution to that. Like making a wiki page where we just mirror our buildpack choices and then have to update it every time we change them (and probably are going to forget and it'll get out of sync) -- does not actually seem like adding any value.  You could just check the heroku buildpack list instead. (Can you do that from CLI too? Probably). 

Maybe we do need a doc where we just list the categories of configuration we HAVE in heroku. Like, mention "Customized buildpacks, to see them go to this part of UI, or execute this CLI command". With some other categories too. 

Just musing on this ticket, maybe the real place to talk about this would be on #875, which is really not just about about terraform (cause I'm leaning against it), but "how do we record our heroku configuration such that we have a handle on it".

jrochkind:
The way you do the vips buildpack... is it pinned to a specific version of vips, or is it always installing the latest it's got, or what?

I'd like us to document that here on the PR, how the vips versioning is going to work with the vips buildpack.

jrochkind:
@eddierubeiz also, can we document here -- did you try using the Aptfile approach with the newer version of ubuntu? That did not work? Cause if it did, I'd prefer that.

eddierubeiz:
The machinio VIPS buildpack has compiled vips 8.10.2 [since November 5th, 2020.](https://github.com/machinio/heroku-buildpack-vips/commit/0b023442317760657967e0b2fabacbade6f15050#diff-cc3b3db023efc81750e416e31d7a4b949db47a02683968e53761662b6138bced)

It's not pinned to the latest version of vips; the version changes whenever the buildpack maintainers decide to change the version of vips in the `Dockerfile` in the repo.

If they decide to update the version in a way that causes a bug, we could pin the version by requesting a particular commit of the Git repo in the list of buildpacks (yes, turns out you can do that), or fork it and create our own buildpack.

eddierubeiz:
Note 2: If you just use the default Aptfile version of vips that comes with the latest Heroku version of Ubuntu (20), you get vips 8.9 -- a version with the bug we're trying to fix.

eddierubeiz:
I did try a couple other buildpacks, including https://github.com/brandoncc/heroku-buildpack-vips, without success.

eddierubeiz:
I also tried requesting a particular .deb file, namely http://ftp.us.debian.org/debian/pool/main/v/vips libvips-tools_8.10.5-1_amd64.deb , directly in the `Aptfile`. On invoking libs, I got the following:

`LoadError: Could not open library 'vips.so.42': vips.so.42: cannot open shared object file: No such file or directory.`

jrochkind:
> Note 2: If you just use the default Aptfile version of vips that comes with the latest Heroku version of Ubuntu (20), you get vips 8.9 -- a version with the bug we're trying to fix.

It *looks* like vips 8.10 should be in ubuntu repos... but I don't totally understand how this stuff works. 

https://launchpad.net/ubuntu/+source/vips

Still interested in trying to pursue this with the maintainers of the heroku Aptfile buildpack, to ask -- should we be able to get vips 8.10 this way? Is there a different way to do it?

PR #963, opened 2021-01-26
handle existing configset

jrochkind:
In case where the digest configset name already exists, but is not actually configured to be used, we were getting an error. Not sure how this can happen with our intended logic, but it did, might as well guard against it.

Rather than do an extra Solr API call (to check if configset name used) 100% of the time for something that we expect only in edge cases, we'll just rescue the exception. Although that does require matching on solr error message.


eddierubeiz:
Thanks -- I was bitten by this error actually.

PR #965, opened 2021-02-01
Correct item count on Featured Topic page.

### LINKED ISSUE #279: (Post launch) Item count on featured topic page is wrong.
Post launch: the number of "items" listed on the top of the featured topic show page (e.g. https://kithe.sciencehistory.org/focus/alchemy ) is not really the total number of items in the featured topic. Instead, it's just the number of items returned by the latest Solr search. 

This is a known bug we inherited from Sufia. One simple fix would involve just hitting Solr more than once per page load: once for the total count of items in the featured topic and once for the actual faceted search results.

See the corresponding functionality on the Collection show page for the ideal behavior: a count of public items, and a second count of all items, only shown to staff users.

jrochkind:
Closes #279

Have to reverse engineer Blacklight a bit to get at the parts that will let us query solr in a way that is restricted to just this 'featured topic' using the mechanisms we've already set up, but without any current query or other additional limits. A bit difficult to get at these parts, and BL tends to change them frequently, so it's possible the code will break in future versions. But I think it's something not too crazy, I think it's the best we can do.

### DIFF
--- a/app/controllers/featured_topic_controller.rb
+++ b/app/controllers/featured_topic_controller.rb
@@ -42,7 +42,23 @@ class FeaturedTopicController < CatalogController
   end
 
   def total_count
-    @response.total_count
+    @total_count ||= begin
+      # We need to fight with Blacklight a bit to get the results which have been
+      # limited by our "slug", but EXCLUDING any current search query or facet limits.
+      #
+      # This is based on reverse-engineering of the very abstracted stuff BL is
+      # doing, and is probably kind of fragile, sorry. :(
+
+      empty_search_state = search_state_class.new(params.slice(:controller, :action, :slug), blacklight_config, self)
+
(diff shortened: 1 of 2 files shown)


jrochkind:
This does mean doing an extra Solr request even when we don't have to -- when you are just looking at collection home page without a user-entered search, which is probably the majority of use -- the old logic (without an extra solr request) worked just fine. 

So we could try to recognize that case and avoid the extra solr request.... but again hard to do so using Blacklight API that isn't changing/deprecated in future versions. And i think the Solr query is probably fast enough that it doesn't really matter. 

I don't know if you have any thoughts @eddierubeiz

eddierubeiz:
I think it's totally fine to make the extra SOLR request. We make two SOLR requests for one page load in at least one other situation that's come up lately, and yeah, I doubt it'll slow anything down since yeah, SOLR is fast, and in this case all you're fetching is an integer anyway.

PR #966, opened 2021-02-02
Display warning to those using IE

### LINKED ISSUE #642: Detect and warn IE users that we don't support IE and it may or may not work?

jrochkind:
Closes #642 

The first implementation here is server-side, the logic checks user-agent server-side and then includes an extra element at top of page. 

One problem with this is that it makes the page no longer globally cacheable, since different people get differnet pages. If in the future we wanted to cache at a CDN for instance. To do that, all users need to get the same page delivered from server. JS can customize once it's in the browser though. 

We could try to replace this with a purely client-side solution. Does it matter?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -103,6 +103,7 @@ gem 'html_aware_truncation', '~> 1.0'
 gem "prawn", "~> 2.2" # creating PDFs
 gem "pdf-reader", "~> 2.2" # simple metadata extraction from pdfs
 gem 'rubyzip', '~> 2.0'
+gem 'browser', '~> 5.0' # browser user-agent detection, maybe only for IE-unsupported warning.
 
 
 # Until oai 1.0 is released...
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -129,6 +129,7 @@ GEM
       ruby-box
       signet (~> 0.8)
       typhoeus
+    browser (5.3.0)
     builder (3.2.4)
     byebug (11.1.3)
     capistrano (3.15.0)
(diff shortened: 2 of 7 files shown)


jrochkind:
The first implementation puts a banner at top of screen above our navbar, looks like this:

Screen Shot 2021-02-02 at 12 07 30 PM

Is that about what was expected, @MDiMeo or anyone else with an opinion?

If you click the "x" close icon on the right, the alert will disappear, and will *not* show up again for you (unless you delete all your cookies; or there's a bug in our code). You are dismissing it forever, so you can continue to use with IE in peace if you want.

jrochkind:
Oh hey, wait, so... currently none of our Javascript is loading on IE, it's broken on IE. That means:

* We can't actually make the alert "dismissable" since that relies on JS. So it'll just keep showing up at the top of every page if our software thinks you are using IE, no way to dismiss it. 
* We also can't do a client-side solution! So the server-side solution is the only one possible, the one that necessarily breaks CDN-cacheability. 

I wonder if we want to reconsider the idea of displaying this warning, again. Microsoft itself no longer supports IE in Microsoft Teams, and by August will no longer support it in any Office 365 app.

CHF-IT-Staff:
That looks good to me


Chuck Hornstein   (he, him, his)
Director, Information Technology
t. +1.215.873.8272
f. +1.215.629.5272

Science History Institute

jrochkind:
user metageeky on code4lib slack suggested one way to still be globally cacheable is to use CSS to hide the bar for all but IE. 

I can maybe look into that, not sure if there's a way to do it that gets all IE not just old versions of IE. Some brief looking around... seems very hacky and unreliable, but one option. 

One thing I worry about is if putting this in the codebase is a future gotcha... someone later decides to use a CDN, and some pages wind up in the CDN with IE warnings, distributed to everyone. I guess that wouldn't be the end of the world, if someone eventually noticed they'd debug and fix it i guess.

Issue #433, opened 2019-10-23
Consider dealing with accidentally committed AWS keys


jrochkind:
We've now accidentally committed AWS keys to git and pushed to public github twice, once each by different developers. 

While github/AWS's automated alert warns us we have done so asap, AWS requires us to change _all_ of our keys when this happens. (Not sure why, it doens't seem necessary, it's annoying). Which is a fairly annoying, disruptive, time-consuming thing to do. 

It will probably happen again, mistakes seem inevitable. 

We can think of things to do to either:

* a) make it less likely to happen (one example, git pre-commit hook to keep you from doing it), or
* b) make it easier/more painless/automated to rotate all keys when we need to

We'd need to brainstorm possibiltiies, decide which to pursue, and then pursue them.

sanfordd:
It's just an AWS policy, in case a leaked key had permissions that were able to escalate to other areas. We have been lucky, and made our luck, in that most of our keys have very narrow permissions. Most of them aren't a huge risk when they leak as long as we catch it quickly.

I'd note we can also talk about using AWS Roles, though those would require a larger change than simply adding a pre-commit hook or automating some of the rotation steps.

jrochkind:
Here is Amazon's support suggestions, I think some of them are not applicable by some may be. 

---

The most common source of compromise is accidental access key exposure through GitHub. To help prevent that AWS has developed a tool to scrape sensitive info from your code before you commit, called git-secrets. The installation files and documentation are available here:  https://github.com/awslabs/git-secrets
  
If you haven't already, you should enable MFA on your account, and for all IAM users, to virtually eliminate the potential for a password-based compromise. MFA is available for API-Access as well which will dramatically reduce the probability of your IAM Access keys being compromised.  

MFA for Users: http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_mfa_enable.html  
MFA for API: http://docs.aws.amazon.com/IAM/latest/UserGuide/MFAProtectedAPI.html
  
Finally, when using IAM you can restrict access to AWS services, regions, and resources so that users are only allowed to carry out their intended tasks. This way, even if the associated Access Key is compromised, the damage that can be done is limited. If you have any IAM users or keys that are not being used, we recommend that you delete them to reduce the total number of ways your account can be accessed.
Here are a few resources I recommend to help you utilize Access Keys and IAM users effectively:
  
[...]

PR #970, opened 2021-02-04
git pre-commit hook included in repo to try and catch AWS keys before commit

jrochkind:
To use, you need to set a git config value in your local copy, to tell git to use hooks that came with the repo. Or you could copy it or whatever.

Anyway, this won't be foolproof, but sometimes helps.

Kind of wrote my own mashing together a few others, in comments.

Ref #433


jrochkind:
One possible downside is we'll totally forget we did this/how we did this, and then if it's not working well (say too many false positives) forget how to find it to disable it haha. But hopefully the line in the README will help.

Issue #973, opened 2021-02-08
Make CC BY-NC-ND 4.0 license an option in the Rights Metadata field


MDiMeo:
In the Rights Metadata field, creation an option that says "Creative Commons Attribution-NonCommercial-NoDerivs". This should be hyperlinked to the license deed: https://creativecommons.org/licenses/by-nc-nd/4.0/

jrochkind:
@MDiMeo  Just double-checking, are certain we should Rights, instead of making a separate field for CC license?

I would have expected this to be be "rights: In Copyright" with a CC-by license.

But, I guess, a CC license does kind of imply InCopyright, that's the only way you can offer a CC license in the first place, so I guess maybe it does make sense as one field after all. Is that what other cultural heritage orgs do, or what PA Digital/DPLA recommends, they want the CC license in the "rights" field?

OK! Just double-checking. 

(And I woudln't necessarily have expected "no derivatives" as a general rule for our CC's, although maybe on OH's. But in general, we're happy when people do things like make the music video derivative, no? Should we add some other CC's while we're at it, or not yet?)

jrochkind:
Actually this will take a bit of work on *display* end, as right now we have things laid out to render the "rightsstatement.org" style logos, and CC is a different style of logo, we'll have to make sure the rendering renders it properly or uses a graphic or what have you. 

but it can be done.

jrochkind:
We'll have to decide how we want the CC license to *look* to the public on the display end too. If you have any thoughts or suggestions, or we'll just come up with something. 

The existing rights statements use a template like this, based on how rightsstatements.org shows them:

![Screen Shot 2021-02-08 at 5 33 28 PM](https://user-images.githubusercontent.com/149304/107289985-d38c5d80-6a33-11eb-9e34-7efd7c6cff67.png)
![Screen Shot 2021-02-08 at 5 33 07 PM](https://user-images.githubusercontent.com/149304/107289986-d38c5d80-6a33-11eb-966e-b4a577c144a0.png)

But CC does not use that same template. They seem to have ready-made icons. We could use those icons? With words also?

MDiMeo:
We were following the example of what we did already for Public Domain, which is technically following the CC rules and offering the Public Doman License instead of the No Copyright rights statement. I haven't done a ton of digging to see what our peers are doing in this regard -- just following what we did with Public Domain.

The logo from CC is ugly but there is on one the license:
[...]

PR #972, opened 2021-02-08
Oral History delivery with approval

jrochkind:

PR #974, opened 2021-02-10
Rights terms can include CC licenses, and display reasonably

jrochkind:
Ref #973

CC licenses displayed in "rights" spot need to display somewhat differently. We came up with a wway that seems reasonable. Refactored component to handle them right. Some slight accessibility improvements (alt tag) for existing stuff. 

Some usage guidelines for how to display rightsstatements.org (which we don't totally follow at present): https://rightsstatements.org/en/documentation/usage_guidelines.html

A usage example for how to display CC license can be found eg https://creativecommons.org/choose/results-one?license_code=by-nc-nd&amp;jurisdiction=&amp;version=2.0&amp;lang=en
We use the shorter (cc) logo instead of the complete one, but have full wording so it should be okay. We also use the rel=license markup they recommend for machine-access to license.


jrochkind:
while this hasn't been reviewed by stakeholders yet (see screenshots at #973), since we aren't yet using this new license term in production data anyway, it can be reviewed and merged into master at any time.

jrochkind:
ready for review/merge

eddierubeiz:
Thanks for dealing with this. Appreciate the removal of the vestigial OralHistoryDeliveryJob, too.

jrochkind:
This PR accidentally ended up including the functionality that was intended to be separated out in  #972 instead. Oh well, now it's merged like this, so be it!

Issue #975, opened 2021-02-11
Modeling for new OH interviewEE biographical metadata


eddierubeiz:
For each oral history, we want to store the following career, education and award metadata for interview**ee**s:

- Place and date of birth and death;
- One or more "schools" (e.g. `{date: "1958", institution: 'Columbia University', degree: 'BA', discipline: 'Chemistry'}`;
- One or more "jobs" (e.g. `{start: "1962", end: "1965", institution: 'Harvard University',  role: 'Junior Fellow, Society of Fellows'}` );
- One or more "honors" (e.g. `{date: "1981", honor: 'Nobel Prize in Chemistry'}`).

Not covered in this issue are:
- Any metadata re: institutions (employers or universities, etc., except for their names, stored as strings.)
- Any changes to the user interface (I'm actually already working on this, but let's keep things compartmentalized for sanity's sake).
- Any metadata pertaining to the interview**er** (ditto)
- Any extra biblio metadata pertaining to the interview itself.

PR #976, opened 2021-02-11
New OH interviewee metadata: new models and corresponding DB migration.

eddierubeiz:
Ref #975.
Keeping everything extra simple, so any and all front-end work will be in a separate PR.
This should allow us to migrate *all* the interviewee metadata present in the Drupal form present at e.g. https://oh.sciencehistory.org/node/5913/edit.


jrochkind:
"corresponding DB migration" -- if we're doing attr_json, do we need a db migration? Hm, maybe one column for the embedded JSON, if a suitable one isn't there already?

Just reacting to PR title, haven't reviewed draft code yet!

app/models/oral_history_content.rb
```diff
@@ -105,6 +118,52 @@ def combined_audio_derivatives_job_status=(value)
     self.combined_audio_derivatives_job_status_changed_at = DateTime.now
   end
 
+  def interviewee_birth_place
+    @interviewee_birth_place ||= begin
+      return nil if interviewee_birth.nil?
+      interviewee_birth.to_h['place']
+    end
+  end
```
jrochkind, line 123: OK, you don't need #to_h here to convert it to a hash and then access it as a hash key -- it's a model object that already has a `place` method. So the right implementation is just `interviewee_birth.place`

If you want to guard against interviewee_birth being nil, you can just use the ruby "safe navigation operator", so:

    def inteviewee_birth_place
       interviewee_birth&.place
    end

There is no need to memoize it, as it's not at all expensive to look up. 

At that point, I think there is no need for these wrapper methods. Code that wants this can just call `oral_history_content&.interviewee_birth&.place` directly. 

Or if you disagree and want the wrapper method I could be convinced, but it should be this simple one-liner. Same for all following similar methods.

app/models/oral_history_content/interviewee_death.rb
```diff
@@ -0,0 +1,14 @@
+require 'attr_json'
+class OralHistoryContent
+  class IntervieweeDeath
+    include AttrJson::Model
+
+    validates_format_of :date, with: /\A\d{4}(-\d{2}(-\d{2})?)?\z/,
+      message: "must be of format YYYY[-MM-DD]"
+
+    # TODO validate `category` string can be either 'birth' or 'death'
```
jrochkind: Let's do that TODO before merging... although wait, I think the TODO is no longer relevant, no need for a "category", instead you just have attributes `interviewee_birth` and `interviewee_death`. The way you've done it is great, so just eliminate old wrong TODO comment. 

    validates :exhibition, array_inclusion: { in: ControlledLists::EXHIBITION  }

app/models/oral_history_content/interviewee_honor.rb
```diff
@@ -0,0 +1,14 @@
+require 'attr_json'
+class OralHistoryContent
+  class IntervieweeHonor
+    include AttrJson::Model
+    validates_presence_of :date
+    validates_presence_of :honor
+
+    validates_format_of :date, with: /\A\d{4}(-\d{2}(-\d{2})?)?\z/,
+      message: "must be of format YYYY[-MM-DD]"
```
jrochkind, line 9: We're re-using this date format validation in a bunch of places, including originally in Work::DateOfWork. Let's find some way to DRY it, the regexp and validation failure message?

jrochkind (changes requested):

app/models/oral_history_content.rb
```diff
@@ -36,7 +36,9 @@
 # associated works, and using kithe techniques to control auto-indexing: batch-updating,
 # or turning off auto-updating.
 #
+require 'attr_json'
```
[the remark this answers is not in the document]
eddierubeiz: Don't need it!

app/models/oral_history_content.rb
eddierubeiz: Removed the memoizers -- keeping the wrapper methods, though.

app/models/oral_history_content/interviewee_honor.rb
eddierubeiz: Created a special validator, but decided against using it for DateOfWork, at least for now.

Issue #977, opened 2021-02-16
Basic display of OH interviewee metadata in a new tab on the work admin section


eddierubeiz:
We'll start out very simple with just a table listing birth / death place and date, education, career, and awards.
No form for editing any of this yet.
None of this metadata will be shown to users on the public-facing pages yet.

jrochkind:
See #996 on "role" in our Dig Coll data model

MDiMeo:
When we're ready to discuss the front-end design of the metadata pages, please reach out to @HKativa for feedback when you need it.

PR #978, opened 2021-02-16
Display and edit OH interviewee metadata in admin.

eddierubeiz:
Basic display of OH interviewee metadata, with a form for updating it.

The metadata is shown on the existing OH admin tab, with a button taking you to a separate form for editing it.

Fields for institutions (schools and employers) offer FAST autocomplete.

Countries, provinces and states are provided from a static list, until we decide which gem to use (if any.) `country` was a bit much for our needs.

I added a simple front-end test.

Ref #977
Relies on #976 .


app/models/oral_history_content.rb
```diff
@@ -108,6 +108,16 @@ def has_ohms_index?
...
 
 
+  # This is required in order to use repeatable_attr_input .
+  # These methods aren't actually called.
+  def interviewee_school_attributes=(some_arr)
+  end
+  def interviewee_job_attributes=(some_arr)
+  end
+  def interviewee_honor_attributes=(some_arr)
+  end
```
jrochkind, line 118: we shouldn't need these, not sure why we do -- compare to analagous embedded models in Work to figure out why it didn't need it there. Might be because of how the controller is handling assigning input.

app/models/oral_history_content/date_and_place.rb
```diff
@@ -8,5 +8,22 @@ class DateAndPlace
     attr_json :province, :string
     attr_json :country, :string
 
+    def update_from_hash(hsh)
+      self.date =         hsh['date']
+      self.city =         hsh['city']
+      self.state =        hsh['state']
+      self.province =     hsh['province']
+      self.country =      hsh['country']
+    end
```
jrochkind, line 17: I don't think you should need this, I am pretty sure .new should do the same thing, you should really be able to do `model.new(hash)` instead of `model.update_from_hash(hash)`, let's debug why not if not.

app/views/admin/works/oh_bio_form.html.erb
```diff
@@ -0,0 +1,149 @@
...
+    <div class="form-group">
+      <%= f.repeatable_attr_input(:interviewee_honor, simple_form_input_args: { :label => "" }) do |sub_form| %>
+        <% if sub_form.object.errors.present? %>
+          <div class="alert alert-danger mr-2">
+            <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+            <%sub_form.object.errors.errors.each do |err| %>
+              <%= err.type %><br/>
+            <% end %>
+          </div>
+        <% end %>
```
jrochkind, line 136: I believe simple_form/kithe should be automatically displaying errors and you shoudln't need to do this manual code. But I'm not totally sure, I could be wrong. Let's compare to the straight work metadata and do it the same.

eddierubeiz: This is correct. Removed

app/models/oral_history_content.rb
eddierubeiz: "If you have nested AttrJson::Models you'd like to use in your forms much like Rails associated records: Where you would use Rails accepts_nested_attributes_for, instead include AttrJson::NestedAttributes and use attr_json_accepts_nested_attributes_for. Multiple levels of nesting are supported."

This solved the problem and made the form much simpler.

jrochkind: Excellent, nice job! This stuff is kind of confusing. I wonder if AttrJson should turn that on by default.

app/models/oral_history_content/date_and_place.rb
eddierubeiz: You were absolutely right -- just using a simple data.values().all? { |x| x.empty? } in the controller was fine.

jrochkind: I think that comment about empty? is about something other than `update_from_hash`?

I have absolutely nothing against providing a `def empty?` that does that, instead of having to do `values.all` every time, a `def empty?` in the model is a nice solution! Either way is fine. 

If you do want to do that line, style is not use parens after the method call, and there's a shortcut for `{|x| x.empty?}`, so it can look like this: `data.values.all?(:empty?)`.

eddierubeiz: Fixed!

app/controllers/admin/works_controller.rb
```diff
@@ -113,6 +113,51 @@ def download_ohms_xml
...
   end
 
+  # Biographical metadata form
+  # GET "/admin/works/ab2323ac/oh_bio_form"
+  def oh_biography_form
+    @work.oral_history_content!
+    @work.oral_history_content.
+      interviewee_birth ||= OralHistoryContent::DateAndPlace.new
+    @work.oral_history_content.
+      interviewee_death ||= OralHistoryContent::DateAndPlace.new
```
jrochkind, line 123: I think you can avoid these `||=` lines by just making this a default in the model `attr_json`, so it defaults to a blank one.

app/controllers/admin/works_controller.rb
```diff
@@ -113,6 +113,51 @@ def download_ohms_xml
...
+  end
+
+  # PATCH/PUT /admin/works/ab2323ac/submit_oh_bio
+  def submit_oh_biography
+    oral_history_content = @work.oral_history_content!
+    oral_history_params = params['oral_history_content']
+    oral_history_content.interviewee_birth = new_item_from_params( nil,
+      oral_history_params["interviewee_birth_attributes"],
+      OralHistoryContent::DateAndPlace)
+    oral_history_content.interviewee_death = new_item_from_params( nil,
```
jrochkind, line 134: This is getting a little bit simpler than last time, but i still think there should be a way easier way to do this, without that `new_item_from_params` that looks more like:

    oral_history_content.interviewee_birth = OralHistoryContent::DateAndPlace.new(oral_history_params["interviewee_birth_attributes"])

Or actually it should be even simpler than that, it should look mostly like:

    oral_history_content.update_attributes(params['oral_history_content'])

It is admittedly confusing how Rails does this stuff, and then made more confusing with our attr_json additions. But I think we should do it the fewer lines of code way. It will also need Rails "Strong Params", have you ever looked at or learned about those? 

That might be the missing piece. 

But the way we do this with the repeated fields in Work and the analagous form there should be the guide. There are repeated multi-field sub-models there; there is no need for a `new_item_from_params` method there. We should try to do it here similarly. 

It admittedly kind of confusing; I can take a stab at it if you like.

app/models/oral_history_content.rb
```diff
@@ -124,15 +131,18 @@ def interviewee_death_place
   end
 
   def interviewee_schools_sorted
-    interviewee_school.sort_by { |hsh| hsh.date }
+    return [] unless interviewee_school.present?
+    interviewee_school.sort_by { |hsh| hsh.date }.compact
   end
```
jrochkind, line 136: These `sorted` methods should be in a presenter or view-level component of some kind rather than the model. 

Or, we could just skip it entirely -- let the staff entering them enter them in the order they should be, and just respect the order they were entered in?  I bet that's what the microsite did, and may be good enough if we just want to be done with this. But if we're going to have extra "sorted" methods, they shouldn't be muddying up the model, but should be in a presenter or view layer.

app/models/oral_history_content/date_and_place.rb
```diff
@@ -8,5 +9,25 @@ class DateAndPlace
...
 
+    def state_name
+      return nil unless country == 'US'
+      ISO3166::Country.new(country).subdivisions[state]['translations']['en']
+    end
+
+    def province_name
+      return nil unless country == 'CA'
+      ISO3166::Country.new(country).subdivisions[province]['translations']['en']
+    end
```
jrochkind, line 20: Is this really the way the `countries` or `country_select` gem, or wherever this comes from, wants you to go from the code to a human name? I am surprised it's not more straightforward than this. 

Do you know what gem he province and state names are coming from? Can you link to the documentation that suggests using it's API like this? I'm just wondering if there's a different way it's recommending.

app/models/oral_history_content.rb
eddierubeiz: Got rid of them -- totally unnecessary in retrospect.

app/controllers/admin/works_controller.rb
eddierubeiz: No, I need them as the code currently stands. I've written tests that cover the scenario.

jrochkind: Can/should you replace em with attr_json defaults? In OralHistoryContent, 

     attr_json :interviewee_birth, OralHistoryContent::DateAndPlace, default: -> { OralHistoryContent::DateAndPlace.new }

app/models/oral_history_content/date_and_place.rb
eddierubeiz: I got rid of this-- we're just using a simple file with text arrays right now.

app/controllers/admin/works_controller.rb
eddierubeiz: No, unfortunately that won't work. The default is only applied for new `OralHistoryContent`s, and if you delete the attribute, it needs to be recreated before showing the form-- otherwise the fields won't show up.

app/controllers/admin/works_controller.rb
eddierubeiz: Thanks for your help on Tuesday!

app/controllers/admin/works_controller.rb
jrochkind: OK, no problem, it's probably good enough at this point, but "if you delete the attribute" -- why would you ever delete the attribute?  Oh right, because you wrote code to delete it if you blank out all the fields!

Ok, it's good enough!

PR #979, opened 2021-02-17
Custom front-page for Oral History collection specifically

jrochkind:
There are a couple ways we could have done this. We *could* have put some conditional logic in the existing CollectionShowController that tried to configure it differently when it recognized a specific collection. 

But instead, I sub-classed the CollectionShowController into namespaced CollectionShowControllers::OralHistoryCollectionController.  Then I used some kind of convoluted tricks in routes.rb to route to that controller *instead* of the default CollectionShowController, when the collectionId for Oral History was noticed. 

It's a bit hacky -- too much sub-classing especially of Rails controllers can get messy to understand/debug. But I think it was going to be messy no matter what, i decided this seemed the cleanest, especially because we are probably going to want to customize the whole "home page" too (views), not just some config. 

For now, all we're customizing is the facet-related configuration. 

This was done using a pattern that can easily be extended to other specific collections, should we want to customize other specific collection landing pages too.  See routes.rb, we just put in a route constrained by other collection IDs and routing to other special-purpose sub-classes.

### DIFF
--- a/app/controllers/collection_show_controllers/oral_history_collection_controller.rb
+++ b/app/controllers/collection_show_controllers/oral_history_collection_controller.rb
@@ -1,0 +1,9 @@
+# A sub-class of collection show controller that is specifically for the Oral History
+# collection -- it's routed to for that collection in our routes.rb
+#
+# It lets us override configuration and views.
+module CollectionShowControllers
+  class OralHistoryCollectionController < CollectionShowController
+
+  end
+end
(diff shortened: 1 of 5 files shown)


jrochkind:
something spec'd but not done yet, is doing something with the "format" facet to transform it into something that lets you limit to OH's that contain "PDF", "Audio files", and/or "Synchronized Transcript". Unclear the best way to implement this right now, also terminology to use unclear.

jrochkind:
Will require a reindex on production after we merge. 

    ./bin/cap production invoke:rake TASK="scihist:solr:reindex"

jrochkind:
Right now this just delivers custom facets, but the architecture is there, with a custom controller, to support other customizations such as view templates, in the future. (If it works out as expected anyway).

eddierubeiz:
Got it. Yes, other snowflake collections with special UI are certainly easy to predict (hello, Bredig.) This all makes sense to me.

Issue #655, opened 2020-03-17
link to audio at specific timecode


eddierubeiz:
I don't know whether our audiences would find this feature useful, but this came up in passing at yesterday's meeting, and I thought it would be interesting to consider in the long term. (Don't sue me.)

If you want to annotate time-based media or use it as a primary source, it's helpful to be able to refer to a portion (a "clip") of an existing audio file using a URL.

A clip URL would be comprised of the following metadata:

- asset friendlier_id
- start offset time relative to the beginning of the audio
- end offset time relative to the beginning of the audio

What happens when you visit such a URL? Again, it depends on how our public would be using this feature.

jrochkind:
What I was thinking was more of just a "start at" time, which when you accessed would just start playing at that time. I'm not sure there's actually a need for an "end at" time. 

I think it would be implemented as simply our existing audio work URL, with a `?s=#{time}` on the end. 

Note that YouTube has a similar feature. 

One additional thing specifically relevant to our "Table of Contents" -- which also matches a feature in OHMS standard viewer -- is we could also have a feature taht alternately puts a specific ToC segment in a query param, to both open up that ToC segment accordion AND start playing at that time.

jrochkind:
Note asset friendlier_id is not actually relevant, since we're moving to a "only play from combined derivative" model. Which is why the simple existing work URL with additional query params works.

eddierubeiz:
A related feature: the ability to link to a particular section of the table of contents. This would pair nicely with #684, section-level outbound links.

PR #993, opened 2021-02-23
link to ToC segment and timecode in OHMS

jrochkind:
Ref #655

Among other issues to consider:

* When we link directly to ToC segment, there can be no SHI branding at all, cause it scrolled off screen! Oops. Consequence of our design. 

* No tests at present, this is not easy behavior to test. 

* Not sure about using #fragmentIdentifier vs ?queryString

* We have implemented ability to link directly to a given timecode in seconds, but have no UI for getting those links yet. (Other than right-click copy-link on a transcript timecode, that works presently)


jrochkind:
Closing, re-did this as #1132

PR #1001, opened 2021-03-02
DRY up our pagination status info

jrochkind:
I noticed when adding admin page for interviewer profiles that this code was copy-pasted in several places, makes sense to DRY it up into one of our presenter viewmodels.

Theoretically we could be using kaminari functionality for this (kaminari is the gem for pagination), by theming it to be how we wanted. Not sure why we didn't, but not looking into that, just taking the code that was there and DRYing it.


eddierubeiz:
Awesome.

PR #1002, opened 2021-03-03
update dependencies

jrochkind:
`bundle update` to update all dependencies within any requirements

Then did a `bundle outdated` to see what was still outdated -- saw a couple that we could probably update by changing our Gemfile, so did that too. Did test a cap deploy to staging after updating bundler-capistrano. 

Our `bundle outdated` is so very small now! Does include blacklight, where we are intentionally locking to an older version, because BL minor releases tend to break our app, or in some cases produce voluminous deprecation warnings that we have no good way to eliminate. 

Except for Blacklight, we are remarkably free of outdated dependency debt, which is marveous, and very atypical for our community from what I've seen. I think it's a testament to our "write our own app" approach.


jrochkind:
Also worth noting, there were a few dependencies which in the past were holding us to older versions of transitive dependencies -- which I went and filed issues and sometimes PR's to try to fix. That's part of what's helped us get to so small a bundle outdated list now, hooray.

eddierubeiz:
Looks great. Thanks!

PR #1003, opened 2021-03-03
OH Interviewer Profiles

jrochkind:
Very limited testing, maybe needs more?

Used a Rails has_and_belongs_to_many ("invisible" to rails join table), even though lots of people say never to do that (instead using the more complicated has_many :through with visible join table).  Because it was just SO much simpler. And this stuff is confusing enough already (especially with the Admin:: namespace that makes all the rails magic name-guessing more complciated always. Maybe it shouldn't have been used here for model not sure, but we already started putting some models under Admin::. Actually wondering if we shoudl't have done that, there's really nothing admin about this model, hmm). Can always be changed later. But I don't think we'll need to, this func isn't going to go anywhere else. 

The staff interface for selecting profiles is a multi-selection select menu which is probably going to be difficult to work with, this may just be a proof of concept.

### DIFF
--- a/app/controllers/admin/interviewer_profiles_controller.rb
+++ b/app/controllers/admin/interviewer_profiles_controller.rb
@@ -4,7 +4,7 @@ class Admin::InterviewerProfilesController < AdminController
   # GET /admin/interviewer_profiles
   # GET /admin/interviewer_profiles.json
   def index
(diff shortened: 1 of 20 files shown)


jrochkind:
@eddierubeiz this is ready for review/merge.

eddierubeiz:
Nothing too crazy. Merging.

Issue #754, opened 2020-05-18
Validate all OHMS transcripts currently in production.


eddierubeiz:
We recently tightened up the rules for OHMS transcripts: bad footnotes are no longer accepted when you upload the XML file for a transcript. We need to go back and ensure that all items currently in production, including those uploaded before we changed the rules, follow the new rules.

jrochkind:
Do we need to? We can ask @lsberry16

eddierubeiz:
I found two (tiny) footnote errors in staging, and was able to check that they also exist in production. It would be nice to fix them.

It's unlikely that there are any production errors that don't *also* exist in staging, but we'll run the test in production anyway later this week.

[Oral history interview with Malcolm M. Renfrew ( 1n79h560c )](https://digital.sciencehistory.org/works/1n79h560c#ohTranscript) has the following errors:
Missing reference(s) to footnote(s): 25

[Oral history interview with Izaak M. Kolthoff ( fb494945p )](https://digital.sciencehistory.org/works/fb494945p#ohTranscript) has the following errors:
Reference to missing footnote 16

MDiMeo:
@lsberry16 will fix OHMS errors

lsberry16:
Done! There was an error in the coding of Kolthoff's footnotes and Renfrew had added some written addenda in which the missing footnote 26 appeared but these addenda had been omitted from the OHMS transcript. I fixed both transcripts in OHMS added them and reuploaded the OHMS XML to the Digital Collections.

eddierubeiz:
Thanks @lsberry16. I ran the check code after you fixed the two problems above and now there are no problems in production.

jrochkind:
ok based on those reports i am closing this issue.

PR #1014, opened 2021-03-08
basic ohms validation on existing oral histories

eddierubeiz:
Ref #754 
`bundle exec rake scihist:check_ohms_xml_data` will run `OralHistoryContent::OhmsXmlValidator.new(model.ohms_xml_text).valid? ` on all oral histories and prints out the results.


jrochkind:
Looks good. 

We validate on input for new things entered/uploaded, right? This is mainly to check the "historical" stuff that pre-dated that validation?

Do you think we'll ever want to run this again? I wonder if we actually want to merge it, or just use it as I think you already have, and then leave it alone. More code in the codebase is just more code that can eventually stop working or confuse developers trying to find something else etc. Thoughts?

jrochkind:
From the first issue:

> I found two (tiny) footnote errors in staging, and was able to check that they also exist in production. It would be nice to fix them.

This task doesn't like automatically fix anything, right? It just identifies them? So the ones you've already identified, you can go ahead and fix now even before merge, yes? 

I see why we might need to merge to run it in production at all though! What do you think about merging, running in production, then removing from codebase? Alternately, you could just copy and paste the five lines to find invalid into console on production, that's what I've done before for things like this. 

It's good that the existing validation code is well-factored so it's so easy to check em all in 5 lines!

eddierubeiz:
I can't really _fix_ this problem without e.g. adding a footnote to the text or removing a footnote. But we can bring it up on Monday, so people know there are two small problems that need to be addressed.

jrochkind:
I moved to `bundle exec rake scihist:data_fixes:check_ohms_xml_data` so we can keep all temporary data fix rake tasks together. Maybe I'll add a README.

Issue #998, opened 2021-03-01
Adjust clean up of temporary ingest files


jrochkind:
### Original feature was a sketch

1-2 years ago we implemented the "Add from cloud" feature, that allows staff users to first copy assets over to a temporary ingest holding pen S3 bucket, and then select files from that "holding pen" for attaching to an ingest. When it does this, it makes another copy of the file in another location, based off of what was in the temporary holding pen. 

At that time, we considered -- should the file in temporary holding pen be automatically deleted after ingest? 

At that time, we decided against this, in my memory based on "Well, if the ingest fails, we want it still to be there to try again."   Also possibly that it was a bit non-completely-straightforward to implement. 

But we don't want to leave the files on that tempoary holding pen forever, because we'd have to pay for their storage there, and because they'd be clogging up the list of files for staff to choose from!

So, at that time -- before this new feature had ever been used -- we came up with the simplest thing we thought could work. Figuring that we might have to revisit that later. And that was, just leave the things in the holding pen bucket for 7 days after they were added there. The idea was that normally they are added right before they are going to be used.  7 days after it's been added, it gets deleted, the logic doesn't know about when it's actually been ingested, because we were using a simple S3 "lifecycle rule" feature on the S3 buccket. 

More recently we expanded 7 days to 30. 

### Bulk load of batch of Oral Histories to temporary ingest holding pen bucket

Recently (approx Feb 24-26?) Chuck added a giant batch of OH files to the ingest holding pen, so they'd be there for the project work of this current batch, to make ingest as efficient as possible. 

Because we might not be done with them all even in 30 days, our current logic is not necessarily great -- what if they accidentally expire when they aren't all ingested yet? We'd have to do it again. 

It is a bit of a pain/cost to "do it again"; staff needs to find/identify the files to copy to holding pen; Chuck needs to run a routine that can take 24+ hours to complete to copy over the 1.7TB; and it costs about $4.00 (four and no cents) to copy over all of those files. 

While the ~1.7TB of these files exist in the temporary holding pen, we pay about $40/month to store them there, if they are all there without being deleted. (This can not necessarily be used to predict our added costs of indefinite storage in digital collections, as there are things that make the permanent storage both more and less expensive than this temporary holding pen; although it can be a rough order of magnitude. This $40 is for an *extra* copy that is just sitting in the temporary holdingpen). 

## What should we do?

[...]

PR #1019, opened 2021-03-09
Ingested items from ingest holding pen bucket get deleted 7 days after ingest

jrochkind:
Ref #998 

We use a promotion callback (provided by kithe/shrine), to hook into promotion of file -- copying from temporary uplaod location to permanent storage. If it's coming from the :ingest_bucket (temporary holding pen), after succesful promotion we create a record in a new table, with a date on it after which it can be deleted. 

There's a rake task that deletes those (and cleans up record), and a cronjob that runs it nightly. 

* There are no tests, I couldn't think of a non-horrible way to do tests, so much is about interaction with S3 and things that are set up differnet in production. 

Also: Around when we deploy/merge this, we have to remember to go turn OFF the 30-day-delete lifecycle rule on the S3 bucket. Some things WILL slip through the cracks -- ingested in the last 30 days, won't be recorded to delete 7-days-after-ingest, but won't be deleted after we turn off lifecycle rule. So be it, that's how it is!


jrochkind:
Hm, I must not have committed a consistent db/schema.rb. I'll get it merged with master and figured out.

jrochkind:
I manually deleted the "delete 30 days after upload" rule in AWS console from the bucket `scihist-digicoll-production-ingest-mount`. 

i will also PR to our ansible-inventory playbook to remove it from the ansible playbook for creating S3. 

I tested, and removing it from the playbook and re-running it is not alone enough to remove it -- the playbook is smart enough to add it if it's not there, but not to remove it once we've removed it from our playbook. One reason ansible might not be the ideal tool for trying to maintain S3 configuration-as-code.

jrochkind:
@eddierubeiz 30-day-delete is now gone from bucket, and this one is ready to be reviewed/merged at your convenience!

lib/tasks/scheduled_ingest_bucket_deletions.rake
```diff
@@ -0,0 +1,17 @@
+namespace :scihist do
+  desc "Delete files from ingest_bucket that have been scheduled"
+  task :scheduled_ingest_bucket_delete => :environment do
+    ScheduledIngestBucketDeletion.where("delete_after < ?", Time.now).find_each do |scheduled|
```
eddierubeiz, line 4: I feel like this should be `>`, not `<` .

jrochkind: Hm.  Let me think/talk through it. 

Let's say it's Jan 1. We ingest some files, and we create ScheduledIngestBucketDeletion object for each of them with a `delete_after` of Jan 8. Meaning, it's okay to delete these after Jan 8. 

Then on Jan 9, we want to find all the `ScheduledIngestBucketDeletion` objects that have a `delete_after` *before* now, because they are safe to delete. 

So I think that is expressed by `delete_after < [now]`. 

If we were instead deleting all the objects with a `delete_after > [now]` that would mean on Jan 9 we'd be deleting an object (eg) with a delete_after of `Feb 1`, which was supposed to mean "okay delete after Feb 1". 

So I think `<` is right. What do you think?

eddierubeiz: You are right  -- it's <. Objection withdrawn! I was really thinking in terms of a  `keep_until` date, not a `delete_after` date.

PR #1026, opened 2021-03-16
OH IntervieweeHonor needs two dates, start and end, not just one date

jrochkind:


jrochkind:
@eddierubeiz OK, honors section displays kind of non-ideally, and differently than than the other sections... but I'm beating my head against the wall on this, let's do this just to get start-date/end-date for honors in there, so it's there in the model at least for data migration and stuff. 

I may make one or more addtional PR's with other improvements to the general form situation.

PR #1027, opened 2021-03-16
Clean up OH biographical form

jrochkind:
Use more correct bootstrap CSS, with complete control of bootstrap grid, by way of a new simple_form wrapper.

Use form-row for tighter spacing. Generally cleaner.


eddierubeiz:
Thanks for cleaning this up!

PR #1028, opened 2021-03-17
For OH access request for ineligible content, log and return 404, instead of raising non-specific error which results in 500 and honeybadger notify.

jrochkind:
We had an instance where a bing search bot triggered this. Maybe the Work had requestable content at some point, at which bing came across the url, and then it was edited to longer have it? Hard to say. But we don't need to get a honeybadger error when this happens, just giving bing a 404 is fine. The page does not exist, there is no page at that URL.


eddierubeiz:
Yeah -- this was my bad. Thanks for fixing it!

PR #1029, opened 2021-03-17
Tell Honeybadger to ignore invalid MIME errors

jrochkind:
Rails 6.1 finally uses a specific identifiable class for this, so we can add it the ignore list.


eddierubeiz:
Hooray!

PR #1033, opened 2021-03-18
Display of OH interviewee biographical info

jrochkind:


jrochkind:
So one thing I don't love here is our our CSS involving headings (h2, h3, etc) has become pretty irrational -- we don't have a simple stylesheet with a progression of less emphasis headings we're using, but instead each use (relevant here for bio info) is like one-off hacked to "look right". 

In a way that will make it harder to fix, as we're doing things like `<h2 class="h4">` (this is an h2, but make it LOOK like an h4, I mean like an h4 currently looks).  

So I had some thought to try to improve that before merge -- but I don't have the time/energy for it. It will have to be just more CSS technical debt for now -- our CSS definitely has some technical debt, I'm not really good at keeping CSS clean and elegant over the lifetime of a growing project.

eddierubeiz:
Looks great. The date range display object will be nice to have; we'll use it a lot, no doubt.

PR #1040, opened 2021-03-23
This file isn't used anymore, remove it

jrochkind:
To avoid it being in the way confusing me into thinking it's the file i'm looking for!

I think this was an experiment that may never have been used in production.


eddierubeiz:
Great -- thanks.

Issue #985, opened 2021-02-22
Better display of non-audio oral histories


jrochkind:
See this one:

https://digital.sciencehistory.org/works/1v53jz08f

It shows up weird, with a giant poster image of PDF cover page. It should be more like the other template we have with files listed in the left column.

jrochkind:
OK, so another one that ends up with some questions! 

Right now oral histories without any audio are displayed using our "Standard" page template that was optimized for image-centered works. So they end up with a gigantic poster image of the PDF cover page, which isn't really very useful:

![Screen Shot 2021-02-22 at 4 21 24 PM](https://user-images.githubusercontent.com/149304/108773443-5cbf8c00-752c-11eb-8971-43c1e94d5bc2.png)

We have another template, that we've been using for oral histories that have assets available only "by request", that instead of a giant poster image has a file list.  

What if we used that for all oral histories without audio too? 

![Screen Shot 2021-02-22 at 4 40 03 PM](https://user-images.githubusercontent.com/149304/108773622-a314eb00-752c-11eb-8cab-c578aa028556.png)

One issue, as you can see, is it's really highlighting the filename of the PDF more. The filename that may not really make a lot of sense to the user. 

a) That's fine, this is fine they'll deal with it. 

b) staff can manually enter a label for the PDF, "Transcript", so that's what shows up. This is potentially a pain on ingest. (We can go probably bulk edit all existing ones). 

c) This is _another_ reason that has me thinking about adding an architectural feature where we can give an ingested file a "role", such as "transcript", so that the software can know to just label it "PDF Transcript" if it has that role. See extensive discussion at #990 

So this is another one where I need a bit of feedback before proceeding. @MDiMeo @lsberry16

jrochkind:
See #996 on "role" in our Dig Coll data model, relevant to challenges here, could automatically label transcript as "Transcript" if we had role in our data model.

PR #1041, opened 2021-03-25
Some general clean-up of OH views

jrochkind:
Motivated by new asset 'roles' feature. 

1. A rake task to assign roles to legacy OH assets, based on exisitng assumptions in logic
2. We had two different "tabular file listing" displays of assets in the two different OH templates being used. They were copy-paste-changed. We DRY them into a new ViewModel component. So they now both look more consistent, which changes how they look a bit. Also tweaked/improved a few things while doing this. 
3. Since it's DRY now, we can, in one place, make them label files "Transcript" or "Front matter and index" (rather than filename or manual label). 
4. This makes it possible for us to use the file-listing template for OH's with *no* audio files at all, instead of default image-based template which wasn't really working. Ref #985
5. We also change filenames used for downloads to the original_filename for both audio and PDF files. This makes sense, and matches microsite. Eg might be "Molina_M_0896_FULL.pdf" -- actually we might still put the ID on the end for enforced uniqueness.

After this goes live, ingest staff will no longer have to manually assign "label" to Front matter, but *will* have to assign "role" to Front Matter and Transcripts.

### DIFF
--- a/app/assets/stylesheets/local/oh_audio.scss
+++ b/app/assets/stylesheets/local/oh_audio.scss
@@ -273,26 +273,6 @@
 
   .other-files {
     border-top: 3px solid $table-border-color;
-
-    .show-member-list-item {
-      display: flex;
-
-      & > div {
-        padding: $table-cell-padding;
-        border-top: $table-border-width solid $table-border-color;
-      }
-
-      .image {
-        width: 54px; // thumb_mini width
-        padding-left: 0;
-      }
-
-      .title {
-        flex-grow: 1;
-        width: 100%;
-        word-break: break-word;
-      }
-    }
   }
 
   .work.chf-attributes {
--- a/app/assets/stylesheets/local/oh_audio.scss
+++ b/app/assets/stylesheets/local/oh_audio.scss
@@ -271,8 +271,8 @@
     }
   }
 
-  .other-files {
-    border-top: 3px solid $table-border-color;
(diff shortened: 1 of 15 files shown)


jrochkind:
Example files for the three types of displays affected. This branch needs to be deployed to staging to demo the new on these staging links. 

1. By-request. [staging (new)](https://staging-digital.sciencehistory.org/works/nb58t1q) vs [production (old)](https://digital.sciencehistory.org/works/nb58t1q)
   * Front matter will be automatically labelled based on role. 
   * Original filename is also there
   * When you download, you get it downloaded using original filename

2. Immediate availability with audio: [staging (new)][https://staging-digital.sciencehistory.org/works/8pmnk1q#ohDownloads] vs [production (old)](https://digital.sciencehistory.org/works/8pmnk1q#ohDownloads)
  * note label is "Transcript", but filename is also there as metadata

3. Immediate access, but no audio file present at all: [staging (new)](https://staging-digital.sciencehistory.org/works/1v53jz08f) vs [production (old)](https://digital.sciencehistory.org/works/1v53jz08f)
   * totally different template, works much better I think

jrochkind:
@eddierubeiz I don't want to merge this before I show it to stakeholders, but you can review it whenever you have a chance!

eddierubeiz (approved):
Looks good. The front end tests on OH pages are pretty extensive, as I recall.

Issue #1025, opened 2021-03-16
index OH biographical info


jrochkind:
I realized that we haven't done the work to make the new OH biographical fields *searchable*, to make queries match on words from them. 

So this is just a ticket to register that, and that we should do it. 

It could be quite simple, depending on what we want to do. Options from simple (10 minutes of work) and less so (several days of work) include:

* Just index it as "ordinary metadata", will match with about the same level of relevance boosting as other "ordinary" metadata, will not show up as "highlighted matches"
* index at more carefully chosen relevancy boostings, perhaps different for different fields. Perhaps "instititution" and "honors" should be ranked more like "subject", higher than low level metadata
* Index in a way that shows up as highlighted matches
* Eventually offer "fielded search" on the OH collection page, so you can search just within "institution" field for instance. 

We might want to start out with something dirt simple, and make a different ticket for doing something more complex. 

@lsberry16 @MDiMeo

PR #1048, opened 2021-03-29
solr index interviewee biographical

jrochkind:
Ref #1025


eddierubeiz:
This will have to be redone as part of The Great Refactor, but merging it now.

Issue #1049, opened 2021-03-30
Adjustments to interviewee bio display


eddierubeiz:
The display of interviewee bio data on the public page could use a couple adjustments in light of the imported data. We may want to modify the OH tab in the admin interface, too.

- In Experience and Honors, if the start and end dates are the same, hide one. "1947 to 1947" should read instead "1947".
- Honors > Award contains arbitrary HTML. We're importing it as is, so the front end should display it as HTML.

I think that's it!

For an example, compare:
https://oh.sciencehistory.org/oral-histories/benedict-manson
https://staging-digital.sciencehistory.org/works/kw52j9080

eddierubeiz:
@jrochkind, here's [a list of 5284 honors](https://github.com/sciencehistory/scihist_digicoll/files/6229719/awards.txt).
 If you (or I, in the import code) get rid of the enclosing `<p>` tags, I suspect many of our problems with this field will go away.

jrochkind:
@eddierubeiz to get rid of <p> while leaving short list of allowed tags, on ingest:

    DescriptionSanitizer.new.sanitize("<p>This is <i>an italicized title</i>.</p>")
    # => "This is <i>an italicized title</i>."

eddierubeiz:
In the `honors` section for e.g. [Manson Benedict](https://oh.sciencehistory.org/node/5367/edit) , if you *do not* select the "Show End Date" checkbox, Drupal will store a copy of the start date in the end date field. I have yet to see any trace in the database of the value of that checkbox, which is frustrating. This may merit a ticket of its own.

jrochkind:
> In the honors section for e.g. Manson Benedict , if you do not select the "Show End Date" checkbox, Drupal will store a copy of the start date in the end date field. I have yet to see any trace in the database of the value of that checkbox, which is frustrating. This may merit a ticket of its own.

Interesting!

So, I've already done the code to display this correctly if we have start and equivalent. 

But, in the interests of keeping our data clean -- I am thinking your ingest should probably leave the end blank if it is identical to the start? Since that's how appears to the editor in the drupal microsite; it's sort of a quirk of the implementation that it fills in the 'end' anyway, but if we import it that way, it will show up that way in our app's editing screens, differently than it shows up in drupal editing screens. 

What do you think?

PR #1050, opened 2021-03-30
Improve interviewee biographic

jrochkind:
1. Ignore end date same as start date in formatting dates. (Still think this might be a data import error too, probably good to look into it)
2. Allow limited sanitized HTML in honors string

Ref #1049


eddierubeiz:
Thanks. I'm also sanitizing the honors string on the way in, in the import.

Issue #1011, opened 2021-03-08
OH Microsite URL Redirects


MDiMeo:
Before we eradicate the microsite, we need to redirect oh.sciencehistory.org to the Oral History Collection landing page on Digital Collections, and make sure that links to oral histories (like from Wikipedia) go to the oral history record in Digital Collections instead.

jrochkind:
Back to backlog to remind us that we need to do a final data sync right before we shut down microsite and re-point it's DNS.

lsberry16:
I Googled site:oh.sciencehistory.org and identified some categories of links that I think are going to need manual intervention, or that we can just declare dead. I have this info in a Word doc if that's more helpful.

MISC:

https://oh.sciencehistory.org/oral-histories/projects
Redirect to new OH Projects page on main site, which I’m asking Clare to create

https://oh.sciencehistory.org/oral-histories/resources
This will be a dead link

https://oh.sciencehistory.org/oral-histories/access-levels
This will be a dead link


BROWSE ALPHABETICALLY:

For example: https://oh.sciencehistory.org/oral-histories/B
These links to lists of interviewees by initial letter of last name will be dead.


INTERVIEW-BY-SPONSOR

For example: https://oh.sciencehistory.org/interview-by-sponsor/54
I had forgotten these existed and to the best of my knowledge there are no live links to these anywhere in the wild. I assume they’ll be dead, and I believe that’s totally fine. I believe we have a creator field for “sponsor” now, so if we want to add a faceted search by sponsor in the future, we could do that.


PROJECTS:

https://oh.sciencehistory.org/project-name/69
redirect to https://digital.sciencehistory.org/collections/gt54kn818?f%5Bproject_facet%5D%5B%5D=Pew+Biomedical+Scholars

https://oh.sciencehistory.org/project-name/36
redirect to
https://digital.sciencehistory.org/collections/gt54kn818?f%5Bproject_facet%5D%5B%5D=Chemical+History+of+Electronics

https://oh.sciencehistory.org/project-name/38
redirect to
https://digital.sciencehistory.org/collections/gt54kn818?f%5Bproject_facet%5D%5B%5D=Mass+Spectrometry

https://oh.sciencehistory.org/project-name/43
redirect to
https://digital.sciencehistory.org/collections/gt54kn818?f%5Bproject_facet%5D%5B%5D=Scientific+and+Technical+Information+Systems

https://oh.sciencehistory.org/project-name/42
redirect to
[...]

PR #1052, opened 2021-03-30
Redirect legacy OH content

jrochkind:
Ref #1011 but this does NOT close that ticket, because we still need to actually generate the old-url->new-url mappings. There are a couple placeholders in the config file at present, that are proof of concept and show the format. 

This assumes we will point oh.sciencehistory.org at the app -- now the app has routing to recognize requests to thathost, and try to handle them. 

1. requests to home page will be redirected with a permanent redirect code to the Oral History Collection page, in digital collections at normal host. 
2. Request to known item URLs in our mapping file wil be redirected to those known items in digital collections at normal host. 
3. OTHERWISE requests to oh.sciencehistory.org not recognized will get a 404 page, with some custom helpful HTML. 

Case 3 is not a redirect, it's actually delivering content at the alternate host. I knew that was possible, but had never done it before. It ended up having all sorts of gotchas. I considered abanodning it, but, it is a neat solution, and it seems to work out okay for this simple page. Had to suppress all font-awesome icons (which ends up effecting footer), to avoid having to fight with CORS in this case, which was going to be hard to do reliably it ends up very deployment-specific config.

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -10,4 +10,23 @@ class StaticController < ApplicationController
 
   def policy
   end
+
+  def oh_legacy_url_not_found
+    Rails.logger.warn("Unknown legacy oral histories url: #{request.url}; referer: #{request.referer} ")
+
+    @original_url = request.url
+    @original_referer = request.referer
+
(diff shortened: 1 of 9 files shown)


eddierubeiz:
I've added code to the import that creates a `yml` file we can use (see reference above.)

Issue #1045, opened 2021-03-29
make within-collection results more compact


jrochkind:
results page should show you less of the collection metadata, to get results on the first page. 

We thought of when evaluating how to make "oral histories" collection page in Digital Collections work as well as microsite. But will apply to all D.C. collection pages.

PR #1053, opened 2021-03-31
Make Collection search results screen more compact

jrochkind:
Don't repeat collection description on the results screen, just show a brief header and results. Ref #1045. 

This is complete, and can be code reviewed, but shouldn't be merged until we get stakeholder feedback. Will leave in 'draft' until then.

### DIFF
--- a/app/assets/stylesheets/local/collection_show.scss
+++ b/app/assets/stylesheets/local/collection_show.scss
@@ -5,6 +5,11 @@
 
   .collection-top {
     display: flex;
+
+    .show-title h1 a {
+      color: inherit; // underline is enough, the bright blue is too much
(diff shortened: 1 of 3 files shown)


jrochkind:
@lsberry16 @MDiMeo @HKativa or anyone else, if the inside-a-Collection search results look like below, is it clear what's going on, and that you're looking at results limited to "search inside a collection"?  Feedback appreciated!

Screen Shot 2021-03-31 at 10 12 26 AM

------

Screen Shot 2021-03-31 at 10 09 31 AM

lsberry16:
I think that looks very good, and it's clear to me that the search was limited to the referenced collection.

Issue #995, opened 2021-02-26
Error in public view of unpublished oral histories


eddierubeiz:
` available_by_request_audio_count`, the method that counts how many assets on a work are "available by request", depends on being able to list the content type of each asset. This info isn't available before the files are characterized during ingest, resulting in an error (see  https://app.honeybadger.io/projects/58989/faults/77050523 ).

The public will never experience this issue: according to Hillary, the only way to reproduce this is:

"If you [meaning a logged in admin user] flip on "unpublished assets available by request" before the files are fully ingested [on an oral history] and look at the public view, you get an error. [...] Once all the files are fully ingested, the public view displays, no error."

For emphasis, a non-logged-in user who visits the public view of an unpublished work is of course denied access before this bug has a chance to occur.

jrochkind:
Agree that ideally we should make this fail gracefully instead of result in an uncaught exception. Even if it's a situation not expected to happen, it's clearly a situation that *can* happen, and thus a bug that it results in an uncaught exception. 

if it's not expected to happen much/ever, it doesn't necessarily need to do anything especially awesome, but still should not result in an uncaught exception.

eddierubeiz:
I agree, absolutely.

MDiMeo:
What are we waiting on here? Do we just need documentation or can we close this?

eddierubeiz:
This is very low-priority, but please don't close it. It's a real bug and we'd like to tackle it. It is not blocking our work on the microsite, however.

eddierubeiz:
```
URL:
https://digital.sciencehistory.org/works/42cy3cq

ERROR LOCATION:
    work_file_list_show_decorator.rb 
    line 68
    in available_by_request_audio_count()


CODE:
    def available_by_request_audio_count
        @available_by_request_audio_count ||= available_by_request_assets.find_all { |asset| asset.content_type.start_with?("audio/") }.count
    end


ERROR:
    NoMethodError: undefined method `start_with?' for nil:NilClass
```

jrochkind:
Capturing more of the top of the stack trace in case honeybadger deletes it for being older than our plan keeps. 

```
[PROJECT_ROOT]/app/decorators/work_file_list_show_decorator.rb:68 :in `block in available_by_request_audio_count`
[PROJECT_ROOT]/app/decorators/work_file_list_show_decorator.rb:68 :in `each`
[...]

PR #1055, opened 2021-03-31
Fix nil error.

jrochkind:
When audio is still in process of being ingested (or ingest fails), it may have a nil content_type. Make sure our code doesn't bomb then, a nil content type means it's not an audio file.

Ref #995


eddierubeiz:
Thanks for fixing this. It's bitten me at least once since we discovered it.

Issue #1056, opened 2021-04-02
Public OH with up-to-date combined audio does not show the audio player on public view.


eddierubeiz:
Compare the oral histories for  [Martin](https://digital.sciencehistory.org/admin/works/llyiryg) and [Duarte](https://digital.sciencehistory.org/admin/works/6yjgubm).

Only the Duarte shows the audio player in the public view; Martin should too, but does not.

Note that:
- Both are public
- Both have all assets public
- Both have up-to-date combined audio.

The only difference we can see is the Martin has a front matter PDF (there may be some other difference we missed).

jrochkind:
There are a few things going on here. 

The test for using the Audio template is:

    @work.is_oral_history? && has_oh_audio_member? && @work.oral_history_content&.available_by_request_off? 

For some reason the test is failing for Martin (llyiryg) Why?

First reason... in the staging data I copied, the audio file *is not published*. So it's correct that we can't show the audio player for a non-published audio. OK, but it looks like it is published in production. 

I think someone tried to switch this thing from being setup for restricted/by-request to being public, and some bug in our admin interfaces means the metadata wasn't fully switched over....

jrochkind:
OK, "available_by_request" mode is still "automatic". 

But since there are no non-published assets, the form for *changing* `available_by_request` mode isn't even shown, so you can't see what it is, or change it to 'off'. 

It needs to be set to 'off' to allow display with the audio player template. 

Potentially we can change the audio player template conditional to actually check for actual selected non-published assets... but potentially a performance issue. I'll look into it.

jrochkind:
Okay, yeah, the request mode was set to "automatic". But there was no way to see it or change it, because there were no non-published assets, so the logic thought it was unnecessary:

<img width="1000" alt="Screen Shot 2021-04-06 at 4 22 19 PM" src="https://user-images.githubusercontent.com/149304/113773800-7986dc00-96f4-11eb-90c3-0244a4ef56d2.png">

I can/have change it so it gives you that form, to let you switch it to "off", even if there are no non-published assets:

<img width="990" alt="Screen Shot 2021-04-06 at 4 22 03 PM" src="https://user-images.githubusercontent.com/149304/113773843-860b3480-96f4-11eb-9017-7c6b863947ab.png">

So you currently need to switch that to 'off' to get the template with the audio player, eg on Martin. 

[...]

PR #1063, opened 2021-04-07
Better visiblity/modifiablity of by_request setting

jrochkind:
Ref #1056

I'm not sure if this is sufficient, see comments on #1056. This is probably a good idea regardless.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -29,7 +29,6 @@ class WorksController < ApplicationController
   private
 
   def decorator
-    byebug
     @decorator ||= if @work.is_oral_history? && @work.oral_history_content&.available_by_request_off? && has_oh_audio_member?
       # special OH audio player template
       OhAudioWorkShowDecorator.new(@work)
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -29,7 +29,8 @@ class WorksController < ApplicationController
   private
 
   def decorator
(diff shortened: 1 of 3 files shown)


app/controllers/works_controller.rb
```diff
@@ -29,7 +29,8 @@ def viewer_images_info
   private
 
   def decorator
-    @decorator ||= if @work.is_oral_history? && has_oh_audio_member? && @work.oral_history_content&.available_by_request_off?
+    byebug
```
eddierubeiz, line 32: Stray `byebug`.

jrochkind: good catch thank you

If this passed tests, I guess tests didn't exersize this method, which is kind of weird too! Maybe I'll see if I can add a test here too.

jrochkind: Hmm, no, the test does exersize this method. I wonder how it passed CI?

jrochkind: If I look at CI, and look at the log... there are a bunch of byebug prompts in the logs! I wonder how the tests pass anyway, maybe after so many seconds `byebug` prompt automatically continues?

Would probably be good to make it raise in tests... but I intentionally use byebug in tests all the time... could make it raise in tests only when run on CI, there is an ENV variable github actions sets or could set. 

But for now I'll just remove the 'byebug.

eddierubeiz: Makes sense. Another way to approach this would be a github pre-commit hook that rejects commits with that string in them. I can't imagine a legitimate scenario in which we would ever want that string in our codebase.

PR #1071, opened 2021-04-13
New place for interviewee_biographies content

jrochkind:
attr_json oral_hisitory_content.inteviewee_biographies, is an embedded attr_json attribute, array of OralHistoryContent::IntervieweeBiography models. Attribute names have changed to no longer have `interviewee_` prefix, since they are embedded in a model named that.

We will be proceeding with trying to change app over to use this data model; doing a separate PR with just the data model so we can work independently on different parts of logic with regard to this. Still possible we will have to give up on this experiment too, or we'll find problems requiring alternate approach to this PR. This PR is a starting point so we can have the data model to both work on.

### DIFF
--- a/app/models/oral_history_content.rb
+++ b/app/models/oral_history_content.rb
@@ -55,6 +55,11 @@ class OralHistoryContent < ApplicationRecord
     succeeded: 'succeeded'
   }
 
+
+  ####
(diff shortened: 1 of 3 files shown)


eddierubeiz:
Great. Thank you! Let's do this. Merging.

PR #1079, opened 2021-04-14
index some oral history biographical info in works

jrochkind:
institution and birth country facets, only for use in Oral History collection. 

Ref #1067


jrochkind:
The controller code to reindex work... I'm not sure a reasonable way to write tests for. Thinking about it, might just skip it.

jrochkind:
OK, did some basic hacky controller specs. Ended up refactoring some tests a tiny bit in the process.

eddierubeiz:
Thanks for cleaning up those somewhat messy tests. Time well spent.

Issue #1083, opened 2021-04-19
[scihist_digicoll/production] ActionController::UnfilteredParameters: unable to convert unpermitted parameters to hash


jrochkind:
From https://digital.sciencehistory.org/catalog?search_field=all_fields&q=photography&range%5Byear_facet_isim%5D%5Bbegin%5D=&range%5Byear_facet_isim

## Backtrace

    line 409 of [PROJECT_ROOT]/app/controllers/catalog_controller.rb: catch_bad_blacklight_params
    line 24 of [PROJECT_ROOT]/app/controllers/application_controller.rb: block in batch_kithe_indexable
    line 23 of [PROJECT_ROOT]/app/controllers/application_controller.rb: batch_kithe_indexable

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/78932313)

PR #1084, opened 2021-04-19
can't count on bad params[:range] having a #to_param

jrochkind:
Ref #1083


eddierubeiz:
Thank you for addressing this!

Issue #1080, opened 2021-04-14
Citation for OH interviews of two people use only one interviewee in the citation


eddierubeiz:
Our citation for the OH [interview of  Isabella and Jerome Karle](https://staging-digital.sciencehistory.org/works/wp988m037) reads:
`Jerome Karle, interviewed by James J. Bohning in Naval Research Laboratory (U.S.) on September 9, 1987. Philadelphia: Science History Institute, n.d. Oral History Transcript 0066. https://staging-digital.sciencehistory.org/works/wp988m037.
`
The citation should mention both interviewees.
_Low priority._

jrochkind:
Incidentally, the library catalog doesn't get this one right either. 

Not sure how to link directly to the library catalog, but it reads:

> Cite As: Isabella Karle interviewed by James J. Bohning and David Van Keuren at the Naval Research Laboratory. (Philadelphia : Chemical Heritage Foundation, Oral History Transcript 0066.)

It's got both interviewers, but it's missing that Jerome Karle is an interviewee.

PR #1088, opened 2021-04-19
Fix citatable attributes formatting for Oral History with multiple interviewers/interviewees

jrochkind:
Ref #1080


eddierubeiz:
Thanks for fixing this! (And for clearing up the spacing.)

PR #1090, opened 2021-04-21
URL mappings from OH microsite to digital collections.

eddierubeiz:
We're probably missing about a dozen of these, but the overwhelming majority of the URL redirects are in the list.
This should allow us to at least start testing the redirect.


jrochkind:
Oh, I've already tested it, but good to have more complete!

Note if you want to test on dev, you can set environmental variable `ORAL_HISTORY_LEGACY_HOST` to whatever, so you can use that in your /etc/hosts instead of the actual `oh.sciencehistory.org`. 

What's the explanation for the missing ~dozen? Not yet in Digital Collections? Other?

jrochkind:
Ref #1011

eddierubeiz:
The remaining few I would call "miscellaneous outliers". Lee, Hillary, Annabel, and Nicole have been collaborating since last week on clearing them up. I defer to @lsberry16 for details, but my sense is we're on the final stretch!

eddierubeiz:
(examples: at least one never actually made it either into the OPAC or onto the library shelves, so wasn't on the initial list of interviews to moved; a couple are actually still in progress per the regular OHC workflow, and so on).

jrochkind:
OK, so the missing ones from this mapping are the ones that haven't made it to Digital Collections at all yet, for whatever reason? Cool.

eddierubeiz:
Correct. That and 3 (I believe) remaining identifier mismatches which are actually already cleared up in production but haven't propagated down to my local machine where I actually match up all the data.

Issue #1076, opened 2021-04-14
Improved UI widget for assigning model associations


jrochkind:
Sometimes we need to assign a model object to be linked to others -- a Work to one or more Collections; or an oral history to an Interviewer Profile; an oral history to a Interviewee Biography. 

Currently, the only UI device we have for letting staff make this assignment is giant HTML select menus listing every possible option. 

![Screen Shot 2021-04-14 at 1 40 11 PM](https://user-images.githubusercontent.com/149304/114754799-f8989780-9d26-11eb-8b40-41f63148e281.png)

![Screen Shot 2021-04-14 at 1 39 55 PM](https://user-images.githubusercontent.com/149304/114754801-f8989780-9d26-11eb-872e-5913f3cbc292.png)

This isn't a great UI. It becomes pretty unworkable once the number of things in the list extends past a couple dozen. In the case of Interviewee Biographies, it's hundreds. And if there's inconsistency between `firstname lastname` and `lastname, firstname`, or you don't know exactly how something was entered, it can be really difficult to find what you want in the list. 

We should develop a better UI, probably involving entering a query on the page and getting a list of matches to click on. This will involve some Javascript. 

Looking around for suitable Javascript libraries, this one seems to me a possible good choice -- modern, doesn't hrequire JQuery (we'd like to move towards eliminating JQuery dependency), maintained, based on earlier legacy things. https://tom-select.js.org/

If we add tom-select.js as a dependency, could we also use it to replace the current auto-complete controlled vocabulary javascript, to reduce our JS dependency size a bit? Maybe.

PR #1091, opened 2021-04-21
add tom-select.js filtered multi-select for large-element association select inputs

jrochkind:
Ref #1076 

tom-select JS and CSS added via npm/webpacker, in the "admin" pack that's added only to admin layout.

For the moment, we are NOT doing any AJAX loading here (although tom-select supports it) -- we are still loading all select options in delivered HTML, just using tom-select for a much nicer UI for it.

Used for:

* Work: select collection
* Work/Oral History: select linked interviewer profile
* Work/Oral History: select linked interviewee biography

### DIFF
--- a/app/javascript/packs/admin.js
+++ b/app/javascript/packs/admin.js
@@ -11,3 +11,5 @@ import '../src/js/admin/simple_uppy_file_input';
 import '../src/js/admin/uppy_dashboard.js';
 import '../src/js/admin/qa_autocomplete.js';
 import '../src/js/admin/queue_status_submit.js';
+
+import '../src/js/admin/tom_select.js' // has CSS
--- a/app/javascript/src/js/admin/tom_select.js
+++ b/app/javascript/src/js/admin/tom_select.js
@@ -1,0 +1,31 @@
+// use https://tom-select.js.org/ for fancy auto-complete text/select boxes.
+
+// tom-select does not require JQuery, and we use no JQuery in this file.
+
+import 'tom-select/dist/css/tom-select.bootstrap4.min.css';
+
+// Not sure how I figured out these paths to load tom-select base and then selected plugins
+// from npm tom-select... but here it is. See: https://github.com/orchidjs/tom-select/issues/57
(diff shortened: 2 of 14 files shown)


jrochkind:
So for what it's worth, for interviewee biography names and interviewer profile names: 

Right now we're loading the whole list of all of them in on the Oral History tab (which actually since it's just a tab is just part of the general Work admin page). 

tom-select would let us move to an AJAX load-on-demand-as-typing-filter, where we don't load the whole list on on page delivery. We're not doing that yet though. 

If we did do that, it looks like it would save 12KB from a currently 125KB page (~10% of page size). So it's definitely a significant amount -- of an already large page, since that multi-tabbed work admin page has a lot on it, all delivered in initial response from server. 

I wonder if I should go ahead and do that now, implement the AJAX load? Not sure. It might be good enough as is.

eddierubeiz:
Noted. It occurs to me that for the average OH work, the"searchable transcript text" also contributes a lot to the size of the HTML.

jrochkind:
Very true, good point. 

That's not a great UI either with regard to HTML DOM size, but it is what was asked for originally -- at that time, I think I still had the unrealistic idea that the oral hsitories with "searchable transcript text" would be the minority edge cases, most would be OHMS-ified. 

Hmm.

jrochkind:
OK, those numbers were wrong too, they were post-gzip numbers -- I got them from Chrome Dev Tools, so they were the difference in *actual* sizes of response, after gzipping that our server does. (We do have a big page). 

Some more careful numbers:

For: https://staging-digital.sciencehistory.org/admin/works/z029p571v

* uncompressed size of "searchable transcript text": 260KB
* uncompressed size of HTML for complete "interviewer profiles" menu: 4KB
* uncompressed size of HTML for complete  "interviewee biographies" menu: 36KB
* total uncompressed size of whole HTML (including full menus) page: 350KB

So I'd say if we care about page size, we should take "searchable transcript text" out. And I think I'll make a GH issue for that. 

interviewer profiles is neglibiel.   interviewee biographies is enough to think about, but agree that don't need to bother with it for now.

Issue #1086, opened 2021-04-19
Create hover over text for Recently Added


MDiMeo:
Use the title (or part of the title, if it's long?) in a box when you hover over an image on the home page under Recently Added.

jrochkind:
We should ALSO add an `alt` attribute with same text, for accessibility.

PR #1092, opened 2021-04-21
Adding tooltips and alt text to the homepage recent items selection.

eddierubeiz:
Ref #1086
Should include tests for the viewmodel that displays the thumbnails


jrochkind:
Hm, I woudln’t say it was “specifically designed” not to have alt text. There isn’t much “design” in it, it’s pretty easy to give it alt text.

The main thing ThumbDisplay does that’s tricky is do lazy loading, reserving the correct amount of space on the page pre-load. While we aren’t currently using lazy loading for Recent Items… it doesn’t seem like there’s any reason we’d never want both lazy loading and an alt tag, it’s not like these are fundamentally incompatible features or anything. So why not use ThumbDisplay?

If we instead go this way, you can get lazy-load OR you can get alt text, but you can't get both... and it doesn't seem to me like there's anything incompatible with ThumbDisplay's design to just add optional alt text -- but I might not be realizing something!

app/javascript/src/js/recent_items_tooltips.js
```diff
@@ -0,0 +1,3 @@
+$( document ).ready(function() {
+    jQuery('[data-toggle="tooltip"]').tooltip();
+});
```
jrochkind, line 3: there's nothing about this that is specific to recent_items, right? It's activating everything with data-toggle tooltip. 

In fact, I'm surprised we aren't already doing this somewhere else? Weren't we already using tooltips?  Oh, I see, ohms_footnotes.js does it in a... customized way. 

The ohms_footnotes.js code is also triggered simply on `jQuery('[data-toggle="tooltip"]')`, but then does custom setup involving container and delay -- I think we want to be more careful these two areas don't collide with each other -- they're both going to try to operate on all data-toggle=tooltips, doing different things -- and they are both loaded on all pages!

Either the ohms one can be triggered by `data-toggle=ohms-tooltip`, or it should apply to only tooltips within an ohms container, or something. 

And generic `jQuery('[data-toggle="tooltip"]').tooltip();` should be in a file just called `js/bootstrap_tooltips_activate` or something, as it is not at all restriccted to recent items.

eddierubeiz: Yup -- narrowed this down so it is only applied to the recent items. (Code is still executed on all pages, though.)

jrochkind: @eddierubeiz the problem is that the existing JS for OHMS will still apply to recent items too.... the recent items still get both JS's trying to apply to them. 

I think you're going to need to put something other than data-toggle=tooltip on *one* of them to avoid having two sets of code apply to one of them. Since this is the one that is a totally generic tooltip, I'd leave this one data-toggle=tooltip, and put something on the OHMS taht makes this one *not* apply to it, only the special OHMS tooltip setup.

eddierubeiz: Good call . The `ohms-tooltip` is a great idea in particular.

jrochkind: Maybe even move em both to the same JS file, since they have such possibility to potentially conflict with each other, they could be right next to each other so the dev can see that, instead of needing to remember "oh yeah I did other tooltips in that ohter file taht might conflict..."

eddierubeiz: Hmmmmmmm yeah good idea.

eddierubeiz:
Decided after all to *not* merge the two activation js files into one, as the ohms footnotes code really belongs in a separate file with other non-tooltip ohms code.
I did move the files around and add comments to make it much harder to get confused about the fact that we have two separate boostrap `.tooltip()` invocations now.

jrochkind:
looking good, thanks!

PR #1095, opened 2021-04-22
tests for case insensitive OH legacy url redirect

jrochkind:
One marked pending implementation. Lets us commit the test now marked pending, nice.


eddierubeiz:
Great. Will remove the 'pending' in due course.

Issue #1098, opened 2021-04-26
Fix OHMS facet


jrochkind:
Should be "synchronized audio" and include transcript and/or table of contents; currently "synchronized transcript" and does not include ones that only have ToC and not sync transcript.

PR #1099, opened 2021-04-26
OHMS facet value should be 'synchronized audio', and on if ohms transcript OR ToC

jrochkind:
Ref #1098


jrochkind:
Will of course require a re-index to take effect.

Issue #1094, opened 2021-04-22
Redirect map legacy OH direct to PDF urls


jrochkind:
There are direct-to-pdf links to old OH PDFs that I find in Google and Wikipedia. Eg:

    https://oh.sciencehistory.org/sites/default/files/manning_rj_0239_suppl.pdf

We didn't intend these ever to be direct links, they skip the request form. :( But they are out there. 

While we were redirecting the "correct" HTML OH lists from old system, we weren't planning on catching these "illegal" direct to PDF links to redirect to new system... but they are out there in the wild quite a bit, and since we seem to have some extra time... 

Maybe we can catch all those to redirect to proper location in new system too? Proper location is of course not a direct PDF link, but just (eg for manning) https://digital.sciencehistory.org/works/pz50gx11n  Then all the people who click on those from google or wikipedia or something would at least still get to the OH. 

Would be some extra work to extract all those urls from the old system. But may be as high priority as anything else we have left in our remaining time? @MDiMeo @lsberry16

lsberry16:
That would be an excellent thing to do, IMO.

eddierubeiz:
I'm going to start a new PR to make these match incoming URLs regardless of capitalization.

eddierubeiz:
And that PR should also include turning on the test (currently marked `pending`) that Jonathan created in #1095. After that is merged, I believe we can close this issue.

jrochkind:
@eddierubeiz The beauty of the rspec "pending" feature is it will actually start *failing* if it... passes. Marking it "pending" reverses the expectation, and rspec is like "Hey, you marked this pending but now it's passing, you better change it to no longer be pending!"  So rspec will help us keep track of that, should lead you to it.

eddierubeiz:
I had no idea. You learn something every day!

PR #1100, opened 2021-04-26
Redirect PDF URLs to the appropriate URLs in the digital collections.

eddierubeiz:
Ref #1011.

This is JUST a list of URLs.

Note: The URLs are still not matched in a case-insensitive way (see https://github.com/sciencehistory/scihist_digicoll/pull/1095 for a pending test for this.)


jrochkind:
Amazing and exciting!

eddierubeiz:
Ref #1094

Issue #731, opened 2020-04-28
Thumbnail image for sharing OH content on social media


lsberry16:
Apparently we don't have a thumbnail image assigned to our oral history content on the Digital Collection for when we share it on social media (or at least Facebook). Do we want to select something other than the default institution logo that shows up now?

jrochkind:
Lee writes in Slack:

> I don't HATE that, but better would be nice. I'm not sure what we pick or what goes into making it the image that goes along with our oral history content. I filed a GitHub issue and we can figure it out. Thanks!

"we" can probably do whatever we want, i"m not sure what is best. Things i can think of are:

1. A better Science History Institute logo. The one it's taking now is white on transparent background, and is just something facebook found that our page is using as a 'watermark' style on the banner, it wasn't meant to be used like this!

2. A logo representing the Center for Oral History. I'm not sure the Center has it's own logo though? A graphic of some kind could be made. 

3. Any sort of graphic we want to represent "oral histories". Compare to the photo used on the [Center web page](https://www.sciencehistory.org/center-for-oral-history) or [oral history digital collection](https://digital.sciencehistory.org/collections/gt54kn818).  Some variation of that formatted to proper size and aspect ratio for facebook could be used. Or we could find appropriate staff (internal or external) to create a different illustration, whether a photo illustration or icon. Or we could find some existing free or licensed image to use. 

Alternately, we could get into a more sophisticated feature that let individual per-oral-history thumbnails be uploaded. Say a portrait of the subject, or other relevant historical photo. 

This would be a more complex feature to develop, especially if we'd want to use that custom per-OH thumb internally to the digital collections app too. It would also require more work on the metadata side, as someone would have to find/select and upload such a thing. I'm not sure we want to create more work on the metadata side, when there is already a large backlog. It might be better to use a generic thumb for all OH's. 

The idea of a per-OH thumb does relate to an idea in our feature brainstorm, to support ingesting graphical content associated with an OH, such as portraits. If we wanted to pursue that for reasons other than thumbnails too, it might make sense to use it for thumbnails as well.

jrochkind:
What it looks like now when shared on facebook:

![Screen Shot 2020-04-28 at 2 29 20 PM](https://user-images.githubusercontent.com/149304/80532967-e9c8a180-896a-11ea-8485-13992686484c.png)

[...]

PR #1101, opened 2021-04-26
og:image social media share image for OH without representative

jrochkind:
Use that Oral History Center thumb we use, with recorder on stack of bound volumes

Ref #731


jrochkind:
We should see if we can get a higher resolution version of this thumb, some social media wants a bigger image.

PR #1103, opened 2021-04-27
OH legacy url redirect is now case-insensitive.

eddierubeiz:
The mapping file (which will never change again after the import) is guaranteed to be all lowercase.
(Try it! `grep '[A-Z]' config/oral_history_legacy_redirects.yml` )
Thus, we can just lowercase the incoming mixed-case URL before looking up the new URL.
Ref #1011


jrochkind:
I expected to see the mapping file change in this PR too; and am bewildered by the tests passing *without* it being in this PR.  Can you explain?

jrochkind:
The answer seems to be that *master* is currently failing, before this PR... how did we let that happen?  I'm very confused.

jrochkind:
OK, this is disturbing. On MASTER, 

`rspec ./spec/requests/oh_legacy_redirects.rb` fails. 

But `./bin/rspec` succeeds. 

Perhaps it is not including "request" specs by default? that's not good!

And it looks like a past update to the `config/oral_history_legacy_redirects.yml` file maybe did break the `./spec/requests/oh_legacy_redirects.rb`  spec, but CI didn't catch it? All very weird.

jrochkind:
Aha, cause `rspec ./spec/requests/oh_legacy_redirects.rb` doesn't end in `_spec.rb`, and it's supposed to! Drat! My fault. Those tests were never being run by CI. 

It's good to verify tests actually *fail* before fixing them to pass. :(

eddierubeiz:
No problem! Will rename the file and put it under the microscope.

eddierubeiz:
I renamed the test file. It fails with the old code and passes with the new code. All is well.

eddierubeiz:
I added one more path to the redirect file, btw.

PR #1104, opened 2021-04-28
Keep long title from pushing 'search' box to next line on full width layout

jrochkind:
This CSS is a bit convoluted! But this seemed to fix it. Tested at multiple widths and compared to production to make sure it didn't seem to mess up anything else. 

before:
Screen Shot 2021-04-28 at 11 02 24 AM

after:

Screen Shot 2021-04-28 at 11 19 51 AM


eddierubeiz:
Thanks!

PR #1105, opened 2021-04-28
preserve tab selection in url as eg #tab=ohTranscript

jrochkind:
We add tab selection to URL #anchor on selection, and on page load switch to that tab. This lets tab selection be reflected in the URL, so when you bookmark you get current tab. Oral History audio navbar is the use case, although it's active on all bootstrap tabs.

We were already doing this prior to this PR, but putting the ID of the selected tab in teh anchor as the entire anchor, eg "#ohTranscript"

we now use a key/value format instead, like "#tab=ohTranscript".  This gives us room to store more than one thing in anchor, in particular upcoming planned feature for linking specifically to a timecode, which might look like #tab=ohTranscript&t=1212" where 1212 is a number of seconds.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -83,7 +83,7 @@ class Admin::WorksController < AdminController
   # PATCH/PUT /admin/works/ab2323ac/submit_ohms_xml
   def submit_ohms_xml
     unless params[:ohms_xml].present?
-      redirect_to admin_work_path(@work, anchor: "nav-oral-histories"), flash: { error: "No file received" }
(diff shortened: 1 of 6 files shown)


jrochkind:
OK, I wasn't totally sure about this change, is it just moving deckchairs around, or even is it a step in the wrong direction?

But as I move forward with implemeting #655, I think this makes sense. It gets weird if sometimes you are trying to edit things in a #fragment with key/value style (#t=12121), but other times you are just using ordinary `#identifier` style. I think. 

Marking this no 'ready for review'. But feedback/thoughts welcome.

eddierubeiz:
I think this is fine!

Issue #696, opened 2020-04-09
OHMS Front End: Scroll behavior when switching tabs while audio is playing (wishlist item)


lsberry16:
This changed to a 'jump to text' button. 

----
It would be extra spiffy if, when you switch to the TOC or the Transcript tab **while the audio is playing** it drops you as close as it can to where you are in the audio.

Jonathan notes: That’s a pretty complex feature. It might or might not be possible to make it work well with the one or two minute timecode granularities we have. It’s not going to be done for this initial release, but we could consider it in the future.

jrochkind:
@MDiMeo @lsberry16  as I start implementing this... I'm honestly still a bit worried about the usability of it. I think users might find themselves confused about why the scroll is jumping around, find it jumping when they don't want it to and not jumping when they do want it to. 

I wonder if we what *really* want is a checkbox "scroll to match audio" which would try to constantly scroll along in transcript and/or ToC (harder to implement), or a button "jump to audio now" (easy but not sure where to put it on screen) -- tying this behavior to tab-switch when audio is playing (not paused) just doesn't seem right to me...

But I could be wrong, Lee may understand the use case better!

Either way,  I'll try to get it implemented and then we can look at it! I will suggest we spend some time examining it though before just making it live.

PR #1106, opened 2021-04-28
CONSIDERATION: Oral history audio player, scroll to media sync button

jrochkind:
Possible solution to #696. But we might try an "auto-scroll" instead. 

@lsberry16 did confirm this is already probably preferable to what's speced in #696

Still some outstanding concern about what the button is labelled, currently "scroll to media".

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -28,5 +28,6 @@ import '../src/js/audio/ohms_search.js';
 import '../src/js/audio/ohms_footnotes.js';
 import '../src/js/audio/accordion_open_on_screen.js';
 import '../src/js/audio/navbar_tabs.js';
+import '../src/js/audio/jump_to_text.js';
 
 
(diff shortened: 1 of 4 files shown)


jrochkind:
@eddierubeiz Ready for review. Even with a basic capybara smoke test. 

My first time trying to use "ESM modules" (export/import) to share reusable utility code in webpacker. I plan to re-use some of those utilities in "link to timecode".

jrochkind:
Hmm, currently seems to actually be broken for me, at least jumpToToc!  Needs more testing.

jrochkind:
This stuff is tricky enough and not in my wheelhouse enough that maybe could use some people banging on it and trying it out and trying to break it and seeing if it works as they expect.

eddierubeiz (approved):
Formally approving this. I tested it in staging too and the functionality looked great to me. Merge whenever you want.

PR #1107, opened 2021-04-28
update webpacker to 5.3.0

jrochkind:
* bundle update webpacker, which modifies Gemfile.lock
* edit package.json manually to specify ^5.3.0, then `yarn install` to modify yarn.lock
* yarn upgrade webpack-dev-server --latest, makes further changes to yarn.lock

https://github.com/rails/webpacker#upgrading

This upgrade may get rid of some of the JS dependency warnings we're getting.


eddierubeiz:
Thanks for documenting how you got this. Merging.

PR #1113, opened 2021-04-29
Don't create IntervieweeHonors if the only metadata is a date.

### LINKED ISSUE #1109: Stray interviewee honors records
Due to a presumed bug in Drupal, a number of interviews have "interviewee honors" records with only a date.

Examples:
- https://oh.sciencehistory.org/node/7656/edit has a 2018 "honor" -- no end date, no description, just "2018".
- https://oh.sciencehistory.org/node/7656/edit also has a 2018 "honor". Ditto.

These are duly imported and displayed in the digital collection, of course; there's nothing wrong with the import code per se here.

Proposed solution: once the import is complete, we write a simple piece of code (a rake task) that iterates over the oral histories and remove these bogus "honors" records automatically.

eddierubeiz:
They are bogus and an artifact of some obscure Drupal bug.
Fixes #1109

### DIFF
--- a/lib/scihist_digicoll/oh_microsite_import_utilities/updaters.rb
+++ b/lib/scihist_digicoll/oh_microsite_import_utilities/updaters.rb
@@ -166,6 +166,7 @@ module OhMicrositeImportUtilities
     end
 
     def self.honors(w, rows, transformations: nil)
+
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Taking yr advice @jrochkind  -- just removing these in the import.

Issue #1123, opened 2021-05-05
redirect existing links to OH searches in new sites


jrochkind:
For basic cases, this is easy to do, so why not. So if there are OH "search forms" for legacy site, they can keep working redirected to OH collection in new site. The basic approach won't support fielded-searches or facets, just simple search redirects. 

There are several variant URLs in legacy site. 

* https://oh.sciencehistory.org/search/site/query (we direct to there from current DC, perhaps incorrectly?)
* https://oh.sciencehistory.org/search/oh/query (main site uses this in it's "search" function; also the "full text search" on legacy site homepage itself)

specific fielded searches on legacy site, can redirect just to general search? May not be necessary, may not be used elsewhere. Not sure why two different patterns. 

* https://oh.sciencehistory.org/search/honors/querystring
* https://oh.sciencehistory.org/oral-history-name-search?title=querystring  [interviewee name search]
* https://oh.sciencehistory.org/interviewer-search-results?field_display_title_value=query [interviewer name search]

PR #1124, opened 2021-05-05
Redirect some existing legacy OH microsite search result links

jrochkind:
Only bothering with the main searches, not fielded searches. Will catch things like the search form on main sciencehistory.org website, if it's left as is. Final decision hasn't been made wrt what to do with OH search on main site; but it's so cheap/easy to do this redirect, let's do it anyway, just for completeness in redirects. Either it'll be used or it won't, doesn't hurt anything by being there.

Maybe i just think setting up redirects in the routing file is fun.

Ref #1123


eddierubeiz:
Hey, you know what, why not. If we get zero hits in Google Analytics, we can always take these out in a few months.

jrochkind:
Seems very unlikely we're going to remember or want to spend time seeing if we get hits on google analytics and then taking these out. But they don't do any harm.

eddierubeiz:
Not disagreeing with you.

PR #1128, opened 2021-05-06
Add oral history projects redirect

eddierubeiz:
Includes two config settings, so we can change the URL easily if there's a website rename or redesign.
Ref #1011  .
This is the last URL we know about.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -30,6 +30,7 @@ Rails.application.routes.draw do
       "#{ScihistDigicoll::Env.lookup!("app_url_base")}/collections/#{ScihistDigicoll::Env.lookup!("oral_history_collection_id")}?q=#{path_params[:query]}"
     }
 
+    get '/oral-histories/projects', to: redirect { 'https://www.sciencehistory.org/oral-history-projects' }
 
     # Does it match one of our OH_LEGACY_REDIRECTS? Then redirect it!
     get '*path', constraints: ->(req) { OH_LEGACY_REDIRECTS.has_key?(req.path.downcase) }, to: redirect { |params, req|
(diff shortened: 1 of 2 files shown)


jrochkind:
I would argue definitely against an env variable for `institute_main_site_projects_page` -- the only place this appears is in the redirect list, it can easily be changed in the redirect list, no need to add indirection with a config variable. 

I am suspicious of a config for `institute_main_site_hostname` too -- I don't see our website changing to be something other than `sciencehistory.org` without a name change, and if it does -- we're going to have so many things to change this tiny little place isn't the half of it, we will certainly be grepping the codebase for the hostname.  This seems like complexity for the sake of complexity. 

Also, our website is `sciencehistory.org`, not `www.sciencehistory.org`. If you try putting in the latter, it just redirects me, are you getting different results? 

I'm thinking keep it simple, it's just a list of redirects, you know?

eddierubeiz:
OK ! It's just a plain string now (my initial commit) but I'm using sciencehistory.org instead of www.sciencehistory.org. (Good catch.)

config/routes.rb
```diff
@@ -30,6 +30,7 @@ def self.matches?(request)
       "#{ScihistDigicoll::Env.lookup!("app_url_base")}/collections/#{ScihistDigicoll::Env.lookup!("oral_history_collection_id")}?q=#{path_params[:query]}"
     }
 
+    get '/oral-histories/projects', to: redirect { 'https://sciencehistory.org/oral-history-projects' }
```
jrochkind, line 33: Cool cool. Oh right, i get it now -- this can't be in our existing yml because our existing yml assumes internal link. 

OK, fine, keep it simple, it's fine!

In this simple case with just string literal though, the redirect can be written as:

    get '/oral-histories/projects', to: redirect('https://sciencehistory.org/oral-history-projects' )

param, not a block. 

https://guides.rubyonrails.org/routing.html#redirection

jrochkind:
We COULD add a test for this one, but I'm just gonna merge it.

Issue #864, opened 2020-09-21
Fix S3 derivative "orphan" file checker, and run checkers regularly as cronjob, with report


jrochkind:
We had created a tool that checks through all files on S3 to find any that *aren't* referenced by our DB. This is a bit cumbersome to do (and has some cost to S3 every run, although probably trivial). But seemed good. 

But we weren't running it regularly. And the derivative file checker specifically has gotten out of date in it's logic, and reports some false positives with the combined audio derivatives being stored in different file paths than the orphan checker expects. 

It could be fixed to not mis-attribute the combined audio derivaitves as orphaned; and also to actually check the combined audio derivatives for real orphans separately. 

Additionally, we probably ought to be running these checkers (not just derivatives but originals too?) regularly, and reporting on things found (both in dashboard and in notifications of orphaned files found?) 

This becomes more important with our work to have public vs restricted derivatives, as oprhaned files could hypothetically be restricted/confidential info that ought not to be on a public bucket. 

This is kind of convoluted stuff, and could be a  week or two to implement.

eddierubeiz:
I've been working on the first part of this: making the orphan derivative checker detect and delete combined audio derivatives correctly.

eddierubeiz:
I split this into 3 sub-tickets, as it's pretty broad.

eddierubeiz:
Yeah, I'm going to close this.

PR #1131, opened 2021-05-11
Handle combined audio derivatives in orphan_s3_derivatives checker.

eddierubeiz:
Ref #1140, which is part of #864.

- No longer flags combined audio derivatives as orphans.
- Correctly detects and deletes orphaned combined audio derivatives.
- Includes a test for the orphan derivative checker.


app/services/orphan_s3_derivatives.rb
```diff
@@ -121,9 +145,38 @@ def parse_s3_path(s3_path)
   end
 
 
-  def orphaned?(asset_id, derivative_key, shrine_id)
-    return true unless asset_id.present? && derivative_key.present?
+  # Combined audio derivatives are stored at an s3_path of:
+  # combined_audio_derivatives/9d3114be-fb66-43a8-9c25-2600aa4ea7b0/combined_be68ab9e9d1ef65e8f4bdf38586db718.webm
+  # The first chunk is parsed by parse_s3_path as the 'asset_id'.
+  def combined_audio_derivative?(asset_id)
```
eddierubeiz, line 151: this doesn't have to use `asset_id` as the argument name. Will fix.

app/services/orphan_s3_derivatives.rb
```diff
@@ -121,9 +145,38 @@ def parse_s3_path(s3_path)
   end
 
 
-  def orphaned?(asset_id, derivative_key, shrine_id)
-    return true unless asset_id.present? && derivative_key.present?
+  # Combined audio derivatives are stored at an s3_path of:
```
eddierubeiz, line 148: This comment is confusing. What I meant is that s3 paths for c.a.d's start in an invariable string `combined_audio_derivatives`.

jrochkind:
I ran this on staging. Even going so far to run the `delete` version. It seemed to work right. 15K orphaned derivatives found, we don't know why there are quite so many on staging, but figuring that out is not this ticket. I spot checked some to make sure they really were orpahned, all checked out. No combined derivs showed up as orphaned as far as I could tell scanning thorugh the list.

eddierubeiz:
No, you definitely removed some combined derivs.
Compare [production](https://s3.console.aws.amazon.com/s3/buckets/scihist-digicoll-production-derivatives?prefix=combined_audio_derivatives%2F0e84c870-0307-4714-99a0-1bde3134cf6f%2F&region=us-east-1&showversions=false#) and [staging](https://s3.console.aws.amazon.com/s3/buckets/scihist-digicoll-staging-derivatives?prefix=combined_audio_derivatives%2F0e84c870-0307-4714-99a0-1bde3134cf6f%2F&region=us-east-1&showversions=false#).

eddierubeiz:
Going to merge this after the tests pass.

jrochkind:
> No, you definitely removed some orphaned combined derivs.
Compare production and staging.

Huh, I scanned the logs of what was being removed, and didn't notice any. But I guess you must be right? Anyway, fair enough -- it's hard to keep track of exactly what's going on with 200K files.

PR #1132, opened 2021-05-11
URLs to timecodes and ToC segments in OHMS

jrochkind:
Still to do:

-  mini SHI icon on navbar (decided to move this to separate ticket)
- [x] link to timecode should take you to transcript at that point too
- [x] 'share' should be button on description with other share icons, not tab
- [x] stop delayed switch to autoplay
- [x] specs for jump to timecode/segment

Ref #655


jrochkind:
@eddierubeiz there's one small part of this left (see checkboxes in description), but it's a monster of a PR, so feel free to begin reviewing whenever you have time. It should be fairly stable at this point just awaiting a small thing.

eddierubeiz:
OK - I did play around with this in dev, and looked over the code on Friday and again today. Looks good; merging. Per today's meeting, if we need to adjust it more, we can make a separate PR for that.

Issue #1085, opened 2021-04-19
[scihist_digicoll/production] TypeError: no implicit conversion of ActionController::Parameters into String


jrochkind:
Bad data not being caught appropriately. 

https://digital.sciencehistory.org/catalog?f%5Bdepartment_facet%5D%5B%5D%5B%3DLibrary%26q%3D%5D=

## Backtrace

    line 15 of [PROJECT_ROOT]/app/views/catalog/index.html.erb: _app_views_catalog_index_html_erb___4283244947029360353_47412569147380
    line 24 of [PROJECT_ROOT]/app/controllers/application_controller.rb: block in batch_kithe_indexable
    line 23 of [PROJECT_ROOT]/app/controllers/application_controller.rb: batch_kithe_indexable

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/78909879)

PR #1138, opened 2021-05-13
handle two more kinds of bad query param that results in Blacklight uncaught exception

jrochkind:
Ref #1085

and 

https://app.honeybadger.io/projects/58989/faults/79191107

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -409,6 +409,14 @@ class CatalogController < ApplicationController
       render plain: "Invalid URL query parameter range=#{params[:range].to_s}", status: 400
     end
 
+    # escaped newline in range value, eg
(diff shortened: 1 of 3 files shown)


jrochkind:
I ended up generally refactoring "bad query param" catching code to be more: We know the "shape" the data should be, make sure it is that shape (a hash of hashes of strings, or whatever). 

Instead of previously it was more like "we know X and Y and Z pattern of bad params, make sure it's not that."

This I think should more generally catch the bad params and hopefully help us avoid this continuing to happen with slightly different kinds of bad data. 

Our pretty good test coverage of this made me comfortble that the factor should still be catching all previous kinds of bad data, without interfering with good requests.

eddierubeiz:
Great. Love the test coverage, yeah. Eager to get this into prod.

PR #1143, opened 2021-05-17
Fixing two test files with incorrect names

eddierubeiz:
Ref #1144

Files in spec directory that contain rspec tests should end in `_spec.rb`.
Two files were not being run as part of our test suite because they were incorrectly named, and have fallen out of date and stopped passing.


jrochkind:
Wow, good catch. And these tests were not ever being run in CI accordingly... and seem to actually fail?

What should we do with this failing PR? I guess we should make an issue to fix these tests?

eddierubeiz:
Yeah, I'll create an issue.

eddierubeiz:
I'm not going to do any more work on this until I have a functioning copy of `vipsthumbnail` on my dev box, which might be a couple weeks, but that's totally fine, as this isn't urgent.

jrochkind:
@eddierubeiz we should probably finish this, either you or me?

eddierubeiz:
This is ready to go, finally.

eddierubeiz:
I'm merging my own PR; not going to make a habit of this but the circumstances are out of the ordinary.

jrochkind:
I'm fine with merging for urgent things, but this doesn't seem that urgent to me?

Before it was checking for calculation of sha256... I guess we just stopped calculating that for derivatives on a refactor?  Did you check to make sure there wasn't a checksum fingerprint somewhere else in the derivative metadata? If there is, I'd still like the test to ensure it.

eddierubeiz:
Clarification: that sha256 was only there as a workaround for an obscure problem which I believe you and/or a new version of Shrine solved long ago. The problem, as I recall, made it impossible to create derivatives without explicitly assigning the sha256 metadata on the derivative file.

By using the `asset_with_faked_file` asset factory trait, the problem went away, and thus the need for the workaround.

The sha256 wasn't being checked before -- it was only there to allow us to test derivative creation.

The only thing this test ever checked was a) the number of derivatives created from a pdf; b) the names of the associated keys; and 3) the width of two of those derivative images as recorded in the file metadata.

Issue #1135, opened 2021-05-12
Featured topic: Oral Histories


eddierubeiz:
This new featured topic will replace the "Science on Stamps" featured topic.

The ingredients for a featured topic are:
- a 700x700 pixel image (provided by Lee)
- a blurb (which we can take from the existing one on the OH collections page)
- a genre or set of subjects. Easy: we'll use the `oral histories` genre. 

Note: this will link to a page like e.g. [water](https://digital.sciencehistory.org/focus/water) and not [the actual oral history collections page](https://digital.sciencehistory.org/collections/gt54kn818), unless we change the behavior of featured topics slightly, which we could do.

@lsberry16 : we will need a square image 700 pixels on a side. Assume the bottom sixth of the image is hidden by the label created in front of the thumbnail. More details [in the wiki](https://chemheritage.atlassian.net/wiki/spaces/HDCSD/pages/1646428165/Creating+a+thumbnail+for+a+Featured+Topic+landing+page).

eddierubeiz:
Oh, we could also change the behavior for the featured topic object so it can just behave like a link if it's created using a URL.

jrochkind:
> Note: this will link to a page like e.g. water and not the actual oral history collections page. Not a big deal, we assume: they're almost the same thing, except for the special OH facets.

Let's explore that, actually. It does not need to be such. 

I wonder if instead of using our normal technology for the "featured topics", the link to OH on home page in that "featured topics" area -- should just link to our OH Digital Collection home page?

Since we have and are spending so much time making it a good OH landing page -- we plan to do more changes to it taking it even further away from the typical "featured topic" page. 

Will require some adjustments to our code, instead of just adding a "featured topic" config for OH, but it's do-able, and I think is probably the way to go

@eddierubeiz one way to organize our code... might be adding to the "featured topic" configuration and class definition, a, like, "just use this url" line. Where instead of linking to a 'featured topic' page, the logic on home page can see "oh, it is configured 'just use this link'", and well, just use that link.  Yeah, what I just saw you post as you were writing this!  We have an existing Env config for "oral history collection ID" so might try to reference that somehow... this maybe gets more complicated than "just do this", but we can work it out somehow maybe. 

Another option would be not using the existing "featured topic" infrastructure at all, instead just adjust the front-page  logic to be like "8 things from 'featured topic' config, plus this one hard-coded thing".  

Whatever is easy.

eddierubeiz:
Exactly!

[...]

PR #1151, opened 2021-05-18
Featured topic pointing at the oral history collection

eddierubeiz:
Ref #1135 
Adds a new optional `path` setting for the featured topic definition. If it's set, the featured topic simply behaves like a link.


app/models/featured_topic.rb
```diff
@@ -226,9 +226,9 @@ class FeaturedTopic
...
       description: "This digital collection features selected visual representations of the periodic table of the elements, with an emphasis on alternative layouts including circular, cylindrical, pyramidal, spiral, and triangular forms. Ranging in date from the 1860s to the 1990s, the materials provide a panorama of the historic evolution of the periodic table following Dmitri Mendeleev's initial 1869 design. Illustrations depicting chemical elements and concepts in atomic theory are also included in this collection."
     },
-    science_on_stamps: {
-      subject: ["Science on postage stamps"],
-      description: "This digital collection features selected postage stamps and other philatelic materials depicting various aspects of science: from molecules and chemical reactions to portraits of Nobel Prize laureates. Materials in this collection, dating from 1887 to 2000, include stamps and first day covers, as well as commemorative and postmarked envelopes and postcards. Most of the items were selected from the Witco Stamp Collection in the Institute's archives."
+    :oral_histories => {
+      title: "Oral Histories",
+      path: "/collections/gt54kn818"
```
jrochkind, line 231: The trick is that ideally this would use our ENV variable for the collection ID, instead of hard-coding it.  Everywhere else we use that variable...

actually that's no trick at all, it should be easy to do that here, since it's an actual ruby file? I was at first thinking it was a YAML where that would be harder, but it's not!  Unless we run into weird problems with order-of-operations on boot, should work fine, please try to use the ScihistDigicoll::Env lookup for id!

app/models/featured_topic.rb
```diff
@@ -279,6 +279,14 @@ def slug
     category_key.to_s.dasherize
   end
 
+  def path
+    if definition.has_key?(:path)
+      definition[:path]
+    else
+      Rails.application.routes.url_helpers.featured_topic_path(slug)
```
jrochkind, line 286: This is a confusing topic, but i feel like I read somewhere that there are performance problems with calling `Rails.application.routes.url_helpers`, even though everyone did it. There is some alternate way to do it, that is like doing an `include` into a class or something....

Alternately, it might not be so bad just to hard-code `/focus/#{slug}` or whatever. Alternately, it's not so bad to have the logic in the view `if path`. 

Wanting view-related logic in a model... is why we have those "ViewModels", hopefully eventually to be replaced by ViewComponent. Once we are using ViewComponent I think since it works more smoothly we'll end up using it in even more places, but for now not necessarily suggesting adding that complexity, just pointing out that it's motivation is related to all this.

spec/models/featured_topic_spec.rb
```diff
@@ -21,11 +21,12 @@
...
-    describe "constructs sensible queries for each topic" do
+    describe "constructs sensible queries for each topic (except those that are really just a link)" do
       FeaturedTopic.keys.each do |key|
         it "translates #{key}" do
-          result = FeaturedTopic.new(key).solr_fq
-          expect(result).to match EXPECTED_SHAPE
+          featured_topic = FeaturedTopic.new(key)
+          next unless featured_topic.path.include? 'focus'
+          expect(featured_topic.solr_fq).to match EXPECTED_SHAPE
         end
```
jrochkind, line 30: better with separate examples ("it" blocks) for path and non-path featured topics. 

Best rspec practice is individual tests for individual things. I know we sometimes violate that for convenience -- or in system/feature specs for performance since each spec example can take a long time to run. But it's best to keep those the exception, and get back to the "rule" of trying to keep rspec "it" examples brief and focused on a specific thing to test, not testing multiple things in one test. It makes the tests a lot more informative and easier to deal with when they fail, among other things. 


See this recent blog post on general topic: https://www.tomdalling.com/blog/mentoring/write-detailed-rspec-example-descriptions/

app/models/featured_topic.rb
eddierubeiz: OK. I had my doubts about it. I'm going with just hard-coding it, along with a comment.

jrochkind: and actually, you can write a test that tests that it *matches*  `featured_topic_path`, so if the assumption ever changes, the test will warn us!

app/models/featured_topic.rb
eddierubeiz: Good call. Works great!

jrochkind:
nicely done!

PR #1160, opened 2021-05-25
Fix error in attempted cap error message

jrochkind:
Error showed up as: NoMethodError: undefined method `join' for #<Set: {"cron", "jobs", "web", "app"}>

We don't have a 'solr' server anymore. Cap wants to warn us. For now, fix the error that happened when it tried to warn us.


eddierubeiz:
Thanks!

Issue #1154, opened 2021-05-20
Deal with empty interviewee bios


eddierubeiz:
On importing restricted interviews such as [Gregory A. Clark](https://digital.sciencehistory.org/works/49j2s7t), it appears the import code wound up creating an empty shell of an interviewee biography. This shows up as an irritating "Interviewee biographical information" title on the page, with no actual bio underneath it.

These objects should be deleted.

jrochkind:
It is likely this applies to all "restricted" interviews. We don't have them clearly identified with metadata, but one way to find them is to search for the common string "At the request of the interviewee, the abstract of the interview and interviewee’s biographical data have not been included.", which sould probably appear in description field for all of them.

PR #1163, opened 2021-05-26
rake task to delete empty interviewee biographies

jrochkind:
[DRY_RUN=true] ./bin/rake scihist:data_fixes:remove_empty_biographies

Has already been run on staging. After merging, needs to be deployed and run on production. 

Ref #1154


eddierubeiz:
Looks good to me.

Issue #1141, opened 2021-05-17
Correctly deal with restricted derivatives in the orphan derivative checker.


eddierubeiz:
This is part of #864, which we are closing as it's covered by the three other newer tickets it mentions.
The orphan derivative checker currently ignores restricted derivatives. They should be added to the orphan checking routine.
The tests should be updated as well.

Issue #1142, opened 2021-05-17
Run S3 orphan checkers on a cron. Create report for it in the admin pages.


eddierubeiz:
This is part of #864 .

jrochkind:
We could also automatically delete orphans instead of just reporting on them... especially for derivatives... but maybe reporting is a first step.

eddierubeiz:
This should be done only after #1140  and #1141 are closed.

jrochkind:
Probably want to fix #1141  as a pre-req to this one.

PR #1178, opened 2021-06-02
New rake tasks for finding and deleting orphaned restricted derivatives.

eddierubeiz:
Ref #1141 .


jrochkind:
Hi, this is ready for review?

Do you think it makes sense to make this a separate rake task, instead of incorporating in the existing derivative checking tasks?

eddierubeiz:
Yes- feel free to sync prod to staging and try it out!

eddierubeiz:
To your question -- yes, I do like separating this out as its own rake task. What I might do as part of #1142 is create a rake super-task that can be called nightly. Its only job would be to call all the nightly orphan checks tasks in sequence (4 of them I think).

app/models/asset.rb
```diff
@@ -109,6 +109,12 @@ def self.all_derivative_count
     ).first["count"]
   end
 
+  def self.all_restricted_derivative_count
+    Kithe::Asset.connection.select_all(
+      "SELECT count(*) FROM (SELECT row_to_json(jsonb_each(file_data -> 'derivatives')) -> 'value' ->> 'storage' as storage_key FROM kithe_models WHERE kithe_model_type = 2 ) as derivatives where storage_key = 'restricted_kithe_derivatives'"
+    ).first['count']
+  end
+
```
jrochkind, line 117: I'm not totally sure if I like this being in Asset model. 

I realize `self.all_derivative_count` is in model.... this new one is a SUBSET of that one, rather than alternate number? It's getting a bit confusing. 

One of the places we use all_derivative_count is for a total for any progress bar taht's operating on all derivatives.... although that's mostly just weird edge case ones at present.  Some of those might be wrong now, cause the total number will be ALL derivatives, but what they DO might just be on public derivatives (so may never get to 100%). 

This thing of putting restricted derivs on a different bucket is a bit convoluted, alas. :(

Not sure where toher than the orphan checker we're going to use this one JUST for restricted derivs... what if we just inline this were we need it?  Or alternately, just put something in a comment making it clear this is a SUBSET of above one. It probably doens't matter too much where it is.

eddierubeiz: I'd honestly rather move it to the rake task. I doubt we will ever use this query anywhere else.

eddierubeiz: Done

PR #1180, opened 2021-06-02
Removing acl and owner info from preservation logical (sql) backup

eddierubeiz:
Per Jonathan's suggestion, let's make the sql backup more portable by not including owner and ACL info in the backup file.

### DIFF
--- a/lib/tasks/copy_database_to_s3.rake
+++ b/lib/tasks/copy_database_to_s3.rake
@@ -38,7 +38,7 @@ namespace :scihist do
         storage_class: "STANDARD_IA",
         metadata: { "backup_time" => Time.now.utc.to_s}
       ) do |write_stream|
-      cmd.run('pg_dump', '-w', '--clean', ENV['DATABASE_URL']) do |out, err|
+      cmd.run('pg_dump', '--no-owner', '--no-acl', '--clean', ENV['DATABASE_URL']) do |out, err|
         write_stream << out.force_encoding('UTF-8') if out
       end
     end


lib/tasks/copy_database_to_s3.rake
```diff
@@ -38,7 +38,7 @@ namespace :scihist do
         storage_class: "STANDARD_IA",
         metadata: { "backup_time" => Time.now.utc.to_s}
       ) do |write_stream|
-      cmd.run('pg_dump', '-w', '--clean', ENV['DATABASE_URL']) do |out, err|
+      cmd.run('pg_dump', '--no-owner', '--no-acl', '--clean', ENV['DATABASE_URL']) do |out, err|
```
jrochkind, line 41: What does the `-w` that we're dropping do? Looking it up myself, it seems to be a synonym for `--no-password, Never issue a password prompt. `? Do we still want the `-w` there then, it doens't seem like it's related to adding --no-owner --no-acl? https://www.postgresql.org/docs/12/app-pgdump.html

eddierubeiz: Oh good. Yeah, I'll add it back in and (this time) document it. I looked for that -w and could not find it anywhere in the documentation, so I decided I'd rather remove it now while we're actively testing backups. As long as we know exactly what it does, we can keep it.

eddierubeiz:
Also use Schihist::env
 not ENV

jrochkind:
I'll mark this "draft" again until those outstanding things get taken care of, just for keeping things orderly so I know not to review untl it's done?

lib/tasks/copy_database_to_s3.rake
```diff
@@ -3,45 +3,45 @@ require 'tempfile'
...
 namespace :scihist do
   desc """
-    Dump the live database and upload it to s3.
+    Dump the live database and upload it to s3:
 
-    BACKUP_AWS_ACCESS_KEY_ID=joe \
-    BACKUP_AWS_SECRET_ACCESS_KEY=schmo \
-    APP=scihist-digicoll \
-    rake scihist:copy_database_to_s3
+    bundle exec rake scihist:copy_database_to_s3
```
jrochkind, line 8: maybe give an example of doing this with:

      eg from heroku where necessary config is defined:
      heroku run rake scihist:copy_database_to_s3

jrochkind:
looks great! Approved after perhaps comment example I mentioned is added?

jrochkind (approved):

lib/tasks/copy_database_to_s3.rake
jrochkind: Also could be worth nothing that this has to be run from where that has `pg_dump`, where it can be run without supplying username/password, to connect to database you want dumped -- it does not use the Rails database config, it just relies on `pg_dump`!

Really, this is meant for running on heroku, it might not really work right from other places, although could if things are set up right. I think probably the description should just say intended for running on heroku though, and mention that it's used by a nightly scheduled job?  Or even put in a `scihist:heroku:copy_database_to_s3` namespace, what do you think?

PR #1190, opened 2021-06-08
Allow for nil content type in "born on this day" OHs.

eddierubeiz:
Quick attempt to fix a bug on the OH landing page.

### DIFF
--- a/app/views/collection_show_controllers/oral_history_collection/_splash_page.html.erb
+++ b/app/views/collection_show_controllers/oral_history_collection/_splash_page.html.erb
@@ -110,8 +110,8 @@ we want the full width of the screen, no margins or padding. %>
       <div class="this-day-body">
         <% born_on_this_day_biographies.each do |bio| %>
           <div class="this-day-card">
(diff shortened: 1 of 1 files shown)


jrochkind:
Thanks @eddierubeiz !

This implies that there was a Work without a `leaf_representative` set, which is kind of bad data maybe? Should we try to identify what that is, and understand why and if it's a problem? 

Like, for instance, if there is a public Work without any members at all, it could cause this -- and would probably be a human error to publish such a Work?  Or if it has members but none of them are leaf_representative, that's maybe a software error, and we should maybe "fix" that specific Work, as well as wonder how it came to be.

jrochkind:
Oho, wait, I did figure it out!

https://digital.sciencehistory.org/collections/gt54kn818

* Has a birthday today so shows up on home page
* Has no members, it's empty
* The work is *private*  though, so it's not a published work with no members

The problem is that our "born on this day" feature doesn't respect non-public works to keep them out!

I'll file a GH issue for us to fix that bug!

eddierubeiz:
Correct!

Issue #1183, opened 2021-06-03
Create two new metadata fields: Transcription and Translation


MDiMeo:
Create two new metadata fields for Transcription and Translation. They should be on the asset level and use Plain Text. 

Once this functionality exists, we agreed to revisit this in one month -- after we've used the feature -- to discuss any issues that have arisen.

jrochkind:
@MDiMeo @apinkney0696 for the 'translation' field I'd like to propose an extension to make it future-proof:

1. Each translation should be labelled with the language the translation is in ("English" for these)
2. The translation/language-label field should be repeatable, to support multiple translations. 

What do you think? Alternately, perhaps I am over-complicating; since right now we plan only a single English translation, maybe we make only a single non-repeatable translation field, without language field, no more complicated than we need at present. We can always do a data migration later if we add translations in other languages and/ore more than one translation.  (In this case, I might call the field "English Translation" to be clear). 

Thoughts?

jrochkind:
I think @MDiMeo doesn't see or doens't have time to give feedback, @eddierubeiz do you have any thoughts on question above, should we make translations repeatable and langauge-labelled for future-proofing, or is that over-complexity for current needs?

jrochkind:
After talking to @apinkney0696 we decided for now keep it simple what we need for our actual use cases:

1. A "transcription" field, it won't have it's own language code, we'll use the existing work-level language assignmetns and not worry about a page-level language code. 
2. A single "English translation" field. That's what we have a need for, that's what we'll create. (It is important to have metadata on what language a translation is in, so we'll name this one "english translation" so if anyone is ever tempted to put eg spanish in there, they'll know we need a data model refactor that tags language on translation. :) )

eddierubeiz:
This is great.

eddierubeiz:
The language of the work itself is already part of the work-level metadata, and if the work is in a particular language, the transcriptions will be, too.

As for the translations, totally fine with hard-coding the word "English" into the name of the field at least for now.

jrochkind:
@apinkney0696  says in slack:

[...]

PR #1194, opened 2021-06-08
simple text-only transcription and translation fields on asset

jrochkind:
Ref #1183


jrochkind:
probably needs some tests

Issue #1195, opened 2021-06-08
get resque background jobs on heroku working with auto-scaling


jrochkind:
Currently a couple problems when I try doing large ingests with auto-scaling. While hirefire autoscaling seems to work, problems are:

1. We get some errors in some of the jobs, that seem to match this: https://stackoverflow.com/questions/50228454/why-am-i-getting-opensslsslsslerror-ssl-read-sslv3-alert-bad-record-mac

2. After hirefire scales down dynos, resque admin still reports more workers than it should. Only way to get them to go away seems to be manually purging resque's worker count and restarting all workers.

jrochkind:
Also related to #570

PR #1197, opened 2021-06-09
enable hirefire worker autoscaling, and other worker tune-up for heroku

jrochkind:
Ref #1195, but we are only auto-scaling workers here, not yet web. 

# 1. hirefire-resource gem

required for integration with hirefire.io, the main thing it does is provide a URL endpoint hirefire can access to get the current job queue count, so it knows if it wants to scale. 

We don't have hirefire web scaling turned on yet. #1180 

The HIREFIRE_TOKEN needs to be set in heroku config in order for connection to work. 

# 2. resque-heroku-signals gem

Provides a monkey-patch to resque, to make it work properly with how heroku tries to shut down dynos. We want our dynos to shut down somewhat gracefully, and it turns out resque gem needed a patch for that, which resque itself hasn't merged. :( This applies to hirefire downscaling, but also to any worker dyno restart, when a worker dyno may be busy -- from code deploys, heroku's automatic nightly restarts, whatever. 

https://github.com/resque/resque/issues/1559

Heroku config vars that we set to effect graceful shut down (https://github.com/resque/resque#heroku):
* TERM_CHILD=1 : Switches resque to a different shutdown mode that respects the following vars....
* RESQUE_PRE_SHUTDOWN_TIMEOUT=20: After getting the signal to shut down from heroku, give busy workers 20 seconds to finish their work, then tell them they better really abort
* RESQUE_TERM_TIMEOUT=8; after telling them they better really abort, give them another 8 seconds to stop and clean up (for instance register as 'failed' with resque) before really killing them hard. 

In testing, this seemed to mostly work. If a job isn't finished in the 20 seconds, it will wind up in the resque 'failed' queue. 

Trying this repeatedly, in *some* cases a worker dies and the resque admin dashboard doesn't know it's dead, see below. 

# 3. another local patch to heroku

When testing heroku dyno restarts and scaling, I was getting intermittent mysterious errors from resque workers:

    OpenSSL::SSL::SSLError: SSL_read: sslv3 alert bad record mac

Googling, I found that someone else had this problem (on heroku too; not sure why it might only effect heroku) and thought they had diagnosed and fixed it. The PR remains unreviewed/merged in resque though. https://github.com/resque/resque/pull/1739

So we patch it in in our app instead, it's a simple patch. 

# Trimming phantom workers

In my tests of restarting worker dynos from heroku, sometimes resque ended up with "phantom" workers in it's list. Workers that didn't really exist anymore, but resque thought they did, so maybe resque thought there were 10 active workers, but really there were only 8, two were phantom. The phantom ones resque might think are idle, or busy. 

Resque code looks like it's *supposed* to trim these now and then, noticing a worker that hasn't sent a "heartbeat" in a while and deciding it's dead. But that didn't seem to be happenign in my observation. 

But running some simple resque code somehow did succesfully prune them, resque is capable of knowing they are dead, it just wasn't pruning them! I dunno, resque is in a bit of a maintenance abandonment. :( 

    
    Resque::Worker.all_workers_with_expired_heartbeats.each { |w| w.unregister_worker }

I made a rake task to do that too:

    heroku run rake scihist:resque:prune_expired_workers

If a phantom worker is believed to be busy, and gets pruned -- it'll get recorded as 'failed'. In some cases it may be recorded as failed even though it really succeeded -- this is one reason we try to make sure our tasks are idempotent!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -119,6 +119,14 @@ gem 'sane_patch', '< 2.0' # time-limited monkey patches
 gem 'activerecord-postgres_enum', '~> 1.3' # can record postgres enums in schema.rb dump
 
 
+# For autoscaling on heroku via hirefire.io service, but hopefully won't cause any problems
+# when running not on heroku.
+#
+# https://help.hirefire.io/article/53-job-queue-ruby-on-rails
+# https://help.hirefire.io/article/49-logplex-queue-time
+# https://github.com/hirefire/hirefire-resource
+gem "hirefire-resource"
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'pry-byebug', platforms: [:mri, :mingw, :x64_mingw]
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ gem 'puma', '~> 5.3'
 # https://github.com/nevans/resque-pool/issues/170
 gem "resque", "~> 2.0"
 gem "resque-pool"
+gem "resque-heroku-signals" # gah, weirdly needed for graceful shutdown on heroku. https://github.com/resque/resque#heroku
 
 gem 'honeybadger', '~> 4.0'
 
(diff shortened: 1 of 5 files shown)


eddierubeiz:
Thanks for the docs. I'm not in a position to critique much, but looks great.

Issue #1182, opened 2021-06-03
heroku, force https


jrochkind:
on legacy architecture, if you tried to connect to `http://`, it would redirect you to `https://`. We should do the same on heroku. 

I just noticed this, cause I connected accidentaly to `http`, and it looks like that broke fontawesome icon font with some kind of CORS error. We could fix that to work properly if you access the app under http -- but we should probably just make the app force https like it used to. 

One way to do that would be with a rails-level config and/or redirect, there's probably a feature built into rails to do this. Although we'd want to make sure it didn't take effect in development and test environments etc. Maybe triggered by a Env variable.

PR #1202, opened 2021-06-14
Force SSL

eddierubeiz:
[Heroku recommends](https://help.heroku.com/J2R1S4T8/can-heroku-force-an-application-to-use-ssl-tls) turning on `config.force_ssl = true` in `production.rb`.
Works fine on Heroku; I also tested it on our non-Heroku staging environment, and it does not affect the setup on the existing webserver. (Makes sense.)
This won't affect dev or staging.
Ref #1182

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -46,7 +46,7 @@ Rails.application.configure do
   # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
 
   # Prepend all log lines with the following tags.
   config.log_tags = [ :request_id ]
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -46,7 +46,7 @@ Rails.application.configure do
   # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = ScihistDigicoll::Env.lookup!(:force_ssl)
+  config.force_ssl = ScihistDigicoll::Env.lookup!(:force_ssl) == "true"
 
   # Prepend all log lines with the following tags.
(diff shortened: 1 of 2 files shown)


jrochkind:
Can we make it based on a ScihistDigicoll::Env that defaults to true in `Rails.env.production?`, just in case someone wants to test something in production without SSL?  I occasionally need to run on my dev machine in "production" mode to test something, and if it forced SSL, I'd have to figure out how to set up SSL on a `rails server -e production` on a dev machine, which I have no idea how to do.

eddierubeiz:
Ah, interesting use case. Of course.

lib/scihist_digicoll/env.rb
```diff
@@ -179,6 +179,8 @@ def self.persistent_redis_connection!
       # it'll just get out of sync.
     }
 
+    define_key :force_ssl, default: Rails.env.production?
```
jrochkind, line 182: Make it into a boolean, where `"true"` turns into `true`, like so:

    define_key :force_ssl, default: Rails.env.production?, system_env_transform: Kithe::ConfigBase::BOOLEAN_TRANSFORM

eddierubeiz: Great - done.

config/environments/production.rb
```diff
@@ -46,7 +46,7 @@
   # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = ScihistDigicoll::Env.lookup!(:force_ssl) == "true"
```
jrochkind, line 49: If it's boolean, then `== "true"` will never be true I think? Because `true == "true"` is never true? Have you tested this? I think it just needs to be `config.force_ssl = ScihistDigicoll::Env.lookup!(:force_ssl)`, no? And maybe test to make sure it's working! Thanks!

eddierubeiz: No, ScihistDigicoll::Env.lookup!(:force_ssl) returns a string, not a boolean. I did test it.

eddierubeiz: OH I see what's going on. Never mind.

Issue #1193, opened 2021-06-08
"translator" field for bredig 


jrochkind:
@apinkney0696 suggested that we need a "translator" field for our Bredig stuff. 

This is a little bit different than most of our metadata, because it does not apply to the "item in hand", it applies to our processing/digital copy only. 

So it's probably not one of our "Creator" fields. We might want to call the field something like "Translator (digital)" (probably  not exactly that). 

It is somewhat comparable to our "File creator" field, for which we don't allow free entry, but have a choice of "registered" file creators. 

We haven't decided exactly how to do this yet, so leaving this as record of conversation, and a placeholder to remind us to. 

@MDiMeo , I won't work on this until someone tells me to! :)

apinkney0696:
Yes! I definitely think it is important to make space for crediting our translators/transcribers in some way.

If not creating another field, we might consider adding another option to the 'Additional Credit' drop-down menu. Currently, we only have 'Photographed by.' We could potentially add 'Translated by' and 'Transcribed by' to that list. However, this field currently removes the default _Courtesy of the Science History_ Institute line, which we would want to maintain. 

If we decide on creating another field, I'll think of some options for what we should name it. 

Another question for consideration: Our File creator field is not publically viewable - would we want the translator/transcriber to be internally facing administrative metadata or do we want it to be public?

eddierubeiz:
I like the idea of just using `Additional Credit`, @apinkney0696 .
Maybe `Transcription by` and `English translation by` ?

(I can't imagine e.g. Jay Muhlin objecting to being credited publicly for photographing the "work in hand". Are there any other cases in which we might somehow compromise anyone's privacy by making the `File Creator` field public?)

apinkney0696:
Hm. I wasn't thinking of it from the perspective of privacy, though that's a good point. My thought is the File Creator field should remain private simply because it might be unfamiliar (and even confusing) to our users and also take up room on the public record which is primarily descriptive. It's important for us to have for admin purposes, but not necessarily adding anything to our user's experience. 

I was thinking about the long term, if we ever have an external contribution of a transcription or translation from someone who is not staff, it might be nice to publically recognize. We have, so far, only publically credited Digitization funding organizations and Donors on the public side of the digital collections (unless I have forgotten something else).

jrochkind:
>  However, this field currently removes the default Courtesy of the Science History Institute line,

That could be a bug, or at any rate can be changed.

eddierubeiz:
[...]

PR #1210, opened 2021-06-16
Adding "Translator" role to our "additional credit" field

eddierubeiz:
This is batch-editable (at least when it comes to *adding* a translator to multiple works, which is the main use case.)

And yes, adding a translator does not make anything disappear. It just appends a "Translated by" line to the metadata.

Ref #1193

### DIFF
--- a/app/models/work/additional_credit.rb
+++ b/app/models/work/additional_credit.rb
@@ -2,8 +2,9 @@ class Work
   class AdditionalCredit
     include AttrJson::Model
 
-    ROLES = ['photographed_by']
+    ROLES = ['photographed_by', 'translator']
     NAMES = [
+      'Jocelyn R. McDaniel',
       'Douglas Lockard',
       'Gregory Tobias',
       'Mark Backrath',
@@ -26,6 +27,7 @@ class Work
     attr_json :name, :string
 
     def display_as
+      return "Translator: #{name}" if role == 'translator'
       "#{role.humanize} #{name}"
     end
   end
--- a/app/models/work/additional_credit.rb
+++ b/app/models/work/additional_credit.rb
@@ -27,7 +27,7 @@ class Work
     attr_json :name, :string
 
     def display_as
-      return "Translator: #{name}" if role == 'translator'
+      return "Translated by #{name}" if role == 'translator'
       "#{role.humanize} #{name}"
     end
   end
--- a/app/models/work/additional_credit.rb
+++ b/app/models/work/additional_credit.rb
@@ -2,11 +2,11 @@ class Work
   class AdditionalCredit
     include AttrJson::Model
 
-    ROLES = ['photographed_by', 'translator']
+    ROLES = ['photographed_by', 'translated_by']
     NAMES = [
-      'Jocelyn R. McDaniel',
       'Douglas Lockard',
       'Gregory Tobias',
+      'Jocelyn R. McDaniel',
       'Mark Backrath',
       'Penn School of Medicine',
       'Will Brown'
@@ -27,7 +27,6 @@ class Work
     attr_json :name, :string
 
     def display_as
-      return "Translated by #{name}" if role == 'translator'
       "#{role.humanize} #{name}"
     end
   end
--- a/spec/factories/work_factory.rb
+++ b/spec/factories/work_factory.rb
@@ -151,7 +151,7 @@ FactoryBot.define do
         [
           Work::AdditionalCredit.new({"name"=>"Douglas Lockard", "role"=>"photographed_by"}),
           Work::AdditionalCredit.new({"name"=>"Mark Backrath",   "role"=>"photographed_by"}),
-          Work::AdditionalCredit.new({"name"=>"Jocelyn R. McDaniel",   "role"=>"translator"})
+          Work::AdditionalCredit.new({"name"=>"Jocelyn R. McDaniel",   "role"=>"translated_by"})
         ]
       }
       physical_container {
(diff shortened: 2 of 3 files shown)


app/models/work/additional_credit.rb
```diff
@@ -2,8 +2,9 @@ class Work
   class AdditionalCredit
     include AttrJson::Model
 
-    ROLES = ['photographed_by']
+    ROLES = ['photographed_by', 'translator']
     NAMES = [
+      'Jocelyn R. McDaniel',
```
jrochkind, line 7: should these be ordered in some way? It looks like they were alphabetical by first name (ie by string) before, let's keep it that way?

app/models/work/additional_credit.rb
```diff
@@ -26,6 +27,7 @@ class AdditionalCredit
     attr_json :name, :string
 
     def display_as
+      return "Translated by #{name}" if role == 'translator'
```
jrochkind, line 30: D you need this special case "return" here? Why isn't the same mechanism already being used to display phographer credit (ie, the next line, `"#{role.humanize} #{name}"`) sufficient?

spec/factories/work_factory.rb
```diff
@@ -150,7 +150,8 @@
       additional_credit {
         [
           Work::AdditionalCredit.new({"name"=>"Douglas Lockard", "role"=>"photographed_by"}),
-          Work::AdditionalCredit.new({"name"=>"Mark Backrath",   "role"=>"photographed_by"})
+          Work::AdditionalCredit.new({"name"=>"Mark Backrath",   "role"=>"photographed_by"}),
+          Work::AdditionalCredit.new({"name"=>"Jocelyn R. McDaniel",   "role"=>"translator"})
```
jrochkind, line 154: Why are you adding this into default "work factory", what were your thoughts/motivation there?

jrochkind:
@eddierubeiz from original ticket, can you confirm you've verified bothof these things:

> Let's double check and make sure it "Appends Credit Line" (as the hint says), and does not replace it. "Appends" is desired, it's a bug if it "replaces".

(that one might could use a test)

From annabel:

> One addendum: may I request we double-check this can be batch edited, please! Then we can add Jocelyn's name to many records at once.

app/models/work/additional_credit.rb
eddierubeiz: Good catch!

spec/factories/work_factory.rb
eddierubeiz: Hmmmm.....  so supposing, in the *existing* tests, there's some weird interaction that causes problems if you add a translator or have more than one "role" in `additional_credit`, etc., this would make us aware of it.

Secondarily, if a bug involving `translator` comes up _later_, then this metadata is there in all the existing tests that use a `Work` object, and we have it to test with.

app/models/work/additional_credit.rb
eddierubeiz: The front end is supposed to be shown as "Translated by"... so we can do this, or change the name of the `role` on the actual model to `translated_by`. Granted, that would be more consistent with `photographed_by`, but for metadata purposes I just like the single noun "Translator".

eddierubeiz: Good idea re: using the i18n -- the early `return` is kinda icky.

eddierubeiz:
Yes!! I double-checked both. Also, honestly, I'd love to get rid of the "Appends Credit Line" altogether. Adding a translator doesn't append a credit line. It just, well, adds a translator.

spec/factories/work_factory.rb
jrochkind: OK, maybe for #1. 

For #2, it's importnat to keep in mind we don't create a generic factory with the data to exersize every possible test we'd want to do -- we create specific data in specific tests for the test. If we want to write a test to (eg) make sure the translator is displayed properly (as below is suggested we might), we'd want to actually write `create(:work, additional_credit: whatever)`, not just rely on whatever is in the default factory. Relying on what's in the default factory makes it too fragile to default factory changes, the default factory should just be a "typical default generic thing". 

Because *most* of our Works won't have translators is what makes me think the default factory shoudln't. (Yes, this could apply to photographer too, it's true). Plus everything we add to the default factory making it heavier-weight is potentially making every single test that uses it slightly slower, adding up. 

Not a big deal either way, just my thoughts.

app/models/work/additional_credit.rb
jrochkind: I would just change the key to `translated_by` honestly. Becuase it's simple, and consistent with the existing `photographed_by`, I don't think it matters too much either way beyond a "preference", why not leave them consistent? 

Is it actually spec'd somewhere by stakeholders that front-end should say "translated by"? I guess that makes sense though, again cause it's consistent with "photgraphed by"!

jrochkind:
>  Adding a translator doesn't append a credit line. It just, well, adds a translator.

You're doing it exactly the same as adding a photographer, right? Great. I think some people considered the "translated by" or "photographed by" line to be an "appended credit line"?  Maybe it was there because *originally* they *were* going to have it replace the "Courtesy of Science History Institute" line, so the hint is reminding you that it does not. 

But just do this just like photographer, I'd say!

app/models/work/additional_credit.rb
eddierubeiz: OK never mind -- I'm capitulating and using `translated_by` as the label.

spec/factories/work_factory.rb
eddierubeiz: OK you convinced me again.

PR #1214, opened 2021-06-17
README link to heroku developer setup doc in wiki

jrochkind:
Check out the wiki doc too eddie! You don't *have* to do things this way, but I think it makes sense, and will probably write other docs intended for developers using the `-r staging` and `-r production` options that are set up thusly. (Also I think it makes sense to make the defualt staging, rather than production, to minimize accidents).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Start a development Solr instance with `./bin/rake solr:start`.
 
 Run app with `./rails server`, it will be available at `http://localhost:3000`.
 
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Yeah -- having a new laptop is a prime opportunity to test developer setup. That's yet another reason to not simply fork over the entire contents of the old laptop -- that and I *royally* screwed up how Python is installed on the old one.

Issue #1170, opened 2021-06-01
heroku: sync prod to staging routine


jrochkind:
And make sure to document https://chemheritage.atlassian.net/wiki/spaces/HDC/pages/1908834305/Heroku

jrochkind:
May be related to #876, in that we will be using postgres backup files?

jrochkind:
Note the postgres sync may be basically built into heroku:

https://devcenter.heroku.com/articles/heroku-postgres-backups#direct-database-to-database-copies

eddierubeiz:
No need to involve the backup process unless we want to. This works fine:
```
heroku maintenance:on --app scihist-digicoll-staging
heroku pg:copy scihist-digicoll-production::DATABASE_URL DATABASE_URL -a scihist-digicoll-staging  --confirm scihist-digicoll-staging
heroku run rake scihist:solr:reindex        --app scihist-digicoll-staging
heroku run rake scihist:solr:delete_orphans --app scihist-digicoll-staging
heroku maintenance:off --app scihist-digicoll-staging
```
If we *want* to involve the SQL backups, we can use a variation on this:
```
heroku maintenance:on --app scihist-digicoll-staging
yes | aws s3 cp s3://$BACKUP_BUCKET/$BACKUP_PATH .
gunzip $BACKUP.gz
heroku pg:psql --app scihist-digicoll-staging < $BACKUP
heroku run rake scihist:solr:reindex        --app scihist-digicoll-staging
heroku run rake scihist:solr:delete_orphans --app scihist-digicoll-staging
rm -rf $BACKUP
heroku maintenance:off --app scihist-digicoll-staging
```

jrochkind:
@eddierubeiz What do you think is best? We had talked about involving the S3 backups as a way of testing they work, but I'm not sure, what do you think?

Where do we currently have this process documented, can you link to it?  Does the current process involve manually running an `aws s3 cp`?    

I was imagining a single command someone could run to "do the sync", but I guess if that involves having `aws` CLI installed anyway, and would be hard to do, then maybe we just update the current docs for a manual process? This would be another thing to have @CHF-IT-Staff (Chuck) demo from docs.

eddierubeiz:
- The old process is documented at https://chemheritage.atlassian.net/wiki/spaces/HDCSD/pages/950730760/Syncing+scihist-digicoll+production+to+staging .

[...]

PR #1218, opened 2021-06-21
Sync prod to staging, via a new rake task.

eddierubeiz:
Ref #1170

Download the latest DB backup, and replace the Heroku staging database with that. Then reindex and delete any orphans in SOLR.

Includes configurable settings:
`BACKUP_REGION` `BACKUP_FOLDER` `BACKUP_BUCKET` `BACKUP_FILENAME` `STAGING_APP_NAME` `UNZIP_CMD`


jrochkind:
Nice! Can I actually check out this branch and run this myself to test it?

re making it more configurable, can be as easy as `BACKUP_FILENAME= ENV['BACKUP_FILENAME] || "heroku-scihist-digicoll-backup.sql"` (and can do that with the other ones too, why not. Can also test if it can actually be BACKUP_SOURCE_KEY, like it could be a path like `path/to/somewhere.sql`). 

So it's still looking up the AWS keys from places like... `ScihistDigicoll::Env.lookup(:aws_access_key_id)` ... so when you say "developer key", you mean the one we have configured there?

That could work. I am still curious about exploring leaving out the `access_key_id` and `secret_access_key` arguments to Aws::S3::Client altogther -- meaning it will just look them up the conventional ways, starting from whatever you have configured in your ~/.aws or failing that shell env, or whatever?

eddierubeiz:
Thanks for these really good suggestions.
1) Yeah, hadn't thought of it, but in fact there's no need to look up the developer key in the code.
2) Yes, you can absolutely check out this branch and it should work provided you have a) the developer key and Heroku command-line access to the staging app. Works for me.

eddierubeiz:
Oh, as part of this review, please also look at [the dev key in the aws UI](https://console.aws.amazon.com/iam/home#/users/dev-key?section=permissions) and take a look at the inline policy `read_only_access_to_backup_db`.

I am by no means an expert at aws permissions, so I tried to keep things as simple as possible.

jrochkind:
> Enabling maintenance mode for scihist-digicoll-staging

Nice job having that printed out to the console, so I noticed it even without having noticed it in source!

What makes you think this is required? We weren't doing it before, were we? We were making database backup on the database that could be in use?  

I guess it maybe makes sense though, hmm. I guess it avoids people being like "Hey, something is weird on staging", when the answer is "yeah, the sync is in progress", which did happen before?

It's going to stay down for a while though, I think the sync can take 20 minutes plus? Mine is still in progress so can't say fo sure. That could be inconvenient to have it down so long. Not sure if it's worth it?

jrochkind:
OK, the DB sync took ~20 minutes -- is that consistent with what we expect? With how long sync took under old system?

But then it errored out when it got to the S3 part. (Good news -- it turned off maintenance mode again anyway!)

```
[f6b9d41a] Finished in 814.165 seconds with exit status 0 (successful)
Time: 00:00:02 Work:ff365541m: |====================================================================================================================================================================================================| 100.00/s 200/200 100% Time: 00:00:02
Deleted 30 Solr objects
[4eea109f] Running aws s3 sync s3://scihist-digicoll-production-originals s3://scihist-digicoll-staging-originals
[4eea109f] 	fatal error: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
[4eea109f] Finished in 6.461 seconds with exit status 1 (failed)
[173b6428] Running heroku maintenance:off --app scihist-digicoll-staging
[173b6428] 	Disabling maintenance mode for scihist-digicoll-staging...
[173b6428] 	 done
[173b6428] Finished in 1.554 second with exit status 0 (successful)
rake aborted!
TTY::Command::ExitError: Running `aws s3 sync s3://scihist-digicoll-production-originals s3://scihist-digicoll-staging-originals` failed with
  exit status: 1
  stdout: Nothing written
  stderr: fatal error: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
/Users/jrochkind/.gem/ruby/2.6.6/gems/tty-command-0.10.1/lib/tty/command.rb:106:in `run'
/Users/jrochkind/code/scihist_digicoll/lib/tasks/sync_prod_to_staging.rake:33:in `block (2 levels) in <main>'
```

I must not have the AWS key expected, the same one you do?  

I'm not totally sure how to figure out the operative AWS key.... I think probably key with`AWS_ACCESS_KEY_ID`  `AKIA...NDP` (that whole thing shoudl not be a secret, not putting the secret part here, but omitting the middle even though it's not secret. Is that not the "Developer key"?  Hmm, looking it up in IAM console, that **IS** the one labelled `dev-key`. Maybe it's not really the one being used? It should be though, it's the result of `echo $AWS_ACCESS_KEY_ID` in my console.... not sure what's going on.  

Hmm, it's NOT the one configured in my `~/.aws/credentials`... that one is key labelled 'jrochkind', my personal one. I'd think my shell ENV AWS_ACCESS_KEY_ID should take priority, but maybe not?  We are not doing great with AWS key management. 

I guess we could add the permission to my `jrochkind` key... if we were doing this right, I think we'd have a "Developer" (or a "read-backups") role/group we assigned to multiple keys.... not sure how much we want/need to get into dealing with our IAM technical debt right now....

jrochkind:
I will work on this more tomorrow, maybe I'll try setting the 'dev key' in my local aws credentials.

eddierubeiz:
SOLR reindex and delete_orphans were pointed at dev, not staging. Also chained the two operations as the same command per your suggestion on Slack.

eddierubeiz:
I timed it, and here's the info I got:

```
Turning on Heroku maintenance
Current Time : 09:52:48
Getting backup file
Current Time : 09:52:49
Decompressing backup file
Current Time : 09:52:59
Recreating the database
Current Time : 09:53:00
Reindexing and removing orphans from SOLR
Current Time : 09:53:32
Syncing originals
Current Time : 09:55:26
Syncing derivatives
Current Time : 09:56:12
Done with the sync
Current Time : 10:02:51
Turning off Heroku maintenance
Current Time : 10:02:51
Finished rake task
Current Time : 10:02:52
```

jrochkind:
Ok, I have the `AKIAU...NDP` key (`dev-key`) in my `~/.aws/credentials` file, AND in my env `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. 

Still, when I try to do the `aws s3 sync` command, I get an access denied. Very odd. I wonder how my setup differs from yours. 

I am just trying it directly from the command line, copying what is in the rake task:

```
$ aws s3 sync s3://scihist-digicoll-production-originals s3://scihist-digicoll-staging-originals
fatal error: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
```

jrochkind:
Try this maybe, can we compare `aws configure list` outputs?

```
$ aws configure list
      Name                    Value             Type    Location
      ----                    -----             ----    --------
   profile                <not set>             None    None
access_key     ****************2NDP              env
secret_key     ****************jAcl              env
    region                <not set>             None    None
```

eddierubeiz:
As part of this ticket, I'm going to create a new AWS group. If your user part of this group, you will have exactly the permissions needed in order to perform the sync rake task, namely:
- `read` from the backups bucket
- `read` the originals and derivatives bucket in **production**
- `write` to the originals and derivatives bucket in **staging**

eddierubeiz:
I created user group https://console.aws.amazon.com/iamv2/home#/groups/details/sync_prod_to_staging but was unable to get it to work.

For now, then, to use this rake task, the user needs to have the required permissions (e.g. via the key pair in a config file in ~/.aws) . For now they are assumed to be in the developer's user key (e.g. `eddie` or `jrochkind`), and at the moment, those permissions appear to be pretty much unlimited.

In the future, we may want to create `eddie_dev` permissions distinct from the ones that let us log in to the aws console, and which would have *just* the permissions required to run this and similar rake tasks, but *not* the permissions required to break everything in the aws0 console.

eddierubeiz:
Merging this. @jrochkind  -- you may want to use `USE_BACKUP=false bundle exec rake scihist:sync_prod_to_staging`, as this will run much faster.

Issue #931, opened 2020-12-03
Scheduled jobs in heroku


jrochkind:
It looks to me like the heroku free scheduler plugin is sufficient after all. 

It only lets you do nightly (or hourly) jobs, not weekly jobs -- we only create the google sitemap weekly, but it doesn't hurt to do it nightly. 

One downside is there scheduled job configuration will live purely on heroku accessible only via human web dashboard, there is no API for it, so we lose being able to track/control that configuration in source code. (We know do it with the `whenever` gem for configuring cron jobs in ruby, that get deployed to cron by capistrano). 

Additionally our code will no longer "know" when the nightly are jobs are scheduled -- only minor place this comes up now is the abliity for the fixity dashboard to say "these run every night at 2:30am", and have that be correct without having to keep it synchronized. That's not a huge deal.

jrochkind:
I guess we just make these by hand. Note that heroku scheduler uses UTC -- have to calculate what time we want in ET ourselves. And remember it will change by an hour when daylight savings goes on/off!  I think this is fine for our tasks, ti doesn't matter if they run at 2:30am during half the year and 3:30am the other half, although a bit annoying. 

Add comment to our `schedule.rb` (currently used by `capistrano-whenever`) that this doesn't effect heroku -- then remove the file entirely after we go live on heroku?

eddierubeiz:
Sounds good.

eddierubeiz:
Actually, instead of adding an actual comment to `schedule.rb`, I just created #1222 to cover the work of removing `whenever` and all references to it. It turns up in at least 2 or 3 places, and that's a distinct task that we can take care of well after launch, if I understand correctly.

eddierubeiz:
The sitemap creation rake task currently runs weekly. but the Heroku scheduler doesn't allow for weekly tasks. We should probably modify the task to end early unless it's Tuesday and run *that* nightly. Until we take care of that I'm fine with running the task manually for a couple weeks. @jrochkind  - other ideas welcome.

eddierubeiz:
I've added all the tasks except sitemap creation to https://dashboard.heroku.com/apps/scihist-digicoll-production/scheduler, so they'll be running nightly from now on.

jrochkind:
Oh, annoying! 

There are other scheduler-add-ons that cost money that are more powerful. 

But also, it runs weekly because it doesn't need to run more often than weekly, but it doesn't hurt it to run nightly. Just run it nightly?

[...]

PR #1224, opened 2021-06-23
Adding rescue_rake: true

eddierubeiz:
See conversation at ref #931 about 
`HONEYBADGER_EXCEPTIONS_RESCUE_RAKE`.

@jrochkind figured out that we were trying to configure HB in a very roundabout way via an env variable set in `config/schedule.rb`, and appended as a prefix to every scheduled task run by `whenever`. `whenever` is going away, so we're taking the opportunity to do this (hopefully) the right way.

See also:
- https://github.com/honeybadger-io/honeybadger-ruby/issues/258
- https://docs.honeybadger.io/lib/ruby/gem-reference/configuration.html

### DIFF
--- a/config/honeybadger.yml
+++ b/config/honeybadger.yml
@@ -11,5 +11,6 @@ breadcrumbs:
 exceptions:
   ignore:
     - 'ActionDispatch::Http::MimeNegotiation::InvalidType'
+  rescue_rake: true
 
 # http://docs.honeybadger.io/ruby/gem-reference/configuration.html


jrochkind:
@eddierubeiz I think you missed my last comment last night? 

After going through some other thoughts, I tested and verified that Heroku scheduled tasks ARE reporting to HoneyBadger on uncaught exceptions, with no further configuration needed. 

https://github.com/sciencehistory/scihist_digicoll/issues/931#issuecomment-866358958

eddierubeiz:
Triple-checked (my initial experiments went the other way, but I must have missed something.) You're right!

Issue #964, opened 2021-01-28
Batch edit does not apply to Collection field


HKativa:
I attempted to add work pr76f4633 and its child records to the Neville Collection using the cart, i.e. batch edit, and the edit didn't take (note: I subsequently edited the individual records to add them to the Collection). I played around a bit in staging and successfully edited other fields using the batch edit, but again, the collection edit didn't take. I have a vague memory that not all fields are enabled for batch editing, but could well be mis-remembering. This is low-priority - ultimately, I'd like to confirm whether or not the batch edit is intended to work for the Collection metadata field.

jrochkind:
I think the developers just forgot about Collections when developing this feature. 

since collection does not have the same metadata as work... it's probably a mistake that it lets you add it to the cart for batch editing.  We just didn't think of that I think.

There probably only 2 or 3 fields in common between Collection and Work... I guess those 2 or 3 could be applied to Collections in batch edit too? And the others ignored for Collections? 

I'm not totally sure how this could/should work.

HKativa:
Sorry, to clarify, I wasn't trying to edit a collection record. I was trying to batch edit works to add a value to the collection metadata field

apinkney0696:
Applying a collection to the Collections metadata field in the batch edit feature still doesn't seem to work and I wanted to follow up with Hillary's initial inquiry. If the Collections field is not intended to work in batch edit, can we enable that functionality? It would be helpful to the ongoing mass dig projects. Thanks!

jrochkind:
@MDiMeo just a ping to remind you that we rely on our Monday meetings to prioritize work, and don't generally work on things in "Inbox" unless you triage it to tell us to (or in some cases we realize it's an emergency on our own!)

PR #1226, opened 2021-06-24
Batch edit for collection

eddierubeiz:
Make collection batch-editable.
Ref #964

### DIFF
--- a/spec/system/batch_edit_spec.rb
+++ b/spec/system/batch_edit_spec.rb
@@ -8,8 +8,10 @@ describe "Cart and Batch Edit", solr: true, indexable_callbacks: true, logged_in
   let!(:collection_1) { FactoryBot.create(:collection, title: "collection_1") }
   let!(:collection_2) { FactoryBot.create(:collection, title: "collection_2") }
 
+  let!(:work0) { create(:work, provenance: "provenance 0")}
   let!(:work1) {
     create(:work,
+      title: "w1",
       provenance: "provenance 1",
       description: "description 1",
       additional_title: ["additional title 1a", "additional title 1b"],
@@ -18,12 +20,17 @@ describe "Cart and Batch Edit", solr: true, indexable_callbacks: true, logged_in
   }
   let!(:work2) {
     create(:work,
+      title: "w2",
       provenance: "provenance 2",
       description: "description 2",
       additional_title: ["additional title 2a", "additional title 2b"],
       creator: [{category: "contributor", value: "creator 2"}])
   }
-  let!(:work0) { create(:work, provenance: "provenance 0")}
+  let!(:work3) {
+    create(:work,
+      title: "w3",
+      contained_by: [collection_2])
+  }
 
   it "smoke test" do
     visit search_catalog_path(search_field: "all_fields")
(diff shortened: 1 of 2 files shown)


jrochkind:
The implementation here seems to me to be more complicated than it needs to be, but I want to spend some time with it. 

First though, a question about the intended spec -- is this intending to *replace* any existing collection ids with the new one(s) specified, or append the new one to the existing ones?

What does it do with existing multi-valued fields, as far as replacing existing values or appending to them?  This should probably do the same, and our test should probably ensure that it does that.  (Unless we have reason to want it to do someting different here? Check with Annabel?)

 So that's something that could be enhanced right now, decide what the 'correct' behavior is with regard to append-or-replace, and make the test ensure it does that.

eddierubeiz:
I'm open to anything but I *do* want to avoid duplicate associations between collections and works. We may wish to modify the Kithe code to enforce that, too, but that would go in another PR.
Otherwise stated, you should not be able to create two identical rows in the `kithe_model_contains` table.

jrochkind:
Yeah, I thought ActiveRecord used normally would prevent you from creating those duplicate join table rows, but I guess not? Anyway, that's a separate thing. But you already have a test ensuring that? Great, with the right tests, including for what I talked about above as far as delete-or-append, we can then refactor to be simpler while ensuring tests still pass.  (step three in TDD red -> green -> refactor!)

eddierubeiz:
Oh, to answer your initial question:
_is this intending to replace any existing collection ids with the new one(s) specified, or append the new one to the existing ones?_
This is intended to append the new collection to the existing collections. But if a work in the cart _already_ belongs to the collection being appended, that work should not change, since... it's already in the collection.

jrochkind:
> This is intended to append the new collection to the existing collections. But if a work in the cart already belongs to the collection being appended, that work should not change, since... it's already in the collection.

Great, is there a test that ensures that's what's happening, it's not overwriting existing collections? If so great! If not, can there be? Thanks!

eddierubeiz:
Let me add that - good call.

spec/system/batch_edit_spec.rb
```diff
@@ -58,20 +66,43 @@
...
+    expect(work1.contained_by.map(&:title)).to eq ["collection_1"]
 
     expect(work2.additional_title).to eq(["additional title 2a", "additional title 2b", "batch edit additional title"])
     expect(work2.provenance).to eq "batch edit provenance"
     expect(work2.creator).to eq([Work::Creator.new(category: "contributor", value: "creator 2")])
     expect(work2.description).to eq "description 2"
+    # Work 2 should have collection 1 added to it.
+    expect(work2.contained_by.first.title).to eq "collection_1"
+
+    # Go back to add collection 2 to both works:
```
jrochkind, line 92: Adding another transaction in really slows down the test, makes it take longer.

Is there really anything we really need to do two batch edits (two submits) in order to test? 

Can't we just set up the initial data such that we can test everything we might want to test in only one batch edit operation? You have an initial condition of some records, you do a batch edit, test the post-edit condition to make sure it did all the things we wanted. Even adding more records if needed (although I'm not sure they are) will usually be faster/simpler than doing two batch edits. 

It's not clear to me what is being tested in this second batch edit -- didn't the first one already test adding a collection? If we want to test the ccondition of a record with a) no collections, b) a different collection, c) already had collection2 in it, can we just make three records with all three conditions for first batch edit?

Alternately, we could make a whole extra test case just for collection editing -- that's what "best practice" of test design would be... but it makes the test slower to do multiple requests like that. But packing multiple requests into just ONE test is the worst of both worlds...

eddierubeiz: Fixed it!

PR #1232, opened 2021-07-06
add newrelic_rpm gem for monitoring with new relic service, via heroku.

jrochkind:
https://docs.newrelic.com/docs/agents/ruby-agent/installation/ruby-agent-heroku/

The free "Wayne" plan on heroku seems to be free forever. The main things it doens't include is any kind of alerting/notifications. It seems to let us look at what we're going to want to look at to see average response time or dig into the slow responses. I think! It'll be easier to see when we have more data. 

https://elements.heroku.com/addons/newrelic

Hmm, let me deploy this to legacy staging first to make sure the newrelic gem alone without configuration is just no-op.... seems to be fine, without license keys etc configured, I think it just does nothing?

No config file is necessary on heroku, as all minimally necessary new relic config is provided via heroku config vars -- automatically when you turn on new relic as a heroku add-on.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,11 @@ gem 'lockbox'
 gem 'rails', '~> 6.1.1'
 gem 'webpacker', '~> 5.0'
 
+
+# New Relic monitoring tool
+gem 'newrelic_rpm'
+
+
 # lock blacklight to current MINOR version. While BL minor version releases
 # are theoretically backwards compat, experience shows they often cause problems.
 # So you need to manually change this spec to allow updates when you want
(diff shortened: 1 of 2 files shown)


jrochkind:
New Relic will do in-browser moniitoring of actual browser page rendering times and things like that via a Javascript agent, but I am not installing that yet here. 

https://docs.newrelic.com/docs/browser/browser-monitoring/installation/install-browser-monitoring-agent/

jrochkind:
@eddierubeiz OK let's merge this simple one in so we can have new relic in production when we go live!

Issue #1201, opened 2021-06-09
Customize heroku maintenance page


jrochkind:
With SHI branding etc. maybe a way to automatically generate it with a speciifc message of expected lenght of outage etc. 

https://devcenter.heroku.com/articles/maintenance-mode#customizing-your-maintenance-page

eddierubeiz:
This breaks down into two parts:
- Part 1 (difficult): Generate branded, customized HTML to use as a maintenance page. The error message needs to be customizable. The HTML can't refer to any assets served by the digital collections, since it's supposed to function properly when the site is down. We could steal a page from the main site, e.g., and customize that.


- Part 2 (easy): 
```
aws s3 cp maintenance.html s3://some_public_bucket/maintenance.html
heroku config:set MAINTENANCE_PAGE_URL=//s3.amazonaws.com/some_public_bucket/maintenance.html --app scihist-digicoll-staging
heroku maintenance:on --app scihist-digicoll-staging
```

Just as a proof of concept I have something up at `https://s3.amazonaws.com/scihist-digicoll-production-derivatives/maintenance/maintenance.html` but the page could really go anywhere, including, honestly, the main Institute website.

jrochkind:
Here is what current draft of our customized "down for maintenance" page looks like, fyi @MDiMeo and @apinkney0696 . This seems pretty good to me, but let me know if you think it has any issues. 

![Screen Shot 2021-07-12 at 2 57 40 PM](https://user-images.githubusercontent.com/149304/125341526-aebd4e80-e321-11eb-9445-7ddd7970febf.png)

PR #1236, opened 2021-07-08
Independent completely static HTML page for a maintenance/outage message.

jrochkind:
For use with heroku. Ref #1201

Only tricky part is deciding where we put this -- on some S3 bucket, yes. But we have a kind of messy practice now. 

Our current "derivatives" bucket just has top-level keys for derivatives, it's all just spread out in there. In retrospect, we should have put derivatives in a prefix `derivatives/`, but at the time we thought that would be an S3 performance problem (either we were wrong, or it no longer is). 

As our only public bucket, and because we already have a LOT of buckets, we have sometimes put other stuff in the derivatives bucket that isn't really derivatives, or isnt' the same as the other derivatives.  Like `__sitemaps/`, and maybe combined derivatives. That then have folders that are on a peer level with individual work derivative folders; what a mess! And requires custom exceptions in our derivative orphan checker. Argh!

It is, btw, difficult/expensive to rename a bucket or 'move' keys into a folder -- basically requires copying all objects, and we have a lot of derviative objects. 

Still, we should probably stop making the problem worse... even though we already have a lot of buckets, maybe we create one more per environment (so two more, prod and staging), like `scihist-digicoll-production-public` and staging equivalent -- to be used for all future PUBLIC (not acccess controlled) S3 content, with different kinds of content in different folderes/prefixes. We can start with this one. Maybe we can later move __sitemaps etc there, but anyway, we aren't making the problem worse. 

We do currently control S3 buckets in ansible playbook, so to do this we should add it to the ansible S3 playbook.  And I guess find what documentation we have of S3 bucket, and add this to it too...

What do you think about that plan, @eddierubeiz? I don't love it, but it seems the least terrible; I don't really feel like adding yet antoher special purpose non-derivative subdirectory the derivatives buckets....


eddierubeiz:
`scihist-digicoll-production-public` and `scihist-digicoll-staging-public` are the way to go, yes.

jrochkind:
OK, this is done! Created new buckets, put HTML in the production one, set heroku config vars on production  to point to it. 

(Staging is currently unset, as I try to demonstrate a bug/problem/unexpected thing with *unsetting* it to heroku support). 

Before closing this, please merge ansible bitbucket PR to S3 playbook to create `public` buckets. https://bitbucket.org/ChemicalHeritageFoundation/ansible-inventory/pull-requests/154/s3-playbook-creates-public-bucket-for

Otherwise, I think this is done.

eddierubeiz:
BitBucket PR merged.

PR #1242, opened 2021-07-12
Suppress render_layout logging too

jrochkind:
We had tried to suppress all view/render/partial logging in production, but missed this one.


eddierubeiz:
Good. We just don't need render layout info in prod.

PR #1243, opened 2021-07-12
use custom logger formatter to avoid logging things heroku is adding in anyway

jrochkind:
or things that are irrelevant/unintelligible on heroku. Help reduce our log data, for things like papertrail that have paid plans baesd on log data.

In fact, all we're adding with the formatter is log level (info, debug, etc) -- maybe we don't even need that, and could use no log formatter at all, but I like seeing it and being able to search on it, so we'll leave it for now.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,3 +1,5 @@
+require 'scihist_digicoll/logger_formatter'
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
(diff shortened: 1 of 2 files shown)


jrochkind:
Example before in papertrail:

     Jul 12 17:07:05 app/web.1 I, [2021-07-12T17:07:05.379494 #1634]  INFO -- : [edbaa9a0-139b-4a9b-8685-4d226c0d737c] Completed 200 OK in 27ms (Views: 7.4ms | ActiveRecord: 2.5ms | Allocations: 13226)


Example after in papertrail:

    Jul 12 17:05:30 app/web.1 INFO: [bbacd828-6ab4-40ba-9639-883d4529ecc6] Completed 200 OK in 109ms (Views: 47.3ms | ActiveRecord: 54.5ms | Allocations: 21668)

Issue #1136, opened 2021-05-12
Fixity check report consistently reports items that have not been checked for 90 days.


eddierubeiz:
The fixity check report https://digital.sciencehistory.org/admin/fixity_report consistently reports the following "problems":

- roughly 4000 items with _"Fully ingested, stale checks (have not been checked for 90 days, or ever)"_
- A _"Fully-ingested Asset with oldest most-recent fixity check"_ with a latest check date of 100 days ago, instead of 90.

We strongly suspect nothing is actually wrong. We need to change either the reporting recipes, or the fixity check routine, or both, so that this page reports the usual behavior of our fixity checks as normal, and only flags aberrant behavior.

See also https://github.com/sciencehistory/scihist_digicoll/issues/633, which this ticket will replace.

jrochkind:
Some nightly logging might be good, especially now that we have accessible logs. 

Logging *every* asset fixity checked will be too much (for our papertrail plan).  But maybe some summary "descriptive statistics"

*  number we are going to check. Within that, how many have *no* previous fixity checks, vs oldest ones?
* total number found not checked in PERIOD days, which may be greater than the number we are going to check
* A *sample* record we're going to check that had  no previous fixity check, including it's created_at date. 
* A sample record we're going to check that did have a previous fixity check , including it's created_at and date of last fixity check.

jrochkind:
We decided we will:

* do an initial fixity check right after ingest for newly ingested files (#1263), which we think will fix the problem with our algorithm checking 1/90th -- the new ones were pushing aside old ones due for a check. So we do new ones right away still, but not in the nightly cronjob, taking up those 1/90th slots, but instead just in the ingest process. 

* While that should take care of it, we also, just in case, adjust the nightly job to check if there any assets past their 90-day required fixity check, and if so, just check them too! Just an additional cleanup step for any problems, not expected to be triggered much. In #1267

PR #1250, opened 2021-07-21
Add 10% of assets to the nightly fixity check

eddierubeiz:
Ref #1136 .

Our current estimate of how many assets to check every night would work perfectly if the number of assets in the collection always stayed the same.

Unfortunately, we are always ingesting new stuff, so that total **increases** over the course of the 90-day check cycle.

Thus, our **current** total is always **less** than the average total over the course of the entire cycle, resulting in a consistent lag.

Let's add 10% more assets to each nightly check to help correct this.

### DIFF
--- a/app/lib/scihist_digicoll/assets_needing_fixity_checks.rb
+++ b/app/lib/scihist_digicoll/assets_needing_fixity_checks.rb
@@ -2,7 +2,7 @@ module ScihistDigicoll
   # This recipe is used in check_fixity.rake .
   #
   # Running fixity checks takes a while, so we want to spread
-  # the job out over a week or so. The current recipe will:
+  # the job out over 90 days or so. The current recipe will:
   #
   #  * Check all the assets if you run the rake task 7 times.
   #  * Check the same number of assets every day.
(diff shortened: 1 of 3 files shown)


jrochkind:
For non-urgent things, I'd prefer if you wait until I get back to merge/close them, if you can!

Can we do the arithmetic/algebra to figure out the actual number, instead of just estimating 10%?  We know how many assets were added when, it should be possible to figure out the formula.  

For something as important as fixity audit, I don't like the idea of just "estimating" 10%, which depending on the actual ingest history could still be too high or too low, still leading to both false negatives and false positives.

jrochkind:
Also, we put things in "review" instead of closing them, in general, right, even after merging? Would appreciate that, in this case so I get a chance to see it!

eddierubeiz:
Sorry, I closed this by mistake! It's reopened now. I don't intend to merge this before you get back-- of course.

jrochkind:
Cool, no worries!

Also as far as figuring out the formula -- if we can figure out the right arithmetic/algebra, we could consider actually INCREASE the number checked each night, calculated based on recent ingests, such that after 90 days the original pre-condition still will be met!  IF that makes sense. I don't really understand what's going on, and the real challenge here is actually figuring out what's going on, specifically, in terms of math, not just the hypothesis that it makes sense that it's got to do with ingests. I want to actually see/understand the math of what's going on, and then we might choose to base the number checked each night on it, or we might choose to change the calculation of what we expect before flagging a problem state, or both.

jrochkind:
When I try to think thorugh it on the back of the napkin... it actually seems to me that it still should work fine, that new ingests shouldn't mess it up. I can't figure out why they would, new ingests increase the total count, so increase 1/90th, so I don't totally understand how they result in anything going more than 90 days without a check. 

I'd try to maybe do a little pen-and-paper exersize pretending like we start with 4 assets and are doing 1/4th each night and all checked within 4 days, and then add 3 on a day, or something, and see what happens when I play it all out. Or maybe 4 isn't enough, have to go to like 10 and 20 existing assets or something. But I really want to actually understand what's going on to ensure that this *really* is the reason it's failing, instead of just guessing/assuming it. And then ideally mathematically model it so we can solve it with exact numbers.

eddierubeiz:
`Also as far as figuring out the formula -- if we can figure out the right arithmetic/algebra, we could consider actually INCREASE the number checked each night, calculated based on recent ingests, such that after 90 days the original pre-condition still will be met! IF that makes sense. `
That does make sense! That's exactly what this PR is doing. The tough part is predicting how many items will be added in the next 90 days. If we can do that, we will have an exact percentage of extra assets to check each night.

eddierubeiz:
I made a little spreadsheet that explains why I think my idea is a good one -- will show you when you get back and we can continue the conversation. Won't look at it (much less merge) until then.

jrochkind:
Closing, because we're not going to do it this way.

PR #1251, opened 2021-07-22
Fix or work around tom_select hack.

eddierubeiz:
Sometime this week, the CI tests started breaking because they were failing on the the rspec helper `tom_select.rb`.
It's unclear why, and it's impossible to reproduce in dev.
In the two files that used the helper, let's just:
- click inside the`tom_select` widget
- adjust the multi-select options by clicking and typing as a user would.

Works fine, simpler, less hacky, and the tests pass again.

Ref #1252


eddierubeiz:
Merging this myself as @jrochkind  is on vacation and I need CI tests to pass.

jrochkind:
nice, thanks! *this* one is of course urgent to merge without me, agreed!  

I wonder why it stopped working, when the versions of everything should be locked! Oh well!

jrochkind:
Can you file a ticket with tom-select though, if it is a helper provided by the gem that broke?

eddierubeiz:
No, we created that helper. A comment in the deleted file mentions:
`This is a really hacky approach using execute_javascript, but it works. Not sure if there's a better way, we could try actually interacting with the on-screen tom-select-provided UI, but we're taking the easy way out for now.
`So I felt fine getting rid of it.

jrochkind:
OK, cool, thanks, merge is fine!

jrochkind:
Except, perhaps, I would have rewritten the helper to do it the "hard way", so tests could still use the helper instead of having to copy and paste that boilerplate for selecting from it. But maybe doesn't matter I dunno.

eddierubeiz:
Absolutely open to doing that!

PR #1254, opened 2021-07-27
Beef up parameter checks in collection show controller #facet method.

eddierubeiz:
Ref. #1204


jrochkind:
Nicely done!

In the past we've tried to add tests for catching these bad params. Can you add a test, perhaps in `./spec/requests/bad_blacklight_requests_spec.rb` (I think best, if it can be done conveniently), or `./spec/controllers/catalog_controller_spec.rb`?

eddierubeiz:
Absolutely; stay tuned.

eddierubeiz:
I noticed a parallel problem with the featured topic controller and decided to fix that as well as part of this PR.

PR #1263, opened 2021-08-06
Check newly-ingested files' fixity right away -- don't wait until the nightly fixity check job.

eddierubeiz:
A first step in fixing ref #1136, after an in-depth conversation on Slack this morning.


eddierubeiz:
Test assets with trait `:inline_promoted_file` needed to have `queue_adapter: :inline` added to ingest-related tests.

Issue #1246, opened 2021-07-19
avoid "upscaled" download derivatives


jrochkind:
In some cases our original TIFFs can be fewer pixels than our standard "large" size download derivative. 

In these cases, the software is currently still making the large download and offering it for download. This is not helpful, a zoomed up image. We should:

* avoid creating this thumb on ingest if "full size" is smaller than it in either dimension
* not offer it for download (this *might* happen automatically if it's not there, it wont' be offered)
* go through existing assets and remove any already created useless downloads

This could hypotehtically happen for "medium" or "small" size too, although it would be unusual, we should prob do same logic for those.

jrochkind:
- [x] avoid creating this thumb on ingest if "full size" is smaller than it in either dimension -- done in #1262 
- not offer it for download (this might happen automatically if it's not there, it wont' be offered) -- indeed it turns out this already happens automatically. 
- [ ] go through existing assets and remove any already created useless downloads -- TBD

jrochkind:
Thought of another part of this -- the WorkPdfCreator tries to use `download_medium` derivatives to make the PDF, but after we do this, they might not be there! 

Need to adjust it to fall back to download_small, or _original. (_original will always be there, so actually if _medium isn't we should just use original)

PR #1265, opened 2021-08-09
Rake task to remove download_* derivatives larger than original

jrochkind:
These are useless and upscaled. We changed code to not create them, but need to go through and delete any previously created before we made that change.

Ref #1246

This has been run on staging -- it actually deleted derivatives from 13,000 assets! (Out of 47,500 total assets, so around 27% of our assets were actually not big enough that they should get a large size download, at least 2880 pixels. Wow, our original asset resolutions really vary a LOT more than we realized! Ramelli originals are like 6000 pixels wide, I thought that was typical, not that other originals would be less than half the resolution!). 

Here's the log of what was deleted: https://gist.github.com/jrochkind/77118d0f8a7e8a5047beab0575bf62ed

The weird restore_attributes line is about some weirdness involving Rails seeing changes to be saved because of default attribute values we added, I just worked around it for now. 

After merge, this needs to be deployed to and run on production.

### DIFF
--- a/lib/tasks/data_fixes/remove_upscaled_download_derivatives.rake
+++ b/lib/tasks/data_fixes/remove_upscaled_download_derivatives.rake
@@ -1,0 +1,34 @@
+namespace :scihist do
+  namespace :data_fixes do
+
+    desc "Remove download_* derivatives larger than original, mistakenly created"
+    task :remove_upscaled_download_derivatives => :environment do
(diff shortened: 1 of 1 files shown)


eddierubeiz:
This has not been deployed or run in production yet. Plan is to do it tomorrow morning Tuesday August 10th.

eddierubeiz:
Deployed and ran the rake task in production.

PR #1267, opened 2021-08-09
rake scihist:checks_fixity:complete_overdue

jrochkind:
Refactor/add method to FixityReport so we can get the actual list of assets, not just the count, that are displayed in dashboard as problematic. FixityReport is a little bit messy, all those existing methods probably should have had a name making it clear they were returning a count. Oh well.

Addresses part of #1136 

Then add rake task that uses this to just... check them. Since planning on doing from a scheduled job, no progress bar. Just outputs to log if we found any stale ones and checked em... where we will see that it happened only if we go looking for it. Does it need any other kind of alerting? Not sure.

Tried running this on staging -- looked like it would take around 2 hours to complete for the ~2400 overdue assets. So haven't completed it yet, so can't say for sure that it gets that report number down to 0, but it should?

After merging and deploying, we just need to change the Scheduler line that says `bundle exec rake scihist:check_fixity` to say `bundle exec rake scihist:check_fixity scihist:check_fixity:complete_overdue`, and then every night after doing it's usual 1/90th, it will ALSO do any extra that are overdue. Or at least that's the intent. 

So if we merge on production without running manually, the first night it's live the nightly fixity check job will take an extra 1-2 hours.

### DIFF
--- a/app/models/fixity_report.rb
+++ b/app/models/fixity_report.rb
@@ -109,6 +109,20 @@ class FixityReport
     ])
   end
 
+
+  # an ActiveRecord relation corresponding to #not_recent_with_no_checks_or_stale_checks, but
+  # returning a relation you can get actual assets from, not just count.
+  def need_checks_assets_relation
+    build_query([
+      stored_file_sql,
+      "#{recent_asset_sql} = false",
+      "(#{stale_checks_sql}) OR (#{stale_checks_sql} is NULL)"
+    ], count_only: false)
+  end
+
+
+
+
   # Of the assets that have fixity checks, which has the OLDEST most recent
   # check, and what is it?
   #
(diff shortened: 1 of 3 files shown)


jrochkind:
merged, deployed, scheduler line changed on prod and staging.

eddierubeiz:
Awesome. Now we wait.

jrochkind:
> outputs to log if we found any stale ones and checked em... where we will see that it happened only if we go looking for it. Does it need any other kind of alerting?

I remembered about an appropriate tool for this!

In production **papertrail**, I saved a search for `complete_stale_checks:` (oops, the log label doesn't match current rake task name but we'll leave it, since I'm created a saved search for it)....  and then use papertrail features to send alerts once a day if this has happened.... to slack and email to digital-tech@sciencehistory.org.  

This way we will be able to keep an eye on how often these "overdue cleanups" are happening (we don't really expect them to happen at all).

PR #1270, opened 2021-08-10
Adding the correct packages for brandoncc's version of heroku-buildpack-vips

eddierubeiz:
Ref #1264
This assumes:

`heroku buildpacks`
returns
```
1. heroku-community/apt
2. https://github.com/brandoncc/heroku-buildpack-vips
3. heroku/ruby
```
and will install `vips-8.10.6`, which is great.

I'm sure this could have been working a long time ago, but I misread the installation instructions for the buildpack and assumed this buildpack was a non-starter until @jrochkind pointed out I had never correctly modified the Aptfile as described in the documentation.

All's well that ends well: with this PR we go from 8.10.2 to 8.10.6 and yes, all derivatives work great.

### DIFF
--- a/Aptfile
+++ b/Aptfile
@@ -1,5 +1,12 @@
 mediainfo
 imagemagick
+
+# These three packages are required
+# by https://github.com/brandoncc/heroku-buildpack-vips
 libglib2.0-0
 libglib2.0-dev
 libpoppler-glib8
+
+# `poppler-utils` is not installed as a dependency of libpoppler-glib8,
+# and we need it for  `pdfunite`, which we use for on-demand PDF derivatives:
+poppler-utils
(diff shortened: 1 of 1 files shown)


jrochkind:
Comment by poppler-utils says 

> we need `pdfunite` command line util from it, so we'll list it explicitly.

Can you confirm we still have CLI `pdfunite` on our dynos without that there? (That's used for on-demand PDF creation). 

Also... if we don't know what's putting it there, I wonder if we want to leave `poppler-utils` as a way of getting it... unless having poppler-utils ruins recipe for brandoncc vips....

eddierubeiz:
It doesn't ruin `brandoncc` vips, and it does make `pdfunite` available, so it's back in.

jrochkind (approved):

Issue #607, opened 2020-01-29
If an ActiveJob job fails, rerun it automatically.


eddierubeiz:
@jrochkind  notes:

> At some point we should probably configure “run any job that fails a second time automatically” for our bg jobs. I think that is possible to do generically. And all our jobs should be idempotent now, so.

jrochkind:
I think this one would actually be a nice one to take care of on further reflection, so I took the liberty of moving this to the top of backlog, to remind us to check in on it again and consider doing it, when we have time.

jrochkind:
There is retry func built into ActiveJob.  Generically, it allows you to do things like "retry in 5 minutes, the second time 20 minutes" etc. 

However, basic Resque that we're using doens't support "schedule in advance".  So we can only do "retry immediately". 

Still, a "retry immediately, only once" would be sufficient to resolve a number of problems -- including "jobs got interupted because of worker dyno restart/shutdown".  And is probably fairly straightforward and worth doing. 

I might like to propose this as moving clsoer to the top of our lists. 

(See also #1245 for switching to a non-resque backend that supports future-scheduled jobs)

jrochkind:
Hm, I knew that resque by default didn't support future-scheduling jobs, but I thought we could use ActiveJob `retry_on` anyway, telling it `wait: 0`, no future scheduling. But even so we get an error:

`Aug 10 13:05:24 app/worker.1 ERROR: [ActiveJob] Failed enqueuing TestFailureJob to Resque(default): NotImplementedError (To be able to schedule jobs with Resque you need the resque-scheduler gem. Please add it to your Gemfile and run bundle install)`

Still messing with things, but this may mean our current stack wont' allow ActiveJob retry. :(  I don't really want to add resque-scheduler into the mix, although that would be an option.

jrochkind:
We could work around this several ways, monkey-patching something, or writing our own thing in terms of `rescue_from` instead of `retry_on`. 

(The definition of retry_on is not too complicated in terms of rescue_from.... https://github.com/rails/rails/blob/94ccd5410d84313be48dc949fdf1d185c98a9d7f/activejob/lib/active_job/exceptions.rb#L58)

I'm not sure if it's worth it.

PR #1273, opened 2021-08-11
ActiveJob retry_on to retry ALL failed jobs once, still with resque

jrochkind:
Ref #607 

* retry once for StandardError (basically any error) in ApplicationJob super-class of all our jobs. The intent is that any failure should be re-enqueued for a retry ONCE, immediately, before giving up. This should resolve intermittent errors or errors caused by the Job dyno being shut down mid-job.

* Basic resque isn't able to schedule future jobs. Patch the resque adapter so it IS willing to `retry_on` with `wait:0`, just by doing an ordinary job enqueue.

* Retry logging is not great. Customize the ActiveJob retry logging by making a custom rails log subscriber for ActiveJob (based on the default one, by subclass), and using our custom one in place of the default one.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,2 +1,17 @@
 class ApplicationJob < ActiveJob::Base
+  # All our jobs should be idempotent, let's automatically retry them all for any
+  # error.
+  #
+  # This can for instance automatically take care of long-running jobs interupted by a host restart.
+  #
+  # Resque, which we ae are using at the moment, by default doesn't support
+  # future-scheduled jobs, so we retry just once, immediately. Could have
+  # a more sophisticated retry pattern with a back-end that supports future-scheduling.
+  retry_on StandardError, attempts: 2, wait: 0
+
+
 end
+
+
+
+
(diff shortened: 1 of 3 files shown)


jrochkind:
Custom logging is in part set up so we can set up papertrail alerts, for instance `[ActiveJob] AND retrying` to have papertrail let us know when/how often retries are happening, with, once you jump into the logs, the necessary context to see what was going on. 

The default logging made it hard to target just these messages ("ActiveJob" didn't appear in the line), and also made it hard to then see it in context -- the job_id didn't appear in the line, with the job_id you can click on it in papertrail to see all the other lines from that job_id.

jrochkind:
Re logging... there is this gem, but I'm not sure how maintained/currently working it is, or if it just handles exception logging.  We could try to update/write it or an alternative ourselves....

https://github.com/armandmgt/lograge_active_job

jrochkind:
Blog post with things discovered and choices made: https://bibwild.wordpress.com/2021/08/23/notes-on-retrying-all-jobs-with-activejob-retry_on/

jrochkind:
I created a Papertrail saved search and alert on `[ActiveJob] AND Retrying`. https://papertrailapp.com/searches/123664271/edit`

It alerts to Slack errors channel (I think?), and email digital-tech@sciencehistory.org. 

This way we can have some sense of how often retries are happening!  The custom logging was in part to make sure we had clear unique things to filter on for this alert, as well as enough info in the log messages so filtered to have a sense of what was going on and be able to find out details!

eddierubeiz:
Thanks for the extra-helpful narrative in the blog post -- makes it a lot easier to see what's going on here. Looks like this was fun to work on!

Issue #1276, opened 2021-08-19
Add extents to and publish a bulk set of records


apinkney0696:
[Booth extent edits.xlsx](https://github.com/sciencehistory/scihist_digicoll/files/7016056/Booth.extent.edits.xlsx)
I have a set of 385 digital records that need to have an extent applied. In the attached spreadsheet I've included the record ID, the extent data, and whether or not the record can be published. Since this is our first time trying something like this, please let me know if you need to have this formatted differently or anything else. Thanks!

apinkney0696:
Hi @eddierubeiz !

Thanks for catching that. You're right, the first measurement belongs to [this letter](https://digital.sciencehistory.org/works/2een6yq), and the second measurement belongs to [this other letter](https://digital.sciencehistory.org/works/nw0789v).

PR #1285, opened 2021-08-25
Rake tasks to add a bunch of extents to the Booth collection, and publish a number of items in it.

eddierubeiz:
Ref #1276


jrochkind (approved):

eddierubeiz:
Sounds good. I need to make a couple changes before this is ready, actually.

PR #1286, opened 2021-08-25
upgrade blacklight to latest version 7.19.2

jrochkind:
Deal with confusing deprecation issues. 

upgrade blacklight_range_limit to 8.0.1.


jrochkind:
OK, we've silenced *all* deprecation notices from Blacklight (and from the `qa` gem too, because it uses the same cbeer "deprecation" gem). 

This is not great -- but we 1) we want to upgrade blacklight 2) even though we couldn't figure out how to get rid of those deprecation warnings and 3) couldn't figure out a better way to silence the "yeah we already know about those and don't know how to resolve them" warnings without silencing all warnings. 

So be it.

jrochkind:
@eddierubeiz I'm cautious to actually merge this, want to test it more, but you can review/approve it starting now, I think it's basically done.

jrochkind:
OK, my unfortunately long blog post about investigations into Blacklight upgrade and deprecations, for background. https://bibwild.wordpress.com/2021/09/02/blacklight-7-x-deprecation-of-view-overrides-paths-forward/

eddierubeiz:
Yeah, I've been keeping an eye on this for some time.

eddierubeiz:
Thanks for the blog post, btw, super helpful to have this context.

PR #1290, opened 2021-08-26
Update method call for ruby 3 in work_pdf_creator_spec

jrochkind:
Another example where ruby 3's more careful distinction between hash arguments and keyword arguments require a differnece in how the method is called, we need to use curly braces `{}` to make it clear we are passing a hash as the first argument to match method signature. Required in ruby 3, works fine either way in ruby 2.7. Let's get ready for 3.

With this and the last change, the branch with upgraded blacklight passes tests in ruby 3, wow!


jrochkind:
fixed.

Issue #1293, opened 2021-08-30
sign up for free for open source Scout APM account


jrochkind:
Switch our production app over to use that instead of heroku-plugin (which we can't get free for open source); remove heroku plugin. 

> Scout APM is free for open source! Contact us for a free account support@scoutapm.com.

https://scoutapm.com/info/pricing

Email from Scout support said:

> Thank you for reaching out! We do offer free plans for open source projects up to our large tier, however we are not able to provide this through Heroku because they have a separate billing system than us. 

> To get set up with an open source project with us you first need to switch to a direct Scout account. We are able to monitor on any environment for your monitoring needs, we would just need a github link to confirm that it was indeed an open source project. 

> To get set up with a direct Scout account here are the steps you would need to follow:
> * Visit Scoutapm.com and login to your new/personal account and input your billing information.
> * Please change your Scout key in Config Var in Heroku to SCOUT_KEY = "" (your new key will become available once your Scout account is created)
> * If you need to maintain the visibility into Heroku for data purposes I would suggest downgrading your account to the free plan.

eddierubeiz:
Got word from Scout administrators that they've applied our open source "coupon" and we should be paying $0 for our `Scout APM Pro` plan from now on.
Note that the [billing settings page](https://scoutapm.com/billing) does not reflect any evidence of this at the moment.

jrochkind:
Sweet! So last step is switching our production instance to use that scout account, de-provision the heroku add-on scout account?

eddierubeiz:
Correct, and I'm going to go ahead and do that this morning.

eddierubeiz:
I de-provisioned the old Heroku add-on from both staging and production.
The `SCOUT_KEY`, 'SCOUT_LOG_LEVEL`, `SCOUT_MONITOR` and `SCOUT_NAME` config vars on [our production instance](https://dashboard.heroku.com/apps/scihist-digicoll-production/settings) are now set to the proper values.
Production should now be reporting to the new account.

jrochkind:
Sweet! How do I log into the new account to see scout? Just go to scoutapm.com and log into my personal account I guess?

Should we document this somewhere? So people (say hypothetical new staff) would know that we have Scout and how to get there?

 Maybe either on https://chemheritage.atlassian.net/wiki/spaces/HDC/pages/1915748368/Heroku+Operational+Components+Overview, or mention scout on there and link to another page? 
 
[...]

PR #1298, opened 2021-09-01
Adding a scout config file.

eddierubeiz:
Ref #1293 
Following instructions on the Scout instructions.
Set log level to `warn`, per current setup.

### DIFF
--- a/config/scout_apm.yml
+++ b/config/scout_apm.yml
@@ -1,0 +1,36 @@
+# This configuration file is used for Scout APM.
+# Environment variables can also be used to configure Scout. See our help docs at https://docs.scoutapm.com#environment-variables for more information.
+common: &defaults
+
+  # key: Your Organization key for Scout APM. Found on the settings screen.
+  # - Default: none
+  key: PKLv05OWC7djNRel9V8O
+
+  # log_level: Verboseness of logs.
+  # - Default: 'info'
+  # - Valid Options: debug, info, warn, error
+  log_level: warn
+
+  # name: Application name in APM Web UI
+  # - Default: the application names comes from the Rails or Sinatra class name
+  # name:
+
+  # monitor: Enable Scout APM or not
+  # - Default: none
+  # - Valid Options: true, false
+  monitor: true
+
+production:
+  <<: *defaults
+
+development:
+  <<: *defaults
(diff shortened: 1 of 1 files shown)


jrochkind:
He we already had Scout totally set up and working, without the Scout config file I guess. 

My expectation was that all we'd have to do to do the changeover to new Scout account is change the `SCOUT_KEY` heroku config var per the support instructions:

> Please change your Scout key in Config Var in Heroku to SCOUT_KEY = "" (your new key will become available once your Scout account is created)

No problem playing around with it in staging to see what you can do though!

But I think when I set it up the first time, I guess I got by without the config file, maybe just accepting defaults?  I maybe followed different instructions that didn't tell me to make the config file, I don't know. 

But we do already have Scout totally working -- I think all we need to do is switch to the new Scout account we are setting up, by changing the heroku config var?

jrochkind:
I think we decided not to do this maybe, or not like this?

eddierubeiz:
This file isn't necessary, but if we do merge this, we can remove the `SCOUT_KEY`, `SCOUT_LOG_LEVEL`, `SCOUT_MONITOR` and `SCOUT_NAME` from our production Heroku settings, and store that info in the codebase instead. I do prefer the file to the settings vars, but I don't feel strongly about it.

jrochkind:
As a file, we'd have to make sure it was only enabled in production, or at least had different values in production to keep the staging data separate from the production data (which is recommended by scout docs and also makes sense to me)... so we wouldn't want to merge this as-is... I guess it would have to be ERB (if that file can even be ERB?), that takes values from ScihistDigicoll::Env to have differences in prod and staging?

As it happens, I am currently noticing how every time you run a rake task in staging, the scout gem logs to stdout complaining that Scout isn't configured... I just figured out the solution to this currently when using env to configure, is simply setting config `SCOUT_MONITOR=false` is staging. Which is convenient. 

The config file just seems more complicated to me, with that additional stuff to figure out...

eddierubeiz:
Oh, if we did want this file we would need to add `monitor: false` under `staging`. (Equivalent to the above.)

eddierubeiz:
Closing for now; if we change our mind we can always merge this.

PR #1302, opened 2021-09-02
S3 orphan report

eddierubeiz:
Ref #1142 . Creates a table in the DB to store the most recent orphan check, and a controller and view in the admin pages to display the data.
Assumes the following task is added to a nightly scheduler:
bundle exec rails runner 'OrphanAuditor.new.perform!'


jrochkind:
Honestly my eyes just glaze over looking at this, I'm having trouble giving it that much review, so sorry if these questions aren't based on a close read, just more general questions. 

Actually creating DB records for the reports seems kind of excessive to me... we dont' really care about keeping these all hsitorically. But I guess you have to put the results SOMEWHERE after running them nightly, for someone to see them. Is there something in place to make sure we odn't keep these historical reports forever? We really only want the last one. 

Is there a rake task that's going to be run nightly? I don't see a rake task in this PR. Was there already a rake task? Or what's going on there? Is there something that needs to be changed in heroku scheduled tasks with this PR, can you specify in PR description?

Does this currently *alert* on orphans, or it's just sitting there for you to look at it manually sometimes? 

Especially if it alerts, but either way, before deploying this, I want to know how many it's going to find -- in staging and production. Have you run in staging and prod, and spent some time looking at the results to understand what they are and that they aren't false positives?  I'm curious if we get any orphans in production. 

Is there something built into this to clean up the orphans? Or you just clean them up manually with the rake tasks we already have?  Can the admin screen include those instructions?  On staging anyway, can we verify that after running the rake tasks to clean up all orphans, the admin screen reports no orphans, that the two things agree?

eddierubeiz:
_Actually creating DB records for the reports seems kind of excessive to me... we dont' really care about keeping these all hsitorically. But I guess you have to put the results SOMEWHERE after running them nightly, for someone to see them. Is there something in place to make sure we odn't keep these historical reports forever? We really only want the last one._

We're only keeping the last one. This is modeled on the storage report, which has similar aims.

eddierubeiz:
_Is there a rake task that's going to be run nightly? I don't see a rake task in this PR. Was there already a rake task? Or what's going on there? Is there something that needs to be changed in heroku scheduled tasks with this PR, can you specify in PR description?_
Done.

eddierubeiz:
_Does this currently alert on orphans, or it's just sitting there for you to look at it manually sometimes?_
It alerts Honeybadger.

eddierubeiz:
_Is there something built into this to clean up the orphans? Or you just clean them up manually with the rake tasks we already have? Can the admin screen include those instructions? On staging anyway, can we verify that after running the rake tasks to clean up all orphans, the admin screen reports no orphans, that the two things agree?_

No, the cleanup is separate and is done manually. The admin screen includes those instructions. Yes, in staging, cleaning up all orphans does in fact result in zero orphans.

eddierubeiz:
Orphans we can expect to find in production:
- Originals: 37 out of 49,663. Based on random checks, all appear to be legitimate orphans (asset missing.)
- Regular derivs: 10 out of 429,369. All are discarded combined audio from oral histories and clearly orphans.
- Restricted derivs: zero out of 2,754.
- DZI: 2 out of 43,898.

jrochkind:
@eddierubeiz Thanks for that investigation and for documenting the findings here! It's good to see we didn't really have very much at all, it's not like we're going to save tons of money from finding and eliminating these or anything! 

My suggestion is to clear production to 0 before deploying. So instead of "orphans we can expect to find in production", let's eliminate them BEFORE we deploy this, so we're starting from a fresh slate with the report. What do you think?

In staging too I guess. (Staging is easier, we can just clear them out without even investigating them).

jrochkind:
I'm honestly still a little bit worried about the financial implications -- scanning all the buckets nightly (!) for orphans has a cost associated with it. Will it be more than we will save from eliminating the orphans?

@eddierubeiz could you try estimating the nightly scanning cost? And estimating the monthly storage cost of *these* orphans you found this time?

eddierubeiz:
If it's OK with you, I would actually like to merge/deploy first and run the **latest** version of the reporting code against the **actual** production buckets before actually deleting anything in production.

The combination of a) syncing prod to staging and running the latest reporting / deleting code in staging, and b) running the version of the reporting code currently on the master branch against production was a helpful one. I believe we will find few or no surprises when we run the reporting in prod. That said, doing things in this order could serve as a helpful check for my work.

eddierubeiz:
Attaching [a list of deleted s3 files from production](https://github.com/sciencehistory/scihist_digicoll/files/7254403/deletions.txt).

PR #1312, opened 2021-09-08
display of transcription and translation

jrochkind:


jrochkind:
ready for review/merge, but it would be good to get feedback on the small-screen (eg mobile) alternative to tabs UI.

Issue #1303, opened 2021-09-03
Convert sets of records into child records


apinkney0696:
Eddie ran a list for me of all the records that use the Source field in the digital collection last week. (#1281)

As I suspected, these were all records created in the early days of the Digital Collections when the parent-child hierarchy capability didn't exist. I would like to convert these into child records under a new parent for each book, then remove any data in the source field. I can do this manually on my end, but it might take me a while and I wanted to see if it could be an easily automated process from the back-end. 

I have already created the new parent works and grouped the records that will become the children in this spreadsheet: 
[Source_Field_Project.xlsx](https://github.com/sciencehistory/scihist_digicoll/files/7105919/Source_Field_Project.xlsx)

I would ultimately like to remove the source field from the metadata form, but I can make another ticket when we're ready for that.

jrochkind:
I just want to note:

I remember when we *added* the "Source" field, and the parent/child hieararchy capability *definitely* already existed. 

Repository managers at that time just chose not to use it for these records for whatever reason -- I remember they specifically asked for the "Source" field *because* they didn't want to put a parent record in, they only wanted to put in isolated page(s) -- maybe they were only going to have one page and wanted to avoid having a parent with a single child?  I'm not sure their motivation. 

It's totally fine for the approach to change of course, I have no problems with it. (Will we actually eliminate the 'source' field afterwards?).  I just wanted to include my memory of the history of it.

apinkney0696:
That's really great to know, Jonathan!! Thank you.

I was told by Hillary when I asked her about it several months ago that the reason why they used source rather than a parent work was because the parent-child thing wasn't in place yet, so that's where I was pulling that statement! I'm glad to know otherwise! There are a lot of partially digitized books in the collection which use the parent-child system and use the description field to describe how much of the work is digitized. There are only 14 books here that don't use that system. My gut tells me that it's okay to go forward with this ticket in order to make how we treat partially digitized works consistent within the collection, but I will chat with Gabriela about it again this afternoon to get a second opinion now that I know the initial use of the source field was intentional.

apinkney0696:
[...]

PR #1313, opened 2021-09-09
Add children to newly-created parents and delete their source metadata

eddierubeiz:
Ref #1303 .


eddierubeiz:
We won't run this in production until next week at the earliest.

eddierubeiz:
Annabel, I tested this in staging, and it looks ok to me based on e.g. [amphitheatrum](https://staging-digital.sciencehistory.org/works/dodh36q). Uploading this [list of changes](https://github.com/sciencehistory/scihist_digicoll/files/7188204/changes.txt).

lib/tasks/fix_parent_child.rake
```diff
@@ -0,0 +1,288 @@
+namespace :scihist do
+  namespace :data_fixes do
+    desc "Genre adjustments, per https://github.com/sciencehistory/scihist_digicoll/issues/1275"
+    task :fix_parent_child => :environment do
+      adjustments = {
```
jrochkind, line 5: I would have called this something descriptive instead of "adjustments", like `child_to_parent` or something. But that's not going to stop me from approving.

jrochkind (approved):

jrochkind:
@eddierubeiz Please move this rake task to `./lib/tasks/data_fixes` -- and I guess put it in relevent rake namespace too. 

We established that directory to keep from having a whole lot of noise in the rake `./lib/tasks` directory from temporary rake tasks we're never going to use again. So let's use it for these!

apinkney0696:
I just looked through some of these on Staging and they look great! And there's no need to publish the new parents in Production either because I'm going to need to edit them a little more before they're ready. Thanks so much!

eddierubeiz:
Awesome. Will let you know when these are ready in production.

eddierubeiz:
Ran this in production. Attaching a list of [changes](https://github.com/sciencehistory/scihist_digicoll/files/7212329/changes_production.txt) we made.

Issue #1306, opened 2021-09-07
reduce logging bytes: by reducing crawls?


jrochkind:
Our app is exceeding our papertrail plan's logging capacity (250MB), when just a few weeks ago it wasn't. 

What's changed? Mainly just that it is logging a lot more events. I *think* this is probably from bots/crawlers, although we dont' have the info to be sure at present. 

Since papertrail archival files come in an hour block, I started out just comparing two one-hour samples. One from when  our log sizes were much smaller, one from more recently when were exceeding papertrail capacities. I made sure to pick hours *before* papertrail's end-of-day limits kicked in and stopped logging. 

Monday Aug 23 10am ET (smaller logs), and Monday Sep 6 10am ET (big logs). 

8123 total requests vs 20393 total requests -- a lot more requests! What are they? 

# Top requests:

## aug 23:

0036: /catalog.html
0060: /hirefire/9d023e59-b030-4ccd-a9db-389556f74044/info
0101: /catalog/range_limit_panel/year_facet_isim
0101: /catalog/opensearch.xml
0143: /catalog/range_limit
0144: /
0247: /catalog/facet/subject_facet
0662: /admin/queues/overview.poll
4115: /catalog

## Sep 6

0059: /focus/water/facet
0060: /hirefire/9d023e59-b030-4ccd-a9db-389556f74044/info
0127: /catalog/facet/creator_facet
0138: /
0159: /collections/4x51hj94t
0340: /focus/women-and-science
0376: /catalog/range_limit_panel/year_facet_isim
0777: /focus/water
0905: /catalog/facet/subject_facet
2129: /catalog/range_limit
12716: /catalog

OK, the extra hits are *mostly* to `/catalog` endpoint, our search results. 

# What do /catalog results look like

What do they look like, do they have a text query, facets? Drilling down just into the larger  Sep 6 set, of 12716 /catalog hits, only *240* have a `q` param.  12463 (almost all of them) have facet params but no `q` text query. 

From this I'm guessing they are mostly crawler bots that are just crawling all our pages, by clicking on every possible facet combination in the search results -- not really a great use of the crawl. 

We aren't currently logging user-agent, so can't be sure they are crawlers, but this is my guess:  The crawler traffic just went up a lot in the past few weeks, and MOST of our traffic is crawler traffic. 

# Suggested mitigation 

[...]

PR #1315, opened 2021-09-09
robots.txt try again

jrochkind:
Ref #1306 #1314


jrochkind:
google's robots.txt tester at https://www.google.com/webmasters/tools/robots-testing-tool?hl=en&utm_campaign=+6062598&siteUrl=http://digital.sciencehistory.org/

PR #1318, opened 2021-09-13
Adjust UA logging, allow more chars of raw UA, and replace spaces with underscores for easier log parsing leaving spaces being field separators

jrochkind:


jrochkind:
Oh weird, just noticed this was failing and was never merged! Looking into it.

jrochkind:
OK green now, please review when you have a minute, @eddierubeiz , thanks!

Issue #1324, opened 2021-09-14
upgrade uppy JS to 2.0


jrochkind:
We use "uppy" JS library for our front-end uploading interface, which interacts with shrine. 

uppy released a 2.0. (We are actually on an old 1.x, we haven't even upgraded to latest 1.x, we don't upgrade our JS dependencies as regularly as we do our ruby ones, or really at all sometimes)

We should upgrade -- making sure it doesn't break any of our custom local code, which it could. 

But we need the associated shrine-plugin to be updated to work with uppy 2.0 first. This... was just fixed! https://github.com/janko/uppy-s3_multipart/issues/23

PR #1326, opened 2021-09-15
update uppy js library to 2.x

jrochkind:
Ref #1324

Ended up being pretty straightforward, uppy was mostly backwards compat with one tiny change needed. 

I forgot we were still including uppy via it's own script tag, from a CDN! kind of janky. some day we will rationalize all our JS handling.


jrochkind:
unexpected test failure!

Probably the way we're trying to hack automating attaching a file in test, it actually works fine as a human demo'ing it! I'll look into it.

jrochkind:
Oh wait, and our COLLECTION thumbnail upload really is broken at present....

jrochkind:
Want to:

- [ ] DRY the uppy attach test code
- [ ] Realize latest uppy versions have native code for selecting a directory, we don't need our custom code anymore!

Issue #1299, opened 2021-09-02
Searchability of text from new translation/transcription fields


jrochkind:
This text should be searchable, and also show up as highlighted excerpts in search results, along the lines of what Oral Histories transcript text does. 

For the 'english translation' field, this may be easily accomplished by just indexing the content it into the exact same Solr field as the OH transcript goes into? And it'll be handled identically, great. 

For the transcription field that may not be English -- the existing field is probably configured for English so that won't work. 

We probably need a new field, configured in the Solr schema.xml config file.  I suggest making it a language-agnostic field, rather than trying to make one or more non-English language-specific fields.  

If the Work 'language' metadata is "English" and *only* "English", the content can be put into the existing field... if it includes any non-English languages, it would be put in the language-agnostic fulltext field. Note that a work can have more than one language designated, "English, German, Polish" -- and we won't know what text from the "transcript" is specifically each language. Is one reason to use just use a language-agnostic catch-all for all not-just-English. 

This new field would have to be configured to be included in search at the same relevancy/boost as the existing one used for OH transcript. And the app would have to consult it for possible excerpt highlighting to display on search results, checking both the existing field used for OH transcript plus this new one for possible excerpt-highlight text. 

I don't recall how all of this was done exactly for OH transcripts, but that should give us a good guide.

PR #1327, opened 2021-09-15
Searchable asset-level transcripts and english translations

eddierubeiz:
Ref #1299.
Requires reindex.
Modifies `schema.xml`.


eddierubeiz:
Note: the tests assert that if a work is in English, its transcription and translation are indexed into the English-language field.
(As an aside, strange as it sounds, I could envision us having a translation for an English work, by the way, e.g. https://digital.sciencehistory.org/works/46k32ki .)

app/indexers/work_indexer.rb
```diff
@@ -169,6 +169,22 @@ class WorkIndexer < Kithe::Indexer
...
       end
+
+      if rec.members.any? {|mem| mem.asset? && mem.english_translation.present?}
+        acc << rec.members.map {|mem| mem.english_translation if mem.asset? }.compact.join(" ")
+      end
+      # Index the transcription into this if we can assume that the work is entirely in English.
+      if rec.members.any? {|mem| mem.asset? && mem.transcription.present?} && rec.language == ['en']
+        acc << rec.members.map {|mem| mem.transcription if mem.asset? }.compact.join(" ")
+      end
+    end
```
jrochkind, line 180: Oh right, we do need to look through all members for this. 

Is this the first time we're needing to look at all members in indexing?  If so, that's definitely going to slow down indexing -- we need to make sure members are being Rails eager-loaded with `includes` to make it as efficient as possible, and then I'm still curious to know how much it's slowing it down, have some numbers on the PR.

eddierubeiz: It does slow things down a bit, although not much. Will quantify.
I'd also rather use `pick`, since we're only looking at one or two values.

spec/indexers/work_indexer_spec.rb
```diff
@@ -162,4 +162,38 @@
...
+    end
+    let(:german_only) do
+      create(:public_work, language: ['de'], members: assets )
+    end
+
+    it "indexes text known to be English in searchable_fulltext" do
+      output_hash = WorkIndexer.new.map_record(english_only)
+      english  = output_hash["searchable_fulltext"]
+      might_not_be_english = output_hash["searchable_fulltext_language_agnostic"]
+      expect(english).to eq(["transl_1 transl_2 transl_3", "transc_1 transc_2 transc_3"])
```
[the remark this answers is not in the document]
eddierubeiz: Re: concatenation: this did occur to me. Will do a bit of exploring.
GOOD CATCH re: order. I always forget this.

app/indexers/work_indexer.rb
eddierubeiz: I did mean `pluck`, yes. (`pick` does exist but that's not what I meant.)

jrochkind:
I hadn't really realized this was going to cross the boundary of needing member info available on indexing!

Batch indexing is one thing. Also just the one-off indexing when a change is made to a work -- like an admin, but also any batch processes we do!  If all of these are going to load all members of all works... I am worried about performance, and making sure we understand the implications.  Maybe we do need to `pluck` when it's a one-off, but *not* when it's bulk re-index... or maybe that's not it. 

I'm not sure, realizing it's a thing, sort of a new boundary we're crossing!

app/indexers/work_indexer.rb
```diff
@@ -169,6 +169,22 @@ class WorkIndexer < Kithe::Indexer
           end
         end
       end
+
+      if rec.members.any? {|mem| mem.asset? && mem.english_translation.present?}
+        acc << rec.members.map {|mem| mem.english_translation if mem.asset? }.compact.join(" ")
+      end
+      # Index the transcription into this if we can assume that the work is entirely in English.
+      if rec.members.any? {|mem| mem.asset? && mem.transcription.present?} && rec.language == ['en']
+        acc << rec.members.map {|mem| mem.transcription if mem.asset? }.compact.join(" ")
```
jrochkind, line 178: you're kind of going through all the members twice -- once to see if it has any member that's relevant, then again to collect all the members that are relevant. Is there a way to do this only once instead?  Might be easier if each element is a separate thing in in the field, no need to join. 

Not sure if this will really make a perf difference but... it might? 

Really not sure how we handle perf issues here.

jrochkind:
I guess the first sanity check would just be seeing how this totally unoptimized implementation compares to master on staging, for a total scihist:solr:reindex.  My guess is it's going to be like 10x slower, but if it's not? Maybe we don't need to worry about it?

app/indexers/work_indexer.rb
eddierubeiz: All good points. I didn't want to spend any time optimizing it before you'd had a look, but agree that keeping indexing efficient is very much in our long-interest.

eddierubeiz:
Yeah - I tried it in staging and it was (sort of to my surprise) not an order of magnitude slower.

eddierubeiz:
The unoptimized full reindex doubles the time needed to index the collection, from `01:09` to `02:00`. Let's see how much I can speed that up...

jrochkind:
doing not much more than adding an `includes(:members)` might do a lot.

jrochkind:
Also weird, I could have sworn I was getting around 2 minute times even normally these days. Of course the difference between 1 minute and 2 isn't bad; 2 and 4 starts getting more annoying. Maybe it depends on how the Solr is doing. 5 vs 10 if our index gets bigger would start getting annoying.

eddierubeiz:
This latest version takes`01:18`. Only adding 10 seconds to the reindex duration is good enough.

spec/indexers/work_indexer_spec.rb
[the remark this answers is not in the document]
eddierubeiz: (Not concatenating. Good call.)

app/indexers/work_indexer.rb
eddierubeiz: Fixed this.

jrochkind:
I had been avoiding pre-loading all the members/children to do a bulk index, but now that we've crossed that boundary it will make other things easier (like if we wanted to index number of members, or file types of members, etc). 

For comparison, here's about how much it slows down our indexing, at least in a sample size of 1. 

Before loading members, total reindex of 11396 items (all Works plus all Collections) is:

* 00:01:07  170.09 records per second

After this PR, pre-loading all members, total reindex is:

* 00:01:22, 138.98 records per second

Not too bad! OK, so it goes.

Issue #742, opened 2020-05-05
consider replacing custom ViewModel with ViewComponent


jrochkind:
We hacked out our own ViewModel architecture based on Draper. 

We're very happy with the overall architectural pattern, but it has some weird edge cases that can make development painful and more time consuming. Like sometimes hiding your actual MethodMessing errors with a fake/wrong error that sends you on a wild goosechase figuring out your bug. 

Recently discovered this fairly mature and well-supported project out of github, that does *very* similar things to our ViewModel. Might be pretty easy to swap out ours for it, almost a drop-in replacement even. Will have to investigate. 

https://github.com/github/view_component

PR #1330, opened 2021-09-21
In progress replace custom ViewModel with standard ViewComponent

jrochkind:
Ref #742


jrochkind:
@Eddie So, the valid_model_type_names we use to do type-checking in our ViewModels? The standard ViewComponent doesn't have that. 

It is not *too* too hard to add it back in... but requires making a couple things start to look different than standard ViewComponent. 

My thought is just to get rid of the type-checking, and make it look like absolutely standard ViewComponent. 

Thoughts? Did you love that type-checking?

eddierubeiz:
I liked it, but it's not a feature I think we necessarily need. I am fine living without the feature for a bit to see how things go.

jrochkind:
My notes on the migration from our custom draper-based ViewModel architecture, to the very similar but more shared ViewComponent gem architecture:

# migrating from our custom "ViewModel" implementation to standard ViewComponent gem

One by one, take ViewModel and translate to ViewComponent. Removing old viewmodel code.

ViewCompoennts live in ./app/components, with their view templates if present also in same directory!


1. Name change from ending in `_display.rb` to `_component.rb`, and same for class name. And move to the ./app/components directory, `< ApplicationComponent` instead of `< ViewModel`. You need to write the initializer in each class for desired args. 

2. We had implemented some simple type-checking with `valid_model_type_names` -- we can keep that,
adding it custom to ViewModel, or abandon it since ViewModel doesn't use it? (Decided to abandon it, for more standard ViewComponent including open to standard flexible keyword args). 

3. We had conventionally had a single positional argument; viewcomponents by convention use all keyword args. We don't have to change that though, sometimes we might leave it for convenience -- or we'd leave it if we wanted to leave `valid_model_type_names` -- Have ended up doing a bit of both, sometimes leaving it how it is for convenience of migration. 

4. Our ViewModels based on draper always had a `model`, and we often used alias_method to call it something else too. In ViewComponent, there's no default "model", we'll just always call it something else. So `alias_method :work, :model` turns into `attr_reader :work`, and make sure
initializer sets `@work` appropriately.

5. ViewModel `display` method is now `call` -- if you are just rendering a template, it doesn't need a specific implementation, use standard ViewComponent. If you were doing inline html, just rename `display` to `call`. https://viewcomponent.org/guide/templates.html#inline

6. In ViewModel templates, we needed to refer to `view.some_method` to call a method from the
ViewModel ruby. In ViewComponent, the `view.` can and must be ommitted, you can just call
`method_from_component`.

7. Changing specs can be kind of a pain. Instead of just doing eg:

        RightsIconDisplay.new(work).display

    you need to do:

        render_inline(RightsIconComponent.new(work: work))

    And you don't need to explicity wrap it in Nokogiri::XML, you get that returned from `render_inline` already. You don't even need to assign it to a variable, although you can.

    If you DON'T want nokogiri obj, you need to put `to_html` on the end!

    See:

      https://viewcomponent.org/guide/testing.html

8. From view

   old:

        <%= RightsIconComponent.new(work).display %>

   new:

        <%= render(RightsIconComponent.new(work: work)) %>


9. Helpers. Using a non-standard Rails helper method "Just Worked" in ViewModel. In ViewComponent, you either need to call `helper.my_helper` instead of just `my_helper`... OR just add `delegate :my_helper, to: :helpers` at the top. May apply to some helpers delivered by gems too, like `can?` from `access_granted`. Or `current_user`.

    * (should we delegate `can?` and `current_user` in ApplicationComponent? ViewComponent wouldn't like this loss of encapsulation, maybe better to maek each component individually do the dirty thing)

jrochkind:
Another pattern: 

If an object was a `ViewModel` solely to get access to Rails route-helpers (eg work_path, asset_path), it can be changed to a plain ruby object with no sub-class, but then include Rails url helper module into the class, with:

```ruby
class SomeClass
    include Rails.application.routes.url_helpers
```

This will let you use `*_path` helpers, but not necessarily `*_url` helpers (that include full url with hostname), because normally hostname is gotten from "current view_context" that a plain ruby object doesn't have. You can manually pass `host` (using our ScihistDigicoll::Env config) to every url call, or... exploring other solutions as I encounter the issue.

jrochkind:
Hm, why is Github actions CI not reporting back?  Indeed it's not running, not listed in https://github.com/sciencehistory/scihist_digicoll/actions

https://www.githubstatus.com/ doesn't list any problems. Annoying!

If it doesn't start working again tomorrow, will file a Github support ticket I guess.

jrochkind:
Ah, Github Actions don't run if there are merge conflicts... which surprises me, I thought the actions were just running on the branch, and should be able to do so even if there are merge conflicts. Annoying! 

https://github.community/t/run-actions-on-pull-requests-with-merge-conflicts/17104/2

jrochkind:
Some of the specs are definitely not standard/conventional ways to do things for ViewComponent, they were just ported over from ViewModel in a way that works. They are fine! But shouldn't necessarily be used as a model for the future, at least not without being familiar with ViewComponent's testing options: https://viewcomponent.org/guide/testing.html

jrochkind:
The ./app/presenters directory will remain for some "plain old ruby objects"  that are used in view/presentation logic. It turns out we have a bunch of those (some newly converted, some existing), this seems as good a place for them as any. I will put a README.md in that directory explaining it's existence a bit.

jrochkind:
I don't think we want to merge this before next week, but otherwise I think it's ready @eddierubeiz!

PR #1334, opened 2021-09-27
Index needs to tolerate nil members, and members with `position` set to nil.

eddierubeiz:
`work_indexer.rb` needs to tolerate nil members (as opposed to []) and members with nil position.

### DIFF
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -170,14 +170,14 @@ class WorkIndexer < Kithe::Indexer
         end
       end
 
-      acc.concat asset_property_array(rec, :english_translation)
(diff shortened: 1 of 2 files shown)


jrochkind:
great! Can you change title of PR -- this is really the "[re-apply] transcript/translate searchability" PR again, just with one improvement, the improvement doens't describe the PR. 

Can you add a test case to work_indexer_spec, just at the top, that just like demonstrates that a Work with no memberes (just the usual way of having no members, no need to try to make multiple weird ways), makes it through the indexing without raising an error?

eddierubeiz:
Done and done!

jrochkind (approved):

Issue #1347, opened 2021-10-04
batch publish/unpublish


jrochkind:
Add batch publish/unpublish to Admin batch "cart" functionality. 

Came out of discussion in #1320

jrochkind:
@apinkney0696 

I could:

1. Add a "publish" field with options "no change", "publish", and "unpublish" to the "Batch Edit" form at https://digital.sciencehistory.org/admin/works/batch_update

2. Or I could add "publish all" and "unpublish all" buttons to the "Cart" page at https://digital.sciencehistory.org/admin/cart_items, like next to "Batch Edit". 

Any preference?

apinkney0696:
Hi! Option 2 sounds great, please. Thanks!

jrochkind:
Oh yeah... our ordinary "publish"/"unpublish" action actually has two choices:  "Also [un]publish all members", "Leave all members as they are". 

I guess this batch one should have the same options...

jrochkind:
Oh and as a further wrinkle... works need to have certain metadata to be publish-able, if they don't have them you get an error trying to publish... I guess when batch publish, any one of the works in your cart might not actually be publishable, in which case you need to get the error message. 

Always a bit more complicated than first anticipated!

jrochkind:
@apinkney0696 

At this moment demo’d on staging is “publish”/“unpublish” function for the “cart”.

It’s a little bit kludgey. Let me know if it does what you need in the way you need it!
Note that some things can’t be published becuase they lack metadata required for publishing — if you try to publish a cart with those, you will get an error message, but it’s a bit messy, and only for ONE item that failed validation even if there were more. (If ALL the items can’t be published, NONE will be published though, it’s all or nothing).

So I recommend trying that case in staging to see how it works.

It’s possible this is still kind of confusing or insufficient, I just did the simplest thing that seemed like it could work, let me know!

apinkney0696:
Hi @jrochkind ! I just experimented with this a little bit.

I really like that when I try to publish a batch where one record is incomplete, it tells me in the error message the name of the record with the issue. Even though it can only list one un-validated item in the error message, I think that's okay. Otherwise, the error message would get huge. I will likely either need to edit one record at a time at that point anyway, or, because it's already in the batch edit tool, it's an easy bulk fix if there is more than one record with the same problem.

I think the pop-up messages at the top are all nice, and I actually think this will work for me as is. Really all I needed was a means to publish many things at once, and this is definitely going to do it!

PR #1350, opened 2021-10-05
batch publish/unpublish from cart

jrochkind:
This is a little bit hacky, I don't love how the WorkController is just getting all these random verbose methods. But it works!

If you publish, there can be an error -- some works arne't publishable. We show just the FIRST work error we encounter, you might have more than one that were unpublishable, but we only show one. If there's an error, we don't make ANY changes, it's all or none -- implemented by wrapping in transaction

Ref #1347

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -410,6 +410,41 @@ class Admin::WorksController < AdminController
     redirect_to admin_cart_items_url, notice: "Updated works in Cart"
   end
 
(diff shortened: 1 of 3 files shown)


jrochkind:
This PR doens't have tests yet. It's kind of a pain to write the tests, don't want to write a really slow "system" test for every possible branch!

Might only ccover a few paths, and maybe with controller tests.

jrochkind:
Reviewed and approved by Annabel.

PR #1351, opened 2021-10-05
Speed up pasting into irb/console by using newer bugfixed dependencies!

jrochkind:
https://github.com/ruby/irb/issues/43#issuecomment-758089211


jrochkind:
OK, I'm going to merge this in on my own, because it's trivial (gemfile change only), it is needed to make my work in other branches more convenient, and I know @eddierubeiz  is busy with other work.

PR #1366, opened 2021-10-12
Revert "Add url_helper include to description_display_formatter", but with a fix

jrochkind:
Reverts sciencehistory/scihist_digicoll#1359


jrochkind:
OK, reverted, then added test that failed on bug, then fixed by replacing rails_auto_link with rinku. 

We actually realized this is pretty low-value functionality we weren't really using at all, but oh well, now it's fixed how it was before!

Issue #565, opened 2019-12-19
Bring the site closer to WCAG 2.0 compliance


eddierubeiz:
How I might tackle this:

- Solicit advice and comments from the accessibility committee and from the librarians at https://libwww.freelibrary.org/lbph/ .
- Pick an automated tool that flags WCAG violations (one I've worked with in the past is https://squizlabs.github.io/HTML_CodeSniffer/)
- Decide on things we can and can't change (for instance, we can't fix violations that would bring us in conflict with our sitewide branding.)
- Focus on the 80% percent of problems can typically be fixed in 20% of the time. 
- Focus on pages that are heavily used by the public.
- Document what we changed, and show the numbers: blog post, report to the accessibility committee.
- Keep a list of items that were *not* fixed and why, so they can be tackled in a future iteration, should we choose to.

jrochkind:
OK, I've gone through all of the HTML_CodeSniffer => level AA => "error" (as opposed to warning/notice)

There are some outstanding things related to img alt text that I need to thinik/reserach/discuss with y'all more. We'll talk today. 

In the meantime, I'm also moving on to: The IBM tool; level AAA; warnings and notices. 

this is a much larger more confusing noisier list, including lots of things that are just like "this might be an issue you should check" and then I check and it's not (but no good way to eliminate it from the report), and lots of things that I don't totally understand what they mean. 

But for things that I do understand what they mean and see a good way to address/improve them, I'll start doing a few of those too.

jrochkind:
Another good tool (FYI @eddierubeiz ) is Google Lighthouse: https://developers.google.com/web/tools/lighthouse

It runs as a Chrome plugin. While it does more than accessibility, you can run it with only the "accesibility" checkbox checked to only focus on that. 

It's flagging things I think are legit that the HTML_CodeSniffer tool didn't find; and is for me somewhat easier to use than the IBM tool.

PR #1375, opened 2021-10-13
Hide rights statement icon from assistive technology consistently

jrochkind:
The text next to it is sufficient, the icon can be hidden.

We were doing that somewhat inconsistently, we switch to doing it consistently, with blank `alt` attriute AND aria-hidden=true on all of them -- both large format and compact format.

From getting advice on code4lib slack, I _think_ this is the right thing to do in all of these circumstances. Doing my best!

Ref WCAG work at #565

### DIFF
--- a/app/components/rights_icon_component.rb
+++ b/app/components/rights_icon_component.rb
@@ -23,7 +23,7 @@ class RightsIconComponent < ApplicationComponent
   # one line, all a link
   def display_dropdown_item
     link_to(rights_url, target: "_blank", class: ['rights-statement', mode.to_s.dasherize, layout_class]) do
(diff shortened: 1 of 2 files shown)


jrochkind:
Oops, I just realized this is wrong, we can NOT actually hide it all the time. It works for "Public Domain" but not for some of our other ones! 

Putting back to "draft", will fix up a differenet way, with more comments!

jrochkind:
Yeah, this was totally wrong, I was about to make it worse!

The problem is not that all the rights statement icons should be hidden with `alt=""` the problem was that a couple were MISSING alt attributes they should have had! 

I will try to comment the code better for intent too, although it gets confusing.

PR #1381, opened 2021-10-14
Make sure multiple nav elements have distingiushing labels

jrochkind:
A nav element can be a <nav> or anything with `role="navigation"`.

IBM checker warned us about this:

> Multiple elements with "navigation" role do not have unique labels
Violation
>  Each element with "navigation" role must have a unique label that describes its purpose

Also I saw we had a nav nested inside a nav, which seemed unhelpful, one big nav for entire top area seems sufficient.

Ref WCAG work at #565


jrochkind:
And, actually, our header masthead shouldn't have role=nav, it should have role=heading, via the <header> element!

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role

PR #1397, opened 2021-10-18
attribute should be just 'role', not 'aria-role', mistake

jrochkind:


jrochkind:
Actually this whole thing is no longer relevant/needed in master branch.

PR #821, opened 2020-07-27
relax guardrails around publishing OH without valid combined audio derivative

jrochkind:
Ref #767

Issue #496, opened 2019-11-15
Do not allow works to be published unless they have a published representative; prevent the deletion of published works' representatives.


yonyitz:
Replicated a process that I encountered on the main site in staging and it appears to be a glitch:

1.  Added new image for this record: https://staging-digital.sciencehistory.org/works/f1881m63m
2.  Deleted original image
3.  One needs to go back into edit metadata screen and click "update work" in order for representative member to appear

eddierubeiz:
Otherwise stated, IF you start out with a parent with two member items, and AND you delete one of them, AND this would otherwise leave the parent without a representative, THEN the parent's representative should automatically be set to its sole surviving member.

jrochkind:
In general, we don't have a lot of code right now for ensuring there's always a valid thumbnail, or automatically updating the thumbnail when things change, so it's still valid. 

There are probably a variety of cases that leave you without a good thumbnail. Deleting (which can maybe be done in various paths, considering the possibility of multi-level children), maybe also making non-public. 

We could decide if we want to prioritize work to make this never the case, depending on how inconvenient/time-consuming/subject-to-error it is to have to do manual fixes of these. 

We could potentially resolve this either by code taht hooks into "update" processes to make sure a valid thumbnail is set in `thumbnail_id ` where applicable/possible; *or* at display time we could notice that there's no thumbnail set (or the `thumbnail_id` is not available), and choose another arbitrary one to be representative. 

I think this latter is actually what chf_sufia did -- if no representative was set, it just chose the "first" public one to use as a representative. The challenge *potentially* is doing that in a performance-respecting way, which chf_sufia probably did not; although it might be straightforward.

jrochkind:
* Let's look at what chf_sufia was doing for background
* Let's also make sure the admin pages are more clear feedback on management pages when there is no thumbnail set

eddierubeiz:
Noting also that from the public's standpoint, if the representative is not published, it's just as if it hadn't been set at all. So we should consider that case as well.

eddierubeiz:
See also #450 which is about possibly paying attention to content type when selecting the representative at the time of ingest.

eddierubeiz:
In Sufia, if you delete the representative (or thumbnail), you end up with nil representative (or thumbnail). It's not automatically changed to the most-likely new value; that is left to the user. So, for what it's worth, this problem already existed in Sufia.

jrochkind:
[...]

PR #1404, opened 2021-10-19
[do not merge] Extra rules to prevent published works without representatives

eddierubeiz:
Ref #496
- Require a published representative to publish a work
- Prevent the deletion of a work or asset if the published parent work would be left without a representative

What this means, practically, to an admin user:
- Pick a published work. Try deleting its representative. You can no longer do that.
- Unpublish the work. Now,  you *can* delete the representative.
- However, you can’t *republish* the work until you explicitly assign it a new representative.

**Design choice**: this PR will not attempt to modify buttons / menu items; I am changing only controllers. In other words, the UI still allows you request the above actions; you'll just get an error message. (This choice is informed by our experience in 2020 with similar guardrails surrounding combined audio derivatives, most of which we ended up [deciding to remove](https://github.com/sciencehistory/scihist_digicoll/pull/821/files). )


eddierubeiz:
Closing this PR; superseded by #1433

PR #1406, opened 2021-10-19
Remove section role=main

jrochkind:
We already have a <main> from layout, which was leading to accessibility checker complaining about multiple elements with role=main.

This section element didn't seem to do anything, everything displays just fine if we remove it completely

Ref WCAG work at #565


eddierubeiz:
OK -- interesting that it was unnecessary. Probably a Sufia atavism.

Issue #1411, opened 2021-10-21
Duplicate footnote references in Oral History transcripts result in illegal duplicate ids in HTML DOM


jrochkind:
It turns out duplicate footnote references are possible in our OH transcripts. See for instance [Smyth](https://digital.sciencehistory.org/works/4j03d097t) , which has duplicate references to footnotes 3 and 8. And conveniently happens to be the sample transcript we are using in our automated tests. 

When duplicate footnote references happen, our code ends up putting multiple elements into the HTML dom with the same `id` attribute -- eg `footnote8`. Because the `FootnoteReferenceComponent` ends up being rendered multiple times with the same `number ` argument. 

This is illegal HTML as well as an accessibility problem. 

<details>
<summary>See aXe checker warnings on Smyth:</summary>
Some labelled "critical" because of use in aria- attributes. 

```
      2) duplicate-id-aria: IDs used in ARIA and labels must be unique (critical)
           https://dequeuniversity.com/rules/axe/4.3/duplicate-id-aria?application=axeAPI
           The following 2 nodes violate this rule:

               Selector: #ohms_line_198 > .sr-only
               HTML: <span class="sr-only" id="footnote-text-3">3. See Beckman Center for the History of Chemistry oral history research file #0042 for a copy of undated typewritten autobiographical notes (pp.13).</span>
               Fix any of the following:
               - Document has multiple elements referenced with ARIA with the same id attribute: footnote-text-3

               Selector: #ohms_line_610 > .sr-only
               HTML: <span class="sr-only" id="footnote-text-8">
               Fix any of the following:
               - Document has multiple elements referenced with ARIA with the same id attribute: footnote-text-8

       3) duplicate-id: id attribute value must be unique (minor)
           https://dequeuniversity.com/rules/axe/4.3/duplicate-id?application=axeAPI
           The following 2 nodes violate this rule:

               Selector: #ohms_line_198 > a[name="footnote-reference3"]
               HTML: <a name="footnote-reference3" id="footnote-reference3"></a>
               Fix any of the following:
               - Document has multiple static elements with the same id attribute: footnote-reference3

               Selector: #ohms_line_610 > a[name="footnote-reference8"]
               HTML: <a name="footnote-reference8" id="footnote-reference8"></a>
               Fix any of the following:
               - Document has multiple static elements with the same id attribute: footnote-reference8
```

</details>

[...]

Issue #1391, opened 2021-10-18
Add some simple automated accessibility checking into our CI


jrochkind:
It *definitely* won't catch everything, but we can add an automated accessibilty checker to our automated tests that get run on every code change, to warn us if we're breaking or introducing a problem with something it knows about, for pages we've told it to look at. 

https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-rspec/README.md

PR #1413, opened 2021-10-21
automated accessibility testing in CI with axe-core-rspec

jrochkind:
Ref #1391 and WCAG overall ticket at #565

https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-rspec/README.md

Using the [axe tool from Deque](https://www.deque.com/axe/) to do some automated accessibility checking as part of our system tests.  This definitely doens't ensure accessibility, it's just one level of check, but it's good to have that level of check happening for future work we do, so it can warn us of things we might want to fix, or take a closer look at and ultimately tell the tool to ignore in some cases. 

We picked several major pages to incorporate an axe check in. If we add additional system tests for new kinds of pages in the future, we should add axe checks in them!

Note this PR will not currently pass becuase of unmerged #1408. Once that's merged, this one can be rebased/merged with master, and it should be green. In the meantime, it gives you an opportunity to see what axe failures look like in CI?

Note also that our axe tests in this PR have some "ignore these things" configured:

* For a Blacklight issue we can't easily fix, need it to be fixed in BL. https://github.com/projectblacklight/blacklight/pull/2491
* In OH transcript page, footnotes issue at #1411
* in OH transcript page, our audio navbar actually fails contrast test, with bright green on dark grey. :( Setting this aside for now, I'm not sure how to redesign it, we're kind of out of colors.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -113,7 +113,7 @@ GEM
       ostruct (>= 0.3.2)
       rails (>= 5.1, < 7)
       view_component (>= 2.28.0)
-    blacklight_range_limit (8.0.1)
+    blacklight_range_limit (8.1.0)
       blacklight (~> 7.0)
     bootsnap (1.7.7)
       msgpack (~> 1.0)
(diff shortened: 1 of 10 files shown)


jrochkind:
I think CI isn't running because of some github stabiilty problems at the moment https://www.githubstatus.com/

PR #1433, opened 2021-11-01
Changes to work and asset controllers to prevent published works w/o a representative

eddierubeiz:
Ref #496

- Require a published representative to publish a work
- Prevent the deletion of a work or asset if the published parent work would be left without a representative

What this means, practically, to an admin user:

 - Pick a published work. Try deleting its representative. You can no longer do that.
 - Unpublish the work. Now, you can delete the representative.
 - However, you can’t republish the work until you explicitly assign it a new representative.

The old PR was getting really cluttered and hard to work with, so this is the same information presented more cleanly.

### DIFF
--- a/app/controllers/admin/assets_controller.rb
+++ b/app/controllers/admin/assets_controller.rb
@@ -17,6 +17,9 @@ class Admin::AssetsController < AdminController
   # PATCH/PUT /works/1.json
   def update
     @asset = Asset.find_by_friendlier_id!(params[:id])
+    if params['asset']['published'] == "0" && would_remove_representative?(@asset.parent)
+      prevent_unpublish_because_no_representative && return
+    end
 
     respond_to do |format|
       if @asset.update(asset_params)
@@ -31,13 +34,13 @@ class Admin::AssetsController < AdminController
 
   def destroy
     @asset = Asset.find_by_friendlier_id!(params[:id])
-
     authorize! :destroy, @asset
-
-    work = @asset.parent
+    if would_remove_representative?(@asset.parent)
+      prevent_destroy_because_no_representative && return
+    end
     @asset.destroy
     respond_to do |format|
-      format.html { redirect_to admin_work_path(work.friendlier_id, anchor: "nav-members"), notice: "Asset '#{@asset.title}' was successfully destroyed." }
+      format.html { redirect_to admin_work_path(work.friendlier_id, anchor: "tab=nav-members"), notice: "Asset '#{@asset.title}' was successfully destroyed." }
       format.json { head :no_content }
     end
   end
@@ -133,7 +136,26 @@ class Admin::AssetsController < AdminController
       :transcription, :english_translation,
       :role, {admin_note_attributes: []}]
     allowed_params << :published if can?(:publish, @asset)
-
     asset_params = params.require(:asset).permit(*allowed_params)
+  end
+
+  def would_remove_representative?(parent)
+    parent&.published? && parent.representative == @asset
+  end
+
+  def prevent_unpublish_because_no_representative
+    @asset.errors.add(:base,  "Could not unpublish asset '#{@asset.title}'. Its parent is published and this is its representative. Unpublish the parent first.")
+    respond_to do |format|
+      format.html { render :edit }
+      format.json { render json: @asset.errors, status: :unprocessable_entity }
+    end
+  end
+
+  def prevent_destroy_because_no_representative
+    respond_to do |format|
+      format.html { redirect_to admin_work_path(@asset.parent.friendlier_id, anchor: "tab=nav-members"),
+        notice: "Could not delete asset '#{@asset.title}'. Its parent is published and this is its representative." }
+      format.json { render json: { error:  "Could not delete asset '#{@asset.title}'. The work is published and this is its representative." }, status: 422 }
+    end
   end
 end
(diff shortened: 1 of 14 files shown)


app/controllers/admin/assets_controller.rb
```diff
@@ -17,7 +17,7 @@ def edit
   # PATCH/PUT /works/1.json
   def update
     @asset = Asset.find_by_friendlier_id!(params[:id])
-
+    return unless representative_check_for_unpublish
```
jrochkind, line 20: This is a confusing way to write it, it makes it look like you are returning from an action method without rendering any response at all, which would be wrong. 

I see when I look there is a render in the method... but it's confusing. 

Also methods that return a boolean result should end in `?`. 

What if we change this to a two-liner that's more clear, also taking an explicit param:

    if would_remove_representative?(work)
       render_refused_remove_representative && return
    end

eddierubeiz: That's great. Will use that pattern throughout- thanks.

spec/factories/work_factory.rb
```diff
@@ -13,12 +13,13 @@
...
     after(:build) do |work|
       if work.representative.nil? && work.members.loaded?
-        work.representative = work.members.first
+        work.representative = work.members.to_a.find {|w| w.published? }
       end
     end
 
     trait :published do
       published { true }
+      members {  [ build(:asset_with_faked_file, published: true) ] } # published works need a representative now
```
jrochkind, line 22: asset_with_faked_file creates a `published: true` Asset already. It's fine if you want to `published: true` there to make it even more clear/explicit, maybe that's even best? I just mention it because I see you adding `published: true` every time you are using the factory, and I wasn't sure if you realized that was not necessary?

spec/mailers/oral_history_delivery_spec.rb
```diff
@@ -16,7 +16,7 @@
     end
 
     let!(:work) do
-      create(:oral_history_work, published: true).tap do |work|
+      create(:oral_history_work, :published).tap do |work|
         work.members = members
         work.representative =  members[2]
```
jrochkind, line 21: I see this pre-dates this PR, so we don't have to and probably shouldn't fix it in this PR, but seeing this.. I really don't love setting the representative to `members[2]`, like why do we care about the 3rd element of the members list being the representative, why is that our test condition?  The test setup should be more clear about the intent of the scenario being set up. If we care about a specific thing being set as the representative, it should be named and referred to by a name explaining it's significance.

spec/presenters/child_count_display_fetcher_spec.rb
```diff
@@ -24,7 +24,7 @@
     end
 
     describe "with no members" do
-      let(:item) { create(:public_work) }
+      let(:item) { build(:public_work, members: []) }
```
jrochkind, line 27: In the branch you abandoned this was: 

    let(:item) { build(:public_work, members: []).tap {|w| w.save!(validate: false) } }

I thought that was necessary to actually get a public_work with no members, since otherwise... aha, this branch eliminates the validation! Right. OK.

spec/system/oai_pmh_spec.rb
```diff
@@ -78,9 +78,8 @@
     end
   end
 
-  # should not create an error, as long as it's possible in DB
   describe "work with no representative" do
-    let!(:work_with_no_representative) { create(:work, :with_complete_metadata, title: "no representative", published: true) }
+    let!(:work_with_no_representative) { build(:work, :published, :with_complete_metadata, title: "no representative", members: [], representative: nil).tap {|w| w.save!(validate: false)} }
```
jrochkind, line 82: Might not need the complicated `save.validate(false)` here either, now that we got rid of the model validation?

eddierubeiz: Nope -- it's still necessary here.

jrochkind: Very confusing! Can you help explain to me why it's needed here, but not above in child_count_display_fetcher_spec.rb that also does a `let(:item) { build(:public_work, members: []) }`? What's the difference?

eddierubeiz: It's because `build` doesn't save the item to the DB. The `save.validate(false)` has the side-effect of turning the `build` into a `create`, which in turn means that the work is then found in the DB by `WorkOaiProvider`.
I'll change the `build` to `create` -- then I can remove the `save`.

spec/presenters/child_count_display_fetcher_spec.rb
eddierubeiz: Yeah - no longer needed...

spec/system/oai_pmh_spec.rb
jrochkind: OK, right, you don't actually need the `validate: false`, then? And therefore there's no reason to use `build` instead of `create`?  This pattern of `build... save(validate:false)` was to do the equivalent of a `create` while skipping the validation.  So if just `create` works here, that's simpler!

eddierubeiz: Nope... you can't use just `create`. OAI is looking for records in the database, so if you just use `build`, it won't find your work!

eddierubeiz: Yes! Create works great, but you also *need* it so that the OAI provider can find the record in the database.

spec/mailers/oral_history_delivery_spec.rb
eddierubeiz: Good call -- fixed

spec/factories/work_factory.rb
eddierubeiz: I didn't give it much thought until this week but now of course I'm hyper-aware of it!

jrochkind:
phew, nice work!

Issue #1432, opened 2021-10-29
Updated NACO Authority Subject and Creator Headings


apinkney0696:
Whenever y'all have an opportunity (no rush), could I please have records with the following two headings updated from

**Bredig, Georg, 1868-** to **Bredig, Georg, 1868-1944** (764 creator entries, 388 subject headings)

and 

**Caruso, David J.** to **Caruso, David J., (David Joseph), 1978-** (34 creator entries)

Thanks!

eddierubeiz:
I ran a draft of this code in staging @apinkney0696  - take a quick look when you have a chance.

apinkney0696:
Looks great to me! I don't see any issues - thank you!

PR #1435, opened 2021-11-02
Change all works to adjust subjects and creators to new NACO authority headings

eddierubeiz:
Spent a little extra time on this in case we want  to use the code again.
Ref #1432

### DIFF
--- a/lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
+++ b/lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
@@ -22,6 +22,7 @@ module ScihistDigicoll
         @metadata_map.fetch(value, value)
       end
 
+      # Update the work, but only if it has something that needs to be changed.
       def process_work(work)
         update_subject(work) unless (work.subject              & @keys_to_check_for).empty?
         update_creator(work) unless (work.creator.map(&:value) & @keys_to_check_for).empty?
--- a/lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
+++ b/lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
@@ -1,0 +1,52 @@
+module ScihistDigicoll
+  module TaskHelpers
+
+    # This is meant to be called by the rake task scihist:data_fixes:adjust_subjects_and_creators
+    # which instantiates this lass *once*, then calls process_work(work) on each work in the collection.
+    # process_work just updates the subjects and authors of a work according to @metadata_map.
+    class SubjectCreatorAdjuster
+      attr_accessor :metadata_map, :keys_to_check_for, :changes, :errors
+      
+      def initialize
+        @changes = []
+        @errors = []
+        @metadata_map = {
+          "Bredig, Georg, 1868-" => "Bredig, Georg, 1868-1944",
+          "Caruso, David J."     => "Caruso, David J., (David Joseph), 1978-",
+        }
+        @keys_to_check_for = @metadata_map.keys
+        raise ArgumentError, "Can't have nil values for @metadata_map" if @metadata_map.values.any?(&:nil?)
+      end
+
+      def change_if_necessary(value)
+        @metadata_map.fetch(value, value)
+      end
+
+      def process_work(work)
+        update_subject(work) unless (work.subject              & @keys_to_check_for).empty?
+        update_creator(work) unless (work.creator.map(&:value) & @keys_to_check_for).empty?
+      end
+
+      # Update all subject headings according to @metadata_map.
+      def update_subject(work)
+        updated_subject = work.subject.map { |s| change_if_necessary(s) }
+        if work.update({subject: updated_subject})
+          @changes <<  [ work.friendlier_id, work.title, updated_subject ]
+        else
+          @errors << [ work.friendlier_id, work.title, work.subject, 'could not change subject']
+        end
+      end
+      
+      # Update all values of `creator`, regardless of category, according to @metadata_map.
+      def update_creator(work)
+        work.creator.each { |cr| cr.value = change_if_necessary(cr.value) }
+        if work.save!
+          @changes <<  [ work.friendlier_id, work.title, work.creator ]
+        else
+          @errors <<  [ work.friendlier_id, work.title, work.creator, 'could not change creator']
+        end
+      end
+
+    end
+  end
+end
--- a/lib/tasks/data_fixes/adjust_subjects_and_creators.rake
+++ b/lib/tasks/data_fixes/adjust_subjects_and_creators.rake
@@ -1,0 +1,22 @@
+require "#{Rails.root}/lib/scihist_digicoll/task_helpers/subject_creator_adjuster"
+namespace :scihist do
+  namespace :data_fixes do
+    desc "Subject and creator changes to match name authorities, per https://github.com/sciencehistory/scihist_digicoll/issues/1432"
+    task :adjust_subjects_and_creators => :environment do
+      progress_bar = ProgressBar.create(total: Work.count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")      
+      adjuster = ScihistDigicoll::TaskHelpers::SubjectCreatorAdjuster.new
+      Work.transaction do
+        Kithe::Indexable.index_with(batching: true) do
+          Work.find_each(batch_size: 10) do |work|
+            adjuster.process_work(work)
+            progress_bar.increment
+          end
+        end
+      end
+      puts "Changes:"
+      pp adjuster.changes
+      puts "Errors:"
+      pp adjuster.errors
+    end
+  end
+end
--- a/spec/lib/scihist_digicoll/helpers/subject_creator_adjuster_spec.rb
+++ b/spec/lib/scihist_digicoll/helpers/subject_creator_adjuster_spec.rb
@@ -1,0 +1,23 @@
+require 'rails_helper'
+require "#{Rails.root}/lib/scihist_digicoll/task_helpers/subject_creator_adjuster"
+
+describe ScihistDigicoll::TaskHelpers::SubjectCreatorAdjuster, type: :model do
+  let(:adjuster) { ScihistDigicoll::TaskHelpers::SubjectCreatorAdjuster.new }
+
+  let(:work_to_update) do 
+    create(:public_work, creator: [
+        Work::Creator.new({category:'author',  value: 'unchanged'}),
+        Work::Creator.new({category:'author',  value: 'Bredig, Georg, 1868-'}),
+        Work::Creator.new({category:'addressee',value: 'Bredig, Georg, 1868-'})
+     ],
+     subject: ['as is', 'Caruso, David J.']
+     )
+  end
+
+  it "updates creators and subjects" do
+    adjuster.process_work(work_to_update)
+    work_to_update.reload
+    expect(work_to_update.creator.map(&:value)).to eq ["unchanged", "Bredig, Georg, 1868-1944", "Bredig, Georg, 1868-1944"]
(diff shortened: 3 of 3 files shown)


lib/tasks/data_fixes/adjust_subjects_and_creators.rake
```diff
@@ -0,0 +1,61 @@
+namespace :scihist do
+  namespace :data_fixes do
+    desc "Subject and creator changes to match name authorities, per https://github.com/sciencehistory/scihist_digicoll/issues/1432"
+    task :adjust_subjects_and_creators => :environment do
+
+      class SubjectCreatorAdjuster
```
eddierubeiz, line 6: Not sure whether just dumping a helper class inside the task scope is such a great plan.

eddierubeiz: Hm... looks like it *is* just scoped locally. I kind of don't want it in `/.lib` : in its current form it has no use outside this one task. But also, referring to the class in the task before it's declared results in a `NameError`.
Don't want to agonize over this too much, but other ideas welcome!

eddierubeiz: All makes sense. I moved stuff around a bit and put the helper class inside the namespace, right after the task.
That way, the task can instantiate the helper class, and the helper class is invisible outside of the namespace.

jrochkind: OK, in my tests the helper class is *not* invisible outside the namespace. For me, it's totally visible outside the namespace, it's global. 

Maybe I'm wrong? What did you look at to conclude it is invisible outside the namespace or not globally referenceable?

But of course if we put it in `./lib` it would be global too. I think it's fine, I'd just give it a very specific name unlikely to conflict. 

And personally I'd just put it at the bottom of the file, so it doesn't *look* like it's going to be non-global/namespaced... if i'm right that it's really not.

jrochkind: This blog post warns you about defining isolated methods with `def` (something I used to do too, mistakenly!), but the logic applies to classes too. 

https://kevinjalbert.com/defined_methods-in-rake-tasks-you-re-gonna-have-a-bad-time/

 And my own testing just now seemed to reveal that it's global:

```ruby
namespace :test do
  task :test do
    Foo.new.call("inside")
  end

  class Foo
    def call(str)
      puts "called #{str}"
    end
  end
end

# Do we have it? if it were invisible here or not reference-able as "foo", we'd get
# an error on loading this file, but instead it manages to call the method and print
# "called outside" just fine
Foo.new.call("outside")


```

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
```diff
@@ -0,0 +1,40 @@
+module ScihistDigicoll
+  module TaskHelpers
+    class SubjectCreatorAdjuster
```
jrochkind, line 3: can you add some comments saying what this thing does? Maybe mention the rake task.

jrochkind: mentioni t only updates subject and creator, etc.

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
```diff
@@ -0,0 +1,40 @@
...
+      attr_accessor :metadata_map, :keys_to_check_for, :changes, :errors
+      
+      def initialize
+        @changes = []
+        @errors = []
+        @metadata_map = {
+          "Bredig, Georg, 1868-" => "Bredig, Georg, 1868-1944",
+          "Caruso, David J."     => "Caruso, David J., (David Joseph), 1978-"
```
jrochkind, line 11: if we were going to re-use this, we'd want this to be a variable argument somehow of course. We could cross that bridge when we come to it, or we could already make it a variable argument, pass in a hash from rake task?

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
eddierubeiz: Yeah of course

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
```diff
@@ -0,0 +1,40 @@
...
+        @keys_to_check_for = @metadata_map.keys
+      end
+
+      def process_work(work)
+        update_subject(work) unless (work.subject              & @keys_to_check_for).empty?
+        update_creator(work) unless (work.creator.map(&:value) & @keys_to_check_for).empty?
+      end
+
+      def update_subject(work)
+        updated_subject = work.subject.filter_map { |s| @metadata_map.fetch(s, s) }
```
jrochkind, line 22: I don't understand what this `filter_map` is doing. Why `filter_map` instead of just `map` (aka `collect`)?  It makes me think it's going to be removing subjects... but it's not intended to, is it? 

Should there be a conditional in here, like only process and `save` the work if it includes a subject we're interested in?  

Or the fancier/quicker way to do this would be only fetch works we're interested in in the first place, but maybe that's not necessary -- definitely not for a one-off, we can do more performance opt later if we end up re-using this maybe.

eddierubeiz: The `filter_map` allows you to specify a null value for a key in `@metadata_map` which does allow you to remove subjects. But I'm glad you mention it because that doesn't work at all with creators! Will simplify this before we run it. Good catch.

jrochkind: I guess filter_map would let you have a Hash with a `nil` valuae meaning "delete this subject"? Is that intended? Should be documented. And come to think of it, if we plan on re-using this thing for possible future data changes (which I think seems not unlikely), there should be maybe some tests of this class? Not exhaustive tests, just like one or two or three simple tests.

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
```diff
@@ -0,0 +1,40 @@
...
+        if work.update({subject: updated_subject})
+          @changes <<  [ work.friendlier_id, work.title, updated_subject ]
+        else
+          @errors << [ work.friendlier_id, work.title, work.subject, 'could not change subject']
+        end
+      end
+      
+      def update_creator(work)
+        work.creator.each { |c| c.value = metadata_map.fetch(c.value, c.value) }
```
jrochkind, line 31: same, let's do some of conditional where we only call `save!` if we actually needed to make a change? Otherwise rails in some cases ends up doing a lot of extra work, calling `save!` on every record in the db.

 If nothing else I've noticed it opening and closing a transaction in the DB (and doing nothing in it if there are no changes to save), but can do worse things, or even error things -- safer to only call save/save! on the subset of records we actually wanted to touch.

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
jrochkind: @eddierubeiz the filter_map is still there?  If we want to support `nil` value, it should be both documented and tested... otherwise remove the filter_map?

I still have the same question as above:

> Should there be a conditional in here, like only process and save the work if it includes a subject we're interested in?

> Or the fancier/quicker way to do this would be only fetch works we're interested in in the first place, but maybe that's not necessary -- definitely not for a one-off, we can do more performance opt later if we end up re-using this maybe.

spec/lib/scihist_digicoll/helpers/subject_creator_adjuster_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
+require "#{Rails.root}/lib/scihist_digicoll/task_helpers/subject_creator_adjuster"
+
+puts ScihistDigicoll::TaskHelpers::SubjectCreatorAdjuster
+
+describe ScihistDigicoll::TaskHelpers::SubjectCreatorAdjuster, type: :model do
+
+  let(:work_1) do
```
jrochkind, line 8: In general `work_1`, `work_2` etc are not great names in a spec. 

What makes these works different from each other  or notable or why are we even testing more than one? `work_with_x` would be a better name. 

Or do we even need more than one work to test, does one work suffice? 

Alternately, if you are using a different work in each test, it's best to instead just put each test in it's own `describe` block, that has it's own `let(:work)` scoped to that describe block. Rather than create a whole bunch of works at the top, and then only use one in each test, and then you have to look back and forth from top to the test. 

Sorry, I'm trying to pay more attention to reviewing specs as code too, after we ran into all that technical debt from the existing specs not being as careful or maintainable as they could be!

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
eddierubeiz: Yeah, there's a conditional there of course -- only relevant works are saved. I'll put in a comment, though, to make it clearer. Sorry I missed your question!

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
eddierubeiz: This method is only called on items that match the keys in `@metatada_map.`. I'll add a comment to make this clearer.

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
```diff
@@ -0,0 +1,56 @@
...
+        @metadata_map = {
+          "Bredig, Georg, 1868-" => "Bredig, Georg, 1868-1944",
+          "Caruso, David J."     => "Caruso, David J., (David Joseph), 1978-",
+        }
+        @keys_to_check_for = @metadata_map.keys
+      end
+
+      def process_work(work)
+        update_subject(work) unless (work.subject              & @keys_to_check_for).empty?
+        update_creator(work) unless (work.creator.map(&:value) & @keys_to_check_for).empty?
```
jrochkind, line 24: Aha, I had missed this, sorry!

jrochkind: This also let's us see if that if subject AND creator were modified... it'll do two separate saves. If this were something we were going to use a LOT in a high-volume way, we might want to avoid that... just for this simple utility on our actual data set? It's fine!

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
jrochkind: okay, sorry for missing that, nice! just filter_map then!

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
eddierubeiz: Totally

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
eddierubeiz: Got rid of it!

spec/lib/scihist_digicoll/helpers/subject_creator_adjuster_spec.rb
eddierubeiz: One work is totally fine!

lib/scihist_digicoll/task_helpers/subject_creator_adjuster.rb
eddierubeiz: Yeah- this can be turned into a reusable thing next time around.

eddierubeiz:
I just [ran this in production](https://github.com/sciencehistory/scihist_digicoll/files/7499695/results.txt).

PR #1437, opened 2021-11-02
update blacklight to 7.20.1

jrochkind:
Which also includes a fix in blacklight that lets us eliminate an exception from our be_axe_clean accessibility testing.


jrochkind:
Hm, seems like something is wrong with the github actions infrastruture unrelated to this maybe. 

```
Error: The process '/opt/hostedtoolcache/Ruby/2.7.4/x64/bin/bundle' failed with exit code 16
```

jrochkind:
oops no, it was an error on my part, hadn't committed latest Gemfile.lock. Just a confusing error message.

PR #1447, opened 2021-11-08
Fix duplicate html ids in transcript footnote references

eddierubeiz:
This is the simplest thing that could possibly work: keep track of which references have already been output, and only issue an HTML id for the first reference to the footnote.
Includes very simple tests.
Ref #1411


eddierubeiz:
This PR also removes the restriction on `axe_clean`, by the way.

jrochkind:
Very nice!  Instead of passing in an argument `is_first_reference`, what if we're just explicit about exactly what we're asking for, and pass in `suppress_dom_id` or something like that?  The implementation would be the same, just changing the name of the arg/variable to be more about exactly the behavior we're asking for. I think maybe that's more maintainable and direct. What do you think?

eddierubeiz:
Very sane!

app/components/oral_history/footnote_reference_component.rb
```diff
@@ -8,11 +8,12 @@ module OralHistory
   # The tooltip hover is based on:
   # http://hiphoff.com/creating-hover-over-footnotes-with-bootstrap/
   class FootnoteReferenceComponent < ApplicationComponent
-    attr_reader :footnote_text, :number
+    attr_reader :footnote_text, :number, :show_dom_id
 
-    def initialize(footnote_text:, number:)
+    def initialize(footnote_text:, number:, show_dom_id:)
```
jrochkind, line 13: sorry to be nitpicking, but show_dom_id should be optional and default to true maybe? Meh, who cares, I'm gonna merge anyway.

Issue #1451, opened 2021-11-09
Error when destroying an asset from admin UI 


jrochkind:
[scihist_digicoll/production] NameError: undefined local variable or method `work' for #<Admin::AssetsController:0x...>

## Backtrace

    line 43 of [PROJECT_ROOT]/app/controllers/admin/assets_controller.rb: block (2 levels) in destroy
    line 42 of [PROJECT_ROOT]/app/controllers/admin/assets_controller.rb: destroy
    line 26 of [PROJECT_ROOT]/app/controllers/application_controller.rb: block in batch_kithe_indexable

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/82497404)

On https://digital.sciencehistory.org/admin/asset_files/kldt99v

PR #1454, opened 2021-11-09
Fix bad variable reference

jrochkind:
Ref #1451


jrochkind:
Merging and deploying without you, @eddierubeiz

jrochkind:
Note also added a test that failed before the fix!

eddierubeiz:
Thank you. I suspect I'm responsible; I was the last person to edit this controller.

Issue #1458, opened 2021-11-11
non-published assets should not have transcription/translation indexed in a work


jrochkind:
Kind of around the way related to #1428

PR #1459, opened 2021-11-11
avoid indexing transcription/translation from non-published assets in parent works

jrochkind:
ref #1458


eddierubeiz:
Thanks for the tweaks to the tests too.

Issue #1356, opened 2021-10-08
Boost Collections in search results


apinkney0696:
Make Collections more likely to show up first in search results.

For example, ideally, if a user were to search for "Booth," "Bredig," "Beckman," etc., (or other keywords that directly relate to a collection) the first search result would be "Papers of James Curtis Booth," "Papers of Max and Georg Bredig," and "Beckman Historical Collection" respectfully.

"Oral History" [search results](https://digital.sciencehistory.org/catalog?search_field=all_fields&q=oral+history&range%5Byear_facet_isim%5D%5Bbegin%5D=&range%5Byear_facet_isim%5D%5Bend%5D=) is a good example.

jrochkind:
Note that currently if you search for `Bredig Collection` it is the first result and if you search for `Bredig Papers` it is the third result. 

so this is not too bad!

If you just search for "Bredig", the collection page is indeed nowhere to be found. Mainly because the word "Bredig" appears in metadata of individual Works many more times than it does in the collection, including appearing *twice* in the highly-boosted title field for many. 

Noticing that "bredig papers" or "bredig collection" does well makes me personally less concerned. Still, we could try to boost collections, see what it does for results -- we'd want to find a way to "test" somehow that it doesn't interfere with quality of other results. 

I wonder if there's a good way to a list of "sample" queries to check to make sure they aren't imparied. @apinkney0696 , do you know if we can get a list of things people searched for in our search box, from Google Analytics?

jrochkind:
Probably use `bq` (boost query) solr dismax param. 

May have to experiment to see how much boost is necessary to get bredig collection to show up first for query `bredig`, which I think is the goal.

PR #1460, opened 2021-11-11
Boost collection search

eddierubeiz:
One-line change; Ref #1356 . Do not merge before we test this in staging.

### DIFF
--- a/solr/config/solrconfig.xml
+++ b/solr/config/solrconfig.xml
@@ -117,6 +117,8 @@
           all_text_timv^10
         </str>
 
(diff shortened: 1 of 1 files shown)


solr/config/solrconfig.xml
```diff
@@ -117,6 +117,8 @@
           all_text_timv^10
         </str>
 
+        <str name="bq">model_name_ssi:Collection^500.0 </str>
```
jrochkind, line 120: I don't think you need to include `.0` in there, I think just `^500` is fine. But I'll merge it.

jrochkind:
@apinkney0696 approved the effect on staging.

Issue #1428, opened 2021-10-28
adding/changing Asset transcription/translation does not cause work to be reindexed


jrochkind:
So changes you make to an asset indexed text aren't visible in the search index until the Work is reindexed, perhaps by saving the work in admin interface.  this is a problem when we go adding translation/transcription to existing works, the added text might not end up searchable. 

We didn't used to include anything from members in the indexing of the work. So didn't have this issue. 

When we crossed that bridge, we forgot it was an issue. 

One workaround would just be to reindex everything every night or something, since our indexing currently only takes a couple minutes. 

Or we could actually solve it. Perhaps with a new fancy feature in kithe. https://github.com/sciencehistory/kithe/issues/121

(Note, I don't know if this is particularly well-solved in samvera-world either, it would be interesting to check with that community though)

jrochkind:
This one is really tricky. I'm sitting on it still for a bit.

PR #1462, opened 2021-11-11
automatic re-index of parent work when asset changes in relevant ways

jrochkind:
Ref #1428

While not a lot of code, this is some pretty "clever" code, dealing with some obscure parts of Kithe, attr_json, and ActiveRecord. I don't love it, but this is kind of the path we're on with our "automatic solr indexing". 

Tried to comment extensively. 

The ugliest part is that extra `if self.destroyed` line in the model. That one isn't actually covered by tests, I just found it trying out in the actual app, and then had to debug it, a real head-scratcher. 

Going to leave this in draft mode until next week, see if I can think of anything else.


jrochkind:
It doesn't yet have a test for the thing I found last minute to fix in an even *more* confusing way, that deleting an asset was *not* causing re-indexing. 

i want to figure out a way to add an otherwise-failing test for that, otherwise it's done. With some sadness because it is, I realize, kind of convoluted (although concise) code.

jrochkind:
Another thing that makes this all confusing is the code around whether when running in our tests we trigger automatic solr indexes. 

We ordinarily don't want to in tests. But here we want to TEST that, so sometimes have to turn it on.... but still don't want it turned on when we're doing our DB setup. Phew!

app/models/asset.rb
```diff
@@ -58,6 +66,54 @@ def dzi_file
...
+  #   * update PARENT in solr on change, we don't index Assets, but do index
+  #     some aspects of assets in their parent Works.
+  #   * and only when allow-listed attributes we know we index on parent have changed
+  def update_index(mapper: kithe_indexable_mapper, writer:nil, **)
+    if should_reindex_parent_after_save?
+      # WEIRD workaround, in some cases the parent still this record in memory
+      # even though it's been, and will use that in-memory list of members for
+      # indexing in Solr! If that's going on, we need to make sure to reset
+      # the association.
+      if self.destroyed? && parent.members.loaded? && parent.members.include?(self)
```
eddierubeiz, line 79: Meaning -- in some cases the parent work still contains this asset in its `members` association even though the child asset has already been deleted, and could attempt to index information from the deleted asset.

app/models/asset.rb
```diff
@@ -58,6 +66,54 @@ def dzi_file
...
+    elsif self.saved_change_to_published?
+      # if published status changed, we have to reindex if and only if we HAVE any indexed attributes,
+      # to include them in index.
+      indexed_attributes.any? { |attr| self.send(attr).present? }
+    else
+      # an ordinary save (including create), did any attributes of interest CHANGE? (Including removal)
+      # then we need to reindex parent to get them updated in index.
+      indexed_attributes.any? { |attr| self.attr_json_changes.saved_change_to_attribute(attr).present? }
+    end
+  end
```
eddierubeiz, line 115: OK, this all makes sense.

eddierubeiz:
OK. The tests are invaluable here as documentation, but so are the comments. Merging!

Issue #1410, opened 2021-10-20
Recently Added items to populate based on "Publication Date"


apinkney0696:
This will put the most recently published items onto the homepage, rather than the most recently edited as it is set at present.

apinkney0696:
I definitely use the sort by "created date" and "last modified date". Both are useful. 

I agree, end-users probably don't need any more sorts than "recently added" and the date of the physical object

eddierubeiz:
Assuming our intent is to keep track of "the date the item first went public", temporarily unpublishing and republishing a work should leave the publication date intact. @apinkney0696 , can you confirm please?

(Context: our workflow when something is wrong with a work's assets  or derivatives often involve temporarily taking a work offline while we modify its contents.)

eddierubeiz:
I also want to make sure that this date never gets confused, in the code or the UI, with the bibliographical publication date. I think it's a moot point as long as the value of the date is never displayed to the public; my understanding of the above implies that we would only use this date to sort items in three places:

1) the recently added items on the home page
2) the "Recently added" sort of [public search results](https://digital.sciencehistory.org/catalog?search_field=all_fields&q=&range%5Byear_facet_isim%5D%5Bbegin%5D=&range%5Byear_facet_isim%5D%5Bend%5D=

I think that's it -- let me know if I'm missing anything.
3) a new sortable column in the [staff work listing](https://digital.sciencehistory.org/admin?q%5Bparent_id_null%5D=true&q%5Bs%5D=created_at+desc)

jrochkind:
I was assuming that unpublishing and re-publishing a work would re-set the publication date... because it's not obvious to me that would be wrong either, and seems more straightforward.  Either one can be implemented. 

How often do staff unpublish a published work anyway?  I didn't think that was a part of workflow, but my understanding of workflow may again be outdated. 

I agree that we don't show this date on the screen anywhere. We *already* offer a publically visible "recently added" sort, we're just making it accurate.

jrochkind:
@eddierubeiz Also, how about doing the setting of the publication_date on the model as a `before_save` hook (or maybe `before_validate` actually), and doing that as a separate PR of it's own?

eddierubeiz:
Agreed - now that I think about it, it belongs in a hook, totally.

apinkney0696:
[...]

PR #1465, opened 2021-11-12
add published_at date

eddierubeiz:
First step in the direction of ref #1410 .
For now, no changes to the UI; just the migration, controllers and a couple tests to ensure the value is being populated.

### DIFF
--- a/spec/controllers/admin/works_controller_spec.rb
+++ b/spec/controllers/admin/works_controller_spec.rb
@@ -136,6 +136,7 @@ RSpec.describe Admin::WorksController, :logged_in_user, type: :controller, queue
         expect(flash["error"]).to be_blank
 
         expect(publishable_work.reload).to be_published
+        expect(publishable_work.published_at).to be_within(1.second).of Time.now
       end
     end
   end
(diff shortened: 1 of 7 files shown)


jrochkind:
Can you call the column `published_at` to match the ActiveRecord `created_at` and `updated_at`  columns? Or something else like that, that ends in `at` and is a datetime field?

spec/controllers/admin/works_controller_spec.rb
```diff
@@ -321,7 +322,8 @@
       let(:work) do
         create(:work, :published).tap do |w|
           w.members += [asset_child, work_child]
-          w.published = false
+          w.published = false          
+          w.published_at = nil
```
jrochkind, line 326: Normally we can just do:

    create(:work, :published, published_at: nil)

I think the members/published stuff is being done in a `tap` because of weird previous stuff around representative validation or something? Not sure why else we do `create(:work, :published)`, and then immediately set `published` to false! We don't need to fix that problem here (although we could), but let's not add to the weirdness with more in the `tap` if we don't need to... but really not sure what's going on this test altogether, it's a weird test where we create a published work and then set it to not published and then... publish it?

Actually... the logic you added is in the _model_, we should maybe test it in a model test rather than a controller test? Maybe leave the controller test alone, but add a test in a work model test?

jrochkind:
Nicely done! 

Feel free to merge after some test adjustments -- I think there should be a model-level test instead of or in addition to the controller-level test.

spec/controllers/admin/works_controller_spec.rb
eddierubeiz: Yeah. I'm adding a model test, and I can totally fix that `tap`. It doesn't need to be there anymore- correct.

PR #1469, opened 2021-11-18
update blacklight to 7.21

jrochkind:
With removal of axe accessibility check exceptions. 

Re-do of #1437 again, with a newer version of Blacklight that won't have the problem we had before.


eddierubeiz:
Awesome!

PR #1472, opened 2021-11-22
New sort for recent items on the home page takes into account published_at, if available.

eddierubeiz:
Included a basic smoke tests for recent_items, which involved a small refactor.

### DIFF
--- a/app/presenters/recent_items.rb
+++ b/app/presenters/recent_items.rb
@@ -9,10 +9,14 @@ class RecentItems
   @@last_refresh = nil
   @@bag = nil
 
-  def initialize
-    @how_many_works_to_show = 5
-    @how_often_to_change = 60 * 10 # ten minutes in seconds.
-    @how_many_works_in_bag = 50
+  def initialize(
+    how_many_works_to_show: 5,
+    how_often_to_change:    60 * 10, # ten minutes in seconds
+    how_many_works_in_bag:  50 )
+
+    @how_many_works_to_show = how_many_works_to_show
+    @how_often_to_change = how_often_to_change 
+    @how_many_works_in_bag = how_many_works_in_bag
   end
 
   def recent_items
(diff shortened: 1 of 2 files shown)


app/presenters/recent_items.rb
```diff
@@ -9,22 +9,39 @@ class RecentItems
...
     @@bag = nil if time_for_a_new_bag?
     # The randomization function is seeded from @@last_refresh,
     # which is a class instance variable.
     # Hence, the shuffled selection is the same for all users at a given moment.
-    fetch_bag.shuffle(random: Random.new(@@last_refresh))[0... @how_many_works_to_show]
+    maybe_fetch_bag.shuffle(random: Random.new(@@last_refresh))[0... @how_many_works_to_show]
+  end
+
+  def fetch_bag
+    Work.where('published = true').
```
jrochkind, line 31: Would be better to write this as `(published: true)`. (I know it wasn't before either).

app/presenters/recent_items.rb
```diff
@@ -9,22 +9,39 @@ class RecentItems
...
+      includes(:leaf_representative).
+      order(published_at: :desc, updated_at: :desc).
```
jrochkind, line 33: The trick is to make sure if published_at is null, those things wind up at the END of the list (after any items with non-null published-at), so updated_at then only really comes into play to break the tie of all those `nil` published_at, at end of list. (Which will only be an issue temporarily until we have enough things with non-nil published_at!)

It may be doing that already -- but the test doesn't seem to test it! If we're going to have a test, probably makes sense for the test actually to test that part? 

PG docs seem to say:

> The NULLS FIRST and NULLS LAST options can be used to determine whether nulls appear before or after non-null values in the sort ordering. By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.

We're doing published_at `DESC`, so `NULLS FIRST` will be the default -- not what we want actually! So may specifically have to put `NULLS LAST` in the `order` somehow? Make a test that fails, then fix it?

https://www.postgresql.org/docs/8.3/queries-order.html

eddierubeiz: Yes -- good thinking!

app/presenters/recent_items.rb
eddierubeiz: Fixed/

app/presenters/recent_items.rb
eddierubeiz: Fixed - using `arel` now.

jrochkind:
nice, thanks!

jrochkind:
I will edit description to add link to issue...

PR #1473, opened 2021-11-22
Log published_at when a work, asset or collection is published, via a Concern module.

eddierubeiz:
This code was previously included in models/work.rb , but we're moving it out to a concern shared by all 3 - `Work`, `Collection`, and `Asset`. Should make our lives less confusing.

Ref #1474 

Moved the old test at `spec/models/work/published_at.rb` to `spec/models/record_published_at.rb` and added tests for the other 2 classes.


jrochkind:
Do we want to DRY this up with a mix-in module? Or not worth it?

(Like, if we decided we wanted to change this to keep the first one, an alternative we discussed, we'd have to remember to do it in both places, or they'd be inconsistent. Does that matter?)

If we don't actually care about collection published_at for any real use case, do we just not worry about this at all?

eddierubeiz:
Yeah -- I don't feel strongly about this either way, but I also didn't want to put this code in an unrelated PR.

jrochkind:
I am coming around to thinking you are right about collections...  without doing published_at for collections, they will never show up in "recently added" sort. Which is maybe fine, nobody asked for it or will probably notice, but does seem a bit sloppy. 

And then as long as we're being consistent, should we add to Assets too?

I think that maybe suggests doing it as a simple mixin, so we can copy it to all three DRY?

eddierubeiz:
I'm going to say we ignore assets, as they don't show up in search results.

jrochkind:
So we'd still have it in the database even if it doesn't show up in searches. This is for the db, not just Solr indexing. 

But even if we were just doing too, I lean towards mixin anyway for DRY? Are you leaning against the mixin and for the repeatition?

I'd say once we have the mixin, might as well do it or Assets too then. It is simple and just keeps things consistent and from surprising us -- which is pretty much the only reason we're doing it for Collections after all, we don't really have a use case demand for it.

app/models/record_published_at.rb
```diff
@@ -0,0 +1,13 @@
...
+module RecordPublishedAt
+  extend ActiveSupport::Concern
+  included do
+    before_validation :set_published_at
+    private
+    def set_published_at
+      self.published_at = DateTime.now if published? && published_changed?
+    end
+  end
+end
```
jrochkind, line 13: Hmm, if the tests pass this must work?  But this isn't the normal way to do it. You have to put the `before_validation` in an `included do` block... but the `def set_published_at` should not be. 

The `ActiveSupport::Concern` docs indeed aren't very clear on this. :( 

But check out [this example](https://blog.appsignal.com/2020/09/16/rails-concers-to-concern-or-not-to-concern.html).   

I think you should be able to move the `def` outside the `included do` and it will still work fine, and be a bit more standard/normal (with possibly implications for how it works). Do you mind trying it?

Except for that looks great! Go ahead and merge after that!

eddierubeiz: Thanks again for catching that. Will merge after the tests pass.

PR #1481, opened 2021-11-24
Use `published_at` for the "Newest Added" and "Oldest Added" search results sorts.

eddierubeiz:
- Add new solr sort using `published_at` (the tiebreaker for the sort, if there's no value, is `created_at`);
 - Use that sort in search results, with the existing labels "newest added"  (which were formerly actually sorting by `created_at`);
 - removed admin-only "oldest added", no longer needed. 
 -  Display `published_at` on admin work metadata show page along with its friends `created_at` and `modified_at`;
 - Add two new sorts (visible to admins only) to explicitly sort works by `created_at`, `asc` and `desc`;
 - We are indexing published_at for collections too, but at the moment that date isn't being set when you publish a collection.

We are also not changing the display of the list of works in `/admin`, which can be sorted by `created` or `last modified`, neither of which is our concern here.

Ref #1410 .

This will require a reindex.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -423,7 +423,7 @@ class CatalogController < ApplicationController
 
     config.add_sort_field("recently_added") do |field|
       field.label = "recently added"
-      field.sort = "date_published_dtsi desc, date_modified_dtsi desc"
+      field.sort = "date_published_dtsi desc, date_created_dtsi desc"
       # will be used by our custom code as default sort when no query has been entered
       field.blank_query_default = true
     end
@@ -431,7 +431,7 @@ class CatalogController < ApplicationController
     # limit to just available for logged-in admins, with `if ` param
     config.add_sort_field("oldest_added") do |field|
       field.label = "oldest added"
-      field.sort = "date_published_dtsi asc, date_modified_dtsi asc"
+      field.sort = "date_published_dtsi asc, date_created_dtsi asc"
       field.if = ->(controller, field) { controller.current_user }
     end
 
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -428,13 +428,6 @@ class CatalogController < ApplicationController
       field.blank_query_default = true
     end
 
-    # limit to just available for logged-in admins, with `if ` param
-    config.add_sort_field("oldest_added") do |field|
-      field.label = "oldest added"
-      field.sort = "date_published_dtsi asc, date_created_dtsi asc"
-      field.if = ->(controller, field) { controller.current_user }
-    end
-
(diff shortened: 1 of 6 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -423,7 +423,7 @@ def active_sort_fields
 
     config.add_sort_field("recently_added") do |field|
       field.label = "recently added"
-      field.sort = "date_created_dtsi desc"
+      field.sort = "date_published_dtsi desc, date_created_dtsi desc"
```
jrochkind, line 426: don't we need to pay attention to "nulls last" here again (and test in tests maybe? Not sure if there are already suitable analagous tests somewhere).

app/controllers/catalog_controller.rb
```diff
@@ -432,7 +432,7 @@ def active_sort_fields
 
     config.add_sort_field("oldest_added") do |field|
       field.label = "oldest added"
-      field.sort = "date_created_dtsi asc"
+      field.sort = "date_published_dtsi asc, date_created_dtsi asc"
       field.if = ->(controller, field) { controller.current_user }
     end
```
jrochkind, line 437: I suspect that admin doesn't really need "oldest added" at all -- they may instead need "created_at" both ascending and descending, now that "recently added" isn't really "created at" 

We could check with @apinkney0696 to be sure, but I think it's a good first bet, based on previous conversations.

app/controllers/catalog_controller.rb
eddierubeiz: Nulls are in fact sorted last in practice; a test couldn't hurt.

app/controllers/catalog_controller.rb
eddierubeiz: That's totally fine. I'd rather this be another PR though once we've checked with Annabel.

jrochkind: OK, but what about adding created_at asc/desc (admin-only) in this PR then? I am 90% sure Annabel wants it.  That's what they used to get from "recently added" but won't anymore. 

How will we remember to bring this up with Annabel next week?

app/controllers/catalog_controller.rb
jrochkind: Oh right, this is Solr not postgres!

If there's already similar tests and a covenient place to put the tests, let's do it. Otherwise, not worth too much trouble, if you tested manually.

app/controllers/catalog_controller.rb
eddierubeiz: Done!

app/controllers/catalog_controller.rb
eddierubeiz: Added tests. (They fail!)

Issue #1452, opened 2021-11-09
When we have a logged in user for a HoneyBadger error, add it to HoneyBadger context


jrochkind:
Then we know who to talk about about the error! 

https://docs.honeybadger.io/lib/ruby/getting-started/adding-context-to-errors/

jrochkind:
https://www.rubydoc.info/gems/honeybadger/Honeybadger%2FAgent:context

PR #1483, opened 2021-11-29
add current user to honeybadger context

jrochkind:
Useful so we can check with the person doing the ingest when an error happens in ingest. Although actually won't work if the error happened in a bg job as it often does, I now realize. Becuase there's no logged in user there. Oh well, can't hurt might help in some situations.

Ref #1452


eddierubeiz:
Staff users who mistakenly blame yourselves for our bugs and consequently don't report them... we are coming for you.

Issue #1449, opened 2021-11-09
Export metadata in .csv format for patron


apinkney0696:
A representative from the Newman Numismatic Portal at Washington University in St. Louis, the funder of the Booth papers digitization, has requested an export of the metadata for [334 items](https://digital.sciencehistory.org/collections/kk91fk97p?q=mint&sort). 

They would like the data compiled in one .csv format document and they have no hard deadline to receive the data.

Please export to P:\Special Collections\External Access\Newman Numismatic Portal so I can share directly with the patron.

Let me know if you need any other information from them, I'm happy to ask!

jrochkind:
I am not sure how to put "the metadata" in a CSV file. 

We have a LOT of metadata, and much of it is structured/complex/nested in format, that doesn't obviously map to a CSV file.  Figuring out how to "flatten" it into CSV is non-trivial. 

It's also not necessarily entirely clear what attributes are part of 'the metadata' or not. Does this include *all* the "adminsitrative" metadata we have, for instance?  Does it include asset-level metadata (administrative or not), or just work-level?  

Someone is going to need to specify what this CSV file format of our metadata looks like. I'm not sure who that is. 

Do we think they actually plan to use this for something, or they just want it to say they have it?  Like how much thought or communication with them should we put into it?

It would be somewhat easier to give them a whole bunch of .json files than it would be to give them one CSV file. 

But if they actually plan to use this data for something, we probably need to hear from them for what and what they want in the CSV file specifically.

apinkney0696:
I figured I might need to get some more info from them. Just wanted to put the request in to figure out what other info I needed from them, thanks for the prompt response!

> Do we think they actually plan to use this for something or do they just want it to say they have it? Like how much thought or communication with them should we put into it?

In their original email, they ask "Is there any possible way to do an export of metadata for these items? Opposed to individually looking at each time." I'm not sure what that means they would like to do with it, something I can ask about.

I'm assuming they only want the work-level descriptive metadata, but again, I will ask to get more details. Is a CSV possible, or is there another format I could recommend that would be more convenient for you? I think they want to be able to see everything at once, so I doubt they would like a whole bunch of .json files.

jrochkind:
I don't really know how to answer if a CSV is possible. 

If they tell me exactly what columns they want in a CSV and what the data in those CSV files is, I can create a CSV to those specs. 

[...]

PR #1484, opened 2021-11-29
Export Booth papers metadata to a csv file

eddierubeiz:
Ref #1449

### DIFF
--- a/lib/tasks/data_fixes/mint_metadata_export.rake
+++ b/lib/tasks/data_fixes/mint_metadata_export.rake
@@ -1,0 +1,39 @@
+require 'csv'
+namespace :scihist do
+  namespace :data_fixes do
+    desc """
+      bundle exec rake scihist:data_fixes:mint_metadata_export > booth.csv
+      heroku run rake  scihist:data_fixes:mint_metadata_export     --app scihist-digicoll-production > booth.csv
+
+      For all Booth collection items that come up on a keyword search for `mint`,
+      extract the four following fields to a .csv:
+        dc:identifier
+        dc:title
+        dc:date
+        dc:description
+    """
+    task :mint_metadata_export => :environment do
+      booth_collection = Collection.find_by_friendlier_id('kk91fk97p')
+
+      # These extra items probably should contain subject heading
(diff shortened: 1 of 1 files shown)


jrochkind:
Looks fine!

If it wasn't a hacky one-off, I'd say: Why are you fetching the entire Collection to look at what has "Mint" in the subject, instead of just doing a SQL query for only things that have "Mint" in the subject?  

That would be the "right" way to do it. It might be worthwhile as an exersize to do it that way, but it also doesn't really matter for this use-once-then-throwaway with the actual size of our DB. 

I wasn't sure how we were going to get the data from a heroku dyno... but if this works, I guess this works? I'm not totally sure it will, I think the rake exec via heroku is going to insist on putting some prefatory stuff at the beginning of the stdout that will mess up the CSV. Tested on staging yet?  Again, for this simple once-then-done we could just manually edit the output to make it the desired CSV. 🤷

eddierubeiz:
@apinkney0696  -- take a look at this [test run from staging](https://github.com/sciencehistory/scihist_digicoll/files/7621232/booth.csv) when you have a minute.
Thanks!

eddierubeiz:
Yup - six lines at the top of the CSV (4 from Scout) need to be removed.

jrochkind:
I wouldn't think they'd want those XML-like tags `<dc:title>` in there, I can't see how they are going to help them at all in dealing with the CSV, I'd just put the values in in the CSV! But whatever they want!

eddierubeiz:
Re: how I got the list of items, I just wanted to make sure I had the 334 items requested in the issue -- see discussion there for details.

jrochkind (approved):

apinkney0696:
Hi Eddie, this looks great to me! I agree with Jonathan that the element tags <dc:title>,etc probably shouldn't be included, but that's a pretty easy search and replace that I can do in excel. I also noticed that there are some unicode hex character codes such as "&#x2019;" within some of the descriptions. Is there any way we can make those their typical punctuation symbols? 
Thanks!

eddierubeiz:
Oh, I got rid of the XML tags already - they won't be there when I run this in production.

PR #1489, opened 2021-11-30
Avoid n+1's in Work#batch_update

### LINKED ISSUE #1488: Batch-edit of collection membership can result in heroku request timeouts (R12)
@apinkney0696 was using the new collection batch edit feature to move oral histories from one collection (OH) into another (Pew biomedical), and reports the problem kicks in at batch sizes between 25 and 50 works.

jrochkind:
This may or may not be enough to resolve #1488, but definitely will help.


jrochkind:
Doesn't solve the problem -- which it turns out is largely slowness in writing to Solr index -- but still worth plugging these holes.

PR #1498, opened 2021-12-02
Batch update does not do solr indexing inline, but sends it to bg job

jrochkind:
This one will actually close off #1488  fully. Should be able to do batch updates of ~100 in just a couple seconds (which is still honestly too long for an http response, but ok for now), with solr indexing sent to a bg job.


jrochkind:
Hmm, this is still slower on staging than I think it should be, need to investigate more.

jrochkind:
Hm, it appears to still be doing the indexing inline. Then it's ALSO sending it in a bg job! But havne't succesfully stopped it from doing it inline. :(

jrochkind:
omg, there was a bunch of dead/unreachable code in Admin::WorksController, accidentally left in by me in https://github.com/sciencehistory/scihist_digicoll/commit/69a578f3cae1d4a68ecb033003ae9129092f5a6e

That was one thing making this more confusing than it should have been! Will fix in this PR too.

Issue #1492, opened 2021-12-02
Remove "Project" Metadata field


apinkney0696:
All oral histories with a project assigned have been moved into their new collections. This field can now be removed from the metadata form, the data can be deleted from the individual records, and the Project facet can be removed from the refinement options on the search page.

MUST have #1503 merged AND deployed, before removing this data

- [x] remove attr in Work and all places it's referenced (edit, show, indexing and facet config) #1505
- [x] remove actual data from DB, from the json_attributes table, "project" key. #1504

PR #1504, opened 2021-12-07
Remove "project" data from json_attributes column in db

jrochkind:
We are no longer tracking data here, using "Collection" feature instead. Ref #1492

Note that #1503  MUST be merged AND deployed before running this -- can't remove the data until we stop trying to use it there!


eddierubeiz:
I approve this PR!

jrochkind:
OK, going to merge and deploy. 

It also doesn't need to run until after #1505 is deployed, it's runnable after that.

PR #1505, opened 2021-12-07
Remove 'project' attribute from Work

jrochkind:
We now track this relationship via Collections instead. Ref #1492


jrochkind:
After merge and deploy, run `rake scihist:data_fixes:remove_project_data` task from #1504, to remove JSON data from DB that is no longer referenced by app.

Issue #1511, opened 2021-12-10
Searchability of OHMSed Interviews in DC 


sunicolita:
I noticed a spelling issue in the Molina interview OHMS segment titles and descriptions; "Berkeley" is misspelled as "Berkley" and shows up 6 times in a search of the [actual record](https://digital.sciencehistory.org/works/8pmnk1q#tab=ohToc). However, when searching for the word "Berkley" over the whole DC, the Molina record does not appear. While we are planning on fixing the spelling, this brought up questions about how OHMSed interviews appear in DC searches. Can we investigate if there is a bug that is impacting these search capabilities, and if there isn't, can we look into ways of making all OHMS fields included in DC-wide searches? 

Thank you!!

eddierubeiz:
As I understand it, the "Berkley" string is only in the descriptions of OHMS "segments" in the table of contents, not in the searchable transcript proper.

Now that I think about it, I feel like we may have consciously decided not to index those descriptions, but my memory is foggy. Jonathan will remember better than me. To be continued!

sunicolita:
The segment detail is correct, Eddie, thanks for pointing that out (and for your help today in general). Looking forward to hearing what Jonathan has to say!

jrochkind:
Yep, it looks like from the Table of Contents/Index/"Segment" section, we are not adding the title and description to the searchable index. 

The only thing we are adding to the searchable index from the OHMS "Index" are the "keywords". 

I don't recall if this was intentional or why. 

It sounds like we should add title and description of the "Index Segments" as well, @sunicolita ? We can do that. 

I guess I'd add them in the same way that the "transcript" text is added -- relativevly low level of "boosting" (matches on other metadata fields like title or subject are considered a "better match" resulting in higher position in search results), as well as with matches showing up in the "snippets" area in search results, as the transcript text does. 

Does this make sense @sunicolita  and @apinkney0696 ?

sunicolita:
That sounds perfect to me, @jrochkind - very helpful. Thank you!

apinkney0696:
Sounds good to me too.

jrochkind:
OK, I'll move this ticket to "Ready" column, so we get to it when we next have time!

PR #1515, opened 2021-12-13
index ohsm ToC segment title and synopsis too

jrochkind:
Ref #1511

Will need to reindex after deploy


jrochkind:
tested in staging, seems to do what's intended.

Issue #1513, opened 2021-12-13
Increase tolerance of tests that insist on time differences < 1 second.


eddierubeiz:
A test I wrote recently just failed on CI:

`./spec/indexers/work_indexer_spec.rb:77`

```
  1) WorkIndexer oral history indexes created, modified, and published dates
     Failure/Error: expect(Time.parse(output_hash["date_created_dtsi"  ].first)).to be_within(1.seconds).of Time.now
       expected 2021-12-13 19:08:04.000000000 +0000 to be within 1 of 2021-12-13 19:08:05.004951380 +0000
     # ./spec/indexers/work_indexer_spec.rb:79:in `block (3 levels) in <main>'

```

Make that 5.seconds. We don't really care if the two dates are a bit over 1 second apart.

jrochkind:
On testing for time in rspec, a more reliable way might be to use https://github.com/travisjeffery/timecop

Better than guessing how much time something might take.

kinduff:
Better alternative without introducing a new dependency:

```ruby
RSpec.configure do |config|
  config.include ActiveSupport::Testing::TimeHelpers
end
```

Then you can use it in a `before` block like so:

```ruby
before { travel_to Time.local(2021) }
after { travel_back }
```

This will freeze time during the test execution.

jrochkind:
Hey thanks for that @kinduff !

kinduff:
I'm pretty curious and always looking for projects that need contributions. If there's something I can help with, please let me know.

jrochkind:
Hey @eddierubeiz , looking for places to improve this (by searching project for the `be_within(1.second)` pattern), I noticed the test file `./spec/models/record_pubished_at.rb` -- note that it's an error that it doesn't end in `_spec.rb`, all spec files need to end in `_spec.rb` for our rspec suite to run them automatically. Will fix that too. 

I'm actually using the Rails time helper `freeze_time` method. 

https://api.rubyonrails.org/v5.2.6/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time

jrochkind:
@kinduff Thanks! While we work in the open, and produce open source, this particular application is really meant only for use of the non-profit Science History Institute (https://sciencehistory.org), and we both work for that org. I'm not sure it would be a good fit for outside contribution. 

[...]

PR #1518, opened 2021-12-20
Change specs around published_at to freeze time to be more reliable as tests

jrochkind:
Ref #1513


eddierubeiz:
Thanks for fixing the test and the filename too.

Issue #1365, opened 2021-10-12
Automatic retries in Blacklight for Solr timeouts


jrochkind:
We configured Solr timeouts in #1345, next step is making BL retry once after a timeout -- the hope would be that our Solr is being temporarily flakey but making the request a second time would succeed. 

Might require a PR to Blacklight to make possible. 

We want to get acess to the underlying RSolr/Faraday HTTP connection object, to configure retries, probably using the Faraday retry middleware (which will require some experimentation to make sure we understand how it works, as it's not very well documented). 

To do that might require changes to Blacklight.

jrochkind:
As far as error conditions we might need to recognize:

When Solr is in process of restarting, it can return 404 or 502. https://app.honeybadger.io/projects/58989/faults/82007907/01FPRGXFSRW09AE24EBAVVEEBQ?page=0

For other Solr errors, I think often plain 400? And in some cases 500?

jrochkind:
Looking at Faraday retry middleware, a bit hard to figure out how to use it, make sure it does what we expect. 

https://lostisland.github.io/faraday/middleware/retry

Found this: https://github.com/lostisland/faraday/pull/1350/files  Although it doens't matter too much for us. 

Currently looking at how to determine retry_amounts with `backoff_factor`, it doesn't seem to be working as I expect, may end up filing more on faraday, we'll see.

PR #1520, opened 2021-12-20
customize Faraday middleware on RSolr client to do retries on Solr

jrochkind:
Unclear how safe this Blacklight API is to use, overriding private method in Repository sub-class. But best thing we found to accomplish Solr retries in Blacklight.

Needs:

- [x] Blog this for Blacklight community
- [x] Test this somehow

Ref #1365

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -170,7 +170,9 @@ end
 
 ##
 # Added by blacklight. Not sure why rsolr isn't just a BL dependency.
-  gem 'rsolr', '>= 1.0', '< 3'
+  # We need RSolr 2.x to get faraday-based connection so we can customize
+  # middleware, so we customize dependency to require 2.x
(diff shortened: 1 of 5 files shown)


jrochkind:
Current failure is unrelated to us specifically, seems to be a docker-hub outage, which ends up effecting github CI, cause the whole internet is a network of fragile dependencies now. 

https://status.docker.com/

jrochkind:
Blogged it https://bibwild.wordpress.com/2022/01/05/blacklight-automatic-retries-of-failed-solr-requests/

eddierubeiz:
Very cool. Thanks for the tests in particular.

jrochkind:
Thanks, those tests took me like over a day to figure out, but seemed worth it in this case!

PR #1522, opened 2022-01-03
even top and bottom whitespace in deai footer

jrochkind:
It was annoying me that the top and bottom padding in the footer had ended up somehow uneven. Fixed, with maybe a hair more breathing room too.

Yeah, I think all of this use of overrides and especially \!important means our css is a bit out of control, but so it goes.


jrochkind:
Hm, test is failing in Solr setup, unrelated to this particular code. That is troubling though, not sure how to fix it. 

~~~
Run bundle exec rake solr:start
Starting solr at http://127.0.0.1:8989/solr/...
rake aborted!
Failed to execute solr create: 
ERROR: 
Core 'scihist_digicoll_test' already exists!
Checked core existence using Core API command:
http://localhost:8989/solr/admin/cores?action=STATUS&core=scihist_digicoll_test
~~~

PR #1523, opened 2022-01-03
When trying to start up solr, if the collection name already exists just keep going

jrochkind:
This came up because when we switched how we're doing CI Solr setup, we ended up having GH Actions cache a copy of solr that already had the collection. I think. Not totally sure why this wasn't always a problem.

Anyway, now if the collection already exists, instead of bombing when we try to create it, we just continue on happy.

Not totally sure if this will end up causing additional problems. This stuff starts getting messy, it's true. This also isn't a very tidy way to catch an exception, `solr_wrapper` doesn't give us specific exception class to match we have to look at message, and hope it doesn't change.


jrochkind:
@eddierubeiz I'm going to go ahead and merge this myself to keep going, but you might want to take a look at it to see... the messiness.

Issue #1525, opened 2022-01-04
Additional Credit Field additions


apinkney0696:
(1) Add "Transcribed by" as a drop-down option

(2) Add "Transcribed by Jocelyn R. McDaniel" to Bredig documents that currently have the metadata "Translated by Jocelyn R. McDaniel" 

(3) Add "Gudrun Dauner" to the name drop-down options

jrochkind:
@apinkney0696 fyi this is live and complete on production!

PR #1532, opened 2022-01-11
Additional Credit Field additions 

jrochkind:
ref #1525

- Add "Transcribed by" as a drop-down option
- Add "Gudrun Dauner" to the name drop-down options
- rake task: Add "Transcribed by Jocelyn R. McDaniel" to Bredig documents that currently have the metadata "Translated by Jocelyn R. McDaniel"

After deployment need to run `heroku run rake scihist:data_fixes:add_transcribed_by_mcdaniel -r production`. 

(Should we come up with a new pattern, possibly using migrations, so these data fixes automatically get run on deploy? Maybe. But I'm not doing it yet).

### DIFF
--- a/app/models/work/additional_credit.rb
+++ b/app/models/work/additional_credit.rb
@@ -5,6 +5,7 @@ class Work
     ROLES = ['photographed_by', 'translated_by', 'transcribed_by']
     NAMES = [
       'Douglas Lockard',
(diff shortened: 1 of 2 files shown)


jrochkind:
@apinkney0696 fyi, this is live and complete on production!

PR #1536, opened 2022-01-13
switch to ruby 3.0!

jrochkind:
Changed .ruby-version file, then chruby switched to that ruby, and ran `bundle install` to update Gemfile.lock with ruby version change only.


jrochkind:
Verified, by looking at github actions build output, that this is testing on ruby 3.0.3 as a result of the .ruby-version file change. 

And all tests green, hooray.

eddierubeiz:
Awesome. Going to merge Tuesday -- no point in doing it now!

Issue #1526, opened 2022-01-04
Funding Credit on Collections pages


apinkney0696:
Related to ticket #1358 

Create collection-level metadata fields to credit an external funder: 

1. a field for the funder name
2. a field for a link to the funder website if applicable
3. and a field to upload an image of the funding organization's logo.

The collections page should display the text "Digitization and cataloging of this collection made possible through the generosity of [name of donor or funder]" (hyperlink the name of the funding agent if possible, otherwise, add the link to the page separately), and the image of the funder's logo.

As far as the page layout goes, see the attached image for an idea. I'm open to different ideas and playing around to make things fit. The mobile/compressed view would simply stack all of the elements.
![IMG_1962](https://user-images.githubusercontent.com/78036563/148087968-9f6b847a-2361-4c99-a3e9-26f53e268d93.jpeg)

jrochkind:
I think we previously discussed that as an implementation efficciency, we might not really allow upload of the image, but just have a fixed set of images in the sourcecode, that can only be changed by developers (me and Eddie). Since we don't expect very many of these to be added very often. 

We also briefly considered having a "funder" object in the database as a separate normalized entity, so you'd edit/add/remove "funders" in a separate place, then just choose from the list of "funders"  in the collection. But I think we decided _not_ to do that, and that it was unlikely funders would be re-used much, and  okay to just repeat data when they were, no need to complicate the data model for it

jrochkind:
@apinkney0696 is "Digitization and cataloging of this collection made possible by the generosity of" fixed text, that would automatically appear for any collection using this feature, the metadata is just "CLIR", then that fixed text is added in front of it on display?

apinkney0696:
> I think we previously discussed that as an implementation efficciency, we might not really allow upload of the image, but just have a fixed set of images in the sourcecode, that can only be changed by developers (me and Eddie). Since we don't expect very many of these to be added very often.

That's right! I forgot about that. So that part of the ticket request should be adjusted to developers fix the logo images to the pages. I'll have to compile a list of the collections that are going to need this work and their corresponding logo images for you.

> We also briefly considered having a "funder" object in the database as a separate normalized entity, so you'd edit/add/remove "funders" in a separate place, then just choose from the list of "funders" in the collection. But I think we decided not to do that, and that it was unlikely funders would be re-used much, and okay to just repeat data when they were, no need to complicate the data model for it

[...]

PR #1537, opened 2022-01-13
collection funding credit

jrochkind:
ref #1526


eddierubeiz:
Quick note: I'd volunteer to write a couple tests for the new model and controller code. It doesn't have to happen now, but I feel it may serve us well in the long run.

jrochkind:
Oh you're right, i think I meant to circle back and write tests but didn't get to it, thanks for catching. It would have been legit to not merge without it I think...  I guess I'll try to do it this week, it's going to feel even more painful if we let it sit I think. 

Man, the little maintenance tasks are really piling up these days... I guess the more features you write, the more time you have to spend maintaining the codebase, I'm realizing. I might not get to the point where I'm ready to really dive into Video next week, with all these things piling up... I guess some of it is that you've been off on EAD for a while too.

eddierubeiz:
C'est la vie, c'est la guerre!

Issue #1539, opened 2022-01-18
ruby 3.0 can't be deployed: error on rake assets:precompile


jrochkind:
Looks like a bug with ruby 3.0 in the `faster_s3_url` gem I wrote. Wasn't discovered until we tried to push to production, because we don't generally run `rake assets:precompile` in dev or test...  I think maybe this has happened before, may we should add a `rake assets:precompile` to our Github CI?

```
remote:        Running: rake assets:precompile
remote:        rake aborted!
remote:        ArgumentError: wrong number of arguments (given 1, expected 0)
remote:        /tmp/build_b21975c4/vendor/bundle/ruby/3.0.0/gems/faster_s3_url-1.0.0/lib/faster_s3_url/shrine/storage.rb:31:in `initialize'
remote:        /tmp/build_b21975c4/lib/scihist_digicoll/env.rb:298:in `new'
remote:        /tmp/build_b21975c4/lib/scihist_digicoll/env.rb:298:in `appropriate_shrine_storage'
remote:        /tmp/build_b21975c4/lib/scihist_digicoll/env.rb:314:in `shrine_cache_storage'
remote:        /tmp/build_b21975c4/config/initializers/shrine.rb:12:in `<main>'
```

PR #1541, opened 2022-01-18
Need to pass args as real keywords in ruby 3

jrochkind:
We were passing a hash to something that wanted keyword arguments. That doesn't work in ruby 3.0 anymore -- we are dynamically constructing the keyword args from a hash, so we need to use the `**` operator to convert them to actual keyword arguments.

Caused an error in production even though tests didn't catch, because tests didn't exersize this Env code that only is used in development and test environments, for differnet shrine setup. Oops, so it goes.

Ref #1539


jrochkind:
I'm going to merge this myself, but you may want to review it anyway @eddierubeiz !

eddierubeiz:
OK - so you can no longer pass a single hash to a method that expects a set of keyword arguments. Makes a lot of sense to me!

Issue #1534, opened 2022-01-12
Handle reported orphaned S3 asset


jrochkind:
We got report of an orphaned asset on S3! https://app.honeybadger.io/projects/58989/faults/83340039

Eddie says:

> Just one, an original. The asset is https://digital.sciencehistory.org/admin/asset_files/9tmkkl7 . The current file was ingested 10 am last Thursday. The orphaned file itself is here: https://s3.console.aws.amazon.com/s3/buckets/scihist-digicoll-production-originals?region=us-east-1&prefix=asset%2Fa989320f-b090-45a3-97c6-07a25c0a0012%2F&showversions=false

Let's see what we can do to diagnose and take care of this particular thing reported (maybe not much), as well as take a step back now that our "orphan checker" has made contact with a problem, to see what we can do to improve our systems to increase our developer ergonomics and chances of diagnosing next time.

- [x] Make absolutely sure the asset is not referenced by present app -- it's easy to see search through DB to make sure the UUID in the asset path isn't there, but also maybe search through entire DB to make sure no asset has a `file_data/id` with the path either, in case something is weird there?
- [x] Actually delete the orphaned asset
- [x] Should we log all deletes the app does?  Then we might have a better chance of figuring out "wait, why is this orphaned asset here? What asset/work was it previously assoicated with, and where did that go?"   Right now we only have the asset path of the orphaned file, which gives us the UUID fo the asset it (used to) belong to, but nothing more. 
   - Log on deletion could be using ActiveRecord lifecycle callbacks on asset model
   - Things to log could include: asset friendlier_id and title; parent_work (if present) friendlier_id and title; logged-in-user doing the deleting if any (actually this may not be easily available from a model callback, so no big deal); the asset path to match up with the orphan report!
- [x] HoneyBadger notification says "Orphaned items found. See report in the admin pages." -- how about including the direct URL to the relevant admin page right there in the HB notification, to make it even easier to get to?
- [x]  the link from the report points to a download link. More helpful would be a link to the place in S3 where you can see the metadata e.g. https://s3.console.aws.amazon.com/s3/buckets/scihist-digicoll-production-originals?region[…]89320f-b090-45a3-97c6-07a25c0a0012%2F&showversions=false
  *  (let's actually make a method on Asset to provide this DRY, since we use it in several places, and it could change if we change platforms or if AWS changes their admin UI)
[...]

PR #1542, opened 2022-01-18
log when assets are destroyed

jrochkind:
Ref #1534


eddierubeiz:
Great. This will save us a ton of time in the long run.

PR #1545, opened 2022-01-19
cleanup orphan report

jrochkind:
Variety of refactoring cleanup, along with features from #1534

- encapsulate JSONB model access to model, simplifying view a bit
- OrphanReport stores s3 URLs (with bucket) of orphaned files, not just path
- make orphan report urls go to AWS console dashboard
- Trust configured shrine storage, don't try to duplicate un-DRY config
- simplify logic for displaying omission mark for more orphaned list
- display more of orphaned file path
- change 'nightly' to 'weekly' in report, accurately describes current situation


eddierubeiz:
Thanks for cleaning this all up! Looks a lot better already. I can put more work into this code in the coming months too.

PR #1560, opened 2022-01-26
Clean up combined audio creator a bit

jrochkind:
- use standard Shrine#download instead of trying to reinvent it

Noticed that downloads of originals was taking a long time. I don't think this actually speeds it up much, but we were kind of reinventing how to "stream from S3 to a tempfile via shrine", when we could have been using the func built into shrine that has had more developer eyes on it. 

- a few more debug log lines, make it easier to see duration of each part of process

Let's try to log a bit more too.

### DIFF
--- a/app/services/combined_audio_derivative_creator.rb
+++ b/app/services/combined_audio_derivative_creator.rb
@@ -54,7 +54,7 @@ class CombinedAudioDerivativeCreator
 
 
   def cmd
-    @cmd ||= TTY::Command.new(output: Rails.logger)
+    @cmd ||= TTY::Command.new(output: TtyLoggerWrapper.new(logger))
   end
 
   def output_file(format)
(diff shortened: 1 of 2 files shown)


app/services/combined_audio_derivative_creator.rb
```diff
@@ -71,14 +72,13 @@ def duration_of_audio_file_in_seconds(path)
...
   def components
     @components ||= begin
+      logger.debug("#{self.class}: downloading original assets")
       result = []
       audio_member_files.each do |original_file|
-        new_temp_file = Tempfile.new(['temp_', original_file.metadata['filename'].downcase], :encoding => 'binary')
-        original_file.open(rewindable:false) do |input_audio_io|
-          new_temp_file.write input_audio_io.read until input_audio_io.eof?
-        end
+        new_temp_file = original_file.download(rewindable: false)
```
eddierubeiz, line 78: Imagine!

Issue #1551, opened 2022-01-20
re-work our AWS auth config so it can use standard AWS auth lookup in dev


jrochkind:
Right now, in dev, to run the app with S3 connections to dev buckets, you need to either have a local config file like `local_env.yml` file , or have ENV variables defined `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID`. 

In general, I consider the ENV variable approach "preferred" for our ScihistDigicoll::Env stuff generally, I don't want local files. 

However, when used with the AWS auth values specifically, having the ENV variables defined is interfering with my ability to use AWS auth the "normal" way, when doing other work (such as terraform or maintenance work with aws cli) -- ability to do things like have [multiple named profiles in your local AWS config](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) and switch between them, which I like to do to avoid a profile with "full" access when doing my normal work. 

Let's find a way to make the app work where it can default to finding AWS credentials in "normal" way, at least when in dev.

PR #1568, opened 2022-02-08
Allow use of standard AWS credential config in development

jrochkind:
We do this by looking up your credentials and then still explicitly supplying them to AWS resources like S3 (via shrine).  Perhaps even better would be moving to "just use the default credentails", in all environments not just development, assume they are set in the environment (env variables or ~/.aws/ whatever), and let the AWS SDK get them from there. That might lead to less fighting with AWS.  I am not sure why we weren't doing that all along -- perhaps we didn't understand the implications, or it was hard to setup in our original pre-heroku ansible/ec3 infrastructure.

But one step at a time!

After this merge, you can (and should) remove any AWS credetnails from any local app config files and/or ENV variables, and just use the standard ~/.aws/credentials in dev.

README has been updated with instructions for getting started. Also removing some other out of date stuff I noticed.

I have tested on my machine and seems to work, but @eddierubeiz please test this branch locally on your dev, and it's instructions, before merging?

Ref #1551

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ This one is based on the [kithe](https://github.com/sciencehistory/kithe) toolki
 
 To set up a development instance on your workstation.
 
-Prequisites:
+### Prequisites
 
 * ruby installed (I like using chruby and ruby-build to install/manage rubies, some like rvm)
   * `gem install bundler` on new ruby installation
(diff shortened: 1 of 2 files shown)


eddierubeiz:
This works fine for me.

eddierubeiz (approved):
Looks good, and works fine for me. Feel free to merge.

jrochkind:
@eddierubeiz Can you, when you have a chance, test it on your local dev box, with AWS credentials removed from your local_env, and confirm it works? Or just clarify you have if you already have. Thanks!

eddierubeiz:
Yes -- I did just that and it works fine.

jrochkind:
Sweet, thanks!

PR #1569, opened 2022-02-08
tag s3 uploads

jrochkind:
We want to tag all S3 objects with a "Content-Type-Base" tag that has a value that is first part of mime content-type: "image", "audio", "video", "application".  This is to let us get categorized reports on our expenses. Ref #1654

This should be do-able with the shrine [upload_options](https://shrinerb.com/docs/plugins/upload_options) plugin, but was not, for two reasons, which we worked around as described:

* We are using several different shrine "storages" -- a separate for restricted derivatives vs public, originals vs derivatives, etc. We really want to tag all of them, and the upload_options plugin doesn't make it easy to do that, without naming them all specifically -- which could change as we re-architect things. 
   * We could have just tagged originals I guess, but I wanted to do all of them. 
   * So I wrote a custom shrine plugin *based on* the pretty simple shrine `upload_options` plugin, included here. 
   * Some more discussion here: https://discourse.shrinerb.com/t/upload-options-plugin-for-dynamic-options-but-covering-more-than-one-storage/560

* Shrine uses an AWS S3 SDK method to *copy* files from one S3 bucket to another in some cases, and that operation doesn't let you assign tags to destination unless you pass a `tagging_directive: "REPLACE"` option. So our tags were being ignored. 
   *   I PR'd a fix to shrine, and it was accepted at https://github.com/shrinerb/shrine/pull/569/, but we need it now, without waiting for a shrine release
   * So the fix is "monkey-patched" in too. 
   
   
Note also:

* We don't actually test S3 shrine storage at all in our tests/CI, our tests use a different shrine storage so we didn't have to figure out how to set up S3 for testing. The downside of this, though, is that none of this behavior is really tested. :(  (I did manually QA it).  not ideal, but....

* The content type is taken from the associated Asset model's metadata. If it's not there... the logic assigns the value "unknown".  Our DZI tiles for instance end up "unknown".  If in reporting we see a lot of unknown it's probably DZI tiles... but you can probably cross-tab with bucket name in reporting anyhow.

### DIFF
--- a/app/uploaders/asset_uploader.rb
+++ b/app/uploaders/asset_uploader.rb
@@ -7,6 +7,33 @@ class AssetUploader < Kithe::AssetUploader
   plugin :kithe_accept_remote_url
 
 
+  # use shrine upload_options plugin to set AWS tagging, only for "store" (not cache),
+  # will only have effect if we are using an AWS store of course.
+  plugin :upload_options, store: -> (io, options) do
+    output_upload_options = {}
+
+    # if you wanted to avoid for derivatives, you might `unless options[:derivative]`
+    content_type = options.dig(:metadata, "mime_type")
+    if content_type.present?
+      # AWS SDK docs: "The tag-set must be encoded as URL Query parameters. (For example, "Key1=Value1")"
+      # Rails #to_query can do it for us.
+      tags = {
+        "Content-Type-Full" => content_type,
+        #{}"Content-Type-Base" => content_type.split("/").first
+      }
+
+      # Need the weird tagging_directive REPLACE for confusing reasons.
+      # https://discourse.shrinerb.com/t/gotcha-on-s3-upload-options-and-tags/559
+      output_upload_options.merge!(
+        tagging: tags.to_query,
+        tagging_directive: "REPLACE"
+      )
+    end
+
+    output_upload_options
+  end
+
+
   # Re-set shrine derivatives setting, to put DERIVATIVES on restricted storage
   # if so configured. Only effects initial upload, if setting changes, some code
   # needs to manually move files.
--- a/app/uploaders/asset_uploader.rb
+++ b/app/uploaders/asset_uploader.rb
@@ -11,27 +11,22 @@ class AssetUploader < Kithe::AssetUploader
   # will only have effect if we are using an AWS store of course.
   plugin :scihist_upload_options, proccessor: -> (io, options, storage_key) do
     output_upload_options = {}
-byebug
     # arguments that might be useful for dynamic processing including, `options[:derivative]`, true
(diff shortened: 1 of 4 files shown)


jrochkind:
I'm also going to provide a rake task to assign this Tag to all existing originals and derivatives.  (not bother with DZI tags, there are too many!)

My count of all originals plus all derivatives is: `508361`

Will this cost a lot? I think it means an S3 PUT operation for each object. S3 charges $0.005 per 1000 operations for PUT requests. 

So to tag em all is `508361 / 1000 * $0.005` is $2.50. 

Since we might need to do it a couple few times to get it right, it might end up costing us, oh $10 or $15. That seems fine, I don't know if I should even bother @CHF-IT-Staff to ask to confirm it's okay to spend that level of money, but I just tagged them!

jrochkind:
OK after adding some concurrency it can finish in two hours, I think that's as good as it gets, and we'll see!

Hopefully we don't have to run it over and over again, as it'll incure a cost.

jrochkind:
OK, the rake task is still taking like 5 hours to run on staging. :(  

We could run it as a "Detached" process on heroku... would want to turn off progress bar.  

But how do we know if it crashed in the middle (say was restarted by heroku?)?  And if it did, we have no way to restart "in the middle".  Keep in mind it can cost a couple bucks every time it's run. 

OK, I'm going to try to make it log every ~10 minutes with what record it's on, and give a way to start again from where it stopped, to make it more suitable for this kind of long-running job it will take to do. 

(Another option is trying to put it into bg job queue on dynos that we spin up just for this... but that takes more work to figure some things out, like what happens if we have 10K items in queue, does that blow out our redis?)

jrochkind:
OK, there's a task in there now, here's the intended hacky script for this somewhat hacky code. 

You run:

    heroku run:detached rake scihist:data_fixes:set_s3_object_content_type_tags

You can then follow the logs with the `heroku logs` line it gives you if you want -- add `--tail` option to tail them. 

It will print to logs (also in papertrail) something like:

    INFO: set_s3_object_content_type_tags: starting

Then every 10 minutes it will log a status report, something like:

    INFO: set_s3_object_content_type_tags: processed 393 assets, up to Asset PK 0aa2714e-362f-40f7-9f37-e8d271069903

When/if it finishes completely (could take 3-5+ hours), it will log:


If it doesn't finish cause it crashed for some reason (you can look in papertrail the next day too), you can find the LAST progress line logged, and tell it to resume from there:

        heroku run:detached rake scihist:data_fixes:set_s3_object_content_type_tags START_AT_PK=0a6adeab-3a75-43b3-aa8d-8df4c7189294

Wash rinse repeat.

jrochkind:
Oy, now I realize tagging might not work for cost categorization anyway, cause of cloudfront being on top. We have to talk more about it.

Issue #1578, opened 2022-02-15
FeaturedTopicController#index doesn't properly handle requests for json.


eddierubeiz:
https://digital.sciencehistory.org/focus/alchemy.json leads to [a 500 error](https://app.honeybadger.io/projects/58989/faults/84034428). Not a big deal but needs to be fixed.

PR #1580, opened 2022-02-15
handle json in featured_topic_controller

eddierubeiz:
Ref #1578

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -557,9 +557,13 @@ class CatalogController < ApplicationController
     end
   end
 
+  # Trigger this by requesting e.g.
+  #   /catalog.rss
+  #   /focus/alchemy.json
+  # See https://github.com/sciencehistory/scihist_digicoll/issues/1578
   def catch_bad_format_param
-    if params[:format] == 'json'
-      render plain: "Invalid parameter: we do not provide a JSON version of our search results.", status: 406
+    if ['json', 'atom', 'rss'].include? params[:format]
+      render plain: "Invalid parameter: we do not provide search results in #{params[:format]} format.", status: 406
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -168,13 +168,6 @@ Rails.application.routes.draw do
   # out stuff we're not using.
   #
 
-      # Route RSS and atom search results to a 404
-      # until we decide to implement them.
-      # Practically speaking, this only affects bots).
-      get 'catalog.rss',  to: proc { [404, {}, ['']] }
-      get 'catalog.atom', to: proc { [404, {}, ['']] }
-      get 'catalog.json', to: proc { [404, {}, ['']] }
-
       # TODO: Can we get away without actuallymounting Blacklight, just using the CatalogController?
       # mount Blacklight::Engine => '/'
       concern :searchable, Blacklight::Routes::Searchable.new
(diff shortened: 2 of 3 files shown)


jrochkind:
Nice solution. 

I was curious to see what other controllers that featured topic controller may have been based on did... but I see you already applied this to CatalogController. 

I thought, gee, does this problem already exist in CatalogController?  But if I go to `https://digital.sciencehistory.org/catalog.json`, I get a nice 404 already.  (not 406 because I guess technically just adding `.json` to the end of the URL is just a completely separate URL without any inherent content-type request, that's just a convention). 

So why is featured topics different? I wonder if this is a workaround, but there's another fix we could do to a more fundamental thing. How do the featured topics differ from catalog controller in the first place? Could there be a routes difference?

```
 $ rails routes | grep catalog
                                                    GET      /catalog.rss(.:format)                                                   #<Proc:0x00007fedf8e8e6b0 /Users/jrochkind/code/scihist_digicoll/config/routes.rb:174>
                                                    GET      /catalog.atom(.:format)                                                  #<Proc:0x00007fedf8e8bbe0 /Users/jrochkind/code/scihist_digicoll/config/routes.rb:175>
                                                    GET      /catalog.json(.:format)                                                  #<Proc:0x00007fedf8e89bb0 /Users/jrochkind/code/scihist_digicoll/config/routes.rb:176>
                                     search_catalog GET|POST /catalog(.:format)                                                       catalog#index
                            advanced_search_catalog GET      /catalog/advanced(.:format)                                              catalog#advanced_search
                                      track_catalog POST     /catalog/:id/track(.:format)                                             catalog#track
                                        raw_catalog GET      /catalog/:id/raw(.:format)                                               catalog#raw {:format=>"json"}
                                 opensearch_catalog GET      /catalog/opensearch(.:format)                                            catalog#opensearch
                              suggest_index_catalog GET      /catalog/suggest(.:format)                                               catalog#suggest {:format=>"json"}
                                      facet_catalog GET      /catalog/facet/:id(.:format)                                             catalog#facet
                                range_limit_catalog GET      /catalog/range_limit(.:format)                                           catalog#range_limit
                                                    GET      /catalog/range_limit_panel/:id(.:format)                                 catalog#range_limit_panel

jrochkind-chf scihist_digicoll (combined_audio_ffmpeg_urls ?) $ rails routes | grep featured
                                    featured_topics GET      /focus(.:format)                                                         featured_topics_list#index
                                     featured_topic GET      /focus/:slug(.:format)                                                   featured_topic#index
                                                    GET      /focus/:slug/range_limit(.:format)                                       featured_topic#range_limit
                                                    GET      /focus/:slug/range_limit_panel(.:format)                                 featured_topic#range_limit_panel
                                                    GET      /focus/:slug/facet(.:format)                                             featured_topic#facet
```

That's a bit weird... what's at `config/routes.rb:174`? 

https://github.com/sciencehistory/scihist_digicoll/blob/master/config/routes.rb#L171-L176

OK, we're handling that at the routes level for catalog... should we do it the same way for featured topics and collections? Or should we remove the routes now that we're handling it at the controller level instead? (and handle .atom and .rss there too, and add a comment there that this is working around underlying Blacklight's wilingness to handle these formats, but where our actual app doesn't make them work)?

Do we have this problem in collection too? Yes we do!

I think your solution is probably better actually because it handles all those sub-classes -- but let's:

* make it handle .rss and .atom too
* remove those lines from routes that are no longer needed
* Add a comment explaining
* Add a test (maybe one test on catalog is sufficient, don't need one on each subclass. Unless we already have a test on catalog for this, which might be sufficient? One way to find out is to remove all the code that handles this, and see if any tests fail!)

jrochkind:
Doing the weekly check, I see you made those decisions/changes. Thanks, merging!

Issue #1585, opened 2022-02-23
Orphan checker for video originals


eddierubeiz:
This is part of ref #1575, but there appears to be enough work here to justify a separate ticket:
- [ ] New rake tasks for detecting orphans and deleting them
- [ ] Tests for new rake tasks
- [ ] Reporting at `/admin/orphan_report`

jrochkind:
you could make it three separate tickets if you want to.  Or just keep tracking it all on the main ticket. 

Either way, it should def be three separate PR's, please!

jrochkind:
oh wait, I see, this is all about the orphan checker, just breaking it down into steps. OK!

jrochkind:
There's already an OrphanAuditor which, one of them, checks originals, derivatives, AND restricted derivatives. 

I think it just needs to check video originals too?

Oh I see, we also have separate rake tasks for each type, even though they're all combined in the auditor. Do we really need that?  I don't know!

In the end, we really just need to run the existing logic we have for originals but also on the video originals bucket. Perhaps some refactoring could make that easier to do, DRY, without copy and pasting all the code?

It seems too bad that "just run the existing logic on another bucket" requires changing large expanses of code.

eddierubeiz:
It is indeed too bad. Unfortunately (as we already know) the existing code is already quite messy, so we're faced with an unpleasant decision here. The easy way out does lead to a great deal of duplicated code.

eddierubeiz:
Cleaned the code up somewhat and added tests.

PR #1589, opened 2022-02-24
refactoring orphan_s3_originals and adding video bucket

eddierubeiz:
Ref #1585
 - Code for actually detecting and deleting orphans is unchanged, but is called on both buckets.
 - Moved some code into helper methods to work better with the new two-bucket setup
 - Added a test for `OrphanS3Originals`
   - Added `files_checked` to the list of `attr_reader` s so we can check it from the test.
 - Added a test for `S3PathIterator`
 - DRY-ed out our mock s3 client code into a new `AwsHelpers::MockS3Client` class under `spec/test_support/helper_ruby/` (only for `orphan_s3_originals_spec.rb` and `s3_path_iterator_spec.rb`.
   - Did not touch `orphan_s3_derivatives_spec.rb` or `orphan_s3_restricted_derivatives_spec.rb`: that belongs in a separate PR.
 - Added code to fetch video and regular asset counts from the DB in one query
 - Added `ruby-progressbar` to the gemfile; no idea how long this has been pending but we do need it.


spec/services/orphan_s3_originals_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+describe OrphanS3Originals do
+  let(:asset_n)  {
+    sleep 0.05
+    create(:asset, :inline_promoted_file,
+      file: File.open((Rails.root + "spec/test_support/images/20x20.png"))
+    )
+  }
+  let(:asset_v)  {
+    sleep 0.05
```
eddierubeiz, line 11: I had a lot of trouble with a race condition here, hence these ugly `sleep`s.

eddierubeiz: Fixed

spec/test_support/helper_ruby/aws_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+  # we need to be able to feed the
+  # checker an array of fake s3 paths.
+  #
+  # This helper allows you to do something like:
+  #
+  #     paths = ['fake/path/1.jpg', 'fake/path/2.jpg']
+  #     allow(orphan_checker.s3_iterator).to receive(:s3_client).
+  #       and_return(AwsHelpers::MockAwsClient.new(paths: paths).client)
+  #
+  class MockS3Client
```
jrochkind, line 12: This isn't *that* complicated code -- the heavy living it done by AWS's built in `stub_responses` feature. And we're only using it in one file at present... I would have just done this inline where needed instead of extracting into a class.   

I think the downside is that when reading tests you have to go to another file to see what's really going on, *and* that this class gives the appearance it can do something other than provide a stubbed `list_objects_v2`. 

But if you think it's worth it, that's cool we can leave it extracted -- but I wonder about being clear in comments that ALL this mocks at present is list_objects_v2 call?  Maybe even changing the name of the class to be clear that's kind of all it does?

jrochkind:
I gotta admit I didn't exhaustively look at all the code, it overwhelmed me a bit. But overall looks good, and looks like a pretty good solution to keeping this kind of crazy code fairly DRY. 

Only had one comment about the test helper extraction -- consider it then do whatever you think is best (including leaving it how it is) and merge!

spec/test_support/helper_ruby/aws_helpers.rb
eddierubeiz: It is used in two other files; will refactor in a separate PR.

Gemfile
```diff
@@ -195,3 +195,5 @@ end
 # we use for data structures for citation models, and for generating citations
  gem "citeproc-ruby", '~> 2.0'
  gem 'csl-styles', '~> 2.0' # Need to load the styles so we can use chicago
+
+ gem 'ruby-progressbar'
```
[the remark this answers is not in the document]
eddierubeiz: Yup.

spec/test_support/helper_ruby/aws_helpers.rb
eddierubeiz: I'm confident this code belongs in its own helper class. When it was inline (aside from being very un-DRY and easy to get wrong) it just served to distract from the main focus of the tests we were using it in. Now that it's out everything just looks more organized and sane.

eddierubeiz: Good call on the comment; will be more explicit.

eddierubeiz:
All right, I'm happy with this. Merging now so I can start working on using `MockS3Client` in another PR.

PR #1611, opened 2022-03-17
URL used for video SRC should expire later

jrochkind:
We are using S3 signed URL for video originals bucket (which is not publicly visible). S3 signed urls by default expire in 15 minutes. That's not enough, if you left your window open for more than 15 minutes, you couldn't play the video embedded in it anymore!

5 days seems fine? The maximum is 7 days. I dunno, this should be more than enough, but I don't think it needs to be any shorter.

### DIFF
--- a/app/components/work_video_component.html.erb
+++ b/app/components/work_video_component.html.erb
@@ -35,7 +35,7 @@
               }.to_json
             }
     ) do %>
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Sure - 5 days works.

PR #1612, opened 2022-03-17
Specs for Video show page

jrochkind:
Some specs for Video show page, along with some cleanup/finishing of the feature that writing specs suggested -- and some refactoring of existing spec in places that got me into trouble trying to write these!


spec/factories/asset_factory.rb
```diff
@@ -88,9 +88,32 @@
       trait :video do
         title { 'Test video' }
         faked_file { File.open((Rails.root + "spec/test_support/video/sample_video.mp4")) }
-        faked_content_type { "video/mp4" }
+        faked_content_type { "video/mpeg" }
```
eddierubeiz, line 91: Thanks for fixing that!

eddierubeiz:
Thanks so much -- in particular for cleaning up that factory.

Issue #1619, opened 2022-03-24
Adjustment to scihist:copy_staging_work:restore_work


eddierubeiz:
Our rake task to copy a work from staging currently expects all the buckets in dev to also be defined in staging;
We use this feature a lot in development;
We do occasionally introduce changes to the shrine definitions that can sometimes come and go in staging depending on which branch happens to be deployed there at the moment;
During these transition times, it's really useful to be able to download a work from staging as long as both environments know about all the buckets that the work's assets and derivatives are stored in.

Let's be more tolerant of mismatches, throwing a warning during the bucket matching process, but not aborting the whole task.

```

PR #1620, opened 2022-03-24
bundle exec rake scihist:heroku:copy_data more tolerant of missing buckets in staging.

eddierubeiz:
Ref #1619

If all  of these are true:

- You want to copy a work from staging to dev.
- Your shrine storages in dev do not match up with the ones defined in staging (because e.g. the code deployed in staging is older than what's in `master`).
- However, all the work's assets and derivs are in Shrine storages that do exist in both staging and dev.

Then:
- Show a warning
- Don't abort the copy from staging to dev operation if there are buckets in dev that staging does not know about.
- Copy the work down anyway.

In theory this should never happen. In practice it's happened enough times already.

### DIFF
--- a/app/models/copy_staging/restore_work.rb
+++ b/app/models/copy_staging/restore_work.rb
@@ -168,6 +168,10 @@ module CopyStaging
     # ]
     def register_remote_storages
       @storage_map.each_pair do |local_key, remote_key|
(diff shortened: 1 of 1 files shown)


jrochkind:
I don't *totally* understand, but seems safe enough!

Issue #1614, opened 2022-03-21
Remove "Original File Type"


apinkney0696:
Remove the "original file type" field from the metadata interface. This information is viewable in the download drop-down menu for public-facing users, and on the back-end.

jrochkind:
For reference, the original motivation to add it seems to be at: https://github.com/sciencehistory/chf-sufia/issues/994

Note that the current display summarizes all members in one line, while the download drop-down is available individually for each asset -- and depends on the particular show page being used for the particular type of work. 

So that actually for instance for video files right now... we aren't including any download dropdowns at all, or individual file download access at all, so the "original" (ingested) file type technical metadata actually won't be available at all in the public UI. 

This is still ok @apinkney0696?

PR #1622, opened 2022-03-25
Remove "original file type" from the public-facing work metadata table

eddierubeiz:
Ref #1614


jrochkind:
Not sure why we don't have CI on this, let's figure out how to trigger CI.

jrochkind:
I merged master and pushed again, just as a way to try to trigger CI again.

Issue #1583, opened 2022-02-16
Disable advanced search


eddierubeiz:
We appear to have Blacklight's out-of-the-box "advanced search" controller turned on; it should probably be turned off until we investigate it further.

eddierubeiz:
Also turning off `#suggest`, by the way.

PR #1624, opened 2022-03-25
Turning off #suggest and #advanced_search in CatalogController

eddierubeiz:
Ref #1583
- Every other route in in `bin/rails routes | grep catalog` now has a reason to be there.
- `autocomplete_enabled` is set to `false`, and the corresponding SOLR suggester is also turned off, so rather than maintain a confusing, vestigial path that always returns an empty array, I did decide to switch off the `#suggest` method.
- Tests confirm that both methods are now turned off. 
- Comments that should save us time and effort later.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -21,6 +21,14 @@ class CatalogController < ApplicationController
   # Not totally sure why we need this, instead of Rails loading all helpers automatically
   helper LocalBlacklightHelpers
 
+  def advanced_search
+    raise ActionController::RoutingError.new('Not Found')
+  end
+
+  def suggest
+    raise ActionController::RoutingError.new('Not Found')
+  end
+
   # a Blacklight override
   def render_bookmarks_control?
     false
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -21,10 +21,20 @@ class CatalogController < ApplicationController
   # Not totally sure why we need this, instead of Rails loading all helpers automatically
   helper LocalBlacklightHelpers
 
+  # This is turned off until/unless we
+  # decide to offer advanced search to the public (unlikely).
   def advanced_search
     raise ActionController::RoutingError.new('Not Found')
   end
 
+  # This functionality is disabled:
+  # config.autocomplete_enabled above is set to false, and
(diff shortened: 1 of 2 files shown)


jrochkind:
Cool. How about some comments over the methods explaining why we're doing it? Then feel free to merge yourself. 

If we ever try to turn on Blacklight's "suggest" feature (or advanced search for that matter!) we might get confused about why it doesn't work. I don't know if the comments would help someone in that situation be less confused, but at least they'd help someone who found the code and wondered "why is this happening"?

(Also there *could* be tests here, but I don't really think it's needed either).

eddierubeiz:
Tests are almost as easy to add as comments in this case; will do both. Good call.

eddierubeiz:
In particular I think this PR might break the "did you mean" spelling suggestion, which doesn't have tests. So I'll add a test to clarify what is going on.

jrochkind:
We don't use "did you mean" spelling suggestion, do we?

jrochkind:
We don't use "did you mean" spelling suggestion, do we? 

I was myself just thinking of tests that requesting those routes results in a 404.

eddierubeiz:
```
    # Configuration for autocomplete suggestor
    config.autocomplete_enabled = false
    config.autocomplete_path = 'suggest'
```
This would *suggest* that we do have it turned on.

eddierubeiz:
Although it's set to false. All very confusing. Will look into it and make sure, document, and confirm via tests as needed, that I'm not breaking anything in this PR.

jrochkind:
Yes, I would say ` config.autocomplete_enabled = false` suggests to me it's not turned on. 

Plus, I don't think I have ever seen it in our app, or remember turning it on, and I generally don't turn it on because it doesn't work very well. 

Of course, if we ever did turn it on, it would be broken if it were overridden like this. I'm not totally sure if it's necessary to disable like this or not.

eddierubeiz:
Did some research / digging and I'm now comfortable with explicitly overriding #suggest to return an error. Added comments (and tests) which explain why I think this is a good idea and are likely to be helpful if we *do* decide to turn autosuggest back on.

jrochkind:
Nice, thanks!

I might have done the test as a "routing" test instead of a controller test, but i'm not going to sweat it.

Issue #1546, opened 2022-01-19
In dev, deleting orphaned assets is not working


jrochkind:
This is after #1545, but I don't think I introduced the problems there.... but should be debugged after that gets merged, or based on that branch. 

This problem could be unique to dev, and the different way dev sets up the shrine storages (one bucket with different prefixes for the different shrine storage). 

* In my dev copy, running the orphan file report says "There are 6941 orphaned public derivatives."
   * I run `rails runner 'OrphanS3Derivatives.new(show_progress_bar: false).delete_orphans'`
   * It says it deleted `14742` -- mis-count!
   * AND I run an audit again, and the 6941 are still there!
   * What did it actually delete? Sorry if I deleted all your dev derivatives @eddierubeiz 

* in my dev copy, the orphan file report says "There are 1 orphaned restricted derivatives."
  * it's a file i put there myself on purpose wrongly, `edelsteincollection.jpg`
  * If I run `rails runner 'OrphanS3RestrictedDerivatives.new(show_progress_bar: false).delete_orphans'`
  * It says `deleted derivative file at: scih-data-dev: jrochkind.jrochkind-chf/originals/restricted_derivatives/edelsteincollection.jpg`, and `Deleted 1 orphaned derivatives`
  * But it's actually still there. And every time I run the delete_orphans command again, it keeps saying it deleted `edelsteincollection.jpg`, but it obviously didn't, since it keeps saying that every time!

eddierubeiz:
Reproduced this.

PR #1631, opened 2022-03-29
(orphan S3 file checker) Use shrine_path to look up object to delete

eddierubeiz:
Ref #1546

- Use `shrine_path` to look up the shrine object to delete; this is not identical to `s3_key`;
- Better documentation;
- Sample report doesn't use `s3_url_for_path` to generate s3 urls for files that aren't really stored on s3;
- A test that actually creates fake files on the filesystem and ensures the `delete_orphans` method actually removes them;

Ran the detect and delete routines in staging to confirm orphans are still detected and deleted as before.


jrochkind:
Looks good not sure why we're currently failing CI?

Issue #1639, opened 2022-04-04
bug in jump to segment in OH audio


jrochkind:
In OH audio list of separate files, clicking on a file should take you to the relevant point in time on the combined stiched-together play-in-browser audio. 

For this work: https://digital.sciencehistory.org/works/hm50ts561#tab=ohDownloads

The last audio segment is listed with a timestamp off the end of the actually existing combined audio-- listed as starting at `02:41:08`, but the whole combined audio is only `02:12:37`. And then unsurprisingly, clicking on that last segment does not take you to the right place. 

What's going on? Is the combined audio deriv being properly created?  Is the calculation of segment start point wrong?

jrochkind:
I think maybe the combined derivative created is corrupt/unfinished. 

The second to last file (#6) is listed as starting at `2:11:56`. When I click on it, it does play audio from the combined derivative that matches the beginning of that individual file. 

Teh second to last file (#6) is actually `29:11.64` long. So the start point of the LAST file (#7) should be around `21:11:56` + `29:11` == `2:41:07` -- that's what it is! 

the problem is that combined derivative ends long before then. In fact, it ends only a couple minutes into the second to last segment (#6), apparently cutting off the majority of file #6, and not including any of file #7. 

In fact, it looks like the total combined audio deriv _ought_ to be.... around `3:08:59`.

Let's try creating a new combined derivative and see what happens....

jrochkind:
After re-generating the combined derivative, it is still only `2:12:23`, which is not as long as it seems it should be. 

Will need to investigate more.

jrochkind:
On *staging* the combined deriv is `3:13:15`, which about what is expected. 

So this can't necessarily be easily reproduced by copying the file locally. Very odd.

jrochkind:
When I copy and paste the `ffmpeg` command (taking sources from S3 URLs) and run it locally on dev machine... I get a combined deriv that's `03:09:00` -- not sure why we get a few seconds different each time, but this is still in the ballpark. 

So why is it creating one that's too small on production?  Trying to run it in a console to see what happens.

jrochkind:
Ran it in a production console, it completed... with a combined derivative that ffprobe still says is duration `02:41:21.50`. 

Ran the exact same command locally... got one with `03:09:00.41`. What? 

Calling `ffmpeg -version` locally, I seem to have `5.0`.  On production... `4.2.3`.  On staging...  `4.2.3`. 

[...]

PR #1640, opened 2022-04-04
Generate longer expirationf for s3 urls used as ffmpeg input'

jrochkind:
Possibly relevant to #1639. A default signed S3 URL is good for 15 minutes. We feed these URLs to ffmpeg to create a combined derivative. What if it takes longer than 15 minutes to create, and S3 starts refusing? You'd think we'd get an error message. But maybe not.

Just in case this is part of the problem, it doesn't hurt to make these S3 URLs we use as ffmpeg input last a lot longer. How about 48 hours.

### DIFF
--- a/app/services/combined_audio_derivative_creator.rb
+++ b/app/services/combined_audio_derivative_creator.rb
@@ -85,7 +85,7 @@ class CombinedAudioDerivativeCreator
 
       audio_member_files.collect do |original_file|
         if original_file.url.present? && original_file.url.start_with?(/https?:/)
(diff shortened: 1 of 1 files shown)


jrochkind:
want to test on staging first

jrochkind:
None of the recent runs I was trying lasted more than 10 minutes. So I don't think this is it, unfortunately. Although it shouldn't hurt.

jrochkind:
Ha, actually we may end up reverting all of this.

PR #1641, opened 2022-04-05
Revert ffmpeg using URLs as input

jrochkind:
It was resulting in problems -- ffmpeg would sometimes find the remote host closed connection, but ignore it and produce an incomplete output mp3. Ref #1639

Revert of the main portion of #1579, although we retain the use of new duration metadata.


jrochkind:
confirmed working in staging. 

In a separate PR I'll make a rake task that looks for any existing bad ones. :(

Issue #1336, opened 2021-09-28
Better indexing for works in German


eddierubeiz:
As part of the Bredig project, we will be ingesting, transcribing and translating a significant corpus of works written exclusively in German. Currently these are indexed in a catch-all field that does not provide stemming for either English or German. Once enough German content accumulates, we may want to tweak our search settings to make it easier to find German words.

This is a broad ticket that covers:
- Gathering some test searches from researchers familiar with German, with current and desired results: `I searched for X and I only got Y, when I had hoped to get Z.`;
- Researching German-language indexing we can get out of SOLR;
- Creating a new German full-text field and indexing all German-only works into that field;
- Incorporating search results from that field into our existing full-text index;
- Running test searches to establish that the new setup actually does a better job matching German words.

eddierubeiz:
The initial results in dev and staging were so promising that I skipped the "gather test searches" step. The fact that the number of useful matches for the term `buch` doubles speaks for itself.

jrochkind:
Also search for `potato` you get a lot of German matches for `tat`, which is unfortunate. 

We should look into fine tuning or other options.

PR #1644, opened 2022-04-06
Adding searchable_fulltext_de field

eddierubeiz:
- Ref #1336.
- Note that `searchable_fulltext` is being renamed to `searchable_fulltext_en`.
- Modifies schema.xml.
- Also fixes a bug in which our indexing code incorrectly assumed that we stored language metadata as ISO codes. Prior to this PR, the `searchable_fulltext_en` field has been empty of any content (as we do not have any `Work`s whose `language` is `en` )
  - All fulltext fields of any language have, until now, been incorrectly indexed in `searchable_fulltext_language_agnostic`.
  - Rather surprisingly, this means this PR also adds English stemming to our fulltext searches, since they will now be added to a `text_en` index.
- The boost for the new German-only field is set to the same values (`^0.5` for `qf` and `^12` for `pf`) as the previous `searchable_fulltext_` fields.
  - In theory this adds relative weight to the fulltext fields.
  - I would argue that the relative starting weights are so small that the change makes no practical difference.
- Requires reindex.
- Adds a `/lang` directory with two new files in it.
- Does not add in any noticeable way to the time it takes to reindex the collection (a little under 2 minutes, give or take).


eddierubeiz:
I added a new ticket #1647 for us to consider storing languages as ISO codes.

eddierubeiz:
After doing some research I settled on the configuration described in
https://github.com/uschindler/german-decompounder .
Credit goes to @uschindler of SD DataSolutions GmbH, and in turn to Björn Jacke (see https://www.j3e.de/ispell/igerman98/index_en.html) .

jrochkind:
You went way beyond what I would have done, I would just have used built-into Solr German algorithmic stemming algorithms, rather than the third-party `uschindler` dictionary-based one you used. 

If I'm understanding what's going on. 

But if you're knowledge of German convinces you that the third-party dictionaries are clearly better than the stock Solr approach, ok, nice job, thanks!

jrochkind:
I see that we still use the "language agnostic" analyzer for things of _mixed_ language content, as well as for things that aren't English or German. If it was for instance marked with BOTH languages English and German... consistent with what we were doing before, we just index in language_agnostic, not in English or German, since we don't know what passages are what, I guess on the theory that indexing in the wrong language analyzer may produce really confusing results. OK, I guess that's sensible, at least for now!

eddierubeiz:
The big difference is by using the decompounder and dictionaries we're now matching very common compounds like "Handbuch" ("handbook") on a search phrase for "Buch" (book), whereas with the stock German stemmer we wouldn't be.

solr/config/schema.xml
```diff
@@ -180,6 +182,23 @@
       </analyzer>
     </fieldType>
 
+    <!-- A text field with defaults appropriate for German -->
+    <!-- https://github.com/uschindler/german-decompounder -->
```
jrochkind, line 186: thanks for the link to source of config in source code, great!

uschindler:
> You went way beyond what I would have done, I would just have used built-into Solr German algorithmic stemming algorithms, rather than the third-party `uschindler` dictionary-based one you used.
> 
> If I'm understanding what's going on.
> 
> But if you're knowledge of German convinces you that the third-party dictionaries are clearly better than the stock Solr approach, ok, nice job, thanks!

Hi, as maintainer of the Solr/Elasticsearch dictionary and config file let me give some explanation: The default Solr language analysis cannot do German decompounding: Splitting words like "Donaudampschifffahrtskapitän" (the captain of a steam powered ship driving on river Donau) is not possible. Those compounds are common in German and depending on requirements for search it may be a good idea to split those words (in English those are naturally separated).

For license reasons Solr does not ship with required dictionaries and hyphenation rules, although the actual algorithms/code is shipped with Solr - that's why I put them for easy usage in a separate repository. The Apache License of Solr does not allow to ship GPL/LGPL licensed files (and hyphentation rules have Latex license and the dictionary is LGPL, both not allowed for Apache projects).

The decompounding works like that: The OpenOffice/Latex hyphenation rules are used to algorithmically split words into syllables (which is too much). In a second step those syllables are recombined by the Solr token filter until the combinations can be found in the dictionary.

eddierubeiz:
Thanks much for the detailed explanation @uschindler !

jrochkind:
Wow, when a celebrity shows up on your github. Thanks @uschindler :)

jrochkind:
@uschindler if you have any easy input to this, I would suggest adding a link to and brief explanation of your project to the Solr docs at https://solr.apache.org/guide/8_11/language-analysis.html#german

That's what I found when googling for Solr and german -- Eddie, a German speaker, did the extra work to find your repo, but I probably wouldn't have, would have (wrongly) stopped at the analyzers alone without the dictionary! Those Solr docs *appear* to be giving you an example of how to do German analysis with Solr, it would be great if they linked to your repo as a better practice!

eddierubeiz:
@uschindler: our collection mixes German and English content, and by default we search both German and English text. But most of our searches are for English words. The new decompounder is doing its job rather enthusiastically, so we are now getting matches for English words in text that is in German. For example:
[potato (just en)](https://digital.sciencehistory.org/catalog?f[language_facet][]=English&q=potato) ← 100% `potato`.
[potato (just de)](https://digital.sciencehistory.org/catalog?f[language_facet][]=German&q=potato) ← since `potato` is an English word, these results are not so helpful.
[potato (mixed en + de results)](https://digital.sciencehistory.org/catalog?&q=potato) ← our users see this by default.
We are considering some user interface adjustments, but would also be curious if you also had any tips for making indexing cleverer (for instance, adding an initial, automatic attempt to guess the language of a search query).
Any ideas or suggestions you might have are welcome; thank you! :)

Issue #1654, opened 2022-04-18
Consider replacing web and mp3 with aac files stored in m4a container.


eddierubeiz:
This will play out as multiple PRs, but this is a broad issue to keep related work together.
We have a workaround for ref #1552, but we currently believe that a new format for combined audio derivatives (CADs) will solve the problem completely: `.m4a` audio files containing a single mono `aac` bitstream.
- [x] Double check that all browsers we support will play this (use the list we maintain in the wiki)
- [x] Create a migration to add a database column for the new m4a derivative.
- [x] Write code to generate CADs in the new format instead of `mp3` and `webm`
- [x] Run the code for all oral histories that have CADs
- [x] Delete all `mp3` and `webm` derivatives
- [ ] Update workflows and documentation accordingly

PR #1663, opened 2022-04-21
m4a combined audio

eddierubeiz:
Ref #1654 .
Includes a rake task to generate the new derivatives; which appears to work fine in staging on a sampling of roughly 50 works.


eddierubeiz:
A note about the audio conversion:

We currently get ffmpeg via the [activestorage buildpack](https://github.com/heroku/heroku-buildpack-activestorage-preview).

Accordingly, this PR uses the built-in encoder `-c:a aac` to convert to aac. The `ffmpeg` docs claim it's the "second-best" encoder available; the [Fraunhofer](https://trac.ffmpeg.org/wiki/Encode/AAC) encoder would be preferable if it were readily available.

However, the Fraunhofer encoder would need to be compiled into ffmpeg. I am hesitant to embark on the subproject of getting `libfdk_aac` compiled into ffmpeg considering the content of the audio and how we are using it.

eddierubeiz:
`libfdk_aac` requires ffmpeg to be configured with `--enable-libfdk-aac` (and additionally `--enable-nonfree` if you're also using `--enable-gpl` (which we are)).

PR #1676, opened 2022-04-28
m4as for combined audio derivatives

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/1654 .

**SCHEMA**
- Migration and schema change for the new DB column.

**MODELS**
- `app/models/oral_history_content.rb`:
    - a new CombinedAudioUploader::Attachment for the new derivative;
    - new methods to set the derivative;
    - remove old methods needed to set the old derivatives.

**JOBS, SERVICES, UPLOADERS**
- `create_combined_audio_derivatives_job.rb` uploads and unlinks the new derivs;
- `combined_audio_derivative_creator.rb` creates new m4a files (this is the most complex change);
- `app/services/orphan_s3_derivatives.rb` will whine about the old derivatives until they are gone;
  - We will not use it to delete the old derivatives, however;
- `combined_audio_uploader.rb`: comments needed updated.

**COMPONENTS AND PRESENTERS**
- These are verbose, but straightforward: if an m4a is available, display or use that. Otherwise, use mp3 or webm.

**TESTS**
- Tests needed extensive changes but were really helpful.
  - New asset factory trait for m4a, used wherever a combined deriv is needed;
  - New silent m4a files in the test_support directory;
  - Redid the existing silent sound files to be a bit smaller (They are still valid silent mono sound files.);
  - Got rid of the old `ice_cube` sound files: confusing, annoying, and required a `README`;
  - Using silent files instead in all tests.

**RAKE TASKS: NOT YET**
- Does not include any rake tasks to create the new derivatives or delete the old ones.


app/components/oral_history/downloads_list_component.html.erb
```diff
@@ -17,14 +17,32 @@
...
+        Web-quality m4a
+      </a>
+    </div>
+    <div class="combined-audio-download-link-size">
+      <small><%= combined_display_m4a_audio_size %></small>
+    </div>
+  </div>
+<% end %>
+
+<% if combined_mp3_audio_url.present? #REMOVE_AFTER_MP3_TO_M4A_MIGRATION %>
```
jrochkind, line 45: Should this have `&& combined_derivatives_up_to_date?` like it used to, and like the above one for `combined_m4a_audio_url` now does?

app/components/work_oh_audio_show_component.html.erb
```diff
@@ -13,11 +13,22 @@
...
+            <audio controls controlslist="nodownload" data-role="ohms-audio-elem">
+              <source src="<%= combined_m4a_audio_url %>" type="audio/mp4">
+            </audio>
+            <% if has_ohms_index? || has_ohms_transcript? %>
+              <div class="pl-3">
+                <button class="btn btn-emphasis btn-sm text-nowrap" data-trigger="ohms-jump-to-text">Jump to text</button>
+              </div>
+            <% end %>
+          </div>
+        <% elsif combined_mp3_audio_url.present? #REMOVE_AFTER_MP3_TO_M4A_MIGRATION %>
```
jrochkind, line 27: Same as above analagous component (boy, we're kinda un-DRY here huh? So it goes!), does this need `&& combined_derivatives_up_to_date?`?

app/services/combined_audio_derivative_creator.rb
```diff
@@ -37,14 +36,13 @@ def generate
     end
     output_files = {}
     # Create the two output files:
-    ['mp3', 'webm'].each do |format|
+    ['m4a'].each do |format|
       output_files[format] = output_file(format)
       ffmpeg_args = args_for_ffmpeg(output_files[format].path)
       cmd.run(*ffmpeg_args, binmode: true, only_output_on_error: true)
     end
```
jrochkind, line 43: Is there a reason to keep this as array.each, when it's only one format?  Repalce with just `output_files['m4a']` etc?

app/components/oral_history/downloads_list_component.html.erb
eddierubeiz: That check is really looking up the current code in CombinedAudioDerivativeCreator.new(work).fingerprint, which should return `true` throughout the process: that method is really not looking at the derivative but at the underlying audio assets, which are not changing at any point. That said, there's no reason *not* to include the check, now that I think about it!

app/components/oral_history/downloads_list_component.html.erb
```diff
@@ -33,7 +51,7 @@
         data-analytics-label="<%= work.friendlier_id %>"
         >
         <i class="fa fa-download" aria-hidden="true"></i>
-        Web-quality MP3
+        Web-quality mp3
```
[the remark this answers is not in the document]
eddierubeiz: I think MP3 and M4A are great! Let's use that.

app/components/work_oh_audio_show_component.html.erb
eddierubeiz: Noted, and yeah I'll leave `&& combined_derivatives_up_to_date?` in.

app/components/oral_history/downloads_list_component.html.erb
jrochkind: Why do you say "it should return true throughout the process"? What process? The creation of the list of download links?

As I understand it, the inclusion of the mp3 URL was guarded by `if combined_derivatives_up_to_date?`, so that the mp3 URL wouldn't be included in the list of download lists offered to the user if it was not up to date. 

I believe without this check, if the mp3 URL (or m4a URL later) is not up to date, then it will still be included in the list of download links to offered to the user. 

So it's not a no-op, it's doing something. At least that was my intention, and why it is present in the code before this PR.  I acknolwedge it's not ideal that removing it didn't cause any tests to fail!

app/services/combined_audio_derivative_creator.rb
eddierubeiz: Getting read of that `each`. You're right!

app/services/combined_audio_derivative_creator.rb
```diff
@@ -37,14 +36,13 @@ def generate
...
     resp = Response.new
-    resp.webm_file   = output_files['webm']
-    resp.mp3_file    = output_files['mp3']
+    resp.m4a_file    = output_files['m4a']
```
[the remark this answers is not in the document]
eddierubeiz: No, on looking at the code again, good call. This doesn't affect any code outside this class. I'm removing the hash. If we go back to more than one deriv, we can change the code again.

app/components/oral_history/downloads_list_component.html.erb
eddierubeiz: Yeah - no problem including the check. Just noting that "up to date" doesn't refer to the mp3 vs. m4a distinction: as long as the published audio assets aren't changed, any and all combined audio derivatives are considered "up to date".
The check method would only hide the mp3 if someone were to change the audio assets in some way before we removed the old derivatives.

eddierubeiz: The fingerprint code is functioning as intended, did not intend to suggest that. And everything else you said is correct: 
"if the original files change, and old combined derivative(s) are present, they can be recognized as out of date, and not shown.". We're on the same page.

app/components/oral_history/admin/combined_audio_derivatives_component.html.erb
```diff
@@ -8,19 +8,19 @@
...
-              The corresponding webm link is provided for reference only:<br/>
+          <% if combined_mp3_audio.present? # REMOVE_AFTER_MP3_TO_M4A_MIGRATION %>
+            <li class="alert-danger">
+              The following MP3 link will be deleted soon. Click "Generate combined audio derivatives below" to create an M4A; use that instead.<br/>
               <small>
-                <%= link_to combined_webm_audio, combined_webm_audio %>
+                <%= link_to combined_mp3_audio,  combined_mp3_audio %>
               </small>
             </li>
           <% end %>
```
jrochkind, line 26: OK, sorry, I thought we'd show the mp3 only if the m4a did NOT exist, and the mp3 did -- cause that's a thing that can happen while we are in intermediate state, right?

PR #1678, opened 2022-04-28
Create HLS with ActiveEncode/MediaConvert

jrochkind:
In this PR:

* We can launch AWS MediaConvert jobs via ActiveEncode adapter, to use Asset original video and produce HLS in video derivatives bucket
* This is wired in to happen automatically on asset ingest (via a bg job)
* We have an ActiveEncodeStatus model/table, which can be used to track current status of the AWS MediaConvert job
* There is a way to poll to update status, and a rake task to do it for all statuses in `running` state. 
  * We will have a heroku scheduled job, that runs (every hour? Every ten minutes?) `rake scihist:active_encode_status:update`
* On the asset admin screen, for video, it can show you any ActiveEncodeStatus and their status, and let you press a button to kick off refresh status for a particular one, instead of waiting for the scheduled job. 

However, these HLS copies won't actually be used yet, that's for a subsequent PR. Also, the testing situation is a bit not great, this stuff is pretty hard to test, I put in some limited tests, that use extensive AWS mocking. 

## Things to be included in later PR necessary to actually have HLS be live

* The HLS copies once created need to be actually used -- probably the "refresh status" task will, if the thing is "complete", do something to let someone know there is an HLS playlist that can be used
* lifecycle management -- deleting HLS when an Asset is deleted, or when a new HLS is created!

## Things to be included in later PR's for full administrative control

* orphan checking on video_derivatives of HLS?
* prod->staging sync coveres video_derivatives bucket?
* dev copy_data copies HLS?  Or maybe skip that; I think it won't actually work since our dev buckets aren't public, and we don't have a way of serving HLS from non-public bucket. This is gonna get a bit messy.

### DIFF
--- a/app/views/admin/assets/show.html.erb
+++ b/app/views/admin/assets/show.html.erb
@@ -48,7 +48,34 @@
       </div>
     <% end %>
 
+    <% if @asset.content_type&.start_with?("video/") %>
+    <h3 class="mt-4 mb-4">HLS Encode Jobs</h3>
+    <% @asset.active_encode_statuses.order(updated_at: :desc).each do |status| %>
+      <dl class='row'>
+        <dt class="col-sm-4">started</dt><dd class="col-sm-8"><%= l status.created_at, format: :admin %></dd>
+        <dt class="col-sm-4">last updated</dt><dd class="col-sm-8"><%= l status.updated_at, format: :admin %></dd>
+        <dt class="col-sm-4">state</dt>
+          <dd class="col-sm-8">
+            <%= status.state %>
+            <% if status.running? %>
+              (<%= status.percent_complete  %>%)
+            <% end %>
+          </dd>
+        <dt class="col-sm-4">job id</dt><dd class="col-sm-8"><%= status.active_encode_id %></dd>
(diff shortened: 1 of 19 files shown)


app/views/admin/assets/show.html.erb
```diff
@@ -48,8 +48,45 @@
...
+    <% if @asset.content_type&.start_with?("video/") && @asset.active_encode_statuses.present? %>
+    <h3 class="mt-4 mb-4">HLS Encode Jobs</h3>
+    <% @asset.active_encode_statuses.order(updated_at: :desc).each do |status| %>
+      <dl class='row'>
+        <dt class="col-sm-4">started</dt><dd class="col-sm-8"><%= l status.created_at, format: :admin %></dd>
+        <dt class="col-sm-4">last updated</dt><dd class="col-sm-8"><%= l status.updated_at, format: :admin %></dd>
+        <dt class="col-sm-4">state</dt>
+          <dd class="col-sm-8">
+            <%= status.state %>
+            <% if status.running? && status.percent_complete.to_i > 0 %>
```
eddierubeiz, line 60: Thanks for this!

Issue #1662, opened 2022-04-21
Create a migration to remove the `webm` and `mp3` combined audio deriv columns from the database.


eddierubeiz:
Once #1654 is successfully deployed and after the columns are no longer used by any code, and the files are no longer in s3, let's also make sure to clean up and remove references to these old derivs.

PR #1687, opened 2022-05-10
Remove code that references mp3 combined audio derivatives

eddierubeiz:
- All oral histories with public audio members have already had their combined audio derivatives generated as m4a;
- The underlying mp3 and webm files, and references to them, are already removed from s3 and the production database.
So it's now time to remove code mentioning these derivatives.
- And a migration to remove the columns. @jrochkind argues that it's better to include the migration removing the DB columns in the same PR that removes the code: "[...] more worried about forgetting to remove columns we were no longer using. In general, I think I always remove the columns in the same PR I remove all the code that uses them!"

Ref #1654
Ref #1662)

### DIFF
--- a/app/components/oral_history/admin/combined_audio_derivatives_component.html.erb
+++ b/app/components/oral_history/admin/combined_audio_derivatives_component.html.erb
@@ -15,14 +15,6 @@
                 <%= link_to combined_m4a_audio,  combined_m4a_audio %>
               </small>
             </li>
-          <% elsif combined_mp3_audio.present? # REMOVE_AFTER_MP3_TO_M4A_MIGRATION %>
-            <li class="alert-danger">
-              Do not use this mp3 link as the OHMS Media URL, as it will be deleted soon.<br/>
-              Instead, click "Generate combined audio derivatives" below to create a new m4a link. Once that's created, use that mp4 link instead.<br/>
-              <small>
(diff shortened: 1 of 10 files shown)


jrochkind:
Looking good!

Just to double-check I checked out the branch and searched for `webm`. Noticed that the schema.rb file still has:

    create_table "oral_history_content"
        # ...
        t.jsonb "combined_audio_webm_data"
        t.jsonb "combined_audio_m4a_data"

Do we need some migrations to remove these now unused columns from the db?  And commit new schema.rb. 

(We also have some data_fix rake tasks that refer to the columns that wont' exist anymore! I guess there's no guarantee old data_fix rake tasks will keep working forever as the code changes, they were meant as one-time things. We should periodically go and just remove older-than-x-weeks data_fix rake tasks).

eddierubeiz:
Absolutely. I put migration / db work in a [separate issue](https://github.com/sciencehistory/scihist_digicoll/issues/1662). You could make a good case that it's not really a separate issue but I really wanted that as a separate PR.

eddierubeiz:
Re: old rake tasks referring to defunct DB fields, I would argue it's OK to keep those around even if they are in some ways obsolete. I know I've found them useful as a source of reusable patterns for new rake tasks with comparable objectives.

Issue #1679, opened 2022-05-02
Heroku R14 Errors for fixity check


jrochkind:
Our scheduled jobs for fixity checks are producing a lot of Heroku R14  (Memory quota exceeded) errors. They are currently running on a Standard-1X. 

R14 doesn't create a problem really -- although it means our process may be using swap and slowing down drastically. 

While not a disaster, it would still be good to clean this up.  We could just bump to Standard-2X -- it's only using a bit too much memory for Standard-1X. 

But it doesn't really seem like fixity check should use a lot of memory, it's worth reviewing code to see if it's doing anything odd -- using custom code to download all bytes of a file to fixity check that might result in a memory leak? 

<details>
<summary>One example of log from R14's on Fixity Check</summary>

```
Apr 21 02:30:29 heroku/scheduler.1499 Starting process with command `bundle exec rake scihist:check_fixity scihist:check_fixity:complete_overdue && curl -s https://api.honeybadger.io/v1/check_in/4dI2v2`
Apr 21 02:30:29 heroku/scheduler.1499 State changed from starting to up
Apr 21 03:05:18 heroku/scheduler.1499 Process running mem=528M(103.3%)
Apr 21 03:05:18 heroku/scheduler.1499 Error R14 (Memory quota exceeded)
Apr 21 03:05:37 heroku/scheduler.1499 Process running mem=544M(106.4%)
Apr 21 03:05:37 heroku/scheduler.1499 Error R14 (Memory quota exceeded)
Apr 21 03:06:00 heroku/scheduler.1499 Process running mem=543M(106.2%)
Apr 21 03:06:00 heroku/scheduler.1499 Error R14 (Memory quota exceeded)
Apr 21 03:06:17 heroku/scheduler.1499 Process running mem=543M(106.2%)
Apr 21 03:06:17 heroku/scheduler.1499 Error R14 (Memory quota exceeded)
Apr 21 03:06:38 heroku/scheduler.1499 Process running mem=544M(106.4%)
Apr 21 03:06:38 heroku/scheduler.1499 Error R14 (Memory quota exceeded)
Apr 21 03:06:57 heroku/scheduler.1499 Process running mem=543M(106.2%)
Apr 21 03:06:57 heroku/scheduler.1499 Error R14 (Memory quota exceeded)
Apr 21 03:07:19 heroku/scheduler.1499 Process running mem=543M(106.2%)
[...]

PR #1696, opened 2022-05-12
Reduce fetch batch size and turn of query cache for bulk fixity check

jrochkind:
Seeing if perhaps this helps us avoid out of memory errors running in heroku standard-1x on our nightly fixity check

Reduce fetch batch size, and try disabling ActiveRecord query cache to see if it helps us avoid memory bloat. 

Ref #1679


eddierubeiz:
Thanks. You never know!

PR #1702, opened 2022-05-17
Sort by full date

jrochkind:
Before we were just sorting by year. 

1. Logic to pull out min and max dates represneted by compelex DateOfWork objects
2. Solr fields to hold full dates, that will sort nulls last
3. Indexer code writes to new solr date fields instead of old integer year fields

This is in draft at present, because we need to make sure to reindex right after deploying (date sorts won't function properly in the interim). So please mark "approved" now, but don't merge. 

Since we have solr schema changes, in your local checkout you will need to run:

    ./bin/rake solr:clean
    RAILS_ENV=test ./bin/rake solr:clean

That _should_ work, but not totally sure we have that working to refresh your local solr for schema changes. Let's find out.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -431,12 +431,12 @@ class CatalogController < ApplicationController
 
     config.add_sort_field("newest_date") do |field|
       field.label = "newest date"
-      field.sort = "latest_year desc"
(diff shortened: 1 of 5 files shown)


eddierubeiz (approved):

jrochkind:
deployed to production, reindexed in production.

Issue #1674, opened 2022-04-26
Adding search for OH interviewer profiles and interviewee biographies


rachellane12:
It is challenging to find an interviewer profile or an interviewee biography due to having to click through multiple pages and scroll down the pages. A search would be helpful so that it's easier to find an interviewer or interviewee.

PR #1703, opened 2022-05-18
Interviewer and interviewee search

eddierubeiz:
Ref #1674 
I'm including two simple request tests, mostly useful for TDD.

### DIFF
--- a/app/controllers/admin/interviewee_biographies_controller.rb
+++ b/app/controllers/admin/interviewee_biographies_controller.rb
@@ -4,7 +4,7 @@ class Admin::IntervieweeBiographiesController < AdminController
   # GET /admin/interviewee_biographies
   def index
     scope = if params[:q].present?
-      IntervieweeBiography.find_by_name_substring params[:q]
+      IntervieweeBiography.by_name params[:q]
     else
       IntervieweeBiography
     end
--- a/app/controllers/admin/interviewer_profiles_controller.rb
+++ b/app/controllers/admin/interviewer_profiles_controller.rb
@@ -5,7 +5,7 @@ class Admin::InterviewerProfilesController < AdminController
   # GET /admin/interviewer_profiles.json
   def index
     scope = if params[:q].present?
-      InterviewerProfile.find_by_name_substring params[:q]
+      InterviewerProfile.by_name params[:q]
     else
       InterviewerProfile
     end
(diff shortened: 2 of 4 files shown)


jrochkind:
Do we really need ransack here?  The main reason I have used ransack in the past is to handle those column headings you can click on to change sort order -- which would be tedious to do without ransack. But I dont' think we have those or need those here? And even when i've done it, I've sometimes regretted it, fighting with ransack to do things that weren't on it's golden path. 

Would it make sense to do this without ransack? What do you think?

jrochkind:
Do we really need ransack here?  The main reason I have used ransack in the past is to handle those column headings you can click on to change sort order -- which would be tedious to do without ransack. But I dont' think we have those or need those here? And even when i've done it, I've sometimes regretted it, fighting with ransack to do things that weren't on it's golden path. 

Would it make sense to do this without ransack? What do you think?

jrochkind:
Re the request specs: I _think_ if you actually want to look at content of response, the full `system` test is probably better. 

Which is, I think, what we've done in the past, used request specs for just really basic looking at http response status and headers (or non-HTML JSON responses).  System tests for anything that's actually an html response we want to look at. 

But the distinction isn't entirely clear I guess.  Either way I don't consider it a blocker for this PR, it's fine. But now I'm curious. 

Rspec says:

> Request specs provide a thin wrapper around Rails' integration tests, and are
designed to drive behavior through the full stack, including routing
(provided by Rails)

vs. 

> System specs are RSpec's wrapper around Rails' own
[system tests](https://guides.rubyonrails.org/testing.html#system-testing).

> System tests allow you to test user interactions with your application,
running tests in either a real or a headless browser. System tests use
Capybara under the hood.

Hmm, so wrapper around Rails Integration vs System specs. What's the difference to Rails? 

> Integration tests are used to test how various parts of our application interact. They are generally used to test important workflows within our application.

vs

> System tests allow you to test user interactions with your application, running tests in either a real or a headless browser. System tests use Capybara under the hood.

What?

It looks like Rails is a bit confused about this too. System specs are _newer_ so maybe the newer replacement for Integration tests? But Integration tests are *faster* because they don't use a real browser. 

Still kind of confused about what Rails intends as the distinction here, who knows. I think generally we should try to stick to System tests for HTML, but also no big deal here.

jrochkind:
Nicely done! Rails forms can get confusing, when you want to just do something simple and not totally buy into Rails abstractions. I think this looks right? Merging!

PR #1710, opened 2022-05-24
The name of a DB scope on a model prevents the CI database from building.

eddierubeiz:
This is a throwaway PR based on #1703. I used it to figure out what is breaking the CI database in that branch. Tests pass in master and fail on this branch, so I will make this test progressively similar to master until the tests pass. See the last commits: the actual name of the scope was, in some unfathomable way, preventing the test database from setting up.


eddierubeiz:
My change to the `interviewee_biography.rb` model are causing the problem.

PR #1721, opened 2022-05-25
Don't remove the properties of birth_attributes or death_attributes if they're blank

### LINKED ISSUE #1673: Removing information from fields for OH interviewee biographies
If a field is incorrectly filled in and then later someone tries to delete information in that field, it is not possible to delete the information for the OH interviewee biographies.

For example, I tried to remove the death date for Florine Wise (because she's not deceased to our knowledge) but couldn't do so. It kept showing the incorrect death date on the PUI. @jrochkind had to remove the death date on the back-end manually.

eddierubeiz:
Fixes #1673; includes simple tests.

- The params method in the IntervieweeBiographiesController was _removing_ blank birth/death attributes from the params hash.
- Because of this, the blank attributes never made it to the  #update method, and it was thus impossible to *remove* e.g. someone's death date.
- I'm keeping the fix simple and not changing the UI because *removing* this biographical info is an action we do very seldom. It took us quite a while to even figure out this was broken in the first place.

### DIFF
--- a/app/controllers/admin/interviewee_biographies_controller.rb
+++ b/app/controllers/admin/interviewee_biographies_controller.rb
@@ -79,9 +79,7 @@ class Admin::IntervieweeBiographiesController < AdminController
         tmp["#{name}_attributes"]&.reject! { |k, v| v.values.all?(&:empty?) }
       end
 
-      tmp["birth_attributes"].reject! {|k,v| v.blank? } if tmp["birth_attributes"]
       tmp.delete("birth_attributes") if tmp["birth_attributes"].blank?
-      tmp["death_attributes"].reject! {|k,v| v.blank? } if tmp["death_attributes"]
       tmp.delete("death_attributes") if tmp["death_attributes"].blank?
 
       %w{birth death school job honor}.each do |name|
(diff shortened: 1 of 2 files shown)


jrochkind:
I think part of the point of that code was to prevent there being _empty hashes_ in the json.  In part because I recalled that was a kind of obsession of yours at one point, @eddierubeiz , heh. 

Like, if you submit a form with ALL the boxes for that sub-object blank, do you get a JSON added to the serialized JSON in the db anyway? Like, one with all keys and all empty string values and/or an empty hash with no keys?

OR, was it just to prevent you from having all sorts of empty strings in the database instead of nil/no key, for missing values? 

I'm curious to see if we can find any info on what it was supposed to do in the commit message... nope I didn't!

Either way... I guess there were no tests of whatever the intention was... and empty strings/hashes in the database doesn't seem like the worst thing in the world, and better than a bug like this, so we'll go with it!

eddierubeiz:
Yup. I am now perfectly content living with `"birth_attributes"=>{"date"=>"", "city"=>"", "state"=>"", "province"=>"", "country"=>""}` in the DB. I suppose I've moved on.

jrochkind:
I kind of want to find a way to fix those at the attr_json level, that won't have these problems. I think it can be done.

Issue #1683, opened 2022-05-09
Adjustments to Digitization Queue


apinkney0696:
- [x] Remove R&R Queue (#1722)
- [x] Combine Photographs and Archives Queues into "Archives" (#1274)
- [x] Combine Museum Fine Arts and Museum Objects Queues into "Museum" (#1274)
- [x] Add/adjust fields to 'New Digitization Queue Item' form: (#1726)
- Add Deadline date field (YYYY-MM-DD)
- Add two checkboxes: one for R&R, one for Digital Collections, and the option to select both.
- Combine Instructions and Additional Notes fields
- Remove the Materials field
- [x] On Queue display page, remove the "Status Changed" date display and replace it with "Deadline" (#1726)
- [x] When creating a new attached work to a digitization queue item, autofill the "Department" field in the new work's metadata based on the digitization queue the item is in. (Modern Library and Rare Books queues = Library, Museum queue = Museum, Archives queue = Archives) (#1728)
- [x] Allow for the ability to delete entries in the queue completely (#1729)

- [ ] Cleanup needed to be done as second phase to avoid database errors with heroku preboot #1723 #1725 #1727

PR #1725, opened 2022-05-26
cleanup 1724

jrochkind:
- Consolidate Digitization Queue collecting areas per #1683
- Cleanup after #1724 is fully deployed


jrochkind:
thanks for the catch on reference, now to figure out which one I meant and fix it!

Issue #1707, opened 2022-05-24
Add Scout info to README.md


CHF-IT-Staff:

jrochkind:
Our Digital Collections repository README.md?

I forget, are we paying for Scout, or getting it for free? This is like a condition of our Scout plan?

If we want to do this, and it’s the Digital Collections README we’re talking about, we should make a GH Issues ticket for it.


__________________________________

Jonathan Rochkind   (he, him, his or they, them, theirs)
Software Developer/Technical Lead
Othmer Library of Chemical History
t. +1.215.873.8239

Science History Institute
Chemistry • Engineering • Life Sciences
315 Chestnut Street • Philadelphia, PA 19106 • U.S.A.
Learn about the scientific discoveries that changed our world at sciencehistory.org/learn<https://www.sciencehistory.org/learn>
From: CHF-IT-Staff ***@***.***>
Date: Tuesday, May 24, 2022 at 8:28 AM
To: sciencehistory/scihist_digicoll ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [sciencehistory/scihist_digicoll] Add Scout info to README.md (Issue #1707)

We were asked to add a logo and a link to the Thanks section at the bottom of our README.md file by Scout.
The link is https://ter.li/h8k29r
The logo's are attached[https://camo.githubusercontent.com/f0f0e576cb0dac502a6bdd01c6fc0d09e22b6798e726c8572ce116de469e3fc3/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3565386234393935373664383934353731313336393933632f64616461383364352d373665382d346564612d626365642d363538386437333934646266]<https://camo.githubusercontent.com/f0f0e576cb0dac502a6bdd01c6fc0d09e22b6798e726c8572ce116de469e3fc3/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3565386234393935373664383934353731313336393933632f64616461383364352d373665382d346564612d626365642d363538386437333934646266>

[...]

PR #1736, opened 2022-06-01
Add Scout credit to README

jrochkind:
We would prefer they give us a them-hosted linkable image for an image, so for now this is just text.

Ref #1707 
@CHF-IT-Staff note that we aren't including the image here; not sure if they require it or not, since their terms aren't written anywhere. it would be easier to include an image if they give us one they are hosting that we can hot-link in, that's what BrowserStack did.


CHF-IT-Staff:
Scout provided a url for the image: https://cdn.buttercms.com/H1XnscpwTVC4gsjtyJ2U

Issue #1737, opened 2022-06-01
BrowseEverything error


eddierubeiz:
Error at:
https://app.honeybadger.io/projects/58989/faults/85794580

First PR that reproduces this error is https://github.com/sciencehistory/scihist_digicoll/pull/1712 . The commit is dc4142fde.

To reproduce in staging:
Create a new work, enter minimal metadata, click "Create new work".
At https://staging-digital.sciencehistory.org/admin/works/[ID]/ingest, click "Add from cloud".
S3 file list doesn't come up; instead browse-everything modal background div covers everything else.

eddierubeiz:
First thing to try is downgrade BE to a less recent version to see if that fixes it.

eddierubeiz:
This wasn't the problem -- BE has been at its current version for over a year. Used git-bisect to figure out the problem -- whatever it is -- originates with our upgrade to Ruby 3.1 last week.

eddierubeiz:
We're on the same page. I'm becoming increasingly convinced that this is just about browse_everything and Ruby 3.1 . For one thing, If you *only* revert `.ruby-version` back to `3.0.4` and leave `Gemfile` exactly as it is in `master`, everything looks good. I'm considering doing that in this PR (and erasing some of these commits in the process). Seems more tractable.

And yes. I believe there is an easy test we can add to detect this bug.

jrochkind:
Reported to browse-everything https://github.com/samvera/browse-everything/issues/389

PR #1738, opened 2022-06-01
Revert to ruby 3.0.4

eddierubeiz:
It looks like browse-everything [isn't ready for Ruby 3.1.](https://github.com/samvera/browse-everything#supported-ruby-releases) Ref #1737 .

The bug is first noticeable in  [commit dc4142fde, the upgrade Ruby 3.1 last week](https://github.com/sciencehistory/scihist_digicoll/pull/1712).

This PR is simple:
- Revert only `.ruby-version` to 3.0.4, leaving Gemfile exactly as it is in master.
- Run `bundle update` on top of that.


jrochkind:
Thanks for figuring this out! 

* This also *updates* the version of many gems in Gemfile.lock, which ideally wouldn't be included in the same PR
* There were a few things we had to *add* to Gemfile to make 3.1 work (things that were extracted to be separate gems in 3.1) -- should we back out of that, if those changes were only needed for 3.1?  Not sure. 

I don't know that this needs to block reverting though, it's probably fine. 

* I would like to identify the exact part of b-e that is not 3.1 compatible, to report to b-e. Honeybadger was hiding the relevant part of the  stacktrace, I guess deciding any stack trace in a gem was not relevant to you. But if it's easily reproducible in dev, it should be findable. 

I can see if I can do that, also of course does not block merge/deploy of this now .

eddierubeiz:
The changes to the `Gemfile` to make 3.1 work I kept because I figured we should make a revert PR as similar as possible to the facts on the ground in production.

jrochkind:
I am wondering if adding the new gems to Gemfile for 3.1 -- will cause a problem in 3.0!  But _probably_ not?  I see your point, let's try it without rolling those back! 

Some of this is my fault for bundling various things in the same PR to update to 3.1 in teh first place, i prob should have kept the bundle update separate, so it'd be easier to rollback just the 3.1 update!  I was just thinking "get the newest version of everything ot maximize chances it supports ruby 3.1"!

Issue #1524, opened 2022-01-04
Download transcription and translation as a .pdf


apinkney0696:
Ticket for changing the download format for the transcription and translation text from .txt to .pdf.

In the tech team meeting yesterday, we discussed the .pdf layout will include a science history branding mark, followed by the transcription/translation text in an Institute branding approved font. The line breaks from the webpage that associate the text sections with the images don't need to be included. 

I was thinking it could be helpful to include the work's citation and the credit "Translated by Jocelyn R. McDaniel" at the end of the document. Is that possible?

apinkney0696:
Some additional inspo for this ticket:
[pdf download layout.docx](https://github.com/sciencehistory/scihist_digicoll/files/8365119/pdf.download.layout.docx)

jrochkind:
Thanks! Pretty similar to what I was thinking but:

* I would have only put logo on first page, not repeated header
* I would have put the citation at the top before the transcription/translation, instaed of at the bottom

Let me know if either of those appeal, otherwise we'll do it your way. 

Two more:

* I don't mind the indent of the text after Translation/Transcription heading, although I wouldn't have done it -- but if we have trouble figuring out how to make that happen we might skip it
* The repeated footer with address does seem like a good idea, but again if we have trouble figuring out how to do it, might skip it.

apinkney0696:
That all sounds totally fine to me! I just wanted to put a visual out there to have something to go off of. I think the citation and credit are important to have, but I'm not picky about where they end up on the page. Thanks!

jrochkind:
I was going to say should we try to do it with institute brand fonts -- but i don't know if we have a license to embed them in PDFs we distribute? We could check with Communications to see if this can be done, or we could just use default fonts.

jrochkind:
@apinkney0696 in our current text-file downloads, we have transcription and translation as separate files. 

Do you think it's preferable to combine them into one file?

I have no real opinion, I don't totally understand how these will be used. 

If combining them especially, I wonder if we want to reconsider where the button(s) to download go.

eddierubeiz:
Possible additional feature: QR code linking back to the digital collections.

jrochkind:
@eddierubeiz Intersting idea! I'm guessing most of these PDFs won't actually be printed out, and a clickable link will be sufficient, but that's just a guess!

eddierubeiz:
[...]

PR #1740, opened 2022-06-02
Download transcription and translation as a .pdf

jrochkind:
ref #1524

The TransTextPdf class creates the actual PDF, using prawn and prawn-html to lay it out in simple HTML and convert to PDF. 

The HTML-creating code is definitely kind of kludgey brute-force; the more readable way to do this would be to use ERB somehow (with a ViewComponent?), but ran into barriers there, so just did something stupid simple that worked. 

We can use very very limited CSS, which we do. 

This all works for this basic PDF, if we wanted to do something more sophisticated, we'd need to refactor some things to be a more sophisticated approach too.


jrochkind:
Note also robots.txt excluding these new PDFs -- I can think of zero reason for any robot to have this, so in light of our recent robots struggles, let's just start off excluding them.

jrochkind:
Gah, this font may be GPL licensed, which doesn't work for us. I could have sworn I found an Open Font License one, but that's not what's in this PR.  Re-checking, may be able to get it from an altenrate source with OFL.

Issue #1634, opened 2022-03-31
change how we do rights statements


apinkney0696:
In our most recent meetings about Video, we brainstormed ways to maximize the clarity of the Institute's copyright policy. The solution we came up with, which has been reported to and supported by Michelle and Gabriela, is adding a tooltip next to the large Rights Statement, beneath the representative image.

The tooltip, something like this ![image](https://user-images.githubusercontent.com/78036563/161065233-ab41b232-8847-4877-b71c-bc83efc5b400.png), would lead users directly to the Copyright section on the About page. 

Here is the idea sketched out:
![IMG_2647](https://user-images.githubusercontent.com/78036563/161065703-f7946ca9-537a-4275-ac4e-6cec4069f05c.jpeg)

In summary, this request is to:
1. Add a tooltip next to the large copyright statement under the representative image. A question mark icon for this would be best.
2. Hovering over the tooltip should bring up a small textbox that says "What is this?"
3. Clicking on the tooltip should take a user directly to the Copyright section of the About page.

jrochkind:
Hi!

Is this just for the video page, or for all pages? 

i'd be inclined to do it on all pages, for consistency?

Is it just for certain copyright status, or all statuses? 

I'm not sure if the "hover to bring up a small textbox that says "What is this?" is a beset pracitce/standard UI/UX:
* For accessibility, we need to support keyboard navigation -- But when you keyboard navigate to activate that element, should it trigger the "hover" message, or the link? Probably the link?  
* Keep in mind that "hover" actions are not triggerable on mobile/touch devices
* I'm not sure if the standard utilities we're using (from bootstrap) for tooltips will support one action on hover and another action on click, but we can investigate. There might be a simple way to do it. 
* Do you have any examples of websites with UI elements like this, that we can use as a model?

apinkney0696:
Based on @jrochkind and my meeting yesterday, we are re-working the copyright information plan as follows:

The copyright statement link located beneath the representative image and the copyright statement link in the metadata display will no longer link to rightstatement.org but instead will send the user to a new page (or a pop-up window?). On this page, there will be a section that includes the paragraph of information about the rights statement from rightstatement.org AND our institutional language around copyright.

There will be one page per copyright statement, excluding Public Domain and CC licensed works, which will continue to link to the Creative Commons website.
[...]

PR #1742, opened 2022-06-07
Local rights explanation page

jrochkind:
Combining text from rightstatement.org with local fixed copyright statement

Only for rightstatement ones, not CC or public domain. 

This ended up involving some significant refactors to the RightsIconComponent, which has gotten somewhat convoluted, sorry!

Ref #1634


jrochkind:
@apinkney0696 This one too is just waiting for final stakeholder review and approval.

jrochkind:
Oh hey, going to throw something in robots.txt too to keep robots from following these pages -- now that we have one unique URL for every work, we really don't need robots spidering and indexing them.

jrochkind:
Honestly I don't love how all this code is structured. Maybe it needs to be refactored to be simpler/more consistent. But for now I'm thinking "good enough" over perfectionism.

jrochkind:
This is now approved by stakeholders, ready for review/merge @eddierubeiz

Issue #1638, opened 2022-04-04
Make searchable UUIDS for all of a work's descendant assets, maybe?


eddierubeiz:
It would be nice to be able to look up a newly-ingested work based on one of its member asset's primary keys without opening a production console and running `Asset.find` .

Typical use case (there are variations):

- A job fails for a particular asset.
- The failed job metadata / logs contain the asset's UUID, but no metadata for the parent work.
- You copy the asset's UUID from the failed job metadata into the search box at the top of `/admin/works` and the work containing the asset matches _(this is currently impossible)_
- If the work only has one or two assets (commonly true) you now have a convenient way to poke at the asset.

I think this could be nice to have. @jrochkind  - curious how you feel.

jrochkind:
Should we have an Assets page in Admin dashboard, listing/searching assets?

jrochkind:
We're (ab)using "ransack" for admin search. Not sure how easy it is to get it to search on an associated PK, but we could probably work around it.

jrochkind:
> Or we could try adding descendant UUIDs to public-facing SOLR search, and show matches for staff only?

That seems awfully complicated. I'd leave admin functions in the admin dashboard, mainly, unless we have a good reason not to?

eddierubeiz:
Yeah, maybe an asset-only search using Ransack is best...

PR #1753, opened 2022-06-21
Simple Asset admin screen

jrochkind:
Keep it simple. No sortable columns, always sorts by created_at desc.

Ref #1638


jrochkind:
eh you could have made me write a test, but I'll take it.

Issue #1751, opened 2022-06-15
Typo in the name of our French translator.


eddierubeiz:
The translator's name should be changed to `Julie Hugonny`.
Thanks for reporting this, Jocelyn!

eddierubeiz:
The change only affects [five works](https://digital.sciencehistory.org/catalog?q=Huggony).

eddierubeiz:
The change only affects 5 works:

- https://digital.sciencehistory.org/admin/works/5bcqdva/edit
- https://digital.sciencehistory.org/admin/works/oddcx7f/edit
- https://digital.sciencehistory.org/admin/works/21bcx0g/edit
- https://digital.sciencehistory.org/admin/works/rkszvss/edit
- https://digital.sciencehistory.org/admin/works/pivfxgt/edit

PR #1754, opened 2022-06-21
Opt-in to attr_json feature for jsonb_contains query

jrochkind:
attr_json gem provides this as an opt-in feature. Not sure why I left it out of kithe and out of here until now! It's super useful, makes it way easier to do queries on our json_attributes stuff without having to work it all out, including in situations with arrays.

https://github.com/jrochkind/attr_json#querying

For instance, ref #1751, wouldn't it be easier to do `Work.jsonb_contains("additional_credit.name": "Julie Huggony")`? Now we can.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -1,4 +1,5 @@
 class Asset < Kithe::Asset
+  include AttrJson::Record::QueryScopes
 
   include RecordPublishedAt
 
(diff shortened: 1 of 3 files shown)


jrochkind:
Oy this PR is not what was supposed to happen. Not sure what's happened here, closing.

Issue #1757, opened 2022-06-27
Store `hls_playlist_file_data` keys and values properly


eddierubeiz:
1) Currently, hls_playlist_file_data key and value are escaped and stored as a string:

```
select distinct jsonb_typeof(json_attributes->'hls_playlist_file_data')
    from kithe_models
    where kithe_model_type = 2
    and file_data ->> 'storage' = 'video_store';
```
will return "string" and null.

This blog post describes the problem: https://dev.to/mrmurphy/so-you-put-an-escaped-string-into-a-json-column-2n67 .

A sample value for this string, as currently stored in the DB would be:
```
 "{\"id\":\"hls/fd660b63-6603-4584-ad7b-1263ecf2b53b/426eac8d0062d9384bd1dfa2588e1991/hls.m3u8\",\"storage\":\"video_derivatives\",\"metadata\":{}}"
```

2) Say you're interested in the `hls/fd660b63-6603-4584-ad7b-1263ecf2b53b/426eac8d0062d9384bd1dfa2588e1991/hls.m3u8`  part of the string. To get at it you need to parse the string inside the DB using an ugly hack like:
```
select (json_attributes->'hls_playlist_file_data' #>> '{}')::jsonb->>'id'from kithe_models where id='SOME_ID';
```
3) Of course, we'd rather not; what we should really do is store the string as jsonb, just like the rest of `json_attributes`.

This ticket would cover the work of:

- editing the code that stores info about video derivatives so that the data is stored as `jsonb`
- writing a migration that converts the (currently very few) existing video deriv records in the DB to the new format
- editing the code in method `get_hls_derivs_from_db` in file  app/services/orphan_s3_video_derivatives.rb (see ref #1748 ) to simplify it, once the hack is no longer required.

jrochkind:
It's already being stored in a jsonb column -- the problem is that it's being escaped first!

I think I may know why and how to fix it; the migration will be trickier, yeah!

PR #1763, opened 2022-06-29
Use correct type to avoid double-escaping hsl_playlist_file_data

jrochkind:
Ref #1757 

With test to ensure it's not double escaped.

Migration of existing data now in rake task: `heroku run rake scihist:data_fixes:unescape_hls_playlist_file_data -r production`
* This is idempotent, it's okay to re-run it on already un-escaped data, it will just leave it alone
* The OLD version of the attribute (with type `:json`)  _can_ still read the unescaped data
    *   So if we deploy this code, then run the rake task before the heroku preboot has actually switched over to our new code --we're good. The fixed data can be read by old still-running version, and will be readable by new version once deployed. 
    * The reverse isn't true, the NEW version of the attribute (with `ActiveModel::Type.new`) can _not_ read old un-migrated data. So if we deploy this code but forget to run the rake task... the new version of app will start raising errors every time it tries to load video content. 

So this is a bit tricky, but, eh, we're not a hospital.  I will leave this in draft until we're ready to actually do the switcheroo, intentionally, so we don't accidentally deploy it and not run the rake task. 

(I could have made the rake task a db migration so it would run automatically on deploy... but there were enough weird things going on here already, I was more comfortable leaving it manual? May not be right call?)

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -23,7 +23,11 @@ class Asset < Kithe::Asset
   # of a db column. We 1) create the `attr_json`, then 2) in the shrine attachment we tell it
   # column_serializer:nil tells shrine not to serialize the JSON, let
   # attr_json take care of that.
-  attr_json :hls_playlist_file_data, :json
+  #
+  # And ActiveModel::Type::Value.new tells attr_json to just pass it through without
+  # transformation (it's sort of the basic no-op or null type), to get serialized to
+  # json by parent column.
+  attr_json :hls_playlist_file_data, ActiveModel::Type::Value.new
   include VideoHlsUploader::Attachment(:hls_playlist_file, store: :video_derivatives, column_serializer: nil)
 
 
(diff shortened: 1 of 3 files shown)


jrochkind:
NOTE -- we might not deploy this for... a while?  With my vacation and such. 

Not sure how that impacts coordinating merge/release of #1748, it's kind of a pain. This PR does not include fixes to #1748 accordingly.

jrochkind:
Oops, my rake task didn't actually work on staging. So another reason this is still in draft, will update here when I have it working.

jrochkind:
OK, fixed the problem so rake task to migrate works with version of PG we are using. 

Only question is now how to synchronize this release, with me being on vacation next week, and with #1748

eddierubeiz (approved):

jrochkind:
Deployed and migrated.

Issue #1769, opened 2022-07-06
Update `related_url` metadata to point at new URLs on hosted ArchivesSpace


eddierubeiz:
With the switch to a hosted ArchivesSpace instance, we are finally turning on the public user interface, which gives each collection its own URL.

Let's point Bredig collection "related url" metadata at the new Bredig collection URL. (There may be others, so let's update all  matching URLS.)

Note: the script to `[produce the mapping of old to new URLs](https://github.com/sciencehistory/archivesspace_scripts/pull/31)`. has been ready since February.

PR #1770, opened 2022-07-06
update related urls

eddierubeiz:
Ref #1769
Includes a test, which we may not want to keep. Just so much easier to do TDD for this sort of thing.
The list of URLs may change until the last minute, but is very unlikely to do so.

### DIFF
--- a/lib/tasks/data_fixes/update_related_urls.rake
+++ b/lib/tasks/data_fixes/update_related_urls.rake
@@ -1,0 +1,23 @@
+require "#{Rails.root}/lib/scihist_digicoll/task_helpers/related_url_updater"
+namespace :scihist do
+  namespace :data_fixes do
+    desc "Changes related urls to the new ArchivesSpace URLs."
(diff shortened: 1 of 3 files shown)


lib/tasks/data_fixes/update_related_urls.rake
```diff
@@ -0,0 +1,23 @@
+require "#{Rails.root}/lib/scihist_digicoll/task_helpers/related_url_updater"
```
jrochkind, line 1: `./lib` should already be on the ruby "load path", so you should be able to simply do:

    require `scihist_digicoll/task_helpers/related_url_updater`

Which would be preferable. If it doesn't work, I'll be curious!

jrochkind:
marking "draft" so we don't accidentally merge it before the new PUI is ready. Looks good though!

jrochkind (approved):

eddierubeiz:
Oh, you already approved. Merging.

PR #1771, opened 2022-07-08
Make exhibition field private

eddierubeiz:
Requires reindex.
Compare results `/catalog?search_field=all_fields&q=Science+at+play` before and after logging in; specifically, take a look at the department facet breakdown.


jrochkind:
looks good to me!

jrochkind:
deployed and re-indexed!

PR #1781, opened 2022-07-13
upgrade blacklight 7.29

jrochkind:
Had to do some things i don't really underestand to get it to work with latest Blacklight, involving permitted search params. 

Ran into actual blacklight bugs/changed behavior, which resulted in lots of blacklight debugging and help from the community fix blacklight,and new blacklight releases. 

Was then able to un-do some workarounds we had put in for Rails 6.1.6.1, that new blacklight/blacklight_range_limit accomodated themselves.  

ref continued cleanup re #1786


jrochkind:
It was a change in blacklight_range_limit, which might get reverted. https://github.com/projectblacklight/blacklight_range_limit/pull/195

jrochkind:
Spent today investigating/understanding another thing that has changed in blacklight/blacklight_range_limit that was causing us problems. It was a mysterious one. https://github.com/projectblacklight/blacklight_range_limit/issues/196

Note that our app DOES set Rails config `config.action_controller.action_on_unpermitted_parameters = :raise` -- only in development and test -- and has for some time. We do this to make it easier to catch params you accidentally have not "permitted" through Rails strong params, which can be a difficult bug to figure out (or even notice!) sometimes. 

but  this may be violating some assumptions Blacklight has that you do NOT set this, Blacklight may be wanting to use StrongParams in a different way where it assumes there WILL be some unpermitted params it _means_ to ignore. 

For now, we've worked around it, but perhaps we'll have to stop doing `config.action_controller.action_on_unpermitted_parameters = :raise` at some point. 

At this point we are only waiting on https://github.com/projectblacklight/blacklight_range_limit/pull/195 -- if it gets merged, this branch should pass again with new release of blacklight_range_limit.  If it does not, we'll have to rewirte portions of our app to go along with the new query string param names for range limit limit params.

Issue #1656, opened 2022-04-18
Rewrite the DZI orphan checker.


eddierubeiz:
We have good reason to believe the DZI orphan checker is not doing its job right.

See #1653 for at least one example in which it did not detect DZI files that had belonged to 3 deleted assets.

~~A rewrite would hinge on our ability to get a list of top-level objects in the `scihist-digicoll-production-dzi` bucket, and nothing else.~~
~~These objects (all "folders", in this case) should line up, one to one, with assets that are supposed to have DZI tiles.~~

(Later edit: it turns out the existing setup works mostly fine; it was looking at the top two levels of the dzi directory but only reporting / deleting the dzi tiles *iff* it found an orphaned dzi manifest file first.)

We do **not** want to attempt a full traversal of the bucket, as there are **lots** of files down there. They must be detected, and deleted, in batches.

eddierubeiz:
A variation on `aws s3 ls s3://scihist-digicoll-production-dzi` should work.

jrochkind:
I _believe_ that `aws s3 ls` will actually need to retrieve every key for every object under the hood, and then it just only prints out the unique prefixes. 

So I believe that could take a while for our 25 million object bucket. If my calculations are correct, it will cost...  12.5 cents every time you do it. (Also my calculations may not be correct). 

There are ways to do the equivalent with ruby SDK API, it should be the same for either.

Issue #1653, opened 2022-04-18
Run DZI orphan checker; investigate whether there are files it should be catching that it isn't.


eddierubeiz:
Run the DZI orphan checker manually and see if it catches anything ( we suspect it might, subsequent to AWS complaining about too many simultaneous deletion requests). Regardless, go over the code; Jonathan notes:

_Looks like [we call delete_prefixed in shrine storage](https://github.com/sciencehistory/scihist_digicoll/blob/master/app/models/dzi_files.rb#L86)..._ 
_Then shrine storage...  yep just calls SDK API with a prefix to make a "collection" of objects, and then [batch_delete](https://github.com/shrinerb/shrine/blob/master/lib/shrine/storage/s3.rb#L203) on the collection.  I don't think there's any AWS API to delete files by prefix in one HTTP call._
_This makes it hard to introduce a delay for not doing too many files at once, since it's several layers of dependency down doing the actual delete! Would have to stop using the shrine method or PR a feature to it I guess._

jrochkind:
Since the checker runs nightly and has not notified us, it seems safe to assume running it manually will NOT find anything. 

That doesn't mean no orphans are there. 

Here are the errors we got that seem to suggest the app tried to delete files on S3, but got rate-limit errors from AWS when trying to delete, suggesting that some files must be left behind, right?

So the hard part will be finding those files.  

**Warning** This may get a lot harder if you wait 7 days and HoneyBadger isn't keeping full historical data past 7 days? You may want to capture any relevant context now?

```
Apr 13 09:46:04 scihist-digicoll-production app/worker.4: ERROR: [ActiveJob] [fa6e0d39-a098-4c42-b448-90fe11e10be6] Stopped retrying DeleteDziJob (Job ID: fa6e0d39-a098-4c42-b448-90fe11e10be6) after 2 attempts, due to a Aws::S3::Errors::SlowDown (Please reduce your request rate.).

Apr 13 09:46:06 scihist-digicoll-production app/worker.5: ERROR: [ActiveJob] [41b8bd74-2c4e-4f6c-a8ac-6938a13dca65] Stopped retrying DeleteDziJob (Job ID: 41b8bd74-2c4e-4f6c-a8ac-6938a13dca65) after 2 attempts, due to a Aws::S3::Errors::SlowDown (Please reduce your request rate.).

[...]

PR #1787, opened 2022-07-19
[WIP] Delete all orphaned dzi tiles, EVEN in the absence of the orphaned manifest file.

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/1656

1. The rapid rate of delete requests generated by a `delete_orphans` job (or a regular asset deletion) [sometimes causes s3 to throw](https://github.com/sciencehistory/scihist_digicoll/issues/1653) a `Aws::S3::Errors::SlowDown (Please reduce your request rate.)`.

2. In the aftermath, you will typically find the `.dzi` manifest files have been correctly deleted, but
the actual tiles in the orphaned tile directories are still lying around.

3. We no longer want to rely on the existence of an orphaned `.dzi` manifest in detecting the presence of orphaned tiles. This is really a bug: on April 18th, we ran the orphan checker and [it failed to detect](https://github.com/sciencehistory/scihist_digicoll/issues/1653#issuecomment-1101563386) folders full of dzi files that were in fact orphaned. (The manifest file had been correctly deleted, but the tiles had been left behind.)

5. We keep the same basic approach: traverse ONLY the top two levels of the dzi hierarchy (assets, then one level down for each asset). However, we're running the check a bit differently:

First iterate over the top level and make sure it belongs to a legitimate asset in the DB.
- If the top level folder is NOT legitimate, report the entire folder as orphaned. Do not bother looking at its contents.
- If the top level folder IS legitimate, iterate over what we expect to be the two items below it (manifest and tile folder).
  - Ensure these both match what we expect. (This includes checking the md5 hash in the s3 path against the one stored for the asset in the DB.) If not, report the manifest file or tile folder as orphaned so it can be deleted.


eddierubeiz:
Cost estimate

Assuming there are LESS than a thousand files per DZI tile folder, we have:

- 26,478,689 tiles [total in the production bucket](https://s3.console.aws.amazon.com/s3/buckets/scihist-digicoll-production-dzi?region=us-east-1&tab=metrics);
- 62,289 assets;
- 142 tiles on average per asset (note this is less than 1000).
- 3 list requests (top level folder; dzi manifest; tile folder and its contents) for each of 62,289 assets = 186,867 list requests
- Current cost of 186867 list requests = roughly [12 cents](https://calculator.s3.amazonaws.com/index.html).
- Assuming we run this weekly, this check should run us about $7 a year.

jrochkind:
Thanks! I want to make sure @CHF-IT-Staff sees above cost estimate. Checking this huge number of files is real money. Not a big deal at $7/year, but if we make some mistake it could easily go up. Say, if we forget and start running it daily for instance, it would I guess be $50/year, which still isn't a huge deal, but you see what I mean.

eddierubeiz:
I could copy 100 assets from staging to dev, then generate 100 sets of dzi tiles in dev, then run the checker on that set of 100 tiles, then multiply those numbers by 623 (because 623 x 100 = 623,000 ≈ the number of assets we currently have in production.

Not perfect but ...  if it takes 20 minutes to check 100 files we definitely have a problem.

jrochkind:
I mean, you could also just generate the 100 sets of DZI tiles on staging if you want! OR copy a selection of 100 from production to staging. 

We don't sync DZI tiles to staging, but if we're talking about choosing to generate some, we could do that on staging, rather than dev. For a more standard and predictable env.

eddierubeiz:
Yesterday I tried a very crude estimate (see above) and created 100 dzi directories in my dev setup.

(Full disclosure: my dev setup is not at all the same as a standard-1x heroku dyno, and the 100 dzi dirs I was testing against may not be a representative sample of our real dzi directories -- just to cite two obvious flaws in the comparison.)

With that out of the way checking 100 dzi directories in dev took 7.6 seconds.
Multiplying that out gives us 78 minutes or 4 cents per week of dyno time to check all the directories,
or an extra **$2.31 a year.** due to dyno time.

jrochkind:
i don't love that the orphan checker takes an extra hour+ every time we run it... but I guess @CHF-IT-Staff really wants all the orphan checking we can do! 

So if there's no obvious way to make it more efficient, so be it I guess!

jrochkind:
Costs will double if we have to make our scheduled tasks use a standard-2x instead of standard-1x, to avoid out of memory errors, which I keep thinking we might have to, as our app seems to have ballooned to take more memory and is giving us such errors.

jrochkind:
Costs would also end up 7x as much of course if at some point we decided to run nightly forgetting to think about costs. 7x2x would be estimate of $32/year, which still doesn't really matter. Just examples of how adding features like this leaves us little potential cost pitfalls waiting to trap us.

eddierubeiz:
We discussed last week and Chuck approved the cost. Merging and deploying.

Issue #1779, opened 2022-07-13
Adjust OAI-PMH feed for Wikimedia Partnership


apinkney0696:
Here is the original email correspondence with Leanne Finnigan from PA Digital:

Hi, Annabel.

> I’m reviewing SHI’s metadata after mapping edm:object for the Wikimedia partnership. There are 70 records with duplicate URLs in edm:object. Here’s the list:
> 
> [https://docs.google.com/spreadsheets/d/1kVAoQfkShz562QN9i-yzE4RJJZzuqTevr4XIoIqh0CE/edit?usp=sharing](https://nam10.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdocs.google.com%2Fspreadsheets%2Fd%2F1kVAoQfkShz562QN9i-yzE4RJJZzuqTevr4XIoIqh0CE%2Fedit%3Fusp%3Dsharing&data=05%7C01%7Cleanne.finnigan%40temple.edu%7Cd838b8cb0018477bb9b708da4af74220%7C716e81efb52244738e3110bd02ccf6e5%7C0%7C0%7C637904723085902789%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=t5PysRw6NCq2mTVTYGoMtyMkHEpbx%2FVZXSC4tkACwrA%3D&reserved=0)
> 
[...]

PR #1792, opened 2022-07-20
OAI-PMH: Include all child images as edm:object

jrochkind:
Ref #1779

* Per PA Digital's instructions for contribution to wikimedia commons -- they want all child images, and this is where they get it from
* if no full-size JPG download deriv is available (say for video), we just don't include, and test to make sure that's fine.
* We don't necessarily handle multi-level child works (like a child that itself has child works); becuase we don't do that much, and it's hard to handle in a performant way. We may leave out some images if that happens though. There are other parts of our app that make this perf optimization too at cost of complete inclusion, it's ok.
* added strict_loading to fetch for oai-pmh export, so we'll raise if we accidentally n+1!
* This does mean that a given entity in our oai xml export can have hundreds of edm:object tags. PA Digital tells us they and wikimedia commons can handle this just fine. It will slow down our export somewhat, but I don't think enough to cause a problem, as far as I can tell, and no other problems for us. This seems to be the most straightforward way to get them the data.

### DIFF
--- a/app/controllers/oai_pmh_controller.rb
+++ b/app/controllers/oai_pmh_controller.rb
@@ -12,8 +12,9 @@
       sample_id (Work.first.friendlier_id rescue "tq57nr00k")
 
       # Note we limit to published true, and pre-load any accessed associations for performance
+      # use .strict_loading to ensure we can't access any n+1 associations we haven't pre-loaded
       source_model OAI::Provider::ActiveRecordWrapper.new(
-        Work.where(published: true).includes(:leaf_representative, :contained_by),
+        Work.where(published: true).includes(:leaf_representative, :contained_by, members: [:leaf_representative]).strict_loading,
         identifier_field: :friendlier_id
       )
     end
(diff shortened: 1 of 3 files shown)


jrochkind:
I discovered a cool tool for testing out OAI-PMH:

https://validator.oaipmh.com/#ListRecords

If you choose the "Download XML" tab, and enter eg `https://staging-digital.sciencehistory.org/oai`, it will page through our entire OAI-PMH feed, and download ALL our records. (Currently 128 separate XML files/pages). 

This is a way of making sure no errors are produced paging through the entire thing -- seems to be the case. 

I also did it on both staging and production to see if there were any obvious indications that adding all these member URLs slowed things down -- just based on a really informal and inexact eyeballing of the individual XML download speeds it gives:  hard to say. Some individual XML files are faster, some slower, some about the same. I'm going to say if it's slower, it's not enough difference to cause any real problems for anything.

PR #1798, opened 2022-07-26
Fix ActiveEncodeStatus#refresh_from_aws when Asset has been deleted

jrochkind:
Ref #1796 which is an error where this situation resulted in an uncaught exception. We actually already meant to handle this and even had a spec for it -- but turns out the spec was unrealistic. Adding the "reload" to the spec made it realistic, and initially reproduced the exception. Marking the association optional then made the spec pass again.

So I think this should work to take care of #1796. Once merged we must re-enable cron job as per instructions there.


eddierubeiz:
Now that this is merged I will take some time tomorrow (Wed.) morning to deploy this and turn the scheduled task back on. If there are any orphaned files I may delete them manually or test the DZI orphan checker on them.

Issue #1783, opened 2022-07-14
Adding MP3 or M4A derivatives to OH requests?


rachellane12:
From Slack (@jrochkind):

I wonder if we should also just create MP3 derivatives, and offer people FLAC or MP3 options (when we have FLACs).  If we’re thinking about improving the OH download experience. Maybe we should make a placeholder Issue for that to consider it too.

1. Decide if we want MP3 or M4A (I haven't identified any other candidates), based on some combo of:
    * interviewing (local?) users
    * google or peer research as to common/best practices, including ensuring common consumer OS's/software can play m4a, that they are accessible

2. Make our ingest process create the mp3/m4a derivative, only for flac originals
    * At first I was wondering if we could use AWS MediaConvert for this (that we use for video transcodes). But while it does support some audio-only transcoding, it actually doesn't look like FLAC is an input format it can recognize! https://docs.aws.amazon.com/mediaconvert/latest/ug/supported-codecs-containers-audio-only.html. Also it's just generally kind of hard to work with, we have to do extra custom things like we did with our video transcode to keep track of it at all. So we'll probably just do ffmpeg in a bg job as we do for say combined audio derivs. 

3 Create the derivatives for existing flac originals

4. Make sure the derivative shows up in the "Download" menu when present.

5. Make sure the derivative is available in the email for by-request OHs.

2, 3, 4, and 5 could be separate PR's.

rachellane12:
From Slack (@sarahschneiderSHI):

I do know that Mia prefers using mp3s (or WAV files) when she's using our collections, rather than trying to work with the FLACs. So this would help her for social media use!

sarahschneiderSHI:
I sent a message to Mia asking her about this and will let you know what I hear back. I believe mp4 is a good format for her, but I am not sure about AAC. Thanks!

sarahschneiderSHI:
Oh, I misunderstood, sorry! That is good to know!

jrochkind:
Btw, just testing to see how much space an m4a (lossy compressed) can save over FLAC, with default settings the m4a is around 10% the size of the FLAC!  Knowing it’s audio and can handle lower-quality lossy probably without anyone noticing, it can probably be even smaller. 

We’ll still be storing the FLAC, so we won’t actually save our own disk space (in fact we’ll add another 10%), but we can make things much easier on people that want to listen to or download them.

sarahschneiderSHI:
[...]

PR #1806, opened 2022-07-28
m4a derivatives for flac files

eddierubeiz:
Ref #1783
Includes tests and a simple rake task to create the derivs, which should run in less than a day in production.


app/uploaders/asset_uploader.rb
```diff
@@ -88,6 +88,16 @@ class AssetUploader < Kithe::AssetUploader
   end
 
 
+  # For FLAC originals, let's create a mono m4a derivative.
+  Attacher.define_derivative('m4a', content_type: "audio") do |original_file, attacher:|
+    # Both audio/flac or audio/x-flac seem to be valid
+    if attacher.file.content_type.match /flac$/
+      Kithe::FfmpegTransformer.new(
+        bitrate: '64k', force_mono: true, audio_codec: 'aac', output_suffix: 'm4a',
+      ).call(original_file)
```
jrochkind, line 97: maybe mention in comment that the output characteristics, including relatively low 64K bitrate, is based on assumption these are Oral History voice recordings. And maybe mention the combined audio derivative, which has matching choices.

jrochkind:
I believe we have an existing rake task that will go and make the new m4a derivs for any that need and don't have. But I can't recall what it is. Can you help find it, and maybe test it?

app/uploaders/asset_uploader.rb
eddierubeiz: Yeah -- haven't put much thought into the bitrate; for now, just copied the settings from the combined audio derivs.

jrochkind: I think they are probably right -- I think I remmeber you DID put a lot of thought into em when we made the combined audio derivs!  I don't know that it needs to be re-visited. 

I just think they would not necessarily be appropriate for musical instead of voice content -- which we don't _expect_ to have... The combined audio derivs actually *only functions* on Oral Histories. But this new feature will function on any FLAC uploaded. We don't have any plans for anything but OH... I was just thinking leaving a comment explaining where the settings came (copied from combined audio derivs, if changed in one place should be changed in both) is a good idea at this point.

app/presenters/download_options/audio_download_options.rb
```diff
@@ -15,6 +15,14 @@ def options
       # We don't use content_type in derivative option subheads,
       # cause it's in the main label. But do use it for original.
 
+      if m4a_deriv = asset.file_derivatives[:m4a]
+        options << DownloadOption.with_formatted_subhead("M4A",
```
eddierubeiz, line 19: How do we want to label this download?

app/uploaders/asset_uploader.rb
eddierubeiz: Definitely - music recorded in stereo would be pretty much ruined in this format. These are mono and the lowest possible quality that's acceptable for listening to a conversation between two people. The use case is "I want to download these quickly and listen to them".

app/presenters/download_options/audio_download_options.rb
jrochkind: How does it come out with what you are currently doing?  What do you think?

eddierubeiz: Currently just "M4A". I'm fine with that.
Screen Shot 2022-08-01 at 10 33 05 AM
.

jrochkind: I suppose we could put something in the sub-head like "smaller file", I dunno. Or "Smaller file (M4A -- 21.9 MB)" vs "Original file (FLAC -- 416MB)".  Just brainstorming. Not sure if everyone knows what M4A is -- but it's really what they ought to be downloading unless they specifically want the FLAC!

eddierubeiz: Good call

app/uploaders/asset_uploader.rb
```diff
@@ -88,6 +88,22 @@ class AssetUploader < Kithe::AssetUploader
...
+  # For FLAC originals, we create a mono m4a derivative.
+  # Typically this deriv is only 5% of the size of the original FLAC,
+  # while still fine for listening, at least for for oral histories.
+  # (These settings would not be appropriate for audio other than recorded speech.)
+  #
+  # See also the settings at app/services/combined_audio_derivative_creator.rb
+  # which are identical (similar use case).
+  Attacher.define_derivative('m4a', content_type: "audio") do |original_file, attacher:|
+    # Both audio/flac or audio/x-flac seem to be valid, so let's check for either.
+    if attacher.file.content_type.match /flac$/
```
jrochkind, line 100: Why not just `== "audio/flac"? ` instead of `.match /flac$/`?

jrochkind: Also may need some `&.` to avoid crashing if file.content_type is missing?

jrochkind:
@eddierubeiz So after deploy, we run `heroku rake scihist:data_fixes:add_m4a_derivative_to_audio_assets` -- do you have any idea how long this will take? 

If it will be many many hours, do we need to come up with a different way to do this, like with bg jobs? Or it's okay to just leave the task running I guess....  if it crashes in the middle and we need to run it again, will it be "lazy", only create derivs if one doesn't already exist?

(I would have just used the existing deriv control rake task for this, cause we know it's already written to be lazy etc.  I suspect any time saved from iterating through fewer Assets will be tiny tiny tiny compared to the actual amount of time it takes to transcode).

app/uploaders/asset_uploader.rb
jrochkind: Ah right that comment is right there. OK but how about `attacher.file.content_type.in?(["audio/flac", "audio/x-flac"])` :)

eddierubeiz:
Should take a few hours (I believe under 8 hours)  based on running subsets in staging.
We can of course enqueue the jobs and run them using special jobs dynos, but that seemed like overkill to me.

app/uploaders/asset_uploader.rb
eddierubeiz: Yeah good call on the `&.` I'll take the whole line.

jrochkind (approved):

jrochkind:
Approved. You want to take care of creating the m4a's, or you want me to?

eddierubeiz:
Cool . I'll run the task and report back.

Issue #1785, opened 2022-07-15
Fixity report doesn't know about video asset s3 buckets


eddierubeiz:
The fixity report at `/admin/fixity_report` incorrectly lists video assets as "not fully ingested".  The assets are ingested just fine, but are stored in a bucket the report doesn't know about.

Easy to fix: update two methods in `fixity_report.rb`
```
  def not_stored_file_sql
    "file_data ->> 'storage' != 'store' or file_data ->> 'storage' is NULL"
  end

  def stored_file_sql
    "file_data ->> 'storage' = 'store'"
  end
```

jrochkind:
Also verify video files are actually being checked for fixity

eddierubeiz:
Fixity checks for video files are working fine, by the way.

PR #1808, opened 2022-08-01
Count assets in both store and video_store as "stored"

eddierubeiz:
Ref #1785


eddierubeiz:
May add a test for this.

PR #1809, opened 2022-08-02
Fix to add_m4a_derivative_to_audio_assets task; create derivs via job in  special_jobs queue

eddierubeiz:
Fixes a trivial error that misled me yesterday into believing this job could run in less than a day.
Also now runs the derivative creation in a series of `special_jobs` jobs.

### DIFF
--- a/lib/tasks/data_fixes/add_m4a_derivative_to_audio_assets.rake
+++ b/lib/tasks/data_fixes/add_m4a_derivative_to_audio_assets.rake
@@ -1,24 +1,22 @@
 namespace :scihist do
   namespace :data_fixes do
     desc """
-      Creates an m4a derivative -- lazily -- for all flac audio assets.
+      Enqueues a job for the creation of an m4a derivative -- lazily -- for all flac audio assets.
       bundle exec rake scihist:data_fixes:add_m4a_derivative_to_audio_assets
     """
 
     task :add_m4a_derivative_to_audio_assets => :environment do
-      scope = Asset.where("file_data -> 'metadata' ->> 'mime_type' like 'audio/%'")
-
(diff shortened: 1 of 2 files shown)


lib/tasks/data_fixes/add_m4a_derivative_to_audio_assets.rake
```diff
@@ -1,24 +1,22 @@
...
     """
 
     task :add_m4a_derivative_to_audio_assets => :environment do
-      scope = Asset.where("file_data -> 'metadata' ->> 'mime_type' like 'audio/%'")
-
+      scope = Asset.where("(file_data -> 'metadata' ->> 'mime_type' like 'audio/flac') or (file_data -> 'metadata' ->> 'mime_type' like 'audio/x-flac')")
       progress_bar = ProgressBar.create(total: scope.count, format: Kithe::STANDARD_PROGRESS_BAR_FORMAT)
-
-      scope.find_each do |audio_asset|
-        unless audio_asset.stored? && audio_asset.content_type.end_with?('flac')
```
eddierubeiz, line 14: BUG: only created the derivative if the file was *not* in s3.

jrochkind:
Hm, you created a new CreateM4aDerivativeJob .  This is only used by the rake task which is a temporary one-time-only thing?

So we want to delete it when we are done running this rake task? I'd rather not leave unused code sitting around unmaintained/untested (don't really want to leave it sitting around tested either!)

I thought we had generic derivatives-creating code that could be used... but may not that uses bg jobs actually. 

So that's fine, but should we PR removing the job again when we're done with the one-off?

eddierubeiz:
We certainly can remove the `CreateM4aDerivativeJob` job class after we're done, yes.

jrochkind:
Cool, can you add it to your task list to remember to do so? Thanks!

Issue #1804, opened 2022-07-27
Expand "related URLs" to be structured sub-models instead of just strings


jrochkind:
A "related URL", perhaps re-named "related Link", should, in addition to having a URL, have a "label" (to be used as link text?), a "category" (possibly to be displayed to the user, but also for our own internal management, as we may display different ones in differnet ways), and maybe more?

Existing "related URLs" will need to be migrated.  For this and other reasons "label" will probably be optional, if not present URL/hostname will be used, as at present. 

Categories may be:

* Distillations Articles
* Distillations Podcasts
* Historical Biographies
* Blog Posts
* Library Guides (LibGuides)
* Archival Description (EAD)  
    * for now; in the future perhaps we can automate this based on identifiers? But they are currently listed in "related URL" and we will probably begin by migrating them?
    * Actually, the fact that not every collection has a public EAD in ArchiveSpace, means we'd probably need something manual to say "yes there's a public EAD to link to", so just manually listing this in related links is probably just fine for a while here -- oh wait, but would we want to link it on the *collection* instead? I think so.... 
* External Links/Other
* Related Works (still using this feature at present)

Use these for "Collection" "related url" too!

Some existing "related URLs" may be cleaned up or deleted, see https://github.com/sciencehistory/scihist_digicoll/issues/1776#issuecomment-1191823073.  
* We may want to get rid of those "provenance" links, perhaps hard-code that link in for any record with provenance info, if that's desired?
* All existing links to OPAC in "related URLs" can I think be deleted, they should all be automatically made from OPAC id?

For ways this might be displayed see #1777 , the "Learn More" section

This has been extracted from #1776 

See also #1803 -- we're not sure if Museum Exhibitions will end up part of this feature, or a different feature.

jrochkind:
@apinkney0696 @eddierubeiz @archivistsarah

Regarding links from Digital Collections to ArchiveSpace EADs....

Historically we haven't done this, for historical reasons. Recently, perhaps mainly/only in Bredig, we've started doing this -- listing links to the ArchiveSpace EAD on individual work records. 

eg https://digital.sciencehistory.org/works/1atx812

[...]

PR #1818, opened 2022-08-09
Collection#external_id attribute, move collection OPAC urls to bibID metadata

jrochkind:
- add Collection#external_id field
- show bibID as link to OPAC on collection page
- rake task to migrate Collection bib_ids

Rake task will actually remove all OPAC urls from Collection#related_url, recording them in Collection#external_id as bibIDs instead. 

Feels risky cause destroying data, but I tested in staging, data seems to have properly migrated. 

Part of #1804

### DIFF
--- a/app/components/collection_metadata_component.rb
+++ b/app/components/collection_metadata_component.rb
@@ -18,8 +18,21 @@ class CollectionMetadataComponent < ApplicationComponent
     @all_count ||= collection.contains.count
   end
 
+  # from related_url (legacy), or from our external_id with bib IDs in it.
   def opac_urls
-    related_url_filter.opac_urls
+    @opac_urls ||= begin
+      # bib_ids are supposed to be `b` followed by 7 numbers, but sometimes
+      # extra digits get in, cause Siera staff UI wants to add em, but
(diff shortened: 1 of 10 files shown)


jrochkind:
@eddierubeiz To be honest, I'm no longer even sure this is a good idea -- is it just making things more complicated/moving things around for no purpose? Was it good enough to just leave them in related_url and recognize OPAC URLs by hostname?

But meh, sometimes you just gotta move and see what happens, I'm kind of stuck in a second-guessing mode. And it will make it more consistent with Work.

jrochkind:
OK, after running migration on staging, did a spot check of a few collections to make sure they had their former OPAC URLs as bib_ids, and the bib IDs showed up as links on the show page. 

Also did a count, on staging after migration:
* 0 related urls, 45 bib_ids

On production (not exactly the same data)
* 46 related urls

OK, I'm feeling confident the data edit worked.

jrochkind:
OK, migration added that will automatically run data_fixes rake task on deploy... for better or worse, since it isn't exactly a migration. 

(I didn't bother to disable the progress-bar, we'll see how annoying it is, if we notice).

Issue #1776, opened 2022-07-11
New "Learn More..." Section on works page


apinkney0696:
Create a new section on the works page for links to the Institute's interpretive resources. 

Each item could have an image, the title of the item, and a category for each. The categories of resources we would want to link to include:
- Distillations Articles
- Distillations Podcasts
- Historical Biographies
- Storymaps
- Blog Posts
- Digital Exhibitions

Not sure exactly how we will become aware of our digital collections being used in new resources or if there is any way to automate this. Probably will want to create a workflow with content creators so that they alert us when the reference or use the digital collections works.

I'm also unsure how this would look in the back end. I wonder if we would want this to function similarly to how interviewer profiles are applied to OH works - staff could create one "Learn More" item and apply it to multiple works. 

See ticket #1777  for an illustration.

jrochkind:
My thinking was this would be an enhancement replacement to the current "Related Links" feature, not an additional new feature with "related links" remaining intact as-is. Does this match your thinking @apinkney0696?

The implementation is of course more than a new section on the page display, it also involves metadata modelling and entry for this new material. 

Is "Storymaps" a thing that already exists at the Institute?

I'm not sure that this is the right/best way to take care of Digital Exhibitions (or physical exhibitions too!). Just a "related links" feature won't give us a way to list/collocate all items from a given exhibition; I was thinking we'd want that, so would want to tackle this differently, something more like the Collection feature for Exhibitions, although it might not be exactly like the Collections feature.

jrochkind:
I'd like to do an analysis of the current URLs listed in "related links" to see what sorts of things are currently listed there; perhaps just listing the unique hostnames with counts would be sufficient to paint that picture, at least for first try.

jrochkind:
Per discussion, I think we will probably split this ticket in two -- one for an extended "related urls" feature to be tackled first, and another for "how will we deal with museum exhibitions" to be tackled later. 

To begin with though, I have some listing of how "related URL" field is currently used in our database:

Here are all all the hostnames found in current production related_urls

{"othmerlib.sciencehistory.org",
 "digital.sciencehistory.org",
 "archives.sciencehistory.org",
 "othmerlib.chemheritage.org",

 "www.sciencehistory.org",
 "guides.othmerlibrary.chemheritage.org",
 "www.encyclopedia.com",
 "archives.dickinson.edu",
 "catalog.hathitrust.org",
 "www.acs.org"}>

[...]

PR #1820, opened 2022-08-09
rake task to remove all Work#related_url to museum provenance note

jrochkind:
We decided not to use these here, we may put them on Collection level instead. https://github.com/sciencehistory/scihist_digicoll/issues/1776#issuecomment-1202840485


jrochkind:
tested on staging.

jrochkind:
Build seems to be failing due to some general unrelated platform problem, we'll see if it resolves itself tomorrow. 

> Err:5 http://azure.archive.ubuntu.com/ubuntu focal InRelease
>  Could not connect to azure.archive.ubuntu.com:80 (40.119.46.219), connection timed out

PR #1821, opened 2022-08-10
Rake task to remove all OPAC urls from Work#related_url

jrochkind:
We have previously confirmed in production that all of these are mirrored in Work#external_id bibIDs. (Well, all but one we are fine deleting anyway).

There are actually 4700 of these redundant OPAC urls! Cleared with @apinkney0696 that it makes sense to delete. 

Ref #1776 and #1804


jrochkind:
With DB migration to automatically apply.

Issue #1828, opened 2022-08-22
Low Priority: Edit Creator of Beckman records


apinkney0696:
Could I please get some help with a minor metadata edit so I don't have to go through and delete and replace one by one? For some reason, there are about 200 Beckman records that list the creator as "Beckman Instruments, inc." instead of "Beckman Instruments, Inc." which is the proper LC creator term (the only difference being the lowercase i). This means that there are two listings in the creator terms facet list. 

Could someone please replace the creator entry for the records in the following link with the correct "Beckman Instruments, Inc." LCNAF term:
https://digital.sciencehistory.org/catalog?f%5Bcreator_facet%5D%5B%5D=Beckman+Instruments%2C+inc

Thanks!!

apinkney0696:
And I just noticed there's the same problem for the subject headings too. Any change we can adjust both at once?

jrochkind:
This is definitely do-able, don't waste your time doing it by hand!

jrochkind:
@apinkney0696 In the subject heading, the lowercase `"inc."` is actually the one MOST used. 

2213 records with `"Beckman Instruments, inc."`, only 6 with `""Beckman Instruments, Inc."`

This made me curious to see what is auto-completed by our auto-complete intended to be from FAST. 

In creator, it seems to auto-complete the upper-case, `Beckman Instruments, Inc.`. 

In subject, it also seems to auto-complete the upper-case, `Beckman Instruments, Inc.`. 

OK, it's a mystery how all those lower-cased ones got in there then!

Just since this will in fact be changing the majority of beckman instrumetns subject headings, can you just confirm @apinkney0696  that this is desired and we should go ahead?

apinkney0696:
Yes, please! My guess is that the authority heading got updated after we described the collection or maybe that the Beckman collection was cataloged before the autocomplete function was added to the application.

jrochkind:
Could be that Beckman Instruments didn't have an authority record when we started, and we added it!

Anyway, ok, will do thanks.

eddierubeiz:
I just ran this rake task in production.

PR #1834, opened 2022-08-24
fix beckman heading data_fixes rake task

jrochkind:
Ref #1828


jrochkind:
tested on staging. 

note that you will need to run rake task after merge and deploy. `heroku run rake scihist:data_fixes:fix_beckman_heading -r production`. 

Note use in code of the very convenient `jsonb_contains` db query-er offered by kithe, that we hadn't previously been including/using.

eddierubeiz:
Yeah --  the `jsonb_contains` is in fact really great.

Issue #1317, opened 2021-09-10
Enhance "related items" feature (automatic similar/more-like-this from Solr)


jrochkind:
Over the past year, the manual "related items" feature seems to have been used a lot more than initially, often listing just other things with the same author. 

We could automatically generate "related items" based on similar metadata -- same creators and/or subjects for instance. This might be valuable if we think "related items" bring viewers to more items of interest and value to them, and want to avoid some of the tedious manual labor of manually listing things mechanistically that have the same author etc. 

We could use the Solr "more like this" feature to implement this. 

@MDiMeo just an idea

jrochkind:
(moving this back to top of backog, as we need more specs before 'ready')

apinkney0696:
Let's move ahead with this when we have the time! Automating related items based on subject headings (and maybe descriptions? other facets?) will be useful, I just want to make sure I can edit the related items and manually select related items when necessary. Any other considerations we need to think about?

jrochkind:
My thought for the simplest way to implement would be -- use automated related items only if the item has no manual related items specified. 

If there are manual related items, the automated related item finder isn't used. 

We'll still have to play with the solr feature and see if we can configure it to produce results we are satisfied with for automated related items at all!  That might be something @eddierubeiz finds engaging, although I'm happy to do it too.

jrochkind:
@eddierubeiz I'm very interested in the recipe:  subjects + creators + description

(Yes, including description text in there!)

PR #1836, opened 2022-08-29
Refactor of work indexer class

eddierubeiz:
This should simplify upcoming changes to the indexer code and make it easier to see what we're doing.
Ref https://github.com/sciencehistory/scihist_digicoll/issues/1317

### DIFF
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -1,71 +1,136 @@
 # Solr indexing for our work class.
 class WorkIndexer < Kithe::Indexer
+
+
+  # instead of a Solr field per attribute, we group them into fields
+  # that will have similar boosting, to use solr more efficiently.
+  # text1 is boosted highest, text2 next highest, etc.
+  field_map = {
+    :model_pk_ssi =>      ["model_pk_ssi"],
+    :title =>             ["text1_tesim"],
+    :additional_title =>  ["text1_tesim"],
+    :creator =>           ["text2_tesim","creator_facet"],
+    :genre =>             ["text2_tesim", "genre_facet"],
+    :subject =>           ["text3_tesim", "subject_facet"],
+    :interviewer =>       ["interviewer_facet"],
+
+    # index description to its own field for highlighting purposes. Fields with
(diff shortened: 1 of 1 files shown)


jrochkind:
Is a given `field_map[key]` used in more than one place? Can you give an example of where one is used in more than one place? Unless this happens a lot, I'm not totally seeing the advantage of the extra level of indirection with the hash...

Or wait, i see, it's tied into the "feature flag" constants you are using for differnet indexing styles.... So, you didn't htink "feature flags" are overkill here after all, that's basically what these are!   

Why not make it an `ENV` instead, so you can change it without having to make another commit and redeploy code? `ENV['MORE_LIKE_THIS_INDEX_MODE']` which can be `OFF`, `SIMPLE`, `MEDIUM`, or `FANCY`. (not sure these are the right terms, but ok!)

In general, if there is a setting that is mutually exclusive, make it ONE variable that can have four states -- now it can only represent one of four legal states. Instead of four booleans where you need to add an extra check to make sure only one is true, and no illegal state is set. 

In general, I see maybe how this is useful for your testing of four different settings. Although another approach would just be having four (or, really, three) separate branches for each setting. But if you think this is easier ot manage, that's fine. But I'm not in fact sure it makes sense to merge into master for perpeutity... once the testing is done, this seems like additional complexity/indirection that makes the code harder to read and harder to modify (when we add more fields) and is un-needed . That's just my first reaction

eddierubeiz:
Yeah, good reaction! I am not at all opposed to just reverting to a slightly modified version of the original indexer class once we're done with the feature. For the next week this refactor will be really helpful though.

jrochkind:
Well, can you just build your work on this temporary branch without merging it into master? Like do other PR's into this branch if you need to?

eddierubeiz:
Yeah, I can totally do that but I'd rather use an actual feature branch in that case! I can close this and start a feature branch.

jrochkind:
I think I don't understand what makes something an "actual feature branch" or not....

eddierubeiz:
Actually - no need to close this. We can just merge it into `more_like_this_feature` instead of `master`. But yeah, this is now ready for review.

PR #1838, opened 2022-08-30
Adding a more-like-this request handler

eddierubeiz:
Add a request handler to get the more-like-this results from.
Ref #1317


jrochkind:
Are you ready for me to merge this? The field configuration for mlt does not depend on what "strategy" you are using? ("simple" vs "advanced" or whatever you were calling them?).

eddierubeiz:
Yes absolutely. And yes, it doesn't depend on anything.

PR #1842, opened 2022-08-31
Switch Work and Collection display from using related_url to new related_link

jrochkind:
Including pulling out related_works and (now) finding_aid links for special handling. 

related_url had special handling of OPAC links, but this is no longer part of related_link, there's no category for opac links, opac links now just come from bibIDs in external_id. 

- don't show related_urls field, we're moving to related_links in other part of page
- get related works from new related_links rather than old related_urls
- show finding aid related_links on Work page in 'physical location' section
- collection show page use related_link instead of related_url
- remove now-unused ExternalLinkComponent
- remove now-unused RelatedUrlFilter -- replaced by RelatedLinkFilter
- remove outdated comment references to related_url


eddierubeiz:
Thanks for the tests.

PR #1847, opened 2022-08-31
Rake task to actually migrate related_url to related_link

jrochkind:
Also adds a new related_link category for distillations_video. 

    rake scihist:data_fixes:convert_related_url

Categories existing related_urls according to their urls, to make them related_links with proper categories. Adds labels for the identified existing related_urls that need em.

Should be idempotent, so we could for instance run in production multiple times if needed, and people are still editing.

PR #1855, opened 2022-09-06
rake task to migrate previously mis-migrated legacy related_works links

jrochkind:
What has become eg category:other_internal url: https://digital.sciencehistory.org/concern/generic_works/wd375w406 in database, should be category:related_work, url:https://digital.sciencehistory.org/works/wd375w406

Ref fixup of #1847

Tested in staging, seems to work okay. 

After merge and deploy:

    heroku run rake scihist:data_fixes:migrate_old_related_works_links

### DIFF
--- a/lib/tasks/data_fixes/migrate_old_related_work_links.rake
+++ b/lib/tasks/data_fixes/migrate_old_related_work_links.rake
@@ -1,0 +1,31 @@
+namespace :scihist do
+  namespace :data_fixes do
+    desc """
+      Migrate legacy related work links with /concern/generic_works/
+
+      They were migrated over to new related_links in somewhat corrupt fashion as other_internal
+    """
+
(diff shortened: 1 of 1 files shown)


eddierubeiz (approved):

jrochkind:
Weird, worked fine in staging, didn't seem to do anything in production. still investigating.

jrochkind:
Oh wait... may not have fully worked in staging either. Phew, this stuff is harder than it looks! May be able to get it to work just by re-running task multiple times.

jrochkind:
For some reason when the record contained multiple related_links that needed to be fixed, only the first one would be succesfully saved. 

Just gonna run the rake task 4-5 times in production, I don't think anything had more than 5 related urls! Kind of messy, but oh well. 

Confirmed with query afterwards that no more un-migrated bad /concern/generic_works data existed afterwards.

PR #1856, opened 2022-09-06
More like this feature

eddierubeiz:
This is pretty much done, so switching to "ready for review".


app/components/work_show_info_component.html.erb
```diff
@@ -197,3 +197,32 @@
...
+        <div class="related-work-text">
+          <%= link_to work_path(work.friendlier_id) do %>
+            <%= work.title %>
+          <% end %>
+        </div>
+      </li>
+    <% end %>
+  </ul>
+  <p>NOTE: the more-like-this section is just a demo. Presentation will change after we evaluate this feature.</p>
+<% end %>
```
jrochkind, line 222: We should DRY up the identical display of works for related items and more like this. In each case we have an actual work, can we make a component that takes a Work as an argument, and displays it in this fashion?

(In the future, I imagine enhancing this section so instead of a title, it may also shows other metadata more similar to a search results listing. Not part of this PR, but if we have a DRY component it's the place to do it).

spec/services/more_like_this_getter_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+# so we are short-circuiting that part of the code.
+describe MoreLikeThisGetter do
+  let(:work_to_match)   { create(:public_work)  }
+  let(:getter) {MoreLikeThisGetter.new(work_to_match) }
+  let(:three_public_work_ids) { [1,2,3].map {create(:public_work).friendlier_id } }
+  let(:matching_work_ids) { three_public_work_ids + [create(:private_work).friendlier_id]} 
+
+  it "returns only published works; displays them in order" do
+    allow(getter).to receive(:friendlier_ids).and_return(matching_work_ids)
+    expect(getter.works.map {|w| w.friendlier_id}).to eq (matching_work_ids[0..2])
```
jrochkind, line 12: looking around at the code, I see that this second line is meant to test that it only returned published works, not private ones. 

That is kind of hard to see from the code though. Is there a way to make it more apparent what the test is testing? The magic `matching_work_ids[0..2]` is cryptic; could it instead just be testing against a `public_works` variable or something?

It's also not clear to me what the "order" is -- is it the order that Solr returns it in, like most "similar" first?  How does this test test that?  Could it be made more clear in code how it's testing that, and/or comments, _what_ "order" is expected, in order of what?

The comments say "Testing the SOLR part of this is slow -- and a pain -- so we are short-circuiting that part of the code" -- I'm not sure what this means. This code isn't _really_ going to solr?

Hmm, can we find a way to actually go to solr?  What makes it a pain, that we aren't sure what solr will return as "more like this"?

I think it would be better if we had just one test at least actually going to solr.

app/services/more_like_this_getter.rb
```diff
@@ -0,0 +1,87 @@
+# MoreLikeThisGetter.new(work).works
+#
+# returns an array of zero to ten
```
jrochkind, line 3: Why ten? What determines that it's 10?  The actual Solr config? Something else?

We don't actually want 10 more-like-this works displayed. What part of the code is limiting to only 3 or X?  I am having trouble finding it. 

Would it make sense for hte MoreLikeThisGetter to actually take an argument to one of it;s methods asking for how many you want at most?  (If limiting to only, say, 3, it should be the 3 that Solr rates as _most_ alike. Do we have a way to make sure that's so, test it?)

app/services/more_like_this_getter.rb
```diff
@@ -0,0 +1,87 @@
...
+    @work = work
+  end
+
+  # Returns an array of up to 10 published works that SOLR deems similar.
+  def works
+    works_in_arbitrary_order = Work.where(
+      friendlier_id: friendlier_ids,
+      published: true)&.
+      index_by(&:friendlier_id) || {}
+    friendlier_ids.map {|id| works_in_arbitrary_order[id] }.compact
```
jrochkind, line 41: This will re-fetch the works from DB every time the method is called. Is that a problem? Is the method called more than once to display the objects? Should it be memo-ized to prevent that?

app/services/more_like_this_getter.rb
```diff
@@ -0,0 +1,87 @@
...
+    RSolr.connect( :url => solr_url, :timeout => TIMEOUT,:open_timeout => OPEN_TIMEOUT)
+  end
+
+  # Returns a RSolr::Response::PaginatedDocSet,
+  # or an empty array if RSolr times out or can't be reached.
+  def more_like_this_doc_set
+    @more_like_this_doc_set ||= begin
+      solr_connection.get('mlt', :params => mlt_params)&.
+        dig("response", "docs") || []
+    rescue
```
jrochkind, line 83: Broadly `rescue` with no Exception class mentioned can be too broad, and rescue things like syntax errors, making debugging harder. Better to rescue the solr-related exception classes that we actually want to rescue. 

Although that can be hard to figure out from Rsolr. But it's worth figuring out. 

Also, when the rescue happens -- it hsould be logged! `Rails.logger.error` with details in the log message (including friendlier_id). 

Also... is there any way to test this rescue, for at least one kind of error?

app/services/more_like_this_getter.rb
```diff
@@ -0,0 +1,87 @@
...
+    }
+  end
+
+  def solr_url
+    ScihistDigicoll::Env.lookup!(:solr_url)
+  end
+
+  def solr_connection
+    RSolr.connect( :url => solr_url, :timeout => TIMEOUT,:open_timeout => OPEN_TIMEOUT)
+  end
```
jrochkind, line 75: OK, the hardest part... I wonder if there'a way to re-use the Solr connection that Blacklight is already using, intead of re-opening one. 

I am worried that re-opening a new one won't always get it quite right, like will it get all password and proxy and other settings we may apply to blacklight right? It'll also maybe be less performant to re-open a new connection every time instead of re-using one; not sure, this may not matter, depends on what Blacklight is doing. 

We need to spend more time reverse engineering Blacklight and trying to re-use as much as possible of it's code for where/how to make/get a Solr connection, and make sure we're doing it exactly the same way in case where we write our own code. This is a pain to do, yes, Blacklight is confusing. One part (although I don't understand where it's called how!) is at: https://github.com/projectblacklight/blacklight/blob/v7.29.0/lib/blacklight/solr/repository.rb#L108

app/components/work_show_info_component.rb
```diff
@@ -128,6 +128,17 @@ def oral_history_number
...
+    @more_like_this_getter ||= MoreLikeThisGetter.new(work)
+  end
+
+  def more_like_this_works
+    more_like_this_getter.works
+  end
+
+  def more_like_this_json
+    more_like_this_getter.json
+  end
```
jrochkind, line 141: I don't love that the component is triggering the call to Solr, so a call to Solr happens when you ask the Component to display. 

In an ideal world, the call to Solr would happen in the controller, and the data would be passed down to the component from the controller, somehow... but there may not be a non-terrible way to do this. It's worth spending some time at least thinking about it though. 

If we can't find a feasible way to do it, comments at teh top of this component should clearly advertise that it triggers a network call to solr for more like this results, via the MoreLikeThisGetter object.

spec/services/more_like_this_getter_spec.rb
eddierubeiz: I was able to figure out how to get SOLR to give us more-like-this results in the test environment, so we do have this now.

app/services/more_like_this_getter.rb
eddierubeiz: I did memoize this, yes -- good call.

app/services/more_like_this_getter.rb
eddierubeiz: I did add a "maximum items" argument to the instantiation method for the getter object.

eddierubeiz: Oh, it's not obvious where the default value of 10 retrieved items is and how it might be possible to ask for fewer results to be returned. (No, you can't use a `mlt.count` parameter).

app/services/more_like_this_getter.rb
eddierubeiz: Fixed this and we are properly testing for it.

app/services/more_like_this_getter.rb
eddierubeiz: I think I have this figured out!

app/services/more_like_this_getter.rb
```diff
@@ -0,0 +1,121 @@
...
+    # is to provide two successive retries on failure, which we don't really want here.
+    @solr_connection ||= begin
+      repo_class = Blacklight::Solr::Repository
+      faraday_params = {
+        :timeout => TIMEOUT,
+        :open_timeout => OPEN_TIMEOUT
+      }
+      repo_class.new(CatalogController.blacklight_config).connection.tap do |conn|
+        conn.connection.params = faraday_params
+      end
```
jrochkind, line 79: Phew, tricky abstract blacklight stuff, nice job figuring it out. 

I think I'd use `Scihist::BlacklightSolrRepository` to get the retries, why don't we want the retries? 

Either way, to me this would be more clear without the local variables, just say what you mean all at once:

```ruby
Blacklight::Solr::Repository.new(CatalogController.blacklight_config).connection.tap do |conn|
   conn.connection_params = {
         :timeout => TIMEOUT,
        :open_timeout => OPEN_TIMEOUT
   }
end
```

But now I also wonder if we could skip the custom timeouts... why not just use the default solr connection with all it's retries and default timeouts, to make it as simple as:

```ruby
Scihist::BlacklightSolrRepository.new(CatalogController.blacklight_config).connection
```

Or if you really want the custom timeout and no-retry-behavior, ok, cool, as above.

eddierubeiz: Yeah, I feel like `more_like_this`  is nice to have, but if SOLR is slow or unavailable I want to prevent any slowdown in the display of the work show page.

jrochkind: ok fair!

app/components/work_show_info_component.rb
```diff
@@ -111,6 +115,12 @@ def related_works
       ).includes(:leaf_representative).all
```
jrochkind, line 115: good catch

app/components/work_show_info_component.rb
```diff
@@ -111,6 +115,12 @@ def related_works
...
   end
 
+  # This triggers a single request to SOLR to retrieve works deemed similar.
+  # This is unusual, but calling the code in the controller turned out to be unduly complex.
```
jrochkind, line 119: can you put a comment at the TOP of this component .rb file too, warning that rendering it will, unusually, trigger a solr fetch?

Other than that, approved!

PR #1861, opened 2022-09-15
Try to better auto-detect Rails development settings for app_url_base

jrochkind:
Our app needs to know the "base url". I realized we were defaulting to assume it was alwasy http://localhost:3000 in development. While you can always locally set APP_URL_BASE to something differnet, can we default to auto detecting if you for instance did `rails server -p 3002`, a totally normal and reasonable thing to do? And annoying if every time you change the port you need to change teh setting of a local APP_URL_BASE too?

I found a way to do it, although it's pretty hacky and uses maybe not entirely public Rails API. But maybe it'll work out?


eddierubeiz:
Yeah, not a good idea to have that hard-coded.

Issue #1758, opened 2022-06-27
Metadata Export API for Max Planck use case


jrochkind:
As a specific use-case for existing #201, we now want to focus on the Max Planck oral history project, create an API that they can use to get our oral history metadata. 

Treat them as a use case/user in making requirements and spec'ing out what we have (although it would ideally be available for all items not just OH, for #201).

jrochkind:
Draft public doc intro for API availability. Not sure where it will go. Re-written from [FAQ](https://digital.sciencehistory.org/faq) "Can I freely download or extract your metadata?". Being edited here for now. 

----

# API access

We strive to make our open data freely available, but the options we provide for machine-readable metadata access are currently limited. If you have a project that could benefit from more convenient machine-accessible APIs for metadata access, please [get in touch](mailto:digital@sciencehistory.org) to share your use case.

## OAI-PMH

We have an [OAI-PMH](https://www.openarchives.org/pmh/) feed which can give access to our metadata in an XML format. The fields are based on the [OAI-DC](http://www.openarchives.org/OAI/2.0/oai_dc.xsd) schema, with extensions suggested by the [DPLA](https://dp.la/) [metadata application profile](https://pro.dp.la/hubs/metadata-application-profile), as this feed's main use case is DPLA use. 

This metadata includes standardized basic descriptive attributes, but does not include all internal, administrative, and relational metadata.

You can bulk harvest via an OAI-PMH 2.0 endpoint at https://digital.sciencehistory.org/oai

You can also get an oai-dc XML representation for any record by adding `.xml` to the end of a record's URL. For instance, https://digital.sciencehistory.org/works/vt150j62m.xml.

## Atom feeds

Any search result is available in the Atom Syndication Format. Just add `.json` to the path of any search results, for instance:
https://digital.sciencehistory.org/catalog.atom?q=chemistry

* These Atom results are paginated, note the pagination links at top
* Individual entries include, title, thumbnail, brief description, and link to HTML page
* For Works, there are also <link> entries to metadata in OAI-DC XML and local json formats

You can also access atom search results within any collection. This includes listing all items in a collection, for instance for the Oral History Collection: https://digital.sciencehistory.org/collections/gt54kn818.atom

## Individual Work metadata

For every "work", you can access metadata in an XML/OAI-DC format, or a local internal JSON format. 

[...]

PR #1862, opened 2022-09-19
Work metadata JSON response

jrochkind:
Not json:api standard, just an in-house json for now. 

Ref #1758


jrochkind:
OK after this PR, a JSON response (with local/custom/non-standard JSON metadata) is available by adding `.json` after a work url, such as eg `https://staging-digital.sciencehistory.org/works/fagrxsp.json`

Some refactoring of existing code to make things neater and DRY.

spec/factories/work_factory.rb
```diff
@@ -217,7 +217,8 @@
 
       oral_history_content {
         OralHistoryContent.new(
-          interviewee_biographies: [build(:interviewee_biography)]
+          interviewee_biographies: [build(:interviewee_biography)],
+          interviewer_profiles:    [build(:interviewer_profile)]
```
eddierubeiz, line 221: Thank you for shoring these factories up a bit.

PR #1863, opened 2022-09-20
Re-enable atom results, with a custom overridden implementation

jrochkind:
Ref #1758


jrochkind:
oops replaced by #1865

PR #1865, opened 2022-09-22
Atom format search results are back

jrochkind:
Ref #1758 

We restore Atom formatted search results -- we don't get the Blacklight ones working again, instead we provide our own Atom results implementation, that displays based on the ActiveRecord models (kithe-style), instead of just Solr response documents (Blacklight style). 

While we potentially could have used the Blacklight framework to config a custom "document partial" for each atom result, to keep using the overall Atom template from Blacklight... this was hard to figure out how to do, would use a lot of Blacklight apparatus we don't actually want (like an additional partial template, when we really want to use a ViewComponent)... and the overall atom template is actually fairly simple. So we just copy-paste-and-modify the whole views/catalog/index.atom.builder partial to our local app, to override the blacklight one. 

Note it uses the `.builder` format for the template instead of `.erb`. (Our AtomEntry view component does too!).  That's convenient for XML. 

The Atom response includes <link>s to the .json and .xml (oai-dc) metadata responses for each Work result. This lets these all be used in concert as a kind of (Frankensteinian) API for getting lists of results, and then getting their metadata (in oai_dc xml or local custom json). Note the Atom search results are paginated, and include pagination <link>s in the <feed> too. (Blacklight already did this; I may have written the implementation myself a decade ago!)

We don't have .xml or .json responses for Collections. Collections can appear in Atom search results, they just won't have these <link>s to get machine-readable metadata, at present. Not part of our current use case, and kind of a pain for Reasons.

### DIFF
--- a/app/components/atom_entry_component.atom.builder
+++ b/app/components/atom_entry_component.atom.builder
@@ -1,0 +1,19 @@
+xml.entry("xmlns:media" => "http://search.yahoo.com/mrss/") do
+  xml.title model.title
+
+  xml.updated (model.updated_at&.iso8601 || Time.current.iso8601)
+
+  xml.id work_url(model)
+  xml.link "rel" => "alternate", "type" => "text/html", "href" => work_url(model)
+
+  xml.link rel: "alternate", type: "application/xml", title: "OAI-DC metadata in XML", href: work_url(model, format: "xml")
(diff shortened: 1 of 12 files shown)


jrochkind:
@eddierubeiz Oops, I thought you had already reviewed this one too, but am going to merge it anyway right now, to get it live so I can email our external partners with it live. 

This is a somewhat big one, so you might want to review it anyway even though I've already merged it? We can always make more changes.

PR #1866, opened 2022-09-22
A couple additional tweaks to Work public json representation

jrochkind:
Ref #1758 

- change name to WorkJsonSerializer since we didn't go with json:api
- add updated_at to Work json serialization
- dates in JSON serialization in UTC


jrochkind:
@eddierubeiz I'm going to merge this myself without your review, so I can get it live before I show API to external partner.

eddierubeiz:
All makes sense!

Issue #1817, opened 2022-08-08
` pg_restore` fails in staging upon attempting to restore from a copy of the production database.


eddierubeiz:
Run `heroku pg:backups:info c044` on the staging app for more details.

The first error (which may or not be relevant) is:

```sql  pg_restore: creating EXTENSION "plpgsql"
 pg_restore: creating COMMENT "EXTENSION "plpgsql""
 pg_restore: while PROCESSING TOC:
 pg_restore: from TOC entry 4148; 0 0 COMMENT EXTENSION "plpgsql"
 pg_restore: error: could not execute query: ERROR:  must be owner of extension plpgsql
 Command was: COMMENT ON EXTENSION "plpgsql" IS 'PL/pgSQL procedural language';
```
Another similar error follows (`must be owner of extension pg_stat_statements`)

Then, pg_restore has trouble installing `pgcrypto`.

Finally, our own PL/SQL code to set up `gen_random_uuid` fails.

To reproduce:

```sh
heroku pg:copy scihist-digicoll-production::DATABASE_URL \
DATABASE_URL -a scihist-digicoll-staging  \
--confirm scihist-digicoll-staging
```

(This is the first step of our regular routine to copying the production database to staging.)

eddierubeiz:
Heroku support got back to us with:  ```We are still working on a fix for this issue, but in the meantime, we have seen the following workarounds temporarily unblock customer's workflows: [...]```

So for now: just wait.

eddierubeiz:
Hi there,

Thanks for reaching out to us and we’re very sorry for the inconvenience caused here.

We are still working on a fix for this issue, but in the meantime, we have seen the following workarounds temporarily unblock customer's workflows:

**Manually migrating Postgres plugins by:**

- pulling the affected database into a local database
- manually migrating the relevant extensions locally
- creating a new database on the app
- pushing the local DB with migrated extensions into the new database
- promoting the new database

**Manually updating schema references. For example:**

Changing all instances of `WITH SCHEMA public;` to `WITH SCHEMA heroku_ext;` and all instances of `DEFAULT public` to `DEFAULT heroku_ext` by using find & replace in a text editor or by changing the output of pg_dump using `sed`

While these workarounds are only temporary they may help to unblock you for the moment. We apologize again for the inconvenience caused and as we work to remediate the issue, we will update you with more information and the next steps.

Please let us know if you have anything else to ask.

Regards,
CJ

jrochkind:
[...]

PR #1868, opened 2022-09-26
[WIP] [Do not review yet] Move extensions to heroku_ext schema

eddierubeiz:
Ref #1817

From now on in Heroku,

"All new extensions will be installed in the heroku_ext schema (except for the list below). You will not be able to install new extensions in other schemas via CREATE EXTENSION ... WITH SCHEMA for new or existing databases."

This means that

1. currently, without the `--extensions` flag we can't restore from a backup of production using `heroku pg:backups:restore`. If you use the standard recipe described in e.g. https://devcenter.heroku.com/articles/heroku-postgres-backups#restore-a-backup, **chaos ensues**.
2. anytime we want to set up a *new* environment in Heroku, we will have to work either with or around the diktat: new extensions go in `heroku_ext`.

In a[ recent FAQ](https://help.heroku.com/ZOFBHJCJ/heroku-postgres-extension-changes-faq ), Heroku support makes the ambiguous statement: "The goal moving forward is for your database and schemas to exist in the same state across all environments of an applications pipeline. This "local consistency" will look different for your individual applications depending on a number of factors and will involve using either the public schema, the heroku_ext schema or a combination of both depending on what fits your needs."

Needless to say, the "depending on a number of factors" clause is less than helpful here.

We turn to [a blog post by Justin Searls](https://blog.testdouble.com/posts/2022-08-15-migrating-postgres-extensions-to-the-heroku_ext_schema/) for some ideas. Searls explains his basic approach as follows: "after attempting a few workarounds, it became clear that the best way out of this problem was through", meaning that the author feels the benefits of permanently migrating extensions to `heroku_ext` in dev, staging and prod outweigh the drawbacks.

I'm not convinced that this is a the right approach, but I wanted to pull together a proof of concept so we have at least one alternative to compare the status quo. This is just a very rough draft, and it contains some ugly hacks at the moment.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,3 +4,26 @@
 require_relative 'config/application'
 
 Rails.application.load_tasks
+
+# An ugly hack inspired by:
+# https://blog.testdouble.com/posts/2022-08-15-migrating-postgres-extensions-to-the-heroku_ext_schema/
+#
+# The goal here is to have `rake db:setup` create a database that has pg_stat_statements and pgcrypto
+# installed in heroku_ext.
+#
+# See https://help.heroku.com/ZOFBHJCJ/heroku-postgres-extension-changes-faq
+# We want to do this *without* modifying `db/schema.rb` directly, since schema.rb is created automatically by
+# bin/rails db:schema:load .
+Rake::Task["db:schema:dump"].enhance do
+  filepath = "db/schema.rb"
+  schema = File.read(filepath)
+  schema = schema.gsub(
+  	/enable_extension "pg_stat_statements"/,
+  	'execute("CREATE SCHEMA IF NOT EXISTS heroku_ext;\nCREATE EXTENSION IF NOT EXISTS pg_stat_statements WITH SCHEMA heroku_ext;\n")'
+  )
+  schema = schema.gsub(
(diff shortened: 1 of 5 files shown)


eddierubeiz:
The goal for this approach is to have following extensions in the following schemas in dev, staging and prod:

```
# \dx
                                            List of installed extensions
        Name        | Version |   Schema   |                              Description
--------------------+---------+------------+------------------------------------------------------------------------
 pg_stat_statements | 1.9     | public     | track planning and execution statistics of all SQL statements executed
 pgcrypto           | 1.3     | public     | cryptographic functions
 plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language
(3 rows)
```

If that assumption is true, we should be able to run some variation on:
```
dropdb scihist_digicoll_development
bundle exec rake db:setup
bundle exec rails db < latest_prod_backup.sql
```
but also, in both staging and prod,
```
heroku pg:psql < latest_backup.sql
```

eddierubeiz:
We may still use a variation of this approach to fix #1817, but now that `pgcrypto` is gone from production and staging (see https://github.com/sciencehistory/scihist_digicoll/pull/1874 ) I'm closing this as it's liable to cause confusion.

Issue #1869, opened 2022-09-28
atom results for a collection do not work


jrochkind:
(at least for oral history collection? May be specific to our singleton-controller collections?)

```
ActionController::UrlGenerationError: No route matches {:action=>"opensearch", :collection_id=>"gt54kn818", :controller=>"collection_show_controllers/catalog", :format=>"xml"}
```

## Backtrace

    line 49 of [PROJECT_ROOT]/app/views/catalog/index.atom.builder: block in _app_views_catalog_index_atom_builder___2453149386422585355_31031960
    line 19 of [PROJECT_ROOT]/app/views/catalog/index.atom.builder: _app_views_catalog_index_atom_builder___2453149386422585355_31031960
    line 26 of [PROJECT_ROOT]/app/controllers/application_controller.rb: block in batch_kithe_indexable

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/88741185)

PR #1870, opened 2022-09-28
Fix atom results for custom controller collections

jrochkind:
By elminating attempt to link to (broken) opensearch description for all collections. Even in cases where it was linking to something, it was not a correct opensearch description, for our custom use of searchable collection controllers

Ref #1869


eddierubeiz:
Noted!

PR #1871, opened 2022-09-28
Limit request rate with rack-attack

jrochkind:
* in production, set up Rails.cache to use heroku memecachedcloud plugin, if available. Rack attack needs a cache, and uses Rails.cache by default
* set up rack-attack to limit to 2 req per sec over a minute
* also log at a lower limit, without limiting
* some kind of fancy obtuse code to keep from logging *every* time a request is blocked or tracked, which is what rack-attack does by default. We only want to log once per "block period'.  https://github.com/rack/rack-attack/discussions/592

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -167,6 +167,8 @@ gem "irb", ">= 1.3.1"
 gem "reline", ">= 0.2.1"
 gem "warning", "~> 1.2" # managing ruby warning output
 
+gem "rack-attack", "~> 6.6" # throttling excessive requests
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'pry-byebug', platforms: [:mri, :mingw, :x64_mingw]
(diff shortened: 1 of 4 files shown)


jrochkind:
Yes, there are no tests here, it's a pain to test this kind of thing and I don't think it's worth it. I have manually tested.

jrochkind:
papertrail configured to send alerts hourly to slack on `rack_attack, so log messages of throttling or rate limit monitoring. 

Not totally sure if I configured this just in production or if it applies to both.

jrochkind:
Using the heroku `Memcached Cloud` add-on to power Rails.cache. 

Added a free one in staging, for now. 

Added a $14/month 100MB one in production. (Smallest non-free one; you can only use one free one per account). 

Documented at https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/1915748368/Heroku+Operational+Components+Overview#Add-ons

jrochkind:
Just realized this might mess up the API access we just rolled out!

I'm going to exempt .xml, .json, and .atom links from the rate-limit, although not from the reported tracking.

Issue #1764, opened 2022-06-29
upgrade postgres to pg 14


jrochkind:
We currently accidentally have different pg versions in staging and prod:  13 in staging and 12 in prod. 

Postgres 14.x is the latest supported by heroku. We should upgrade both prod and staging to PG 14. 

This may involve some downtime. 

https://devcenter.heroku.com/articles/upgrading-heroku-postgres-databases

eddierubeiz:
```heroku pg:copy scihist-digicoll-production::DATABASE_URL DATABASE_URL -a scihist-digicoll-staging  --confirm scihist-digicoll-staging``` is currently failing, which may well be connected to this issue.

jrochkind:
Let's actually prioritize this... getting to latest postgres may avoid the need to install pg extension for UUID, which is what gets is in trouble at #1817. 

If we can get to latest supported pg, we may be able to fix our `schema.rb` file to no longer try to enable the `pgcrypto` extension, which may avoid the issue at #1817.

eddierubeiz:
Just noting we upgraded staging to 14 last night without issue, and plan to do production next Monday, Oct. 3rd, in the late afternoon.

PR #1874, opened 2022-09-30
[WIP] [Do not review yet] remove pgcrypto

eddierubeiz:
It does appear that it may be possible to drop `pgcrypto` entirely from our database once we move to postgres 14 in staging and prod.
Do not run this migration on postgres < 14.
Ref #1817 
Ref #1764

### DIFF
--- a/db/migrate/20220926155354_move_extensions_to_heroku_ext.rb
+++ b/db/migrate/20220926155354_move_extensions_to_heroku_ext.rb
@@ -1,0 +1,18 @@
+class MoveExtensionsToHerokuExt < ActiveRecord::Migration[6.1]
+
+  def up
+    execute <<~SQL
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Let's edit .github/workflows/ci.yml ...

eddierubeiz:
And that worked.

jrochkind:
Ooh very nice, thanks! I wasn't sure the "ALTER TABLE" was necessary, thanks for figuring that out.

jrochkind:
Maybe you could add a comment to the migration explaining what is going on, but I don't know how likely anyone is to ever look at the migration again, so whatever.

jrochkind:
And/or a comment in the commit message would have been another way to document it, although again not sure if anyone would ever find it.

PR #1880, opened 2022-10-12
Update rubygem dependencies

jrochkind:
Before starting work to start using vite.js, I felt like updating dependencies. 

Update from Bootstrap 4.3 to 4.6 justifies some extra time manually reviewing to make sure our styling is still ok.


jrochkind:
We also update `redis` gem from 4.x to 5.x here... also could potentially trigger bugs in resque etc. This update PR ends up being  potentially risky in a few ways so we should be careful... but also might as well get it done, I think?

jrochkind:
Everything seems to be working testing manually on staging, but all this upgrading makes me a bit nervous. 

I am not sure if we should try to rope anyone else in for manual testing, any thoughts @apinkney0696 and @eddierubeiz ?  We could talk about it Monday, but I'm recalling @apinkney0696 won't be at meeting on Monday.

PR #1881, opened 2022-10-13
Rack-attack re-configure

jrochkind:
The code is a bit convoluted, sorry. Once a day per IP trigger a different log line that says `rack_attack: ALERT`, so we can configure papertrail alerting only on that, not on every rack-attack log line. 

Also bump up rack attack limits just a bit. 

- increase rack-attack limits
- speical alert-triggering rack-attack logging
- that request_id is not Rails request ID and is not useful


jrochkind:
Still no tests, this is kind of hard stuff to test, although it's also kind of weird code, so that's not great. But I did test manually in dev.

PR #1882, opened 2022-10-13
Vite

jrochkind:
* Totally replace webpacker with vite
* Move *most* JS that was previously in sprockets also over to vite -- only a little bit of JS remains in sprockets.  (See #1889)
* Remove some legacy polyfill-for-IE JS that is no longer needed. 

This PR will leave CSS and static file assets (like images) in sprockets for now. 

Also we will leave *blacklight* JS in sprockets for now, even though theoretically `blacklight-frontend` would allow us to shift it over, there are just some complexities we're going to handle separately. Including what it's dependencies do to our node_modules tree (https://github.com/projectblacklight/blacklight/issues/2849), and the desire to move CSS over to vite too, so we can have Blacklight CSS and JS both from the same `blacklight_frontend` version. 

This will, hopefully, leave us in a place no worse than our previous webpacker situation  (no more confusing or hard to maintain), so can be merged and deployed, and even keep living like this for a while.


jrochkind:
Vite-rails has some special instructions for deploying with heroku... but they don't appear to me to be necessary right now. And based on what they do, it definitely seems better to avoid them if not necessary. 

https://vite-ruby.netlify.app/guide/deployment.html#using-heroku

jrochkind:
I was having some mysterious problems that went away with `rm -rf node_modules ; yarn install` -- it's possible you'll have to do that in development.

jrochkind:
While we will leave this in draft mode while we do more vite work, this initial step is ready for a review, @eddierubeiz! And would be convenient to get one before i start doing more vite work. 

Note: * While things should work without it, you may choose to run the vite dev server in development. Just run: `bundle exec vite dev` in it's own terminal window. If you are working on modifying JS, this may give you quicker iterative builds and better error messages.

eddierubeiz (approved):

Issue #1872, opened 2022-09-28
Google Classroom Integration


apinkney0696:
**I'm pasting an email thread with Erin M. Below. Let's chat about this during our meeting on Monday.**

From: McLeary, Erin <EMcLeary@ScienceHistory.org>
Sent: Wednesday, September 28, 2022 3:53 PM
To: Pinkney, Annabel <apinkney@sciencehistory.org>
Subject: interesting (to me) examples of Google Classroom integration
 
Hi Annabel, 

A question for the digital tech team! I've been seeing some websites of organizations similar to us offering integration with Google Classroom, so a teacher who wants to assign a resource from that site can do so directly from the resource page. Here's an example from [DPLA ](https://dp.la/primary-source-sets/space-race)(under "cite this set") and here's an example from [Science Buddies](https://www.sciencebuddies.org/news/article?id=1261752) (top right corner, green and yellow icon). I'm curious about you all's opinion on how difficult it is to add this sort of integration, if this is something we might be interested in doing down the road. I like the Science Buddies example because it's not a full lesson plan (like the DPLA example), it's just an article that an educator might want to assign or use. It made me think of your work to add related resources to certain items in the digital collection and how something like Google Classroom integration could be very appealing for an educator looking at such a set of related resources. (Of course, the educator has to find our resources, but we can disaggregate the PR challenge from the technical challenge for the moment!)

This is just my curiosity at this point, but I'm interested in any thoughts the digital tech team has!

--Erin
__________________________________

From: Pinkney, Annabel <apinkney@sciencehistory.org>
Sent: Wednesday, September 28, 2022 4:08 PM
To: McLeary, Erin <EMcLeary@ScienceHistory.org>
Subject: Re: interesting (to me) examples of Google Classroom integration
 
Hi Erin,

What a neat idea! I will put in a GitHub ticket now and discuss this with the tech team at our weekly check-in meeting this coming Monday and get back to you with some initial reactions. 

To make sure I understand, Google Classroom seems to be a way for educators to compile online resources in one place to direct their students. If we were to enable Google Classroom integration, then our record pages would be able to be linked from these classroom sites. Let me know if this sounds accurate to you.

Best,
Annabel

Annabel Pinkney, M.L.I.S.   (she, her, hers)
Digital Collections Librarian
Othmer Library of Chemical History 
t. +1.215.873.8216

__________________________________
Erin McLeary   (she, her, hers)
Director of Strategic Initiatives
Office of the President 
[...]

PR #1888, opened 2022-10-17
Share to google classroom

eddierubeiz:
Ref #1872

The design is still very much in draft form. I'm using an inline icon for now, just to make editing easier. The geometrical paths in the svg tag are taken from edu.google.com: [Image](https://edu.google.com/images/svg/gsuite-icons/classroom.svg). Use "view source" to see the geometry of the image.


jrochkind:
Let's leave the icon an SVG if we can, so photoshop wouldn't be the right tool. 

I was imagining simply taking a single-color black google classroom SVG icon I found, and changing it's color to our brand dark grey?  Let's see if I can find that again... Ha, now I can't! But I think it's out there somewhere.

eddierubeiz:
Erin's notes in response to a first draft:

- What do you think of using the yellow and green version of the icon? As a google classroom user, the grey version didn't visually scan "Google Classroom" to me and the logo itself is so tiny I'm not sure I would recognize it without the colors (as compared to the larger logos on the social media buttons, which are much easier to recognize).

- When I click the icon it does take me through the steps to create a post in my classroom, but the object itself is not linked.

jrochkind:
> When I click the icon it does take me through the steps to create a post in my classroom, but the object itself is not linked.

OK, we tried to do the same thing as Erin's DPLA exmaple she gave us: https://dp.la/primary-source-sets/space-race.  So... this is disturbing if it's not working right. 

Can we ask Erin to test DPLA again, and see if it works as she expected, and if our demo is working the same or differnet as the DPLA example?  we tried to copy the DPLA example she gave us, so if it's not working the same as the DPLA example, we didn't succeed, and have to figure out why. But if it is working the same as the DPLA example, but it turns out the DPLA example isn't in fact a good example, it doesn't do what is needed... we'd have to do a lot more work to figure out how to integrate. 

Let's get the integration working in a way we think will satisfy users, before worrying more about how/where it goes on page.

eddierubeiz:
Erin has confirmed the share button is working as it should, by the way.

jrochkind (approved):

eddierubeiz:
Annabel & Erin agreed on this as good enough to deploy, with the understanding that we can always change the design later if we don't like it:
Final version

PR #1891, opened 2022-10-18
have to update scout_apm gem to support redis 5, that we already upgraded to.

jrochkind:
Without this upgrade, we get lots of errors in log from scout

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -578,7 +578,7 @@ GEM
       sprockets (> 3.0)
       sprockets-rails
       tilt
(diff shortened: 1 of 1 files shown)


jrochkind:
@eddierubeiz I'm going to go ahead and merge this one myself and do a quick deploy. The old version of the scout gem is resulting in lots of errors on heroku deploy, which are annoying even if they aren't errors in actual run.

jrochkind:
In fact, I think I'm even going to do it without Github CI reporting back, as it seems to be oddly stalled.

Issue #1893, opened 2022-10-18
hirefire can't fetch worker queue data after dependency upgrades


jrochkind:
After #1880, hirefire.io auto-scaling service is complaining:

   Couldn't fetch queue data. An error occurred 

Only thing in logs is (and lots of em) 

    Oct 18 10:41:46 app/web.1 2022-10-18 10:41:46 -0400 Rack app ("GET /hirefire/[secretkey]/info" - (44.192.8.45)): #<NoMethodError: undefined method `+' for nil:NilClass>

(wish we actually had a backtrace!)

I suspect the upgrade to `redis` gem may have done it, may try downgrading. 

Have also already reported to hirefire support.

jrochkind:
Downgrading redis gem to 4.8.0 per associated PR did seem to resolve the hirefire problem. Have reported problem to hirefire. We'll leave redis gem locked at 4.x until hirefire resolves.

PR #1894, opened 2022-10-18
try rolling back redis to 4.x to fix hirefire

jrochkind:
Ref #1893


jrochkind:
Downgrading redis gem to 4.8.0 per associated PR did seem to resolve the hirefire problem. Have reported problem to hirefire. We'll leave redis gem locked at 4.x until hirefire resolves.

PR #1897, opened 2022-10-19
allow redis to redis 5 again

jrochkind:
with temporary hirefire branch that works with redis5

hirefire believes they have tested this branch and it works, but has asked us to test before they do a full release. While I don't love testing on production (we don't run hirefire on staging), I guess I'm willing to. At worst, the hirefire auto-scaling will break again, until we roll back again, like it did yesterday.


jrochkind:
Seems to be working. I even artificially created some queue events to trigger a worker autoscale from hirefire. Everything seems to be working fine!

Let the hirefire guy know; when there is a new release of hirefire-resource, we will switch back over to that, instead of the branch from github.

PR #1901, opened 2022-10-20
update version of intermediate JS dependency @xmldom/xmldom

jrochkind:
To resolve security warning.

Did it by manually removing the '@xmldom/xmldom' section from `yarn.lock`, then running `yarn install`, and committing changes to yarn.lock file. (Only change was correctly updating xmldom version).

That seems to be the only way to update intermediate dependencies with yarn, which is weird.

Not sure why dependabot couldn't do this update itself.


eddierubeiz:
Thanks for looking at this.

PR #1905, opened 2022-10-24
Add oral_history_assets to work json api response

jrochkind:
Instead of providing a general solution for API access to work members, we just go with a kind of hacky custom-fit solution that meets Max Planck needs for oral histories. Just for OH's, we put transcript and front_matter PDFs (identified by asset #role) info in the respresponse, for now only for public assets, including a download URL. (local app url that will redirect to S3).

We also include the original_filename... testing this out for expansion to non-public assets, we might just include metadata including original file name (but no download url), to help Max Planck cross-reference to files supplied via another source.

Ref #1758


jrochkind:
Realized an id (friendlier_id) for each asset in JSON response should really be present, to allow consumers to keep track of what they have, and be able to report any problems or errors back to us referring to a specific known thing.

PR #1911, opened 2022-10-25
try using custom github action to install Apt packages; with caching, hopefully faster

jrochkind:


jrochkind:
**Does** seem to speed things up!  

Even before the cache exists, just by downloading and using the `apt-fast` tool, is avoiding problems we are currently having with ordinary `apt-get` having some kind of network problem, where it is taking it forever to get bytes of downloaded packages. 

However, this third-party github action package triggers github warnings. :( https://github.com/awalsh128/cache-apt-pkgs-action/issues/58

Not sure if we should actually use until that goes away; otherwise, we could find our builds broken when github finally removes this func, which the warning says will happen "soon". 

See also #2087 where we update our own use of actions/cache to latest version without that same deprecation.

jrochkind:
on the other hand, currently normal `apt-get` is crawling so slow it's hard to use, like taking 10-20+ minutes trying to install apt dependencies.  Maybe we gotta switch to this regardless of current deprecation warnings?

jrochkind:
The upstream provide of the `awalsh128/cache-apt-pkgs-action` action has updated it to no longer produce github actions deprecation notices. 

Looking at recent past runs prior to this PR, the "install apt dependencies" step can take between 30-60 seconds. With this PR, it seems to be more reliable and quicker at around 10 seconds. 

So, seems good.

eddierubeiz:
Fantastic.

PR #1917, opened 2022-10-26
Survey sparrow widget

eddierubeiz:
I haven't spent a ton of time testing this but I wanted to get some feedback on you before too long, @jrochkind . Annabel is fine with the basic behavior, so I'm requesting tech (for me) and UI (for Annabel) feedback now. After we have consensus on all that, I still would like to request a couple testing iterations in staging before we actually go live with this.

- Even after going over the documentation carefully and tweaking a bunch of settings, I still had a lot of trouble with the survey showing up repeatedly after I filled it out, so I decided to try setting our own `surveyAlreadyShown` cookie and checking it before triggering SurveySparrow's JS code.

- We also decided on a semi-transparent bluish background instead of the default black background that hides the d.c. site entirely. The color is somewhat arbitrary; feel free to suggest another.

- The js launch code is de-minified and included in our own `survey_widget.js` file. I could see a going with a couple other variations on this but for now, as a proof of concept, this seemed like a decent choice.


app/frontend/javascript/survey_widget.js
```diff
@@ -0,0 +1,88 @@
...
+  event.preventDefault();
+  var expiratonStr = new Date(new Date()
+    .setFullYear(new Date()
+    .getFullYear() + 3))
+    .toString();
+  document.cookie = "surveyAlreadyShown=true; path=/; expires=" + expiratonStr;
+  window.console.log("Set the cookie!");
+}
+
+// This is taken from the survey sparrow site.
```
jrochkind, line 51: Can you provide a URL, where did you get this from?  Among other things, if it *changes* in the future from SurveySparrow's end, we want to be able to track down where to get a new version. 

Is this actually a supported thing to do, by SurveySparrow , to copy this code from them? Or an unsupported hack? If the latter, I don't love it... there's no supported by SurveySparrow way to do what we need? Can we think about if there is a different way to do this supported by SurveySparrow? If not, we should definitely file an issue with them telling them what we want to do; and there should also be comments here explaining what we're doing.

app/views/application/_survey_widget.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if current_user%>
+	<div class="alert alert-primary">Survey hidden because you are logged in.</div>
+<% elsif  Rails.application.config.hide_survey_sparrow %>
+	<div class="alert alert-primary">Survey hidden because hide_survey_sparrow is turned on.</div>
+<% else %>
+<div id="ss_survey_widget"></div>
+	<script id="SS_SCRIPT">
+		// moved this to the js file
+	</script>
```
jrochkind, line 9: Does this empty <script> block need to be here then?

jrochkind: The point of this file is basically just to put a `<div id="ss_survey_widget"></div>` target on the page? 

Do we really need/want those alerts in there? Or are they just temporary for debugging?  I don't think we want actual public users seeing the alert `Survey hidden because hide_survey_sparrow is turned on`... _ever_, right?

If this is going ot be just a single div, does it really need to be a separate partial?

jrochkind: I wonder if there's a way to only put the div on the page for users who have not already taken the survey, so we don't need custom JS?

app/frontend/javascript/survey_widget.js
jrochkind: Generally linking to relevant Survey Sparrow docs URL in a comment _somewhere_ in this PR  is also probably a good idea. As well as putting it in the Github PR descirption itself; if you disappear and someone who isn't you (or me!) has to maintain this, leave them cookie crumbs to figure out what they're doing.

app/frontend/javascript/survey_widget.js
```diff
@@ -0,0 +1,88 @@
...
+// Despite our best efforts,
+// the survey was not registering
+// the fact that users had dismissed it
+// or already filled it in.
+
+// Note that SurveySparrow has its own
+// mechanisms in place to attempt to ensure
+// that users only fill out the survey once;
+// we just don't trust them.
+jQuery( document ).ready(function() {
```
jrochkind, line 16: Ideally we try to avoid using jQuery in code we write, becuase of a fantasy of eventually getting rid of it.

Not sure how long this code will end up staying around for... but sometimes even if we think short term, we end up keeping it for a while. Can we rewrite this without jQuery-isms? I don't think it uses many. Note we already have a `domready` thing that we use in other places to replace the `ready` itself. Then you just need to use built-into-browser function to find the `#ss_survey_widget` element, maybe that's it?

jrochkind:
@apinkney0696 @eddierubeiz 

Is the _only_ way to not see the survey presented on every page you visit to take it? Ethically, I would argue we really need to give a user a way to say "no thanks" and not see the survey in their way on every single page view, if they don't want to take the survey. If I were an IRB board, that's what I'd say. 

How long do you plan to leave the survey active?

app/frontend/javascript/survey_widget.js
```diff
@@ -0,0 +1,88 @@
...
+function sparrowLaunch(opts) {
+  var e = "ss-widget",
+    t = "script",
+    a = document,
+    r = window,
+    l = localStorage;
+  var s, n, c, rm = a.getElementById('SS_SCRIPT');
+  r.SS_WIDGET_TOKEN = "tt-5zUYbALWMcKM1wqSzFDcLi";
+  r.SS_ACCOUNT = "sciencehistory.surveysparrow.com";
+  r.SS_SURVEY_NAME = "digitial-collections-survey";
```
jrochkind, line 61: You want to hard-code these in here?  Not sure what the alternative is honestly. but it seems weird. Seems at least you oughta put them in a JS `const` at top of file, with comments explaining where they come from. 

Do we plan to take out this code entirely when survey period is complete? Or are we going to leave it in for future survey use?

config/application.rb
```diff
@@ -77,9 +77,14 @@ class Application < Rails::Application
...
     # Show the GDPR "I accept" cookies banner by default.
     # This setting is overridden in test.rb,
-    # so the tests don't have to click "I accept".)
+    # so the tests don't have to click "I accept".
     config.hide_accept_cookies_banner = false
 
+    # Show the survey sparrow survey by default.
+    # This setting is overridden in test.rb,
+    # so the tests don't to fill out the survey.
+    config.hide_survey_sparrow = false
```
jrochkind, line 86: You're putting this in Rails `config` instead of our `ScihistDigicoll::Env`?   If we use `Env`, we can turn it on/off from heroku config vars, without having to redeploy code to turn it on/off. 

I feel like the variable should probably be `show_survey_sparrow` instead of `hide_survey_sparrow`, reverse the meaning. 

The code as is seems to assume that as long as this code is in the repo, we will be showing surveys, and to stop showing surveys we actually need to remove it from the repo.  Is that a safe assumption, or do we actually need config to turn them on and off?  Because we might want to turn them on and off at specific times that aren't the same times we deploy;  because we might leave the code in for future surveys?

app/frontend/javascript/survey_widget.js
```diff
@@ -0,0 +1,88 @@
...
+  // Any click on the main survey div is
+  // interpreted as "user has started survey."
+  // Once they start the survey, never show
+  // this user the survey again for an entire year.
+  jQuery ('#ss_survey_widget').click(userHasStartedSurvey);
+});
+
+
+function surveyAlreadyShown() {
+  return document.cookie.match(/surveyAlreadyShown=true/) != null;
```
jrochkind, line 38: Is this really the official way to check cookies from JS?  The `match` seems weird to me. Googling... maybe it is? Very weird.

app/frontend/javascript/survey_widget.js
eddierubeiz: Thanks for the `domready` tip - yeah, I'm optimistic I can get rid of all references to jQuery.

eddierubeiz:
Note: hitting the "X" button allows a user to a) not take the survey and b) never see it again. But it may not be easy to see - it's on the left hand side of the sidebar, which is weird.

And maybe we should also look into whether we could have a button on page 1 saying "I do not want to take this survey".

apinkney0696:
> Is the only way to not see the survey presented on every page you visit to take it? Ethically, I would argue we really need to give a user a way to say "no thanks" and not see the survey in their way on every single page view, if they don't want to take the survey. If I were an IRB board, that's what I'd say. It doesn't seem to me like treating the users very well to say "Unless you consent to take this survey, you have to see this survey pop up every single time you click in digital collections and then dismiss it every time".
> 
> How long do you plan to leave the survey active? 



Yes, as Eddie mentioned, the user is able to exit the survey if they don't want to take it - we're certainly not forcing them. 
I couldn't find a way to add another button that says I don't want to take the survey, but I added this line to the welcome page:

"To exit the survey at any time, press the X at the upper left of this window."

I basically plan to leave the survey active from the time we get it posted to the time our quarterly subscription runs out on January 18th. I can pay for another quarter of the survey if we need it, but we'll see how our data looks at that point.

config/application.rb
eddierubeiz: Good idea -- yeah, let's move this to an ENV var.

app/frontend/javascript/survey_widget.js
eddierubeiz: It's what we use for the "accept cookies" footer, and it does work. I agree it's very counterintuitive.

app/views/application/_survey_widget.html.erb
eddierubeiz: I'll remove the console.log notices before we merge this. They do help at this stage in the process, though.

jrochkind: Cool makes sense! To be clear, I have no problem with any console.log, even remaining in production. My concern is only with the on-screen HTML div alerts.

eddierubeiz: Oh, yeah, those are definitely going away before we merge.

eddierubeiz:
I checked in with SurveySparrow to confirm we are on the right track. The key excerpt after I explained the situation:
**Sriram**
	Ohhh, yes. May i know the workaround implemented?
**Eddie**
	Yes of course
**Sriram**
	Did you restrict by session?
**Eddie**
	Exactly
	Hold on I can give you the JS code in a minute...	https://github.com/sciencehistory/scihist_digicoll/blob/161fd223110eef71b08b958dffbe28a173616cef/app/frontend/javascript/survey_widget.js
**Sriram**
	I got you. Thought of giving the same workaround.
**Eddie**
	OK, well I feel much better now that we've checked with you! Thanks for confirming that we were not missing any other ways to do this! :)
**Sriram**
	Yes. The only way, for now, is to restrict by sessions. Because if we show the survey after the user logs in, we can capture the email address and mark the response which will work accordingly with "Limit submissions per user"

app/frontend/javascript/survey_widget.js
eddierubeiz: Definitely. Edited the PR and put in some basic notes in the code as well.

app/views/application/_survey_widget.html.erb
eddierubeiz: Yes, the script block needs to be there, or else there will be an error in the canned JS we get from SurveySparrow. I could edit that JS to avoid the error, but we don't want to go any further down that road.

config/application.rb
eddierubeiz: Also reversed the name of the var and its behavior to `show_survey_sparrow`  -with a default value of "false". Anything other than a value of "true" in the env var will suppress the survey.

app/frontend/javascript/survey_widget.js
eddierubeiz: Got rid of JQuery in this code yesterday.

app/views/application/_survey_widget.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<% elsif ScihistDigicoll::Env.lookup!(:show_survey_sparrow) == "true" %>
+	<div id="ss_survey_widget" />
+	<script id="SS_SCRIPT"><%# 
+		This script tag is only here to avoid an error in the the canned JS we need to put on the site.
+		For the actual JS code, see the GitHub PR above.%></script>
+<% else %>
+	<%#
+		<div class="alert alert-primary">Survey hidden because show_survey_sparrow is not turned on</div>
+	%>
+<% end %>
```
jrochkind, line 19: can you remove the commented-out stuff? 

So it could just be `if !current_user && ScihistDigicoll::Env.lookup!(:show_survey_sparrow) == "true"`...  I'm not personally convinced we need to hide this for logged in users... can't all 6 of us just close the survey once without taking it? But if you want to, also fine.

jrochkind (approved):

app/views/application/_survey_widget.html.erb
eddierubeiz: Valid points, but I am leaving the code as is; my arguements are a) this is temporary code and b) there's a real chance we may be debugging it again soon. My guess is that having all the extra comments might just save us a bit of work later on.

Issue #1916, opened 2022-10-26
FAST autocomplete is broken


jrochkind:
Maybe broke as part of our vite migration?

When I type in the should-be-autocompleted `subject` field on production, I get this JS error in dev console:

```
5qa_autocomplete.js:56 Uncaught TypeError: o.autocomplete is not a function
    at Vi (qa_autocomplete.js:56:14)
    at HTMLInputElement.<anonymous> (qa_autocomplete.js:124:7)
    at HTMLBodyElement.dispatch (jquery.js:5430:27)
    at N.handle (jquery.js:5234:28)
    at Object.trigger (jquery.js:8745:12)
    at Object.simulate (jquery.js:8814:16)
    at HTMLDocument.n (jquery.js:8848:17)
```

jrochkind:
Hm, the file that implements autocomplete does a:

`import 'devbridge-autocomplete';`

Under webpacker, that was enough to make a jquery `.autocomplete()` function available.  Under vite it appears not to be, not sure why or how to fix, looking into it.

jrochkind:
https://github.com/devbridge/jQuery-Autocomplete/issues/843

jrochkind:
Hm.  We could copy and paste that jquery plugin to a local file, and edit it's loading code to work. 

We could switch to a whole different autocomplete implementation (that one isn't very actively maintained, and requires jquery; although it works well).  Blacklight just switched to: https://github.com/github/auto-complete-element

That one uses "web components" so requires some change to markup, a little bit of work to switch to it... although probably it would work?

jrochkind:
Huh, while it also hasn't been maintained in forever, there may be a non-jQuery autocomplete alternative from the same source that may import properly and require few code changes....

https://github.com/devbridge/Vanilla-Autocomplete

jrochkind:
Another one we could maybe switch to in future: https://github.com/digitalfortress-tech/typeahead-standalone

PR #1918, opened 2022-10-26
Fix vite breakage of jquery.autocomplete

jrochkind:
Ref #1916

We copy and localize the devbridge jquery.autocomplete to get it to load properly. We don't love this solution, but at least it temporarily restores autocomplete func for staff. 

Note I first committed unmodified source of vendor plugin, then made another commit with changes, so local changes are more clear. 

Medium term: Still don't have any options we love! Extensive comment docs are left in source with tech analysis of options and issues.

### DIFF
--- a/app/frontend/javascript/vendor/jquery.autocomplete.js
+++ b/app/frontend/javascript/vendor/jquery.autocomplete.js
@@ -1,0 +1,1013 @@
+/**
+*  Ajax Autocomplete for jQuery, version 1.4.11
+*  (c) 2017 Tomas Kirda
+*
+*  Ajax Autocomplete for jQuery is freely distributable under the terms of an MIT-style license.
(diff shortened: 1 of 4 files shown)


app/frontend/javascript/vendor/jquery.autocomplete.js
```diff
@@ -21,7 +32,10 @@
     } else {
         // Browser globals
         factory(jQuery);
-    }
+    }*/
+
+    // Science History Institute force loading into window.jQuery
+    factory(window.jQuery);
```
eddierubeiz, line 38: Noted.

eddierubeiz:
OK. I took a look at the file you changed -- hard to be absolutely sure what's going on without actually tinkering with it, but merging now considering the circumstances.

Issue #1927, opened 2022-11-04
export metadata for Annabel collection blog


apinkney0696:
Hi! Could I please have the following metadata exported from all (published only) records in a .csv or .xlsx format?

Title
Creator 
Place
Date
Genre
External ID
Department

Going to create some data visualizations for my January collections blog and wanted to get a head start with the holidays coming up.

apinkney0696:
Ohh! Love it. 

My initial plan was to make a story map and geographically place objects. I just wanted to request more metadata than necessary to have ample info to play around with. 

Chat more on Mon.!

jrochkind:
let's just manually export the data Annabel needs right now, even if we talk about creating some more complex built-in feature? I'm always a fan of waiting until we've done a manual export ~3 times before talking about automating it.

eddierubeiz:
OK @apinkney0696 - here's [production](https://github.com/sciencehistory/scihist_digicoll/files/9963149/Production.metadata.xlsx).

eddierubeiz:
Here's [an updated spreadsheet](https://github.com/sciencehistory/scihist_digicoll/files/10043910/extra_metdata.xlsx) which includes thumbnail links for all published works, and a column you can use to easily filter out non-Bredig rows.

PR #1931, opened 2022-11-07
Collections blog metadata export

eddierubeiz:
Ref #1927
```bundle exec rake scihist:reports:collections_blog_metadata```


apinkney0696:
This is perfect Eddie. Just what I needed to start playing with. When you have time, I'm ready for a final set of data from production!

Issue #1883, opened 2022-10-13
Mysterious solr index miss for "video ISI"  (due to German language field included)


jrochkind:
I expect to be able to search "video ISI"  to find this one: https://digital.sciencehistory.org/works/otbju0w

But I can't... not really sure why not, something weird is happening in solr.

eddierubeiz:
Eh? We do index `genre` in `text2_tesim` and in `genre_facet`.
And https://digital.sciencehistory.org/catalog?search_field=all_fields&q=Video+ISI seems to work fine for finding that item. 
The genre for that work is "Video Recordings", for what it's worth.

eddierubeiz:
Correction --  there's more than one item with that title, and the item is not in the list of results.

jrochkind:
Weird, I'm very confused what's going on then, but does seem not right!

jrochkind:
These all find it:

* https://digital.sciencehistory.org/catalog?q=%22video%22+%22putting+scientific+information%22
* https://digital.sciencehistory.org/catalog?q=%22video%22+ISI

Note that last one puts `"video"` in double quotes... when I drop the double-quotes, I get entirely different results... not how we expect Solr to work!

https://digital.sciencehistory.org/catalog?q=video+ISI

Changing title to "Solr mysterious lack of hit"....

jrochkind:
Still very mysterious... Trying to reproduce with smaller and smaller solr case....

I have determined that the solr `qf` (query fields) property matters.. With our full multi-field qf, the failure reproduces. 

If I trim it to _just_ `text1_tesim^1000 description_text4_tesim^50` the failure does not reproduce. 

Somehow taking a query that produces a hit and _adding_ search fields to it... causes it to fail to hit. That's the mystery. Trying to get minimum fields that reproduce, so maybe it can be compared with more debugging.

jrochkind:
Weird, no, I'm having trouble reproducing the _hit_ with the smaller `qf`; my actual app in dev does it, but I'm having trouble doing it against solr directly to more easily debug.

eddierubeiz:
The plot thickens.

jrochkind:
OK, if I redefine `searchable_fulltext_de` as `text_en` -- the same file type as the other two fields -- it _still_ seems to reproduce the problem to include it in `qf`. 

This makes little sense. 

What makes it different? That it contains no values for this document?  That may be, adding other fields to dismax that include no values for the document seem to also reproduce maybe? 

[...]

PR #1934, opened 2022-11-08
Remove German hyphenization from solr text_de

jrochkind:
Differing number of tokens from text_en broke solr dismax when used together, Ref #1883

Also replaced the tokenization and lowercasing with Solr analyzers more like text_en, just for consistency, just in case.

We believe that the German normalization and stemming filters should not cause a problem, although we aren't 100% sure, we're trying it.

### DIFF
--- a/solr/config/schema.xml
+++ b/solr/config/schema.xml
@@ -187,13 +187,22 @@
     <!-- https://github.com/uschindler/german-decompounder -->
     <fieldType name="text_de" class="solr.TextField" positionIncrementGap="100">
       <analyzer>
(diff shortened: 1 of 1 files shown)


jrochkind:
Will need solr reindex after deploy!  Person who does the merge should do so when prepared to immediately do a deploy and run solr indexing please, otherwise just mark 'approved'.

jrochkind:
tested in dev and staging, all looks well and `video isi` search is returning expected values.

jrochkind:
Deployed and reindexed.

Issue #1936, opened 2022-11-08
PDF and Zip downloads are broken


jrochkind:
Choosing option gives you an error in a JS modal. 

No HoneyBadger error is generated. :(. Also we have no failing tests. :(.  Look into improving both of those aspects along with fixing underlying problem.

jrochkind:
So... somehow the move to vite meant our scripts are being interpreted in a JS `strict` mode where all variables have to be declared before use. 

And some of our scripts don't do this, so produce errors under strict mode. Or at least one of our scripts does, in this case `scihist_on_demand_downloader.js`

I can fix it to declare it's variables that are production errors... but I wonder what other scripts might be broken too. 

I wonder if we should figure out how to get these _not_ to be interpreted under "strict" mode, since they weren't written for it. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

I should be able to add a failing test, at least, too, though. 

Not sure about getting HoneyBadger to report Javascript errors, there is an HB library to do that, that we haven't been using, should we? https://docs.honeybadger.io/lib/javascript/integration/browser

jrochkind:
Aha. 

> Vite cannot handle and does not support code that only runs on non-strict mode (sloppy mode). This is because Vite uses ESM and it is always strict mode inside ESM.

https://vitejs.dev/guide/troubleshooting.html#others

OK.... I wonder if we have any other gotchas waiting for us. 

I wonder if there's a good way to analyze our code statically looking for likely strict mode violations.

jrochkind:
I did a bit of regex searching through our local javascript source, to try to find variable assignments without `var`, and then manually make sure the variable used had been declared previously. 

I didn't find any others. No guarantee there aren't any other gotchas waiting for us, but I think we just keep going.

jrochkind:
I experimented a bit with adding the honeybadger JS library for JS error reporting... it's do-able...

But it wouldn't have caught this particular problem with code as-is, because our JS code caches the error and reports it in a nice error message to user, instead of being an uncaught JS error for a HB error reporter to catch. Heh. 

Still thinking about it, but may not bother with this yet for now.

PR #1937, opened 2022-11-08
Fix broken Zip/PDF downloads

jrochkind:
Ref #1936 

- correctly failing test for broken on-demand deriv download
- fix to work under JS strict mode


eddierubeiz:
Good old `var` keyword. Still breaking our code after all these years.

PR #1939, opened 2022-11-10
Fix for nil height or width bug

eddierubeiz:
Ref #1930

@jrochkind - I'm inclined to just add in this check and not pursue this further, but open to suggestions.

My underlying assumption is that the file wasn't fully characterized by the time the collection edit page was loaded, and thus the value of the file's width and height evaluated to`nil`, causing [this error](https://app.honeybadger.io/projects/58989/faults/90140357).

That said, I wasn't able to reproduce the bug in dev, so it's difficult to say for sure whether this will fix the problem.

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -81,6 +81,26 @@ class Admin::CollectionsController < AdminController
     end
   end
 
+  def representative
+    @collection&.representative
+  end
+  helper_method :representative
+
+  def representative_is_image?
+    representative.file.present? && representative&.content_type&.start_with?("image/")
+  end
+  helper_method :representative_is_image?
+
+  def representative_dimensions_correct?
+    [ representative.width.present?,
+      representative.height.present?,
+      representative.width == representative.height,
+      representative.width >= CollectionThumbAsset::COLLECTION_PAGE_THUMB_SIZE * 2,
+    ].all?
+  end
+  helper_method :representative_dimensions_correct?
+
+
   private
     # Use callbacks to share common setup or constraints between actions.
     def set_collection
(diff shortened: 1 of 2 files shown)


jrochkind:
Yes, just putting in more guards to avoid the particular error is fine, if we can't reproduce otherwise. 

I'm having trouble following the logic in the boolean expression that's become really complicated with `!` in it though... are you sure it's doing what you want?  I feel like maybe the `!`s should not be there... am I wrong? is there a way to simplify the logic to make it more clear?

eddierubeiz:
I thought about this a bit more and decided maybe the template was doing too much work. I moved some of the gnarliness back into the controller in the form of 3 extra helper methods. I think the resulting template is a lot more legible. Maybe overkill in response to such a tiny bug, but I do think it's somewhat better.

app/controllers/admin/collections_controller.rb
```diff
@@ -81,6 +81,26 @@ def destroy
...
+    representative.file.present? && representative&.content_type&.start_with?("image/")
+  end
+  helper_method :representative_is_image?
+
+  def representative_dimensions_correct?
+    [ representative.width.present?,
+      representative.height.present?,
+      representative.width == representative.height,
+      representative.width >= CollectionThumbAsset::COLLECTION_PAGE_THUMB_SIZE * 2,
+    ].all?
```
jrochkind, line 99: Moving to a helper was probably a good idea. 

I don't think `all` adds any clarity here over using `&&`, and possibly is slightly less efficient -- `&&` "short circuits", so if one is false the rest won't be calculated, while this formulation with an array and `all?` will necessarily calculate them all. 

But I don't think it's likely to be signfiicant, I'll just go ahead and approve anyway.

eddierubeiz: [You're correct](https://stackoverflow.com/questions/20674612/do-all-and-any-guarantee-short-circuit-evaluation)- all? does not short circuit.

jrochkind: Right, the way ruby works there is no way for it to, the array has to be constructed first before `all?` can be called on it. Just a ruby language syntax/semantics issue, the code to create the array, and all it's members, will be executed before ruby even gets to `all?`.

Issue #1943, opened 2022-11-14
Browse_everything S3 ingest does not work when running vite dev server


jrochkind:
A weird case where different outcome depending on if you are running vite dev server; or instead using vite autoBuild to do a vite build in production mode. 

In dev mode, I want to add files to a work, and choose "Add from cloud". I check a box next to a file, and choose "Submit". 

When it works, the file I selected appears in the right column list of files to attach. 

But when vite dev server is running, the file does not appear, and there is an error in console:

     POST http://localhost:3000/admin/browse/resolve 422 (Unprocessable Entity)

Stack trace includes: `browse_everything/behavior.source.js:333`. 

Not really sure what's going on; and why would it fail only when vite dev server is running, how could that make a difference?

jrochkind:
Aha, noticed in logs in failing case:

```
Processing by BrowseEverythingController#resolve as JSON
  Parameters: {"authenticity_token"=>"[FILTERED]", "context"=>"", "selected_files"=>["s3:digcol-staging.dump"]}
Can't verify CSRF token authenticity.
Completed 422 Unprocessable Entity in 5ms (ActiveRecord: 0.0ms | Allocations: 784)



ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken):
```

OK, the Jquery AJAX post somehow automatically includes the "CSRF authenticity token". 

But somehow fails to do so only when vite dev server is running, even though it should be the same JS code either way...

jrochkind:
OK, it wasn't the authenticity_token in the form... when it _works_ it's using a separate "csrf_token" that is taken from whole page <head> section, and sent with request _headers_. 

JQuery.AJAX is patched to get this value and send it. 

Rails-ujs is doing the patching (actually configuring) of jQuery to send the rails csrf_token on every .ajax fetch. 

For some reason the patching was not taking effect in `vite dev` mode, but was in `vite build` mode.... changing the _order of imports_ to put the `rails_ujs` import _after_ we import and  set up JQuery fixed things. 

I don't entirely understand what is going on;  in old pre-vite sprockets, the order of `require`s we had, we had require rails_ujs _before_ require jquery... and it still worked.  If it is order dependent because of how we changed things a bit in vite... why only in `vite dev` and not `vite build`?

I can't answer these things, but the fix seems to work, so.

PR #1945, opened 2022-11-14
Fix jQuery sending of Rails csrf_token under vite

jrochkind:
Noticed in #1943 that browse_everything JS was causing Rails CSRF authenticity token failures after our move to vite.

Discovered that the rails_ujs JS package was not succesfully configuring JQuery.ajax to send Rails CSRF token on every .Ajax call.

Changing the ORDER of our imports in our vite file fixes this. This whole thing doesn't make any sense -- it was only broken in `vite dev` not `vite build`, why was only one of those setups so order-dependent? Also our pre-vite setup using sprockets was still "require"ing rails_ujs *before* jquery -- we left the order the same when we moved to vite, but it somehow became order-dependent requiring a different order? (That last one might be more explainable because of the hacky way we're setting window.jQuery in the export itself).

Don't completely understand what's going on, but somehow found a a fix anyway. Left warning comments in source code that does the imports where order has become important.

### DIFF
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -28,11 +28,18 @@ console.log('Visit the guide for more information: ', 'https://vite-ruby.netlify
 // import '~/index.css'
 
 
-// We're still using rails-ujs for now. Rails-ujs 6.x will auto-start itself on import.
-import Rails from '@rails/ujs';
+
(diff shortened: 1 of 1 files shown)


eddierubeiz (approved):
Thanks for troubleshooting this.

PR #1950, opened 2022-11-15
Move everything else we can over to vite, including CSS

jrochkind:


jrochkind:
OK, doesn't appear to be reproducibly slower in production on staging (which made no sense anyway). 

Some performance changes I investigated yesterday though.  "webpacker" means the version before we introduced webpack at all. "master" means before this PR -- vite for most JS, but still using sprockets for CSS. 

The heroku deploy process gets _faster_ with this branch, it speeds up the assets precompile phase. Sprockets version ~60-70s;  master 25s; this branch 30s. (So this branch slows things down a bit compared to master, but still a big win from webpacker). 

In development  -- if you _aren't_ running the dev server, but using vite "autoBuild", it is substantially slower than webpacker, when it detects a change requiring recompile.  (Both webpacker and vite autoBiuld work similarly, they only recompile when detecting files have changed).  

* Webpacker this would generally take only 4 seconds. 
* Master this is up to 10.8 seconds.  
* Initially in this branch it was up further to 12+ seconds. 
* But after we avoid building gzip and brotli variants when under `autoBuild`, this branch is down to 8-9 seconds
* Still twice as slow as webpacker, waiting 4-5 seconds longer, which is pretty annoying
* But you can always use the dev server if you want to avoid this -- you probably will want to if you are actually doing any CSS or JS development anyway (only get SASS sourcemaps with dev server too). 
* Compared to alternatives like jsbundling-rails esbuild where you _have_ to use a dev server, oh well.

PR #1958, opened 2022-11-21
fixing paths when asset.parent is not a work

eddierubeiz:
I'm just trying to do the minimum amount of work to fix the bug described in #1956 .

If an asset is a collection thumbnail,
- Don't show the form that allows a user to change its storage to "restricted" bucket. Currently the form is broken for these assets. If we decide to fix the form, we can show it for collection thumbnails, but that's out of scope for now since I can't imagine wanting to store a collection thumbnail in the restricted bucket;
- Fix the links associated with it on the asset index page;
- don't try to look up its parent's genre in `work_is_oral_history?` because collections don't have genres;

### DIFF
--- a/app/controllers/admin/assets_controller.rb
+++ b/app/controllers/admin/assets_controller.rb
@@ -155,7 +155,7 @@ class Admin::AssetsController < AdminController
   end
 
   def work_is_oral_history?
-    @asset.parent.genre && @asset.parent.genre.include?('Oral histories')
+    (@asset.parent.is_a? Work) && @asset.parent.genre && @asset.parent.genre.include?('Oral histories')
   end
   helper_method :work_is_oral_history?
 
--- a/app/views/admin/assets/index.html.erb
+++ b/app/views/admin/assets/index.html.erb
@@ -34,8 +34,16 @@
       <tr>
         <td><%= thumb_image_tag(asset, size: :mini, image_missing_text: true) %></td>
         <td><%= asset.friendlier_id %></td>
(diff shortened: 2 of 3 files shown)


eddierubeiz:
This is a bug in production, so I reserve the right to merge and deploy my own fix Nov. 22 in the morning.

jrochkind:
Oh yeah, I had been assuming parent was always a work -- I forgot the weird situation where we modelled Collection thumbnail like this. OK, thanks for fixing. 

I see how it applies to that `work_is_oral_history?` method -- looking at that now, I wonder if we want to make that method secure against parent being nil too, just in case that happens? `work_is_oral_history?` should be false if the parent is nil!

I'm not totally following how it extends to the other controller stuff and the derivative storage thing....  I don't totally love scattering "parent is a work" checks all over, that's messy not great OO code, but I guess maybe the modelling we've done has left us there?  It would have been more ideal to make the various things agnostic toward the parent I think, like instead of an if/else condition in the template to put the label `parent work` vs `in collection`, why not just write out `parent: [title] [class]`, so it could be the same code for either kind of parent? And extend this approach in general to the other places -- I don't think that would have really been more work necessarily?

This fix should have had an additional spec, simply testing that the pages can be rendered without error for assets whose parents are collection. We should probably actually go back and at least add that -- in general, every bugfix should have a test ensuring the bug is fixed, bugfixes like this are a place it's really easy to do TDD.

eddierubeiz:
Agree! (and see also #1959)

PR #1962, opened 2022-11-23
Properly catch attempted delete of Dig Queue Item with attached works

jrochkind:
This is just a quick and easy solution to a staff-facing thing that doesn't happen very much.

* There is still potentially a race condition if someone attacches a work RIGHT when someone else is deleting. That will just result in an uncaught exception still. Not gonna happen, not worth writing for.
* If we really wanted to polish this, would maybe make the delete button disappear or become inactive or not even submit in case of attached works -- but not needed, it's fine that it goes to server and comes back with a notice.

We also changed the configuraiton of the relationship to TELL Rails that we want to prevent delete with attachements, on top of the underlying DB constraint ensuring such.

Ref #1952

### DIFF
--- a/app/controllers/admin/digitization_queue_items_controller.rb
+++ b/app/controllers/admin/digitization_queue_items_controller.rb
@@ -43,12 +43,20 @@ class Admin::DigitizationQueueItemsController < AdminController
     end
   end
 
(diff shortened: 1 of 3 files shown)


eddierubeiz (approved):
Thanks for looking into this!

Issue #1953, opened 2022-11-17
Oral History Keyword Display


apinkney0696:
@rachellane12 and the oral history department did some informal user testing with Michelle and family and friends to look at the use of OHMS interviews. They are subsequently reworking their guidelines for keyword use in the table of contents and need a little help with the development. 
(Ticket 1 of 2)

**1. Make the Keywords look less like buttons.** 
The keywords are meant to quickly signify to a user what a section is about; keywords will also be highlighted when searched, so should help the user locate sections about a certain topic.

OH staff received feedback from users that there are too many keywords per section, and are therefore overwhelming. They are going to decrease the number of keywords applied in their OHMS workflow going forward, but would also like the appearance to be less button-like, and simply appear like another text section, delineated by commas or semicolons.

jrochkind:
Yep, when I designed this feature i was definitely thinking they would apply only a few keywords per section, because that's all the OHMS examples I had seen. I was surprised when they started applying so many instead, and agreed that the design was not appropriate for that! Glad we're circling back to fix it. 

That's exciting they did some user testing!

PR #1963, opened 2022-11-23
Make OH keywords no longer look like capsules

jrochkind:
Ref #1953

Not really sure how well this works presently, let's get feedback before merging.


jrochkind:
@apinkney0696 I'm not really sure how well this works....

-----------
Screen Shot 2022-11-23 at 10 32 49 AM

------------
Screen Shot 2022-11-23 at 10 32 43 AM

------------

I feel like it kinda ends up looking like a wall of text, a scramble of punctuation. 

I'm not totally sure what else to do though, if abandoning the "capsule" style. 

Curious your thoughts. Trying to think of other things that could be done with color or whitespace.... maybe bullets separating the items plus more whitespace?  

I can show you (or others) this live in staging too. Not sure at what point we should ask for OH team feedback.

jrochkind:
OK, I've tried it like this....

--------
Screen Shot 2022-11-23 at 10 57 36 AM

-------
Screen Shot 2022-11-23 at 10 57 21 AM

-------
Screen Shot 2022-11-23 at 10 57 05 AM

--------

I think that's probably better?  Note there can be a line wrap in the middle of a "keyword" phrase... I experimented with preventing that, keeping a "keyword" phrase all on one line, but that ended up causing other oddities in some cases.

apinkney0696:
I love the second option!! Undoubtedly better! I think the wrapping issue is pretty negligible and fine as is. I'll check in with Rachel to make sure she's happy with this and let you know if she has any feedback.

rachellane12:
I agree 100 percent with @apinkney0696. This looks great! The bullet points between the keywords separate them without giving the button look we had before. I also think the wrapping issue isn't a big problem. Really appreciate your work on this, @jrochkind!

PR #1969, opened 2022-11-28
vite gzip assets: condition needs to be on string

jrochkind:
We wanted to create gzip and brotli assets only in produciton build not vite autoBuild -- but the value we were checking to determine this is, it turns out, string 'true' or 'false', so our condition was always matching and we were never getting gzip and brotli.


jrochkind:
@eddierubeiz I'm going to go ahead and merge and deploy this, but worth a look for a wacky vite-ruby/JS gotcha.

PR #1970, opened 2022-11-28
Use meta name=description as well as property=og:description

jrochkind:
While facebook and some other social media uses <meta property="og:description"> for description, standard HTML <meta name="description"> is more standard, and some things may use that. Google lighthouse flagged me for not having a <meta name="description">.

Conveniently, we can put BOTH a `name` AND `property` on the same tag, and have this tag do double-duty. https://stackoverflow.com/questions/20401085/is-it-possible-to-use-the-same-meta-tag-for-opengraph-and-schema-org/20429551#20429551

So easy enough to just add the `name=description` too, and potentially have our description picked up by more things looking for one.


eddierubeiz:
Your summary is a lot more digestible than the stackoverflow discussion . Thanks - I almost never look at this file, so it's a good reminder for me that it even exists.

Issue #1948, opened 2022-11-15
Reorganize account types


eddierubeiz:
This is part of ref #1946

| Who      |  ✅ |  ⛔ |
| ----------- | ----------- | ----------- |
| Admin      | ✅ Everything       |  |
| Editor   | ✅ Create ✅ Edit ✅ Publish        | ⛔ Delete |
| Viewer   | ✅ View ✅ Add to digitization queues        | ⛔ Edit ⛔ Delete |

jrochkind:
Recommend spending some time with access_granted README and this wiki page, for orientation: https://github.com/chaps-io/access-granted/wiki/Role-based-authorization-in-Rails

(I am reviewing it too)

PR #1973, opened 2022-11-30
Add user_type column to users table

eddierubeiz:
No reference to it anywhere in the code. Just setting the stage for more work.
Ref #1948 .

### DIFF
--- a/db/migrate/20221122214935_add_user_type_to_users.rb
+++ b/db/migrate/20221122214935_add_user_type_to_users.rb
@@ -1,4 +1,6 @@
 class AddUserTypeToUsers < ActiveRecord::Migration[6.1]
+  # This user_type column will eventually allow us to have more than
+  # two user types. See https://github.com/sciencehistory/scihist_digicoll/issues/1948
   def change
     add_column :users, :user_type, :string, default: 'editor'
   end
--- a/db/migrate/20221122214935_add_user_type_to_users.rb
+++ b/db/migrate/20221122214935_add_user_type_to_users.rb
@@ -1,0 +1,5 @@
+class AddUserTypeToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :user_type, :string, default: 'editor'
+  end
+end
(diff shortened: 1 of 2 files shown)


jrochkind:
I know I encouraged small PR's, but maybe not this small! 

Is there a reason to get this merged first, or should we just go ahead and have a PR that includes this migration as well as use of it?  

Ah, I guess maybe a reason to deploy this first is so the column is there for any migrations we might want to write to move over users?

eddierubeiz:
Yeah -- I generally find it's nice to have this part out of the way. A lot more coming soon, obviously.

eddierubeiz:
Sounds good -- added a comment to the migration just in case.

jrochkind:
@eddierubeiz You had talked about making this an enum instead of a string? Or you meant just on the Rails side rather than pg side?

There is such thing as a pg enum... which is theoretically more efficiently stored...I forget if we've used it before. String is probably fine too, I honestly forget if there is a reason a pg enum is a pain to use in Rails or not or if we've ever used it before. Just checking for your intent.

jrochkind (approved):

db/migrate/20221122214935_add_user_type_to_users.rb
```diff
@@ -0,0 +1,7 @@
+class AddUserTypeToUsers < ActiveRecord::Migration[6.1]
+  # This user_type column will eventually allow us to have more than
+  # two user types. See https://github.com/sciencehistory/scihist_digicoll/issues/1948
+  def change
+    add_column :users, :user_type, :string, default: 'editor'
```
jrochkind, line 5: you sure you want to make the default "editor"? Where does that come from?  I guess we can change the default later, let's not forget to if needed!

eddierubeiz: Of course. I figure when you create a new user you will probably not want to make then an admin by default.

jrochkind:
I approved it!

PR #1974, opened 2022-11-30
Add :see_admin_pages permission

eddierubeiz:
In preparation to some upcoming changes to our permissions scheme, this adds a new `:see_admin_pages` permission to our access policy.
All logged in users have this permission (and we don't actually foresee that changing). But instead of checking whether we can find a logged-in user, we'll be checking for this permission instead, which is tidier.

### DIFF
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -1,7 +1,3 @@
 class ApplicationComponent < ViewComponent::Base
-  delegate :current_user, to: :helpers
 
-  def can_access_staff_functions?
-       AccessPolicy.new(current_user).can? :access_staff_functions
-  end
 end
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -1,7 +1,7 @@
 class ApplicationComponent < ViewComponent::Base
   delegate :current_user, to: :helpers
 
-  def can_see_admin_pages?
-       AccessPolicy.new(current_user).can? :see_admin_pages
+  def can_access_staff_functions?
+       AccessPolicy.new(current_user).can? :access_staff_functions
   end
 end
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -1,3 +1,7 @@
 class ApplicationComponent < ViewComponent::Base
+  delegate :current_user, to: :helpers
 
+  def can_see_admin_pages?
+       AccessPolicy.new(current_user).can? :see_admin_pages
+  end
 end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -6,7 +6,7 @@ class AdminController < ApplicationController
 
   # For now, admin controllers allow anyone who is logged in
   def authorize_access
-    unless current_user.present?
+    unless can? :see_admin_pages
       # raise the error from `access_granted` to be consistent.
       raise AccessGranted::AccessDenied.new("Only logged-in users can access admin screens")
     end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -6,7 +6,7 @@ class AdminController < ApplicationController
 
   # For now, admin controllers allow anyone who is logged in
   def authorize_access
-    unless can? :see_admin_pages
+    unless can? :access_staff_functions
       # raise the error from `access_granted` to be consistent.
       raise AccessGranted::AccessDenied.new("Only logged-in users can access admin screens")
     end
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -52,6 +52,13 @@ class CatalogController < ApplicationController
     }
   end
 
+  # Should be equivalent to:
+  # ->(controller, field) { controller.can?(:access_staff_functions) }
+  # but was not able to get this syntax to work.
+  def can_see_all_search_facets?
+    can? :access_staff_functions
+  end
+
 
   self.search_service_class = Kithe::BlacklightTools::BulkLoadingSearchService
   Kithe::BlacklightTools::BulkLoadingSearchService.bulk_load_scope =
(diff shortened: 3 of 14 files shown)


eddierubeiz:
Note that in the catalog controller we define:
```
  def can_see_extra_sort_orders?
    can? :see_admin_pages
  end
```

on the assumption that if you're allowed to see the admin pages you might also expect be able to sort using our heretofore "staff-only" search orders. This *could* be a separate permission but it seemed overkill to me.

app/components/application_component.rb
```diff
@@ -1,3 +1,7 @@
 class ApplicationComponent < ViewComponent::Base
+  delegate :current_user, to: :helpers
 
+  def can_see_admin_pages?
+       AccessPolicy.new(current_user).can? :see_admin_pages
+  end
```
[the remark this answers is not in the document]
jrochkind: (I realize I may have suggested this inadvertantly originally. But either I had something else in mind, or, now that i see it...)

app/controllers/admin_controller.rb
```diff
@@ -6,7 +6,7 @@ class AdminController < ApplicationController
 
   # For now, admin controllers allow anyone who is logged in
   def authorize_access
-    unless current_user.present?
+    unless can? :see_admin_pages
```
jrochkind, line 9: Right, see, like this! So I don't think we need the `can_see_admin_pages?` helper method (that isn't being used here anyway)

app/controllers/catalog_controller.rb
```diff
@@ -456,31 +463,31 @@ def active_sort_fields
     config.add_sort_field("date_modified_desc") do |field|
       field.label = "date modified \u25BC"
       field.sort = "date_modified_dtsi desc"
-      field.if = ->(controller, field) { controller.current_user }
+      field.if = :can_see_extra_sort_orders?
```
jrochkind, line 466: Again I would cosider skipping the cover method, and just doing:

    field.if = ->(controller, field) { controller.can?(:see_admin_pages) }

But... I see how you're using it in multiple places, maybe it makes sense to add the extra indirection here, I dunno, up to you. But note that the above should work fine. I prob would have done that. Just like `controller.current_user` was repeated previously, without a cover method, which was fine.

app/policies/access_policy.rb
```diff
@@ -32,6 +33,7 @@ def configure
       can :read, Kithe::Model # whether publisehd or not
       can :update, Kithe::Model
 
+      can :see_admin_pages
```
jrochkind, line 36: I don't _think_ you need to repeat `can :see_admin_pages` between multipe roles -- someone with the admin role should _also_ have the "lesser" roles. 

Let's see if we can get that to work, just have `can :see_admin_pages` once, at the least role that has it.

config/routes.rb
```diff
@@ -4,10 +4,9 @@
   concern :range_searchable, BlacklightRangeLimit::Routes::RangeSearchable.new
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 
-  class LoggedInConstraint
+  class CanSeeAdminConstraint
     def self.matches?(request)
-      current_user = request.env['warden'].user
-      !!current_user
+      AccessPolicy.new(request.env['warden'].user).can? :see_admin_pages
```
jrochkind, line 9: OK, I guess since we aren't in a controller or temlate, you really do have to do this the "hard" way?  Good job figuring out the hard way!

jrochkind:
Looking good, see minor comments above. 

i'm thinking about naming... is `can_see_admin_pages` the right/best for this?  I'm thinking maybe `access_...` -- its' not just about _seeing_ them, it's our generic privilege for general accessing/using/modifying them. 

And then I'm thinking "pages".... when it applies to things like sort functions, it's not just pages. 

And then I'm thinking "admin"... is a role we have for our "super-users", which is _not_ what this is about.

Ha, now I've challenged each word. 

How about... `:access_staff_functions`?  Or `access_staff_func` to save typing?  With a comment by it in the policy.rb file explaining it's kind of our catch-all for the lowest-level of common "staff" func that all staff has access to, for things we haven't broken out.

app/policies/access_policy.rb
eddierubeiz: You're right, no need to repeat it. Got rid of it.

app/controllers/catalog_controller.rb
eddierubeiz: No, you're 100% right.

app/components/application_component.rb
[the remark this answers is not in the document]
eddierubeiz: Got rid of it!

app/controllers/admin_controller.rb
eddierubeiz: Yup.

eddierubeiz:
`:access_staff_functions` it is. `sed` is my friend!

app/controllers/catalog_controller.rb
```diff
@@ -52,6 +52,14 @@ def search_service_context
     }
   end
```
eddierubeiz, line 54: I want to get rid of this cover method - will see if I can figure it out.

app/controllers/catalog_controller.rb
```diff
@@ -52,6 +52,14 @@ def search_service_context
...
 
+  # Should be equivalent to:
+  # ->(controller, field) { controller.can?(:access_staff_functions) }
+  # but was not able to get this syntax to work.
+  def can_see_all_search_facets?
+    can? :access_staff_functions
+  end
```
jrochkind, line 60: ok no big deal to leave this, but let's change name to `access_staff_only_facets?`, or someth8ng liek that, what do you think? Sorry for being picky over naming.

app/views/layouts/application.html.erb
```diff
@@ -70,9 +70,7 @@
       <%= render 'ie_unsupported_warning' %>
     <% end %>
 
-    <% if current_user %>
-      <%= render 'front_end_admin_navbar' %>
-    <% end %>
+    <%= render 'front_end_admin_navbar' if can? :access_staff_functions %>
```
jrochkind, line 73: honestly I think this would be more clear if you leave it three lines, instead of the trailing `if`, but no big deal.

jrochkind (approved):

PR #1975, opened 2022-11-30
Use new user_type column to determine permissions

eddierubeiz:
- Decide if someone is "staff" (now renamed "editor") or "admin" based on the new column.
- And allow edits to the new `user_type` column in the admin screens.
- I already updated the values of the column in prod and staging to reflect the current contents of the old admin column, so we can push this to prod whenever.
- Does not modify the behavior of the app in any way, other than to provide a way to set a user to "editor".

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,7 +24,7 @@ class User < ApplicationRecord
 
 
   def admin?
-    raise RuntimeError.new("Deprecated. Use admin_user?")
+    raise RuntimeError.new("Don't use this method. Use admin_user? instead.")
     user_type == "admin"
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,9 +12,30 @@ class User < ApplicationRecord
   devise :database_authenticatable,
          :recoverable, :rememberable, :validatable
 
-
   has_many :cart_items, dependent: :delete_all
   has_many :works_in_cart, through: :cart_items, source: :work
+
+  # This will correspond to a "role" in the AccessPolicy class.
+  # "editor" will replace the current "staff" role.
+  # A new "reader" type will be added in a future PR.
+  USER_TYPES = %w{admin editor}.freeze
+  enum user_type: USER_TYPES.collect {|v| [v, v]}.to_h
+  validates :user_type, presence: true
+
+
+  def admin?
+    raise RuntimeError.new("Deprecated. Use admin_user?")
+    user_type == "admin"
+  end
+
+  def admin_user?
+    user_type == "admin"
+  end
+
+  # This is the same as the old "logged-in but not admin" user type.
+  def editor_user?
+    user_type == "editor"
+  end
 
   # Only used by devise validatable, we want to allow user accounts
   # to be saved with nil password, means they won't be able to log in
--- a/spec/factories/user_factory.rb
+++ b/spec/factories/user_factory.rb
@@ -3,7 +3,11 @@ FactoryBot.define do
     email { "no-reply@sciencehistory.org"}
 
     factory :admin_user do
-      admin { true }
+      user_type { "admin" }
+    end
+
+    factory :editor_user do
+      user_type { "editor" }
     end
   end
 end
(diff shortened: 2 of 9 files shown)


app/models/user.rb
```diff
@@ -12,10 +12,31 @@ class User < ApplicationRecord
...
+  # This will correspond to a "role" in the AccessPolicy class.
+  # "editor" will replace the current "staff" role.
+  # A new "reader" type will be added in a future PR.
+  USER_TYPES = %w{admin editor}.freeze
+  enum user_type: USER_TYPES.collect {|v| [v, v]}.to_h
+  validates :user_type, presence: true
+
+
+  def admin?
+    raise RuntimeError.new("Deprecated. Use admin_user?")
```
jrochkind, line 27: if you are actually raising, that's not just a deprecation -- and it will never ever get to the next line 28, the exception will stop it!

A deprecation would just maybe print a warning but actually work...

But maybe there's no need to have a deprecation phase in our own app, just remove the method? What do you think?

spec/factories/user_factory.rb
```diff
@@ -3,7 +3,11 @@
     email { "no-reply@sciencehistory.org"}
 
     factory :admin_user do
-      admin { true }
+      user_type { "admin" }
+    end
+
+    factory :editor_user do
+      user_type { "editor" }
     end
```
jrochkind, line 11: If this isn't actually being used ANYWHERE yet, maybe it's not needed? 

I could see actually *changing* the existing `:user` factory to `:editor_user`, and changing all uses of it in all code to be that... since all current `create(:user)` calls are really meant to create an editor user? Maybe in a next PR?

jrochkind:
Nice targetted PR!

There’s a rake task for making a user an admin, does it need to be fixed too?

Re UI:

the point of the "admin" bubble next to name in user list was to make sure you realized who was an admin, since it’s an important role.  The same blue bubbles for both with “editor” or “admin” maybe doesn’t succesfully do that anymore, they all kind of blend together. 

Maybe we can  tweak the UI a bit…  maybe only “admin” should be in a bubble, and other roles like “editor” are just words or something? In that same place or another line? Or something else?

Looking at bootstrap badge styles, maybe if you make non-admin roles have the "light" badge style that is basically just text but in the same font, that'd work! https://getbootstrap.com/docs/4.0/components/badge/

app/models/user.rb
eddierubeiz: I actually want the error (happy to not call it a "deprecation"). I definitely want to remove the method later on in the process -- after we drop the old `admin` column though.

eddierubeiz:
I did tweak the UI a bit in response to your comment (see https://staging-digital.sciencehistory.org/admin/users ) -- maybe I got carried away :) https://staging-digital.sciencehistory.org/admin/users

spec/factories/user_factory.rb
eddierubeiz: No you're right. We'll need it later but including it in this PR was a mistake.

app/models/user.rb
jrochkind: I'm not following why we want the error. 

With this error, if any code calls `admin?`, it'll get a `RuntimeError.new("Deprecated. Use admin_user?")`.   

Without this error, if any code calls `admin?`, it'll just get a `NoMethodError` I think?  Or will something else happen?  

Is it just that you prefer the former error becuase you or I might be confused by `NoMethodError`? Thoughts:

1. Nothing in the code should be calling `admin?` if it does it's a bug -- so this method should actually never be called.  Do we really want to leave a method in that should never be called?

I would personally remove it, becuase we control this code completely, there should simply be no code that still calls `admin?` anywhere, if we want to remove the method. If there accidentally is, the `NoMethodError` is good enough, compared to a custom error.  But no big deal. 

3. If it is to remain, "deprecated" is the wrong word -- something that is deprecated still works but will disappear in the future. This one as here does not still work. Also there is no reason to leave line 28 there, which is impossible to ever be reached. If the method is just going to raise, that can be the entirety of the method body.

jrochkind: Ah right, because the `admin` column means Rails does provide an `admin?` ethod, and we want to turn that into an error!

good call!

OK fine. just don't call it "deprecated" and remove the rest of the method body! :)

jrochkind:
That UI works for me for now, nice!

PR #1976, opened 2022-11-30
Add :see_unpublished_items permission

eddierubeiz:
In preparation to some upcoming changes to our permissions scheme, this adds a new :see_unpublished_items permission to our access policy.
All logged in users have this permission (and we don't actually foresee that changing). But instead of checking whether we can find a logged-in user, we'll be checking for this permission instead, which is tidier.
In addition, I made a small refactor of `ViewerMemberInfoSerializer`, which was taking `current_user` as an initializer argument. I think the new structure just makes more sense.

### DIFF
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -1,3 +1,8 @@
 class ApplicationComponent < ViewComponent::Base
+  delegate :current_user, to: :helpers
+
+  def can_see_unpublished_items?
+	AccessPolicy.new(current_user).can? :see_unpublished_items
+  end
 
 end
--- a/app/components/collection_metadata_component.html.erb
+++ b/app/components/collection_metadata_component.html.erb
@@ -1,5 +1,5 @@
 <div class="show-metadata">
-  <% if current_staff_user? %>
+  <% if can_see_unpublished_items? %>
     <p class="show-item-count">
       <%= "#{number_with_delimiter(public_count)} public #{'item'.pluralize(public_count)}, #{number_with_delimiter(all_count)} total" %>
     </p>
--- a/app/components/oral_history/downloads_list_component.rb
+++ b/app/components/oral_history/downloads_list_component.rb
@@ -1,7 +1,7 @@
 module OralHistory
   # Content of the "Downloads" tab for Oral History pages.
   class DownloadsListComponent < ApplicationComponent
-    delegate :format_ohms_timestamp, :current_user, to: :helpers
+    delegate :format_ohms_timestamp, to: :helpers
 
     # our combined_audio_derivatives helper methods
     delegate :m4a_audio_url, :derivatives_up_to_date?, :m4a_audio_download_url,
@@ -24,7 +24,7 @@ module OralHistory
     def all_members
       @all_members ||= begin
         members = work.members.includes(:leaf_representative)
-        members = members.where(published: true) if current_user.nil?
+        members = members.where(published: true) unless can_see_unpublished_items?
         members.order(:position).to_a
       end
     end
--- a/app/components/work_video_show_component.rb
+++ b/app/components/work_video_show_component.rb
@@ -9,7 +9,7 @@
 # it a separate class instead of trying to use lots of conditionals in one class, betting
 # that will be simpler overall, and allow them to diverge as more features are added.
 class WorkVideoShowComponent < ApplicationComponent
-  delegate :construct_page_title, :current_user, to: :helpers
+  delegate :construct_page_title, to: :helpers
 
   attr_reader :work
 
@@ -34,7 +34,7 @@ class WorkVideoShowComponent < ApplicationComponent
     return @video_asset if defined?(@video_asset)
 
   	@video_asset = (work.leaf_representative &&
-      (work.leaf_representative.published? || current_user.present?) &&
+      (work.leaf_representative.published? || can_see_unpublished_items?) &&
       work.leaf_representative) || nil
   end
 
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -121,7 +121,9 @@ class Work < Kithe::Work
   # with proper pre-fetches.
   def ordered_viewable_members(current_user:)
     members = self.members.includes(:leaf_representative)
-    members = members.where(published: true) if current_user.nil?
+    unless AccessPolicy.new(current_user).can? :see_unpublished_items
+      members = members.where(published: true)
+    end
     members = members.order(:position)
 
     # the point of this is to avoid n+1's, so let's set strict_loading, which
--- a/app/policies/access_policy.rb
+++ b/app/policies/access_policy.rb
@@ -22,6 +22,7 @@ class AccessPolicy
 
       can :admin, User
 
+      can :see_unpublished_items
       can :destroy, Admin::QueueItemComment do |comment, user|
         comment.user_id == user.id
       end
@@ -31,6 +32,8 @@ class AccessPolicy
     role :staff, proc { |user| !user.nil? } do
       can :read, Kithe::Model # whether publisehd or not
       can :update, Kithe::Model
+
+      can :see_unpublished_items
 
       can :destroy, Admin::QueueItemComment do |comment, user|
         comment.user_id == user.id
--- a/app/serializers/viewer_member_info_serializer.rb
+++ b/app/serializers/viewer_member_info_serializer.rb
@@ -9,11 +9,11 @@ class ViewerMemberInfoSerializer
 
   THUMB_DERIVATIVE = :thumb_mini
 
-  attr_reader :work, :current_user
+  attr_reader :work, :show_unpublished
 
-  def initialize(work, current_user:nil)
+  def initialize(work, show_unpublished: false)
     @work = work
-    @current_user = current_user
+    @show_unpublished = show_unpublished
   end
 
   def as_hash
@@ -40,7 +40,7 @@ class ViewerMemberInfoSerializer
   def included_members
     @included_members ||= begin
       members = work.members.order(:position)
-      members = members.where(published: true) if current_user.nil?
+      members = members.where(published: true) unless @show_unpublished
       members.includes(:leaf_representative).select do |member|
         member.leaf_representative &&
         member.leaf_representative.content_type&.start_with?("image/") &&
@@ -52,8 +52,6 @@ class ViewerMemberInfoSerializer
   def download_options(asset)
     DownloadOptions::ImageDownloadOptions.new(asset).options
   end
-
-
 
   def thumb_src_attributes(asset)
     derivative_1x_url = asset.file_url(THUMB_DERIVATIVE)
--- a/app/components/work_file_list_show_component.rb
+++ b/app/components/work_file_list_show_component.rb
@@ -11,7 +11,7 @@
 # a single download button.
 #
 class WorkFileListShowComponent < ApplicationComponent
-  delegate :construct_page_title, :current_user, to: :helpers
+  delegate :construct_page_title, to: :helpers
 
   attr_reader :work
 
@@ -24,11 +24,7 @@ class WorkFileListShowComponent < ApplicationComponent
   def member_list_for_display
     @member_list_display ||= begin
       members = all_members
-
-      if current_user.nil?
-        members = members.find_all(&:published?)
-      end
-
+      members = members.find_all(&:published?) unless can_see_unpublished_items?
       # omit "portrait" role
       members = members.find_all {|m| ! m.role_portrait? }
 
--- a/app/components/work_oh_audio_show_component.rb
+++ b/app/components/work_oh_audio_show_component.rb
@@ -2,7 +2,7 @@
 # show up in a fixed navbar.
 #
 class WorkOhAudioShowComponent < ApplicationComponent
-  delegate :construct_page_title, :current_user, to: :helpers
+  delegate :construct_page_title, to: :helpers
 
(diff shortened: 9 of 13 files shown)


app/components/application_component.rb
```diff
@@ -1,3 +1,8 @@
 class ApplicationComponent < ViewComponent::Base
+  delegate :current_user, to: :helpers
+
+  def can_see_unpublished_items?
+	AccessPolicy.new(current_user).can? :see_unpublished_items
+  end
```
jrochkind, line 6: You should be able to delegate `can?` and then just do `can? :see_unpublished_items`, I think that's better?

jrochkind:
I feel like this maybe needs a few specs showing that the gates really are there, that someone without the permission could not see the unpublished items?

Or were those specs already there? If you REMOVE the checks, and run all tests, do some tests fail?

app/components/application_component.rb
jrochkind: But again, the bigger issue is -- why have this helper method at all, can we just do `can? :see_unpublished_items` everywhere we'd call this helper method? As per other PR. 

But actually, ok, I'll put a comment in the ticket as a whole about something more generall....

jrochkind:
OK, I feel like the way `access_granted` really wants us to do this is considering the specific "item"....

`can? :view, work_or_asset`

Based on you defining the role to say that anyone can view a work that is published, but only certain roles can view a work that is not published.  (Hypothetically it could be based on other characteristics of the work too, of course -- or even, like, some people can see some unpublished works but not others -- but we don't plan to do that)

In fact, looking at our policy, we already _have_ that in our policy as `:read`!

The public can "read" only if published:

https://github.com/sciencehistory/scihist_digicoll/blob/ae358c93c643ec95a80a3f30f3aea34602e5a8c1/app/policies/access_policy.rb#L41-L43

Staff can read everything:

https://github.com/sciencehistory/scihist_digicoll/blob/ae358c93c643ec95a80a3f30f3aea34602e5a8c1/app/policies/access_policy.rb#L31-L32

So, oops, we already had this permission defined. So... we don't want to do it twice, differently. Refactor this to use `can? :read, actual_item` everywhere instead of `can? : :see_unpublished_items`, see how that goes? 

I think `:read` as the name specifically used like that is kind of the standard thing for access_granted, it's in the access_granted README.  And using access_granted like this, where only admins can see private posts, is one of hte examples in the access_granted README I think!

I won't in-detail review the other code, cause this requires enough refactor. 

The quesiton will be -- how do we check the permission we don't have a specific item then, we want to know in general if they can?  Perhaps we need a different permission name for that... but I wonder if access_granted has a solution for us... I feel like I've seen one once...  but I'm not seeing it in docs now. Do note there are some more docs in wiki though... https://github.com/chaps-io/access-granted/wiki/Role-based-authorization-in-Rails

app/components/application_component.rb
eddierubeiz: You can't delegate to an object, only to a class. But point taken -- I adjusted the code in the spirit of your idea, which is a good one

jrochkind: I think we're misunderstanding each other. Above where it already says `delegate :current_user, to: :helpers`? If you change this to `delegate :current_user, :can?, to: :helpers`, you should then be able to write simply `can? :whatever` in a ViewComponent. No? I'm not sure what you mean by "you can't delegate to an object only to a class", so I maybe missing something

eddierubeiz: You weren't missing something. I was missing something. Great idea- totally works.

app/components/application_component.rb
```diff
@@ -1,3 +1,3 @@
 class ApplicationComponent < ViewComponent::Base
-
+  delegate :can?, to: :helpers
```
jrochkind, line 2: Before when we were delegating current_user,. we just did it specifically on each component that needed it, instead of on ApplciationComponent. Have to write it each time, but it's more clear where it's coming from, right where you're using it.  Maybe stick with putting it on each component, what do you think?

app/components/collection_metadata_component.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="show-metadata">
-  <% if current_staff_user? %>
+  <% if can?(:read, Asset) && can?(:read, Work) %>
```
jrochkind, line 2: looks like we no longer use `current_staff_user?` anywhere... is there a method defined somewhere that needs to be removed too?

app/components/oral_history/downloads_list_component.rb
```diff
@@ -24,7 +24,8 @@ def initialize(work:, decorator:)
     def all_members
       @all_members ||= begin
         members = work.members.includes(:leaf_representative)
-        members = members.where(published: true) if current_user.nil?
+        show_all_members = (can? :read, Asset) && (can? :read, Work)
+        members = members.where(published: true) unless show_all_members
```
jrochkind, line 28: I think avoiding the temporary variable is a lot more clear:

```
unless (can? :read, Asset) && (can? :read, Work)
   members = members.where(published: true) unless show_all_members
end
```

Although yeah, i guess that condition is a bit confusing too, I don't totally understand how access_granted works hwen you pass that class instead of an isntance... or why we need/want to check both...    For the second question, should we just replace it everywhere with `can? :read, Kithe::Model` actually?  

But for the first question... `can :read, ModelClass` is magically true only of those accounts that have the role where we say they can read regardless of published status?  

Hmm, now I'm wondering... maybe for this one we _do_ want to make a helper method in `ApplciationHelper`, like `can_see_unpublished_records?`, since it is a bit weird/unclear, document it there, and delegate to it from the view components? I dunno if that improves things or not, just kind of writing as I think.

app/components/work_video_show_component.rb
```diff
@@ -32,9 +32,9 @@ def video_src_url
   # the representative, if it's visible to current user, otherwise nil!
   def video_asset
     return @video_asset if defined?(@video_asset)
-
+    show_all_members = (can? :read, Asset) && (can? :read, Work)
```
jrochkind, line 35: Yeah, as you repeat this `show_all_members` thing over and over... I'm leaning toward, yeah, let's make an ApplciationHelper helper method for this, and document it?  (and maybe the definition is just `can? :read, Kithe::Model` if that works.

jrochkind: OR a method on `User`?  `current_user.can_see_unpublished_records?` ?

app/policies/access_policy.rb
```diff
@@ -29,8 +29,15 @@ def configure
...
-      can :update, Kithe::Model
+
+      can :read, Work
+      can :update, Work
+
+      can :read, Collection
+      can :update, Collection
+
+      can :read, Asset
+      can :update, Asset
```
jrochkind, line 40: Oh here you are repeating these again... it's _supposed_ to work with the super-class... I'm wondering if it's breaking becuase of the thing I don't totally understand with access_granted where you can ask for a class instead of an instance?  If we change all those places you ask for a class to `can? :read, Kithe::Model` instead of `can?(:read, Asset) && can?(:read, Work)`.... maybe we can leave the policy in terms of just the single superclass? And I think it reads better anyway?

app/serializers/viewer_member_info_serializer.rb
```diff
@@ -9,11 +9,11 @@ class ViewerMemberInfoSerializer
...
 
-  attr_reader :work, :current_user
+  attr_reader :work, :show_unpublished
 
-  def initialize(work, current_user:nil)
+  def initialize(work, show_unpublished: false)
     @work = work
-    @current_user = current_user
+    @show_unpublished = show_unpublished
   end
```
jrochkind, line 17: I guess I can see how this is better, to just pass in the the boolean instead of the current_user.... until/unless this thing needs to determine something else baed on the current_user too, some more complicated permission....  I guess we can always pass in a second argument for that specific thing then. I think it's fine to pass in current_user, but also I can see the idea to just tell the presenter what it should do and not make it figure it out for itself... I dunno, ok, fine, what you chose!

app/components/collection_metadata_component.html.erb
eddierubeiz: Yup - with the other branch we were working on merged, we can in fact get rid of it.

app/components/application_component.rb
eddierubeiz: Oh, I can't get it to work; raises an `undefined method` error. Sort of annoying since several of the subclasses are delegating other methods to `:helpers`. For now I'm just leaving this as is with a comment.

app/components/oral_history/downloads_list_component.rb
eddierubeiz: Unfortunately, AccessGranted is not smart enough to deduce that e.g. `can? :destroy, Kithe::Model` in the access_policy file implies `can? :destroy, Work`.

I also found your argument in https://github.com/sciencehistory/scihist_digicoll/pull/1976#issuecomment-1334473404 convincing . I find using the basic can? syntax in the various parts of the code where it's needed is actually helpful in this part of the process. So I'm keeping everything simple -- if a little verbose -- as of this refactor. 

My feeling here is that pnce we actually get the various permissions changes sorted out and tested, we can of course explore more concise and elegant formulations in a separate PR.

app/policies/access_policy.rb
eddierubeiz: It would read better but AccessGranted won't let us do that, as far as I can tell.

app/components/oral_history/downloads_list_component.rb
jrochkind: OK, so:

a) I still think factoring this confusing check out into re-usable helper is a good idea either way, but

b) I want to play with access_granted a bit more.  
   * i believe you that if the policy says `can :read, Kithe::Model`, then checking `can? :read, Work` won't succesfully return true -- but I want to investigate this more and see if we can submit a PR to fix it in access_granted
   * But even notwithstanding -- okay, butr what if we do `can :read, Kithe::Model` in the policy, but just _check_ `can? :read, Kithe::Model` instead? That should work. And I think right now every single place does `can?(:read, Work) && can?(:read, Asset)` anyway -- it seems _preferable_ to say `can? :read, Kithe::Model` there _anyway_, plus put `can :read, Kithe::Model` in the policy.

eddierubeiz: (Summary -- this is what we ended up doing.)

app/components/work_video_show_component.rb
eddierubeiz: We did create an ApplicationHelper helper method, which in turn looks at a method in the policy class.

app/policies/access_policy.rb
```diff
@@ -23,9 +23,26 @@ def configure
...
-      can :read, Kithe::Model, { published: true }
+      can :read, Kithe::Model do |mod, user|
+        # mod could be any of Kithe::Model, Collection, Work, Asset,
+        # *or* an instance of any of the above classes.
+        #
+        # If mod is any of the above classes, definitely return false:
+        # the public is not allowed to read *all* possible instances of any of these classes.
+        #
+        # If mod is an instance, return true only if the instance is published.
+        mod.class < Kithe::Model && mod.published?
```
jrochkind, line 36: Better way to write `mod.class < Kithe::Model` is just `model.instance_of?(Kithe::Model)`

app/models/work.rb
```diff
@@ -120,8 +120,13 @@ def all_descendent_members
   # All DISPLAYABLE (to current user) members, in order, and
   # with proper pre-fetches.
   def ordered_viewable_members(current_user:)
+
+
     members = self.members.includes(:leaf_representative)
-    members = members.where(published: true) if current_user.nil?
+
+    show_unpublished = AccessPolicy.new(current_user).can_see_unpublished_records?
+    members = members.where(published: true) unless show_unpublished
```
jrochkind, line 128: just make this an if/unless instead of using a temporary local variable

app/serializers/viewer_member_info_serializer.rb
```diff
@@ -40,7 +40,7 @@ def as_hash
   def included_members
     @included_members ||= begin
       members = work.members.order(:position)
-      members = members.where(published: true) if current_user.nil?
+      members = members.where(published: true) unless @show_unpublished
```
jrochkind, line 43: we have a method `show_unpublished`, so should access that instead of the instance variable. Our "style guide" is kind of never to access instance variables directly outside of setting them in intializer, caues it's easy to make a typo that doesn't get caught and other reasons.

jrochkind (approved):

app/models/work.rb
eddierubeiz: Done!

app/policies/access_policy.rb
eddierubeiz: Done.

eddierubeiz: Oops - not quite.
```
irb(main):004:0> goat = Asset.new()
=> #<Asset:0x00000001190f2070
...
irb(main):005:0> goat.instance_of? Asset
=> true
irb(main):006:0> goat.instance_of? Kithe::Model
=> false
irb(main):007:0> goat.class < Kithe::Model
=> true
```

app/serializers/viewer_member_info_serializer.rb
eddierubeiz: Fixed.

app/policies/access_policy.rb
jrochkind: oops, you're right, `kind_of?`.  I could have sworn they were synonyms, but I guess I got confused! Good job double-chekcing me, thank you!

PR #1981, opened 2022-12-05
gemfile update to rails 7

jrochkind:
* With new version of kithe with rails7 bugfix

* update to a new questioning_authority (qa) gem release with rails 7 support.

* DescriptionDisplayFormatter#format_plain, make html_safe? false consistent with past pre-Rails7 behavior, although not sure if it matters. 
   * This did make me realize some things about how `<` and `>` are escaped/double-escaped in our code that I don't like, but a legacy issue not new for  rails 7 update -- ended up fixing here too. 

Ref #1980


jrochkind:
Noticed an unrelated area of code that could be improved because it happened to be near something that had to be fixed for Rails 7. 

Fixed it here in this PR because it's simple, and would conflict with this PR if a separate PR, so just seemed easiest. 

* DescriptionDisplayFormatter#format_plain should not produce html entities in output

jrochkind:
There does seem to be some weirdness with ingest going on, that is not caught by our CI. :(. Will need to investigate further.

jrochkind:
OK, in process of upgrading things, we ended up incidentally upgrading to a just-released `net-protocol` gem -- which ended up breaking shrine. I noticed issues filed on shrine repo. 

So we also lock to net-protocol `< 0.2.0` in this PR.  Links to relevant shrine issues in code comments by requirement lock in Gemfile.

jrochkind:
OK, while it makes me nervous, I think I'm ready to say this is ready for merge!

Let's not deploy until next week though, so maybe not merge until then.

PR #1983, opened 2022-12-06
Adding tests for our access policy

eddierubeiz:
As we work towards ref #1948 , so far we've made only minor changes to the access policy (compare [Current](https://github.com/sciencehistory/scihist_digicoll/blob/master/app/policies/access_policy.rb) versus [2019](https://github.com/sciencehistory/scihist_digicoll/blob/c126e84e/app/policies/access_policy.rb) ) 

Before we move to ref #1976 and subsequent PRs, let's make sure we agree how `access_granted` 's hierarchical permissions work with single-table inheritance. We may want to submit a PR to `access_granted` in the process.

I'm adding a test which illustrates what I'm getting at here.


spec/policies/access_policy_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+    end
+    it "can read a Collection" do
+      skip "Policy mentions Kithe::Model, not its subclasses."
+      expect(policy.can?(:read, Collection)).to be true
+    end
+  end
+
+  describe 'public' do
+    let(:policy) { public_policy }
+    let(:user) { staff_user }
```
jrochkind, line 118: why is user staff_user if policy is public policy?

jrochkind:
Just as far as code style for writing tests. Instead of putting all these at the top:

```ruby
  let!(:admin_user)  { FactoryBot.create(:admin_user,  email: "admin@b.c") }\
  let!(:staff_user) { FactoryBot.create(:user, email: "staff@b.c") }

  let!(:admin_policy) { AccessPolicy.new(admin_user) }
  let!(:staff_policy) { AccessPolicy.new(staff_user) }
  let!(:public_policy) { AccessPolicy.new(nil) }
```

And then referring to them in individual blocks:
```ruby
    let(:policy) { admin_policy }
```

Please just define them in place when you need them.  In the block about an admin user, just write, for instance:

```ruby
    let(:user) { create(:admin_user,  email: "admin@b.c") }
    let(:policy) { AccessPolicy.new(admin_user) }
```

Avoiding the indirection makes it easier for a reader to actually see what data is being used with what test, put the definition closer to where it's being used. There's also no reason to define all these things for all tests -- when each block will only use one of them.  A problem made even worse by using `let!` (exclamation point) instead of `let` -- the exclamation point means all of these things will be created for every test, whether they are used or not, which will have performance probelms, especially for `create`, which goes to the db. 

So also use `let` instead of `let!` unless there is some reason you need the blocks to be executed before they are first referred to.

spec/policies/access_policy_spec.rb
eddierubeiz: Yup, sorry!

eddierubeiz:
Thanks for your comments! Cleaned up a bit more too and added some sub-`context`s.

jrochkind:
Things can be simplified when this PR to access-granted is merged and released: https://github.com/chaps-io/access-granted/pull/56

jrochkind:
Even an admin can't delete a queue item comment by someone else?? That's actually probably an oversight, I feel like that should be true of ordinary staff but admin should be able to delete anything! but so be it!  writing the tests exposed it, so good there.

PR #1984, opened 2022-12-07
Delegate can? on individual components only where needed instead of globally on ApplicationComponent

jrochkind:


eddierubeiz:
I'm FLUMMOXED as to why this didn't work for me! But.... many thanks! I'll take it! :)

PR #1988, opened 2022-12-12
Adding create permission

eddierubeiz:
Staff and admin are allowed to create works, collections and assets. This was always true but we're making it explicitly true (and will be checking against it soon.)


eddierubeiz:
No problem! I opened https://github.com/sciencehistory/scihist_digicoll/pull/1991 , which does make use of it.

PR #1989, opened 2022-12-12
use can_see_unpublished_records? helper method in CollectionMetadataComponent

eddierubeiz:
Ref #1948 
Makes this component consistent with the others.


jrochkind:
nice cleanup thanks

PR #1991, opened 2022-12-13
Check CRUD permissions on admin works controller and its views and components

eddierubeiz:
Ref #1948.

Public-facing pages are unaffected by this PR.
This does not change the functioning of the app for any of its current users.

- In `app/controllers/admin/works_controller.rb`, add `authorize!` where warranted.
- In `spec/system/work/demote_to_asset_spec.rb`, use an `admin` user, since the test involves deleting a work.
- In `app/views/admin/works/*` and `app/components/oral_history/admin/*`, and `app/helpers/admin_menu_helper.rb`,  grey out (but don't hide) various UI components based on permissions.

We're making liberal use of tricks like:
```class: "btn btn-primary btn-danger #{'disabled' unless can?(:update, @work)}",```
or:
```submit_tag "Upload", class: "btn btn-primary", disabled: !can?(:update, @work)```
This means we're calling `can?` a lot in the template, instead of only once or twice in the controller. I actually feel fine about this given the context.

### CODE REVIEW

File: app/components/oral_history/admin/combined_audio_derivatives_component.html.erb [hunk 1]
```diff
@@ -39,9 +39,9 @@
       </div>
     <% end %>
 
-    <% if work_available_members? %>
+    <% if work_available_members?%>
```
jrochkind, line 42: removing the space here is not an improvement but no big deal who cares just saying heh

jrochkind:
I'm still worried that keeping this up is going to be a dev burden, remembering to wrap like every single function we ever had from here on out -- it seems a significant burden, easy to trip up and forget one here or there. 

 But we''ll see how it goes, I count on you to help review my PR's and try to catch if I forget to guard something!

jrochkind (approved):

PR #1995, opened 2022-12-13
"Review requested" feature for Works

jrochkind:
- basic review_requested toggle on admin screens
- simple filter for review requested

We want to wait for stakeholder review to merge, but ready for code review.

### DIFF
--- a/app/views/admin/works/index.html.erb
+++ b/app/views/admin/works/index.html.erb
@@ -54,7 +54,7 @@
     </div>
 
     <div class="admin-filter">
-      <label for="q_parent_id_null">Child Works In Results</label>
+      <label for="q_parent_id_null">Child Works</label>
       <%= select_tag "q[parent_id_null]",
             options_for_select({"Yes" => "", "No" => true}, params[:q][:parent_id_null]),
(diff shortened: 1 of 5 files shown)


jrochkind:
@eddierubeiz Ready for code review, if you have time this week. 

I definitely felt some brain fog when writing this one, so if there's anything you think can/should be done different or have questions about, please feel free.

app/views/admin/works/index.html.erb
```diff
@@ -54,12 +54,20 @@
     </div>
 
     <div class="admin-filter">
-      <label for="q_parent_id_null">Child Works In Results</label>
```
eddierubeiz, line 57: Thanks for the label change. Don't know why I got so confused last week but this is easier to understand.

eddierubeiz (approved):

app/views/admin/works/index.html.erb
jrochkind: Ha, to me the new label is _harder_ to understand, I changed it just to save space to add another control in, worried that it would be harder to understand!

PR #1997, opened 2022-12-14
Don't hardcode any user type labels into the list of users

eddierubeiz:
Ref #1948 
Admin is the exception here, but there's really no need to hard-code non-admin user types' labels in this template.


jrochkind:
makes sense, thanks. approved.

jrochkind (approved):

PR #1998, opened 2022-12-14
upgrade yarn jquery to recently released 3.6.2

jrochkind:
Yarn makes it harder than it should to do this reliably. I first ran:

   yarn upgrade jquery

But that resulted in "split" jquery dependencies in yarn.lock that I didn't really want -- indirect dependencies on jquery were not upgraded in yarn.lock. And I'm not sure what that does.  So I had to manually delete the split line from the yarn.lock, then run `yarn install` again, to resolve it.

Probably should document the weird tricks I've learned for upgrading "indirect" dependencies in yarn. Very frustrating.


eddierubeiz:
Thanks for dealing with this headache!

PR #2000, opened 2022-12-16
Remove old admin column from users table

eddierubeiz:
Ref #1948
Remove the now-obsolete `admin` column and the corresponding `admin?` user method.

This is our first migration since we moved to Rails 7, so this PR also includes some unrelated changes to `schema.rb`. Notably, the default precision for `datetime` columns is changing from 0 to 6. ([more info](https://rubyonrails.org/2022/2/11/this-week-in-rails-rails-7-0-2-schema-versioning-based-on-the-rails-version-and-more-cbcb0592))

See also: https://github.com/rails/rails/pull/44286


jrochkind:
Thanks for figuring out the precision/Rails 7 stuff, and for link. 

You didn't make manual changes to the schema.rb, that changes are the result of letting Rails migrate it, correct? So you're confident it accurately represents our current database?

Approved assuming that's so.

jrochkind:
Ideally, the commit message for the schema change would include more information about how those changes were made, as well as a link to the relevant Rails docs there. 

That's more a note for the future than a requirement you rewrite the commit message now. 🤷

Issue #2001, opened 2022-12-19
Fix broken images referenced in CSS


jrochkind:
When we moved all our CSS and JS over to vite, we forgot that we were breaking sprockets-style references to Image assets in CSS files. 

These can be found by searching for `asset-url`  in our CSS files. The method of breaking is "silent" -- before these were interpreted by sprockets to resolve to a proper image URL, now just the literal `asset-url(something)` winds up in CSS delivered to browser, which the browser does nothing with -- just missing or broken images in browser. 

Here is my search for `asset-url` in ./app/frontend:

```
app/frontend//stylesheets/local/oral_history_splash.scss:    background-image: asset-url("oral_history_wide_banner.jpg");
app/frontend//stylesheets/local/oral_history_splash.scss:    background-image: asset-url("oral_history_collection_splash.jpg");
app/frontend//stylesheets/local/scihist_masthead.scss:  background-image: asset-url('scihist_top_nav_sym.png');
app/frontend//stylesheets/local/scihist_viewer.scss:      background: asset-url('spinner.svg') no-repeat center;
app/frontend//stylesheets/local/scihist_viewer.scss:        background: $brand-image-placeholder-color asset-url('static-progress.svg') no-repeat;
app/frontend//stylesheets/local/work_show.scss:         background: $brand-image-placeholder-color asset-url('static-progress.svg') no-repeat;
app/frontend//stylesheets/local/footer.scss:  background-image: asset-url('scihist_foot_sym.png');
```

To fix these, we will need to move at least _these_ images over to vite instead of sprockets.

Ideally, we might not want to split our images between sprockets and vite, as it's confusing. 

But per conversation in team meeting today, we decided to start by moving over only the images referenced in CSS, then maybe in January work on moving the rest over. 

I may do this move as a COPY (we will now have two copies of those images), in case they are also currently referenced by Rails sprocket helpers somewhere.

jrochkind:
Having a bit of trouble figuring out the best way to do it, asked on vite-ruby github discussion https://github.com/ElMassimo/vite_ruby/discussions/321

PR #2002, opened 2022-12-19
Move images referenced in vite-controlled CSS url() to vite

jrochkind:
Ref #2001

I think we won't actually move ALL iamges anytime soon at all, we'll just leave it like this for a while?


eddierubeiz:
Yeah, I can live with this. Thanks for updating docs, too.

PR #2006, opened 2022-12-20
Add new "staff" user type

eddierubeiz:
Ref #1948 .

More important changes:
- Update the user model to include the new user type
- Update the policy (and attending tests) to define the new user type's permissions.
- Update the factories to take it into account
- Made the user list view to be more tolerant in the case of a user not having a user type. (this might occur right after we deploy).

The bulk of this PR, however, is just updating tests to specify the user type for tests where the user is meant to be logged in.
For now, we are not adding any tests.

TODO:

- [ ] Can we DRY up the user policy a bit without breaking the test? Certainly seems like we should be able to;
- [ ] Write a test logging in as the new user type and attempting to break various things;


eddierubeiz:
**Q:** Can we DRY up the user policy a bit without breaking the test?
**A:** Apparently not: in our app, an `Admin` can't also be a `Staff` or an `Editor`, and [roles inherit from less important roles below them, but only from roles which apply to the given user](https://github.com/chaps-io/access-granted/wiki/Role-based-authorization-in-Rails). Not the end of the world. Added some explanatory notes in the policy. If there's some clever way around this, it probably belongs in another PR.

jrochkind:
> Q: Can we DRY up the user policy a bit without breaking the test?

> A: Apparently not: in our app, an Admin can't also be a Staff or an Editor, and [roles inherit from less important roles below them, but only from roles which apply to the given user](https://github.com/chaps-io/access-granted/wiki/Role-based-authorization-in-Rails). Not the end of the world. Added some explanatory notes in the policy. If there's some clever way around this, it probably belongs in another PR.

I think you can. 

```
role :admin, proc { |user| user&.admin_user? }

role :editor, proc { |user| user&.editor_user? || user&.admin_user? }

role :staff_viewer, proc { |user| user&.staff_viewer_user? || user&.editor_user? || user&.admin_user? }
```

We just have to say that someone with admin role in DB gets `editor` and `staff_viewer` too. What do you think?

eddierubeiz:
Love this idea.

eddierubeiz:
Worked great!

spec/system/admin/staff_user_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'rails_helper'
```
jrochkind, line 1: So our system specs have become a bit disorganized anyway, but.... in an ideal world I feel like we'd have different system specs for different _pages_ (or different specific workflows/use cases), and for each one have a separate describe/context block for staff users, that made sure they couldn't do what they coudln't do. 

In some cases if we didn't already have such a thing, becuase the page was so simple it didn't need testing, maybe the "as a staff user" block would be the only one. 

Having it like this... is kind of a departure of _most_ of the other ways we've done system specs. And may be hard to maintain as the UI changes, we have kind of one giant spec here that cuts across all sorts of things. 

But this is kind of a pain to manage -- as evidenced that our specs are already kind of inconsistent (note we have a bunch of admin-related system specs in `./spec/system/work`, that maybe should be in `./spec/system/admin/work`?) -- so I make this comment, but then leave it to you -- if you think leaving this as a "staff user [system] spec" is in fact way the best way to make this maintainable and manageable, fine. 

BUT, one thing.  This probably should be multiple `it` blocks either way, not one giant it block. In general, there `it` blocks should be focused on one targetted thing. One reason we disobey that for _system_ (capybara) specs specifically is for performance, each page load is very expensive. But in this case...  you're kind of just going to a bunch of different pages un-related to each other without preserving any state.  I think each page visit can and should be a separate `it` block, and there will be no performance trade-off, and it will be easier to read and have more targetted spec failures. Each separate page taht you have separated as a comment -- make a separate example (`it` block).  Multiple interactions with the same page can still be within the same example block.

eddierubeiz: Good call -- done.

jrochkind (approved):

PR #2007, opened 2022-12-20
Use cannot? instead of !can?

eddierubeiz:
One of several refactors before we actually deploy #1948  to prod.


jrochkind:
@eddierubeiz I don't understand what "actually deploy to prod" means when we are merging and deploying PR's as we go. 

At what point are you saying we need to do this? Before we merge other PR's that will make the "Staff viewer" role a thing?

jrochkind:
!can is fine with me, but this is fine iwth me too, sure why not.

Issue #2009, opened 2023-01-04
Move user type test methods into access policy


eddierubeiz:
To preserve consistency, controllers and views should be asking about specific permissions (e.g. `can? :publish, @collection`). Giving controllers and views a way to ask the user object about its role doesn't make sense to me, and is reminiscent of `user.logged_in?` and similar patterns, which we've finally gotten rid of.

Let's make them private methods, and move them into the user policy, in the same class that's actually calling them.

jrochkind:
While I understnad what you mean, I think this is unnecessary purity, and how it is now is nice and simple.

PR #2010, opened 2023-01-04
move user type test methods into access policy

eddierubeiz:
Ref #2009 
More consistent and sane, I believe.


jrochkind:
Has a conflict that has to be resolved, but I'll mark it approved!

jrochkind (approved):

Issue #1949, opened 2022-11-15
Remove Exhibition metadata field


apinkney0696:
To be completed once the backfilling of exhibition collections is finalized.

Issue #1920, opened 2022-10-31
Changes to support "Exhibition"  type of collection


jrochkind:
We're going to use the existing "Collections" feature to support our aggregation of digitial items featured in an exhibition, by making the following small tweaks:

1. Add a new Collection "department" called "Exhibition" (even though it's not really a department, hopefully this won't be confusing to metadata staff)
2. On the Exhibition page itself, the little genre title "Collection" should say "Exhibition" instead.  It should _not_ be a link to the collections list page.  I guess it will not be a link at all?
3. "Exhibition" type collections will NOT show up on collections list page https://digital.sciencehistory.org/collections
4. On individual work page, there will be a new line _under_ "Collections" labelled "Exhibited in" with (one or more!) "Exhibition" type collections it's assigned to. 

@apinkney0696 please review to make sure this matches your understanding and let me know if i'm missing anything!

jrochkind:
(I don't know how this wound up an "epic" in Zenhub, I didn't mean to use that feature which I'm not familiar with -- but don't know how to turn it back!)

apinkney0696:
This all looks good to me and covers everything we discussed earlier. Thanks!

PR #2015, opened 2023-01-09
Remove Work#exhibition attribute

jrochkind:
We don't record exhibitions that way anymore, instead we use a feature built on Collection model. Ref #1949, #1920.

Wow there were a lot of places here and there that referenced it! Hope I got em all, tests pass anyway. 

Includes a rake task that must be run to remove all data from this attribute in DB. Rake task operates through #json_attributes only, without referencing exhibition attribute itself, so can be run even after this is merged and exhibition attribute does not exist. 

        ./bin/rake scihist:data_fixes:remove_exhibition_attribute_data

Person who merges should be prepared to deploy and run rake task. 

- [ ] rake task run in production

### DIFF
--- a/app/components/work_show_info_component.html.erb
+++ b/app/components/work_show_info_component.html.erb
@@ -110,8 +110,6 @@
 <table class="work chf-attributes">
   <%= render AttributeTable::RowComponent.new(:department, values: [department], link_to_facet: "department_facet") %>
 
-  <%= render AttributeTable::RowComponent.new(:exhibition, values: exhibition, link_to_facet: "exhibition_facet", alpha_sort: true) if can? :access_staff_functions%>
-
   <% if public_collections.present? %>
     <tr>
       <th scope='row'>Collection</th>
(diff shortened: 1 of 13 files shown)


eddierubeiz:
There's an extra backup file in here ( `spec/policies/access_policy_spec.rb-NEW` ).
I'm also curious what `spec/args_template.rb` is doing; you don't appear to be using the `allow [...] to receive` recipe anywhere in the tests of this PR. Am I missing something?

jrochkind:
Gah, thank you Eddie, good catches!

There were so many changed files that I ended up accidentally committing some leftover temp files that weren't even from this specific work! They were just leftover temp working files. Will try to clean up!

jrochkind:
OK, removed those two files, I think they were the only ones that snuck in! Thank you for your review!

eddierubeiz:
No worries. I think it's just `Procfile.dev` - other than that it looks great and I approve it!

jrochkind:
Thank you for catching that again, I will remove Procfile.dev! Hooray for code review!

jrochkind:
Rake task to rehabilitate the data was having some problems on prod... may run more manually....

jrochkind:
resolved

Issue #2014, opened 2023-01-09
Cart option in Production Admin Menu


apinkney0696:
The cart option in Production Admin Menu has disappeared. This ticket is to bring it back, please.

jrochkind:
@eddierubeiz My guess is this was an accidental thing in your refactoring of all those menus for the new staff role access controls?  (The maintenance costs of trying to do this fine-grained access control are real!) Can you take this one?

PR #2016, opened 2023-01-09
Permissions adjustments

eddierubeiz:
After pushing the first round of changes for our new permissions(ref #1948 ) we found a couple bugs; this PR fixes them.
* Editors should be able to publish and unpublish (per the PR).
* Fix a bug in the left-hand admin menu. (ref #2014 )
Adding tests for both these bugs.


jrochkind:
Merging, and I'll go ahead and deploy to prod right now.

PR #2018, opened 2023-01-09
Do not tell solr fulltext fields to omitNorms, not sure why we thought this made sense

jrochkind:
OK when I try this out in staging... it is expected to only effect relevancy ranking of things matching our "fulltext" fields. 

Comparing staging to prod. In everything I think of to search...  it does not effect the first few results. It effects the later results, but minorly... like certain results are in swapped order, etc. 

It doesn't make a huge difference... I have no real way to judge if it's better or worse.  I'm not sure it's worth taking some of @apinkney0696 's time to test it, it's just, like, tiny changes that it's hard to say if they make a difference. 

I'm inclined to merge the change just to make these fields more normal and typical, not have a non-standard configuration that makes a change we actually don't have a reason to think is an improvement.

### DIFF
--- a/solr/config/schema.xml
+++ b/solr/config/schema.xml
@@ -266,13 +266,13 @@
 
          https://lucene.apache.org/solr/guide/8_0/highlighting.html#Highlighting-SchemaOptionsandPerformanceConsiderations
        -->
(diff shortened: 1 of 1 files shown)


jrochkind:
Actually, maybe I found a search where I think this PR makes it worse!

So maybe we should play with it a bit. @eddierubeiz ,you interested in doing any experimentation with it?

jrochkind:
Decided we actually like it better the way it is, not doing this.

Issue #2019, opened 2023-01-10
Bug in new Request for Review feature


apinkney0696:
I can't seem to remove any items from the review list. When I toggle off the review button on a record, it is still appearing for me in the Works menu with the "Requested" review filter applied.

jrochkind:
I think I've reproduced; I'll look into it!

jrochkind:
@apinkney0696 should be fixed! thanks, sorry!

PR #2020, opened 2023-01-10
Fix review requested filter

jrochkind:
Ref #2019

We were looking filtering for "not nil" but in fact our code left json `true` or `false`. We can use `jsonb_contains` to do the casting for us, we actually want to filter for records where value is `true` only, not `not nil`.

Still no tests, sorry, I'm being lazy. Tested manually.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -596,7 +596,7 @@ class Admin::WorksController < AdminController
       end
 
       if params[:q][:review_requested].present?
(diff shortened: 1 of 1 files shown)


jrochkind:
@eddierubeiz Will merge myself on your day off, to get the bugfix deployed!

jrochkind:
(Yes, I realize if I had written tests in the first place this bug wouldn't have been released... and I still have written no tests. Terrible.)

PR #2026, opened 2023-01-18
update general dependencies

jrochkind:
General update of dependencies, including major versions of a couple.

Ran `bundle outdated` after `bundle update` for candidate that need major version bump in our Gemfile. 

- bundle update followed by bundle exec vite upgrade -- now we're on on vite 3.0
- fix for small backwards incompat change in device_detector 1.1.0
- update puma to 6.0
- update alba (json serializer) from 1.x to 2.x


jrochkind:
One major dependency we're outdated on is rack:  rack 3.0 is released, but we're still on rack 2.0 -- looks like Rails isn't necessarily compatible with rack 3.0 yet, so we'll keep sitting on that. https://github.com/rails/rails/pull/46594

jrochkind:
OK, smoke-tested on staging, seems ok.

ioquatix:
Thanks this is helpful to me too.

jrochkind:
@ioquatix Glad it is! You aren't using our app somehow are you?! (I mean, anyone is welcome to, it would just be surprising!).

ioquatix:
Testing all the updated dependencies in real world applications is useful because it helps us learn about issues, I don't use this app specifically but as you linked through, I'm working on Rack 3 which should generally be a good update. If you have time to test Rack 3 on a Rails branch, let me know.

Issue #1968, opened 2022-11-28
OHMS TOC keywords -- Solr boost higher than fulltext


jrochkind:
* also applies to subjects from #1954 
* They DO still need to be included in search results highlighting

PR #2036, opened 2023-01-25
solr indexing boost OH keywords/subjects higher than other fulltext

jrochkind:
ref #1968


jrochkind:
tested in staging compared to prod, seems to do what is intended.

eddierubeiz:
Noted. Merging.

Issue #1959, opened 2022-11-21
More tests relating to collections thumbnails in the admin pages.


eddierubeiz:
Subsequent to bug #1956, I feel our tests should **try** to prevent us from changing the admin pages without taking into account that some of them have a `Collection` as a parent. I'd like tests of the the asset list, view and edit pages to explicitly deal with this case.

Ideally the tests of the admin pages should ensure that if an asset happens to be a collection thumbnail:
- it doesn't break the display, at a minimum;
- any references or links to its parent use appropriate copy;
- any UI that is specifically geared towards non-collection-thumbnail assets either behaves gracefully or is hidden.

PR #2037, opened 2023-01-25
Better coverage for admin assets controller

eddierubeiz:
Should make it more difficult in the future to introduce bugs arising from the slightly different behavior of regular assets and collection thumbnails.

Ref https://github.com/sciencehistory/scihist_digicoll/issues/1959


jrochkind:
ok seems reasonable, thanks! I like that it's relatively faster controller spec instead of so slow system spec.

Issue #1761, opened 2022-06-28
upgrade heroku to heroku-22 stack


jrochkind:
**(This is blocked by #2090, we want to do that first)**

Heroku released a new stack (like the OS etc that runs on dynos), heroku-22. 

There's no rush to upgrade, we are on `heroku-20`, which is [supported until April 2025](https://devcenter.heroku.com/articles/heroku-20-stack#support-period)

Additionally, at present heroku-20 is still the default stack, if docs are correct? https://devcenter.heroku.com/articles/stack#default-stack

But it's good to stay up to date, filing the ticket to just put it on our radar.

The downside is some things may not actually work -- main potential challenge would be our buildpacks and custom software we install. It's possible some of our custom buildpacks are not yet for heroku-22, not sure. 

https://devcenter.heroku.com/articles/heroku-22-stack
https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack

jrochkind:
If anything would possibly cause a problem here, it'd be our non-default buildpacks to get things like ffmpeg and vips installed.

jrochkind:
Need to make sure all our extra software is properly installed https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/2088271873/Derivative+generation+on+Heroku

jrochkind:
## Testing

* [x]  Went through diagnostic script at https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/2088271873/Derivative+generation+on+Heroku  => all wekk
* [x] Ingest image
* [ ] Ingest video
* [ ] Re-generate combined audio derived for OH
* [ ] Do an on-demand whole-work PDF derivative

jrochkind:
OK, so ffprobe/ffmpeg are segfaulting when called with a URL input argument -- which our code does, intentionally, as a performance enhancement. 

(ffprobe does not need to download the whole file, it only downloads the prefix it needs when given a URL)

I've filed this as an issue on the heroku buildpack: https://github.com/heroku/heroku-buildpack-activestorage-preview/issues/27

We can (again) investigate if we can find another succesful way to get ffmpeg/ffprobe installed on heroku-22.

jrochkind:
Aha, interestingly trying to use this buildpack to get ffmpeg:

https://github.com/jonathanong/heroku-buildpack-ffmpeg-latest.git

I am still getting the same segfault! From eg

[...]

PR #2038, opened 2023-01-25
Move collection thumbnail code out of templates

eddierubeiz:
Ref #1761 

- New asset model method `collection_thumbnail?()`;
- `parent_path(asset)` controller helper method;
- Simplify the templates accordingly;
- When you view a collection thumbnail:
  - Show the storage type, even if you can't edit it (I figure there's no point in *not* displaying it).
  - Show an edit button, which takes you to the edit form for the collection. (Do we need this?)

### DIFF
--- a/app/controllers/admin/assets_controller.rb
+++ b/app/controllers/admin/assets_controller.rb
@@ -23,6 +23,8 @@ class Admin::AssetsController < AdminController
   def show
     @asset = Asset.find_by_friendlier_id!(params[:id])
     authorize! :read, @asset
+
+    @edit_asset_path = @asset.collection_thumbnail? ? edit_admin_collection_path(@asset.parent) : edit_admin_asset_path(@asset)
     if @asset.stored?
       @checks = @asset.fixity_checks.order('created_at asc')
       @latest_check   = @checks.last
@@ -170,9 +172,17 @@ class Admin::AssetsController < AdminController
   end
 
   def work_is_oral_history?
-    (@asset.parent.is_a? Work) && @asset.parent.genre && @asset.parent.genre.include?('Oral histories')
+    !@asset.collection_thumbnail? && @asset.parent.genre && @asset.parent.genre.include?('Oral histories')
   end
   helper_method :work_is_oral_history?
+
+
+  def parent_path(asset)
+    return nil if asset.parent.nil?
+    asset.collection_thumbnail? ? collection_path(asset.parent) : admin_work_path(asset.parent)
+  end
+  helper_method :parent_path
+
 
   private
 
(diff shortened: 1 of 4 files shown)


app/controllers/admin/assets_controller.rb
```diff
@@ -23,6 +23,8 @@ def index
   def show
     @asset = Asset.find_by_friendlier_id!(params[:id])
     authorize! :read, @asset
+
+    @edit_asset_path = @asset.collection_thumbnail? ? edit_admin_collection_path(@asset.parent) : edit_admin_asset_path(@asset)
```
jrochkind, line 27: can we make this a controller method instead, that is made available to views with `helper_method :name_of_controller_method`?  

instance variables to communicate state between controllers and views... while Rails kind of suggests it... experience of many shows that they are just messy and hard to keep track of and easy to mess up.

app/models/asset.rb
```diff
@@ -281,6 +281,10 @@ def ensure_correct_derivatives_storage_after_change
     end
   end
 
+  def collection_thumbnail?
+    parent&.is_a? Collection
+  end
```
jrochkind, line 286: I don't love adding this to the model actually... the Asset really doesn't itself behave any differently.... 

If the only place we currently need this is in the Admin::AssetsController... what if we just make it a controller method in that controller (also exposed as a helper)?  it maybe could even be:

```
def asset_is_collection_thumbnail?
  @asset.parent.is_a? Collection
end
helper_method :asset_is_collection_thumbnail?
```

Like that it's about "the" asset, at `@asset` is even baked in?

Also you don't need `&` on `thing.is_a?` cause even `nil` has an `is_a?`, everything has an `is_a?`! But it also doesn't hurt so no big deal.

app/views/admin/assets/show.html.erb
```diff
@@ -1,20 +1,16 @@
...
+<div>Managing an Asset <%= "(this asset is a collection thumbnail)" if @asset.collection_thumbnail? %></div>
 <h1><%= @asset.title %></h1>
 <%if @asset&.parent %>
-  <% if @asset.parent.is_a? Work %>
-    <p>In work: <%= link_to @asset.parent.title, [:admin, @asset.parent] %> </p>
-  <% else %>
-    <p>Thumbnail for collection: <%= link_to @asset.parent.title, collection_path(@asset.parent) %> </p>
-  <% end %>
-
+  <p>Asset is a member of <%= link_to @asset.parent.title, parent_path(@asset) %> </p>
```
jrochkind, line 4: Hmm, so now we only can tell if this is a collection thumbnail or not by recognizing the title, or cicking on it to see where `parent_path` takes you?

I think I prefer the former way this was very explicitly labelling this as a collection thumb here.  I appreciate simplfying/DRYing the logic as you are doing in this PR, but can we leave the UI really distinguishing with explicit labels what's going on?

Issue #1826, opened 2022-08-19
Tweak to `app/services/s3_console_uri.rb`


eddierubeiz:
Not a big deal, but would be nice to fix anyway.

In the DZI orphan checker, a typical path to an orphaned directory might be:
```ruby
s3_path = "f359798d-500a-47fc-8d09-21b25008079b/"
```
Its public url gets added to the list of sample URLS like this:
```ruby
@sample << s3_url_for_path(s3_path)
# same as @sample <<  shrine_storage.bucket.object(s3_path).public_url
```
This results in a URL like
``` ruby
url = "https://scihist-digicoll-staging-dzi.s3.amazonaws.com/f359798d-500a-47fc-8d09-21b25008079b/"
```
In the orphan report, we create an s3 console URL from this:
```
console_uri =  S3ConsoleUri.from_uri(url).console_uri
# => "https://s3.console.aws.amazon.com/s3/buckets/scihist-digicoll-staging-dzi?region=us-east-1&prefix=/&prefixSearch=f359798d-500a-47fc-8d09-21b25008079b"
```

Instead of the above, I'd like the output to be 
```
https://s3.console.aws.amazon.com/s3/buckets/scihist-digicoll-staging-dzi?region=us-east-1&prefix=12716a18-d835-4c1b-972e-ba36a46537ba/
```

jrochkind:
I think we use a utility method for creating links to S3 console in multiple places; not sure if making this change would effect other places and if we'd want it to, maybe so?

PR #2041, opened 2023-01-27
Use prefixSearch only when needed

eddierubeiz:
Ref #1826 .
Tested the fix in both dev and staging.
Note that the buggy method is strictly used as a convenience method, to provide URLs to orphaned files or directories.
`s3_console_uri.rb` could use some refactoring, but that's out of scope; we just want to fix a bug here.


jrochkind:
I don't totally understand it, but I guess the whole thing is pretty hacky, merging!

Issue #2044, opened 2023-01-30
Combined audio derivatives take a long time to fail when the originals are clearly bad.


jrochkind:
This is not a bug per se: neither ingested audio file was playable, so the software was unable to create a combined audio derivative.

## Backtrace
    line 93 of [PROJECT_ROOT]/app/services/combined_audio_derivative_creator.rb: +
    line 93 of [PROJECT_ROOT]/app/services/combined_audio_derivative_creator.rb: block in calculate_start_times
    line 93 of [PROJECT_ROOT]/app/services/combined_audio_derivative_creator.rb: map

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/93002133)

jrochkind:
The honeybadger error said `{"job" => "#<CreateCombinedAudioDerivativesJob>"}`, but frustratingly seemed to lack the actual Job arguments/parameters that would tell us which OH work this was. 

(I am going ot send an email to HoneyBadger support asking about this)

But Annabel figured it out from knowing what Rachel was working on: https://digital.sciencehistory.org/admin/works/mrwh24g#tab=nav-oral-histories

That has two audio members, and one of them, `malcom_s_1117_1_1.flac` was identified in our staff UI as having size `0 bytes` and type `application/octet-stream`

So I believe that file is corrupt (Is "corrupt" the right term when it's 0 bytes? "Missing", I guess!)

Of course, since it's not known by our system to be audio, it shouldn't have even been included in the `CreateCombinedAudioDerivativesJob` -- our app thinks there's only one segment. Maybe there's a bug in `CreateCombinedAudioDerivativesJob` if you try to create one with only one segment? 

We can/should fix that if so (I think you can still create one with only one segment, to convert eg FLAC to the proper format for web play?).  

But it wouldn't resolve the underlying issue of course, of one of those audio files not being present. 

Still needs further debugging.

eddierubeiz:
I'm going to start looking into this!

eddierubeiz:
I have pretty good handle on the problem and have created a draft PR that explains what's going on and fixes it (after a fashion).

For now, it's important to note that this was not a bug. The files were indeed corrupt, and the question is whether we can catch the problem in a more helpful and convenient way.

Issue #1652, opened 2022-04-14
Catch corrupt original files (graphics and sound) earlier and more cleanly


jrochkind:
Sometimes our "Capture One" digitization pipeline produces TIFFs that are corrupt in some way. 

They will make our ingest pipeline fail with an error such as:

* on `vips dzsave`, magick2vips: libMagick error: Sanity check on directory count failed, zero tag directories not supported. `TIFFFetchDirectory' @ error/tiff.c/TIFFErrors/604. (https://app.honeybadger.io/projects/58989/faults/84983034)
* On `vipsthumbnail`:
  *       TIFFFetchDirectory: Sanity check on directory count failed, zero tag directories not supported
           TIFFReadDirectory: Failed to read directory at offset 41751608
           magick2vips: libMagick error:
           magick2vips: unable to read file "/tmp/shrine20220414-5055-9zmz7n.tif"
  * https://app.honeybadger.io/projects/58989/faults/84983035

We're not exactly sure what's up with these TIFFs.  `file` and `mediainfo` both report a sample as image/tiff, without complaint. 

However, imagemagick `identify` will say:

```
identify  ~/Desktop/soda_fountain_beverages_ncta7o9_156_4oaxl62.tiff
identify: Sanity check on directory count failed, zero tag directories not supported. `TIFFFetchDirectory' @ error/tiff.c/TIFFErrors/596.
identify: Failed to read directory at offset 41751608. `TIFFReadDirectory' @ error/tiff.c/TIFFErrors/596.`
```

And `ffprobe` (which it turns out, weirdly, you can use on a TIFF), will output to stderr:

```
[tiff @ 0x122f043d0] IFD offset is greater than image size
[tiff_pipe @ 0x122e05a00] Could not find codec parameters for stream 0 (Video: tiff, none): unspecified size
Consider increasing the value for the 'analyzeduration' (0) and 'probesize' (5000000) options
```

Every time this happens, we get an error on failed ingest/derivative creation, and need to debug it a bit to figure out why it failed -- oh, corrupt TIFF. 

* Should our ingest process actually catch this more explicitly and flag it as a corrupt TIFF?
   * catch the specific error message from the failure for this specific case, and consider it a corrupt tiff?
   * run `identify` or `ffprobe` on it in the early ingest pipeline, and abort ingest intentionally if that reveals it as this weirdness?
   * In either case, we need a way to make it clear in the Admin UI what happened, and to alert/report so it can be fixed
[...]

PR #2049, opened 2023-01-31
Don't download audio files or attempt to create combined derivatives unless we have sensible audio metadata

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2044

This code checks for audio metadata obtained during file characterization (at ingest time). If the metadata doesn't look right, we don't kick off the process of attempting to download the files and stitch them together.

Note that the situation described in #2044 is not a bug per se; the files ingested were not viable audio. So there's really no rush to deploy this.

I also improved the tests.


jrochkind:
What happens if it decides not to kick off the process?  What does the staff user see, when/where if anywhwere?

Right now, if they click the "create combined audio button" and an unexpected error is raised, they see a "failure" message (after refreshing the page). Is that still true in this case?

Is it logged too?

eddierubeiz:
As of the current (draft) code, it doesn't kick off the `CreateCombinedAudioDerivativesJob.perform_later(@work)` but forwards the user back to the controller with a flash error. It's better in that it's instantaneous, but I'm not sure it's the best we can do.

jrochkind:
Cool, that is pretty good. 

I'm having trouble wrapping my head around the complete process, and if using "metadata missing" as a proxy for "corrupt video" is really even right...   

In the cases we had of corrupt videos, do you know what happens whebn the earlier process tries to extract that format metadata, and fails?  I wonder why we didn't see _that_ error in HoneyBadger in our real world case we had! Or does it just log but not actually raise? Or what?

eddierubeiz:
One thing we have to address before merging this (assuming we want to) is making sure that all audio in the collection is characterized with duration info.

We didn't always have the current setup, so I believe there are files currently in production with nil duration info. If you made a change to those works, then tried to recreate the audio, you wouldn't be able to with this new code.

eddierubeiz:
In response to https://github.com/sciencehistory/scihist_digicoll/pull/2049#issuecomment-1411055532  --
I took a good look at the code and the logged output of `app/uploaders/asset_uploader.rb` -- see:
https://github.com/sciencehistory/scihist_digicoll/blob/ce1dd64d5c1eeb83bec4d8a612aef0826cf017e8/app/uploaders/asset_uploader.rb#L22

Currently `asset_uploader.rb` does not raise or log when
- you ingest a zero-length file with a `flac` file extension or
- you ingest an audio file with gaps in its metadata which render it unplayable.

I do think this is something we could address in a separate PR. I'd put this under ref #1652.

jrochkind:
I'd still love to have some general solution for flagging corrupt files at _ingest_, halting ingest, and notifying/labelling for admins. As per #1652. 

But failing that, or maybe even with that, as an extra layer of defense...  I guess the derivative creator refusing to create derivatives if the files look problematic is good. 

It looks like the CombineAudioDerivativeCreator already had a way of reporting errors in creating the deriv, that was already somehow visible to admin users? This just adds a new category? 

Awesome, thanks!

PR #2053, opened 2023-02-02
Google Analytics 4

eddierubeiz:
Approved by Clare & Annabel.

Merging this code should change nothing about the behavior of the site until we define`google_analytics_4_tag_id` in production.

Our plan is to run UA and GA4.0 concurrently for a few months, and only turn off UA at the last minute sometime this summer.

Accordingly, we define a new `google_analytics_4_tag_id` `ENV` setting, which can and will exist alongside the old `google_analytics_property_id` setting. These two settings can be turned on or off independently, and control whether or not we add Google's code into our `<head>` tag and thus which of the two setups we report to.

Rather than declare noop functions in the `<head>` tag, we check whether methods `ga()` and `gtag()` are defined before calling them in `custom_google_analytics_events.js`.

There's no need to change any of our other `html.erb` templates, as sending over event info works pretty similarly. See https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs#measure_events_with_the_default_tracker for migration details.

I've tested this on the staging server against a bogus GA4 property and the information appears as it should.


app/frontend/javascript/custom_google_analytics_universal_events.js
```diff
@@ -0,0 +1,26 @@
...
+      // e.g. "download" or "transcription_pdf" or "english_translation_pdf" or "download_original"
+      // This becomes the second argument of `gtag` in GA 4.
+      e.target.getAttribute("data-analytics-action"),
+
+      // This is currently always the work's friendlier_id.
+      // This will become event_label in GA 4.
+      e.target.getAttribute("data-analytics-label"),    
+
+      // Our code currently sends a null value for this param.
+      e.target.getAttribute("data-analytics-value")
```
jrochkind, line 23: This comment seems confusing to me. If what you mean is "Our HTML never actually defines a data-analytics-value attribute", say that?

app/frontend/javascript/custom_google_analytics_4_events.js
```diff
@@ -0,0 +1,23 @@
+// Send a custom Google Analytics event with data-attributes on a link.
+// Details about migrating to the GA4 event tracker are here:
+// https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs#measure_events_with_the_default_tracker
+// Note that app/frontend/javascript/custom_google_analytics_universal_events.js
+// contains a method that is trigged by the same click.
+$(document).on('click', '*[data-analytics-category]', function(e) {
+  if (typeof gtag === 'function') {
```
jrochkind, line 7: I'd just add a comment that this is checkking to make sure `gtag` is defined, which it should be only if GA4 is enabled. And same on universal GA.  at first I missed this subtlety and wasn't sure we were properly guarding.

jrochkind:
If we're wrong and we ARE using `data-analytics-value`, we'll have to make sure it's okay that GA4 only support three elements of data for the custom event, and then go ahead and remove all the `data-analytics-value` attributes when we drop universal-GA, to avoid confusion. 

If we're not using it anyway, then it doesn't matter... and maybe we shoudl even drop it from our universal-ga custom events now?

eddierubeiz:
We are not using `data-analytics-value`, and I might also remove it from the current file in a future PR.

app/frontend/javascript/custom_google_analytics_universal_events.js
eddierubeiz: Javascript has a `null` but your comment stands!

app/frontend/javascript/custom_google_analytics_4_events.js
eddierubeiz: Done!

PR #2061, opened 2023-02-07
Be willing to re-run OnDemandDerivativeCreatorJob idempotently on error

jrochkind:
Our overall ActiveJob setup was retrying on error, but our code in the OnDemandDerivativeCreator was complaining about this. Make it work, while adding some more tests ensuriung our desired (but a bit confusing) behavior.

We're trying NOT to re-run this infinitely all the time if there is some error that is going to keep re-occuring, but in a nice clean way where the user gets a clean error message and we get the actual error reported too. It's kind of messy, not sure if there's a better design here... we're dealing with the tools straight ActiveJob gives us.

Ref #2055


jrochkind:
(basically, the job was not idempotent on error, and would error if run a second time after an error!)

Issue #1494, opened 2021-12-02
Medium field autosuggest from Getty AAT


apinkney0696:
Medium field to auto-suggest from the [Getty AAT](https://www.getty.edu/research/tools/vocabularies/aat/) standard vocab.

jrochkind:
Just for our records in attempts to debug, some technical info on other issues with Getty in the qa gem is in this open ticket on qa here: https://github.com/samvera/questioning_authority/issues/84

jrochkind:
Filed a ticket on the qa gem at https://github.com/samvera/questioning_authority/issues/346

jrochkind:
Yeah, this isn't going to "just work" for our needs from the existing code library we were working. 

We could invest time in "fixing" it for our needs -- ideally contributing our improvements back, but if we can't get agreement with the community on that, alternatively as a custom "adapter" for the `qa` gem. 

It would probably take a few days or a week of work. 

We may be able to do it as part of our "samvera maintenance pledge".  

For now, I'll move this back to "backlog"!

jrochkind:
We decided the default QA behavior was good enough, and implemented that.

PR #2066, opened 2023-02-13
Use qa gem to auto-complete medium field to Getty AAT

jrochkind:
The QA getty AAT autocomplete doesn't actually work very well, it's somewhat unrepdictable when and what it will autocomplete -- some things I believe are valid terms just never autocomplete?

But apparently some institutions are using it anyway? Northwestern is. But I haven't actually talked to metadata staff there to see if they find it useful, just software developers who confirm they are using a getty AAT search that works wonkily like this one does.

Let's demo this and see if it's "good enough"/"better than nothing". The other option is spending more development time trying to improve the samvera qa getty AAT search -- I think it's just not written to do a search of getty AAT as well as it could be, although it may be hard to make it work how we want because of limitations in getty api. 

Ref #1494, @apinkney0696
.

### DIFF
--- a/app/views/admin/works/_work_fields.html.erb
+++ b/app/views/admin/works/_work_fields.html.erb
@@ -41,7 +41,13 @@
     <% end %>
 
 
-    <%= f.repeatable_attr_input :medium, build: :at_least_one %>
+    <%= f.repeatable_attr_input :medium, build: :at_least_one,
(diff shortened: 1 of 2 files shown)


jrochkind:
Annabel said in slack this looked useful as-is, and worth deploying to see how it works out.

Issue #2057, opened 2023-02-07
0 results page text update


apinkney0696:
The 0 results page in the digital collections currently looks like this: 
![image](https://user-images.githubusercontent.com/78036563/217338314-011fb57a-c371-405c-bfae-aa15d6cef9f1.png)

Instead, here's an alternative layout:
![Capturealjkshdal copy](https://user-images.githubusercontent.com/78036563/217346461-95b01ad1-19d4-4109-8b23-4b87d0103de9.png)

Changes are:

- Update the text to be more concise and helpful.
- Add other search options for archives space and for past perfect online.
- Add a link to contact information (digital@sciencehistory.org)
- Add an image, just for fun

The text color and font were just what I had loaded in photoshop, so don't pay attention to those!

jrochkind:
Great!

I'm not sure if the image in the screenshot is the one you'd like to use or just a stand-in, but either way, if you have an image picked out can you make the image available to us, maybe on P drive with link here? Thanks!

apinkney0696:
Sure - I'm not tied to the image, but you can find it here if you like it:

https://files.sciencehistory.org/navigate/file/000b26c1-8870-4e63-9b0b-77a86dfa9b2b

![0_results_Page copy](https://user-images.githubusercontent.com/78036563/217362925-b45d788d-5665-411b-bd66-a3813c5feeb7.png)

jrochkind:
@apinkney0696 I have no preferences to image, just let us know if you'd like us to use a different one, otherwise we've got that one!

jrochkind:
I guess we _could_ promote/point to the "other searches" more in the 0 hits case, in case what they are looking for _would_ be there?  I'm not sure it's worth it, I'm not sure it will actually have the desired effect, seems like just kind of grasping at straws.

eddierubeiz:
Love this dude!

eddierubeiz:
For ASpace, @archivistsarah points out we may want to label the search differently - something more like what we have at https://www.sciencehistory.org/archives: "archives database" or "full-text archives finding aid search".

jrochkind:
@eddierubeiz Please discuss labelling of new "other search tools" over on #2059, but agreed we should work out something that Sarah thinks is appropriate!

jrochkind:
@eddierubeiz (Also please make them separate PR's, this issue is only about the "no results" page, thank you!)

eddierubeiz:
[...]

PR #2067, opened 2023-02-13
Customized zero results page

eddierubeiz:
Overriding a Blacklight template here, and turning off the default "No entries found" message.
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2057


jrochkind:
@eddierubeiz Do you remember what you were going for using bootstrap `container` and `row` classes here? 

They are adding extra whitespace that I don't really like and that isn't consistent, despite the attempt at the `no-gutters` class there's still extra whitespace in there. 

Maybe just as a way to get two columns side by side? I think just using CSS flex without bootstrap does the job here. I've been trying to avoid using bootstrap column/row stuff lately, honestly. Flexbox can often do the trick. 

Also I totally see your point in removing the "No entries found" message as redundant -- but there's a bit white space there where it use to be, that the pecularities of Blacklight make it almost impossible to remove without overriding a TON of BL stuff. So I'm considering just putting it back as better than whitespae, but maybe not, thinking about it.

jrochkind:
```ruby
          <%
            no_results_work = Work.find_by_friendlier_id('9etr86c')
            no_results_link = no_results_work ? work_path(no_results_work.friendlier_id) : "#"
          %>
          Image from: <%= link_to '"Ever Bumped by Dead Weight?"', no_results_link %>
```

There is no need to take a trip to the database just to create a link. I think we can just hard-code the link, it's true it will break if the work with the friendlier_id doens't exist, but I don't think it's worth taking a trip to the database just to verify this -- although i see you seem to have copied this pattern from the home page!  

https://github.com/sciencehistory/scihist_digicoll/blob/3ae38aaa64f6dc267eb38dd7263157f1b81d0b40/app/views/homepage/_hero_image_and_search_form.html.erb#L8-L9

I don't love it there either. I'm just going to change this to:

```ruby
          Image from: <%= link_to '"Ever Bumped by Dead Weight?"', work_path('9etr86c') %>
```

Issue #2059, opened 2023-02-07
Add options to "other search tools" menu


apinkney0696:
Re: #2057 

> I see your screenshot adds "Museum Collections" and "Archives Guides" to the "other search tools" menu, but that's a separate ticket I'd imagine!

Museum Collections = past perfect online search
Archives Guides = archives space

archivistsarah:
I'd prefer if we can call the ASpace option "Archives Collection Guides" instead of just "Archives Guides" (as in the mock-up over on #2057). If there isn't space for that I have back up ideas.

jrochkind:
Shorter back-up ideas would be good, it would be preferable if we could fit it on one line, in space available, like present options. That's presumably how we wound up with "Archives Guides" as those seemed to be the two most significant words there. So, yes, please share possible alternatives, thank you!

archivistsarah:
OK, then let's go with "Archives Database" which is how we refer to it elsewhere online. It's tricky to find something that is short but indicates that this is a database of descriptions of archival materials, not digitized archival materials.
"Archives Catalog" is my third place choice, but I'd rather not confuse the temporarily complicated (for users) relationship between the library catalog and ArchivesSpace.

jrochkind:
Thanks, we'll see if the long one can fit in first!

jrochkind:
@apinkney0696 While we're at it... should we REMOVE "Oral Histories" from here, since it's from when OH was a separate search, now it's already included in Digital Collections?

jrochkind:
@archivistsarah "Archives Collection Guides" doesn't quite fit, so we'll go with "Archives Database" for now, thanks for the shorter suggestion.

apinkney0696:
> @apinkney0696 While we're at it... should we REMOVE "Oral Histories" from here, since it's from when OH was a separate search, and now it's already included in Digital Collections?

Good question! I think it's okay to leave and link it directly to the Oral History Collection in the DC. It's not readily apparent to an outsider that the Oral Histories are all in the DC or only available through the DC, so think it makes sense to have links to all the collecting areas represented to reduce confusion

jrochkind:
@apinkney0696 

I worry there too many things in the list adding these two more, but leaving Oral Histories.

And I also think it’s just not necessary, the point of “more search tools” is to get things that weren’t included in your search in digital collections.

Like, there might be things at SHI you aren’t going to find here, we want to help you find them.

[...]

PR #2068, opened 2023-02-13
Add two additional 'other search tools'

jrochkind:
ArchiveSpace and PastPerfect, ref #2059

now that it's longer, needs some bottom margin to avoid butting up directly against footer ribbon


jrochkind:
I'm going to go ahead and remove "Oral Histories" as "Other Search Tools" -- they're now included in Digital Collections search, we don't need to link out as an external search. 

Subject to approval from annabel.

jrochkind:
@apinkney0696 signed off after some discussion

Issue #2050, opened 2023-01-31
Use audio lengths calculated at ingest time when creating combined audio derivatives.


eddierubeiz:
When we first created combined audio derivatives, our file characterization routine did not yield audio duration metadata.

This has changed since: every newly audio file *does* have audio metadata associated with it, including the duration.
We should use that duration to simplify, or eliminate, CombinedAudioDerivativeCreator#calculate_start_times .

~~Important: we can't deploy this before making sure that all audio files do in fact have duration metadata - probably by just running a mass characterization in a rake task. So we need to take care of~~ https://github.com/sciencehistory/scihist_digicoll/issues/2054 ~~before we do this.~~ 
Later edit: ⬆️ not needed! all files have duration metadata in producition.

PR #2079, opened 2023-02-24
Remove obsolete & unused method from combined_audio_derivative_creator.

eddierubeiz:
Ref #2050 
I thought we were still using `duration_of_audio_file_in_seconds(path)`. We're not.

Instead, we've been using metadata from the ingest characterization routine (`Kithe::FfprobeCharacterization.characterize_from_uploader(source_io, context)`) in the `AssetUploader` class.

This is the way it should be; we're just removing the unused method now.


jrochkind:
thank you!

PR #2093, opened 2023-03-03
[WIP] Experment: check audio files on ingest

eddierubeiz:
Ref #1652  Ref #2044
I'm moving the audio metadata check code out into its own service, and using it on asset ingest.
~~Actually seems to work pretty well.~~ ◀️ Not so much : the validation errors are not shown to the user.


eddierubeiz:
Duplicate of https://github.com/sciencehistory/scihist_digicoll/pull/2406 . Closing!

PR #2099, opened 2023-03-07
site-wide http basic auth if ENV variables are present

jrochkind:
meant for use on staging


jrochkind:
I've already deployed this to staging. You can see the shared name/password with `heroku config:get HTTP_BASIC_AUTH_NAME HTTP_BASIC_AUTH_PASSWORD`

Let's merge it into master and try not to deploy to staging anything that doesn't have this merged/rebased into it.

Issue #2081, opened 2023-02-27
Try alternate heroku memcached plugin


jrochkind:
We are using memcached currently only for keeping track of request statistics for our `rack-attack` implementation of rate-limiting and rate-monitoring. (It could in the future be used for other "cache" functions). 

We have the timeout for contacting memcached fairly low, to avoid a slow memcached slowing down our whole app, since it our app does read/write to it on every request for `rack-attack` purposes. Looks like 150ms https://github.com/sciencehistory/scihist_digicoll/blob/ce1dd64d5c1eeb83bec4d8a612aef0826cf017e8/config/environments/production.rb#L93

While memcached is supposed to be really fast and 150ms is in fact a long time (making all of our requests 150ms slower is a lot!)... our current memcached server may still not be able to do that?

We're regularly getting `Dalli::RingError: No server available` errors reported by honeybadger. I _think_ this can mean the timeout was reached, rather than the server was just not there at all, perhaps it would have been there if we waited longer. But 150ms is enough!  These are particularly annoying in HoneyBadger because HB for some reason doesn't even recognize them all as the "same" error, it ends up reported as like a dozen different errors with different stack traces -- not sure what's going on there. 

This is very annoying. If we're just using memcached for rack-attack, we _could_ just ignore these errors (and set honeybadger to ignore them).  

We could also stop using rack-attack, not sure it's providing us value justified by how much energy it's taking. 

But as another easy option, let's try switching to a different heroku memcached plugin? 

* We are currently using [Memcached Cloud](https://elements.heroku.com/addons/memcachedcloud) plugin. 
   * in production at the 100MB $14/month package. (no connection limits)
   * in staging at the 30MB free package
   * Even in production we are currently only using ~4MB, so 100MB is plenty, but the free one is missing production-ready features

* The other heroku plugin option seems to be [Memcachier](https://elements.heroku.com/addons/memcachier)
   * it has a pretty similar 100MB play for $15/month (no connection limits)

* So we can try switching to that. We will likely need to change what ENV variables we get connection details from in setup, at   https://github.com/sciencehistory/scihist_digicoll/blob/ce1dd64d5c1eeb83bec4d8a612aef0826cf017e8/config/environments/production.rb#L89-L96

jrochkind:
Let's write the ruby code with `||` so it can use `ENV` variables set by either of those two memcached plugins.

jrochkind:
[...]

PR #2102, opened 2023-03-08
Add `memcachier` store, if it's available.

eddierubeiz:
Ref #2081 
Seems to work without issue in staging.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -73,25 +73,33 @@ Rails.application.configure do
   # Prepend all log lines with the following tags.
   config.log_tags = [ :request_id ]
 
-  # Use a different cache store in production.
-  #
-  # On heroku we use memcachedcloud plugin which provides `MEMCACHEDCLOUD_SERVERS`
-  # env. https://devcenter.heroku.com/articles/memcachedcloud#using-memcached-from-ruby
-  #
   # Caching is not normally turned on in test or dev, but config/environments/development.rb
   # has some standard Rails code that lets you turn it on in dev with in-memory store,
   # with `./bin/rails dev:cache` toggle.
   #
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -73,25 +73,33 @@
...
+    # https://devcenter.heroku.com/articles/memcachedcloud#using-memcached-from-ruby
     config.cache_store = :mem_cache_store, ENV["MEMCACHEDCLOUD_SERVERS"].split(','), {
       :username => ENV["MEMCACHEDCLOUD_USERNAME"],
       :password => ENV["MEMCACHEDCLOUD_PASSWORD"],
-      :socket_timeout => (ENV["MEMCACHE_TIMEOUT"] || 0.15).to_f, # default was 1
-      :socket_failure_delay => false # don't retry failures
+      :socket_timeout => (ENV["MEMCACHE_TIMEOUT"] || 0.15).to_f,       # default was 1
+      :socket_failure_delay => false                                   # don't retry failures
     }
   end
```
eddierubeiz, line 105: Syntax could be DRYer, but I found using conditional assignment here made everything harder to read.

jrochkind:
I had imagined not conditional assignment, but more like `ENV["MEMCACHIER_USERNAME"] || ENV["MEMCACHEDCLOUD_USERNAME"]` every time you mentioned an ENV variable. 

but perhaps that would be harder to read and/or more error prone, not getting it right. 

This seems fine! I suppose we _could_ extract the timeout into a constant to make sure it's the same in both places, but, meh, it's good enough. 

Nicely we can deploy this whenever, and now switch our memcached servers whenever we want without a code change. Sweet. I will merge.

PR #2108, opened 2023-03-16
Bug fix: in interviewee job dates, if start and end dates are both blank, don't show anything

eddierubeiz:
Oops - oversight on my part; quick fix.
Ref #2096


jrochkind (approved):

eddierubeiz:
This is a bug fix; will merge it myself assuming the tests pass.

Issue #2106, opened 2023-03-14
"no results" pages blank in all collections searches (including oral history)


sarahschneiderSHI:
I noticed that, when searching specifically in the oral history collection for something that garners no results, the results page is empty. It would be great to have the results page updated to match what comes up through a "no results" search of the general Digital Collections. (Including the neat image!)

Here is a screenshot of what the oral history page looks like currently when no result is found.
![image (2)](https://user-images.githubusercontent.com/103284115/225021758-cb55bf76-eba1-4609-b014-1c6d276fd473.png)

Thanks,
Sarah

eddierubeiz:
To reproduce:
https://digital.sciencehistory.org/collections/gt54kn818?q=dfg3

jrochkind:
Huh, this may apply to all collection search pages, not just OH? 

Was this an accidental side effect of our no results page customization, do you think, @eddierubeiz ?  We should add `system` specs for no-results page on OH specifically, and generic collections pages.

eddierubeiz:
Yeah - I'll take a look at this and see what I can figure out. Pending @apinkney0696's approval, I'd like to consider treating it as a bug, meaning I might not wait until next week to get started on a fix.

apinkney0696:
Yes, please! I think this is definitely worth prioritizing if you are able to this week.

eddierubeiz:
Sounds good!

PR #2111, opened 2023-03-20
Show no-results content on collection and featured topics pages

eddierubeiz:
The friendly no-results page we created in https://github.com/sciencehistory/scihist_digicoll/pull/2067 was not getting used on the collection or featured topics search, because we aren't calling  render 'search_results'.

See ref https://github.com/sciencehistory/scihist_digicoll/issues/2106 .

I'm not thrilled about explicitly including ```<%= render 'catalog/zero_results' %>``` here. But we're constrained here by
- Blacklight
- the fact that these two searches show the **title** of the search -- which we feel needs to show up _above_ the search constraints.
- the reality that we don't want to mess with the generic blacklight search and how it's shown in `application.html.erb`, for what's ultimately a pretty small problem.

I was able to work out a sort of compromise in a different attempt (see https://github.com/sciencehistory/scihist_digicoll/pull/2112/ ) but we decided against it, given the risk of more problems later, due to Blacklight shifting again under our feet.

Cleaned up the template too - it was quite messy and had some broken HTML in it.

Includes basic tests.

### DIFF
--- a/app/views/collection_show/index.html.erb
+++ b/app/views/collection_show/index.html.erb
@@ -1,124 +1,99 @@
 <% provide :page_title, construct_page_title(collection.title) %>
-
 <div itemscope itemtype="http://schema.org/CollectionPage" class="collection-show">
-
-
-  <div class="collection-top">
-    <div class="collection-desc clearfix">
-      <% unless has_search_parameters? %>
-        <div class="collection-thumb">
-          <%= render ThumbComponent.new(collection.leaf_representative, thumb_size: :collection_page, placeholder_img_url: asset_path("default_collection.svg")) %>
-        </div>
-      <% end %>
-
-      <div class="show-title">
-        <header>
-          <div class="show-genre">
-            <% if collection.department != Collection::DEPARTMENT_EXHIBITION_VALUE %>
-              <%= link_to "Collections", collections_path %>
-            <% else %>
-              Exhibitions
-            <% end %>
-          </div>
-          <h1>
-            <%= link_to collection.title, collection_path(collection), class: "title-link" %> <%= publication_badge(collection) %>
-            <% if can? :update, collection %>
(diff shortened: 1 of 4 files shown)


jrochkind:
I'm very confused what's going on, and how this was possibly working _before_ the customization of the zero-results page. 

Would you mind finding and linking to the PR that did the recent customization of the zero-results page here, so i can compare, and see if it can make sense?

jrochkind:
Also we definitely need some rspec `system` specs testing no-results page for both generic Collection search, and specific Oral Histories collection search.

eddierubeiz:
The PR you mention is here:
https://github.com/sciencehistory/scihist_digicoll/pull/2067/files

The only way the PR affected the no-results page is by changing the value of `no_items_found` from `No entries found` to a blank string in `config/locales/blacklight.en.yml`.

jrochkind:
I wonder previous to that chagne, what the relationship was between the collection no-results pages and the default original `app/views/catalog/_zero_results.html.erb` template. 

Was that template being shown? 

Is that original template still being shown instead of the overridden one?

eddierubeiz:
In other words, a blank search on the collections search before PR 2067 contained the string "No entries found" but did not include the Blacklight `_zero_results` template (which we also override in 2067).

eddierubeiz:
So the challenge here is to backtrack through some of the history of the customization of `app/views/collection_show/index.html.erb` and figure out how it diverged from `app/views/catalog/index.html.erb`. I'll figure it out.

jrochkind:
I'm having trouble looking at the diff because a) my brain is addled, and b) the extensive formatting/indenting changes that (were needed but) really increase the size of the diff. 

In the future we could consider doing separate PR's for formatting changes only, to make the diff's easy to review. 

But for now, I'm just going to trust you and merge it!  Thanks for the specs!

Issue #2095, opened 2023-03-06
OCR management features


jrochkind:
Once #2092 is finished/merged, the functionality will be there to create hOCR and attach it to an Asset model. But this won't be wired up to ever actually get called. 

We need management functionality first. This is one Issue to specify, but may likely be multiple PR's. 

* staff-editable fields on both `Work` and `Asset` models, for "should OCR" (or "OCR desired", or I'm not sure what it should be called). We want to set it on Work, but be able to over-ride it (in either direction) on individual idiosyncratic asset. 
   * On the `Work` model, two possible values: Yes or No. (Default: "no")
   * On the `Asset` model, we actually have three separate values:  Ordinary same as set on work; Over-ride "Yes" for this specific Asset; Over-ride "No" for this specific asset. (Default: "inherit from work")
   * it's possible we want a new "tab" on Work management page for OCR, because there will be more management stuff too (below)
   * But we also need the `Work` level one to be editable on our "batch" edit specifically (I forget if this happens automaticall or what)

* Code to make those selections actually result in the OCR happening. Keep in mind that there are a variety of possible life-cycle orderings in time, all of which should be supported:
  *  At the point a Work is set to OCR, it might already have assets attached, that will need to be OCRd
  * But a new Asset might be attached after a Work has already been set to OCR=yes, and it will also need to be OCR'd
  * After assets have OCR, the Work as a whole or the individual Asset may have their settings changed such that OCR is no longer desired; it should be removed
  * We've talked about implementing this as a regular cron job that checks if things need their state changed -- I'm honestly not sure what other options there might be or what the best way might be, but that's definitely a contender! 
  * Actual OCRing should definitely happen in bg job, using the job created in #2092 

* Additional management functions (perhaps on an "OCR" tab)
  * At Asset level, echo back the actually applied "OCR desired" value, combining both work and asset settings.    
  * Ability to see if hOCR is _actually attached right now_.  
     * At an individual Asset admin page, but maybe also at the Work level as "X out of N" or whatever?
  * Ability to see the TEXT (and/or the complete raw hOCR?) for an existing attached OCR, to get a sense of how accurate it is.  
  * Maybe?: Ability to forcibly _remove_ OCR if it's already attached, or forcibly trigger 'OCR _right now_' for an OCR that would be scheduled at some future point?

eddierubeiz:
I would also like to be able to filter works by "have at least one asset with OCR" or "OCR requested" in the list-of-works admin screen

[...]

PR #2114, opened 2023-03-23
ocr_requested attr_json

eddierubeiz:
Ref #2095
We're not serializing this - it's internal metadata.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -84,6 +84,11 @@ class Work < Kithe::Work
   attr_json :file_creator, :string
   attr_json :admin_note, :text, array: true, default: -> { [] }
 
+
+  OCR_REQUESTED_OPTIONS = %w{yes no}.freeze
+  attr_json :ocr_requested, :string, default: "no"
+  validates :ocr_requested, inclusion: { in: OCR_REQUESTED_OPTIONS, allow_blank: false }
+
   # filter out empty strings, makes our forms easier, with the way checkbox
   # groups include hidden field with empty string. Kithe repeatable
   # input normally handles this for us, but we're not using one for this one.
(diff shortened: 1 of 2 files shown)


jrochkind:
I'm kind of regretting spec'ing that asset should have this value at all...  do we really need that, right now, or should we start out just doing it on work? Was I prematurely overcomplicating, should we start with the simplest thing that will work and add only if we need it?

If we DO go with that... the spec in #2095 didn't have _work_ being able to inherit from parent.  I know works can have parents, but... I'm not sure we need the complexity of inheriting the ocr_requested value from parent, with the unbounded number of SQL queries to fetch all the parents, unless you use fancy postgres CTE that we don't... If we _do_ still have a value on `asset` at all ... can we stick to the spec where WORK does not inherit, it just has true or false? (and then can be a `:boolean` value too, probably?). 

Also, can we call the value `ocr_requested`?  HOCR is an implementation detail, we could store our OCR in some other format in the future, what this value means is that staff has marked it for OCR, not any specific output format.

app/models/work.rb
```diff
@@ -84,6 +84,11 @@ class Work < Kithe::Work
   attr_json :file_creator, :string
   attr_json :admin_note, :text, array: true, default: -> { [] }
 
+
+  OCR_REQUESTED_OPTIONS = %w{yes no}.freeze
+  attr_json :ocr_requested, :string, default: "no"
+  validates :ocr_requested, inclusion: { in: OCR_REQUESTED_OPTIONS, allow_blank: false }
```
jrochkind, line 90: Should we make this `:boolean` instead of `:string`?  Then it's already either true or false, and no need to validate further. 

I am not sure if there are any unanticipated problems with using :boolean in json, which we don't do very much? But let's try it?

jrochkind: Thanks! If we need to convert to String later, we can one way or another -- I _think_ that the :string value might even automatically be able to read the boolean properly as `"true"` , or `"false"`, and if you switched to string and then just went through every object and read it in then `save`d it again, it would convert to String.

eddierubeiz: Great - it's a boolean now!

PR #2115, opened 2023-03-24
[Probably not going to use this] Adding OCR tab to admin work screens

eddierubeiz:
Add a tab to the admin work screen that contains a listing of OCR content with a summary.
See [the UI in staging](https://staging-digital.sciencehistory.org/admin/works/vt150j93m#tab=nav-ocr).
Includes a rudimentary test.
Ref #2095


app/views/admin/works/_ocr.html.erb
```diff
@@ -0,0 +1,73 @@
...
+    <tr>
+      <th class="thumbnail">Thumbnail</th>
+      <th class="type">Type</th>
+      <th class="id">Title</th>
+      <th class="title">OCR</th>
+    </tr>
+  </thead>
+
+  <tbody>
+    <% work.members.order(:position).includes(:leaf_representative).each do |member| %>
```
jrochkind, line 24: I think we talked about paginating here?  I am worried about just putting all 700 (and up!) pages on here, for large pages.  Note these will load with _every_ load of a work/show page -- it's got all tab content on it on every load, whether you choose the tab or not. 

Once we start applying pagination logic here too, I think it starts to give us a hint even more that -- this should be a ViewComponent (that can have it's own methods with logic), not just a partial.  Note that some of hte other tabs on that page are already view components.  In general, we should start always at least considering view components over partials, for all but the simplest stuff. 

Once it's a view component some other stuff can possibly be put into ruby methods in the view component isntead of the template too.  Like the calculation above of `@work.members.count {|m| m.asset? && m.hocr.present?}`, can be in a ruby method in the ViewComponent, `assets_with_hocr_present_count` or something like that.

jrochkind:
Can you give me a screenshot of what the OCR tab looks like, with some asset with HOCR in a work?  I know I coudl check out the branch myself, but I'd have to then create some data etc, and maybe you already have this all set up and can easily add a screenshot? Thank you!

app/views/admin/works/_ocr.html.erb
eddierubeiz: Good call - will refactor!

eddierubeiz:
@jrochkind: The branch is currently deployed to staging. See https://staging-digital.sciencehistory.org/admin/works/vt150j93m#tab=nav-ocr .

jrochkind:
OK, and I'll include a screenshot here. 

Screen Shot 2023-04-03 at 4 02 19 PM

this is really making me think some thoughts though -- sorry for suggesting a larger rethinking here.  First,  if we _do_ have a tab like this, clearly the _table_ needs to be laid out better -- why is it scrunching all the columns into just half of my wide browser window, making titles wrap, while leaving most of my window empty whitespace? If this is going to be here, it needs to have styles or attributes that make it make more efficient use of space. 

BUT...  now I'm wondering if we really need this separate tab at all.  It is almost identical to the existing "members" tab, right?  The only difference in the listing is the extra "OCR" column with a button in it?  Which shows the OCR content on-page (it's all been loaded inline). 

So.... do we really need to duplicate all this?  It turns out that the "members" tab isn't actually paginating -- all 700 rows are loading for Ramelli for instance. But duplicating it here makes the problem twice as bad, duplicating the bytes in response (actually more than twice as bad, since it includes all the HOCR too). And maybe members tab _should_ be paginating and we'd want to fix it in the future... now we have TWO places to fix it if we did. 

Does it actually make more sense to just include this on the "members" tab?  Is there a way to fit the "OCR" info in? Maybe a badge that shows up after the "title" already?  Maybe when you click on it it takes you to another page... do we really need to load ALL the hOCR content on _every_ admin work show page load? (even if they never click on the "OCR" tab, or never click on an "OCR" button!). 

What do you think, is there a feasible way to combine this with the existing "members" tab, instead of making antoher tab that is almost identical with a slight difference? Is there a way to avoid loading all those bytes of hOCR on ever work show page load, only load them when someone actually wants to look at them?

eddierubeiz:
I think you're on to something. I mean, since we're already displaying the OCR at the asset level, all we may need to add is a link, honestly.
Conversely, if you really want to see all the OCR, that could be a different (and simpler) UI that's not shoehorned into a table like this.
Let's discuss tomorrow! I really don't mind if we have to try a couple different things.

jrochkind:
Yeah, maybe a seperate page that actually shows ALL of the page OCR in sequence -- or 100 pages at a time, paginated. A seperate page, instead of a tab, so it's not loaded on every "admin work show" page, but only when you click to see it. (which you can also accomplish in a tab using JS/AJAX to load it only when clicked, but which we haven't done that yet in this app I dont' think, it requires some extra apparatus etc).

eddierubeiz:
Or we could just allow the user to download a file, or a number of zipped files.

eddierubeiz:
I'm moving this to draft and working on a much simpler version.

eddierubeiz:
Not using this! #2121 was just better.

PR #2116, opened 2023-03-29
Allow user to set / unset `ocr_requested`.

eddierubeiz:
Ref #2095
Two ways of setting the `ocr_requested` bit:
1) via the standard metadata form
  - Mostly useful for batch edit
  - Close to the top of the form per Annabel's request

Notes:
We're not doing anything with the value of `ocr_requested` or `hocr` in this PR yet.
No tests; they don't seem necessary here at first glance.

### DIFF
--- a/config/locales/work.en.yml
+++ b/config/locales/work.en.yml
@@ -94,5 +94,6 @@ en:
         representative: "Select the file with media that represents this Work and is used as a thumbnail."
         contained_by: >
           The sub-Collection within the Digital Collections Application to which the digital work belongs. Select the named collection that the work is part of, if applicable. If a desired entry is not available, contact the Digital Collections Librarian. Required if applicable.
-
+        ocr_requested: >
+          We add and remove OCR on a nightly basis. Check this box if you'd like OCR to be created for this work next time the job runs. Likewise, uncheck the box if you want to remove existing OCR.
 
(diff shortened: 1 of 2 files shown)


jrochkind:
@eddierubeiz  Have you talked to Annabel about desired UI?  Confirm that she asked for it on the standard metadata form after all?

> an on-off switch modeled after the “request review” switch. (Not sure if we really need this switch.)

if we aren’t sure if we need it, should we skip it in the PR? It does require a custom controller action and such. if it is going to be there…. does it need a spec?  Does it need consideration for all that complex stuff involving permissions we added, where not all staff users are allowed to do all things, have we considered which are allowed to do this new action? I can't even remember what's going on there, but I remember we discussed "OK, now that we've added this, from here on out every time we add a new staff function we have to consider how it fits in here", no?  How is the new controller action gated to make sure only those who are auth'd can access it?

eddierubeiz:
Yeah - confirmed with Annabel that she wants this in the form! I removed the switch - we can add it later if it turns out to be necessary or useful.

config/locales/work.en.yml
```diff
@@ -94,5 +94,6 @@ en:
         representative: "Select the file with media that represents this Work and is used as a thumbnail."
         contained_by: >
           The sub-Collection within the Digital Collections Application to which the digital work belongs. Select the named collection that the work is part of, if applicable. If a desired entry is not available, contact the Digital Collections Librarian. Required if applicable.
-
+        ocr_requested: >
+          We add and remove OCR on a nightly basis. Check this box if you'd like OCR to be created for this work next time the job runs. Likewise, uncheck the box if you want to remove existing OCR.
```
jrochkind, line 98: Do we actually plan to REMOVE existing OCR as a result of check?  Is this second sentence an implementation detail they may or may not need?

I'm still not really sure about this "OCR requested" verbiage in the first place... it seems weird to say that turning off the request for something means it gets deleted....  But I guess this is good for now?

eddierubeiz: Done!

PR #2117, opened 2023-03-29
Show OCR on the asset admin screen

eddierubeiz:
With a show/hide control.
Includes a helper class to show only the parts of the HOCR we really want on the page.


jrochkind:
I feel like we are trying to move away from the "presenters" directory to ViewComponents in the `components` directory. Would this make sense as a ViewComponent instead?  I think so maybe?

app/presenters/hocr_helper.rb
```diff
@@ -0,0 +1,19 @@
+# Some static utility methods for printing friendly versions of HOCR .
+class HocrHelper
+  # Just the text content of the body. No tags.
+  def self.text_content_as_string(raw_hocr)
+    doc = Nokogiri::HTML(raw_hocr) { |config| config.strict }
+    doc.xpath('//@*').remove
```
jrochkind, line 6: (Deleted previous comment, trying again)

I don't understand what this line is doing. Could you comment it?

Wiat, we're really stripping all tags... if you just want text content of body, are you sure you don't want `doc.css("body").text`?  What is this this doing that's different than that?

Also... are we using this `text_content_as_string` method anywhere?  Do you have plans for it?  Do we need it if we're not using it?

PR #2120, opened 2023-03-31
Create a rake task that adds OCR where needed

eddierubeiz:
The rake task itself is pretty simple and calls `WorkOcrCreatorRemover.new(work).process` on all works.
For a given work, `WorkOcrCreatorRemover` does all the heavy lifting:

- Figures out which assets are images
- If we want OCR on the work, call OCR via `CreateAssetHocrJob.perform_later`
- If we don't want OCR on the work, remove all OCR.

Ref #2095

This could be more efficient at the cost of a bit more complexity. Instead of brute-force `Work.find_each`, I can imagine a more sophisticated version that only looks at works that have EITHER `ocr_requested` set to true OR child assets with OCR. But I'm going with the simplest thing that could possibly work here - at least for now.


app/services/work_ocr_creator_remover.rb
```diff
@@ -0,0 +1,47 @@
...
+    return if @this_is_dev && !asset.file.exist?
+    add_ocr(asset) if @overwrite_existing_ocr || asset.hocr.blank?
+  end
+
+
+  def add_ocr(asset)
+    @do_now ? AssetHocrCreator.new(asset).call : CreateAssetHocrJob.perform_later(asset) 
+  end
+
+  def image_assets_we_can_download
```
jrochkind, line 40: Nice job limiting to just images as a guard. 

I'm not following the name of the method. What does "we can download" mean, are there other image assets we can't download that are being left out of the return value? I don't think so?  Is this really just `image_assets` or maybe `ordered_image_assets`?

app/services/work_ocr_creator_remover.rb
```diff
@@ -0,0 +1,47 @@
...
+      image_assets_we_can_download.each { |a| a.update!(hocr: nil) }
+    end
+  end
+
+  private
+
+  def maybe_add(asset)
+    # tolerate missing files in dev; fail immediately otherwise.
+    return if @this_is_dev && !asset.file.exist?
+    add_ocr(asset) if @overwrite_existing_ocr || asset.hocr.blank?
```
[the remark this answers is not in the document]
eddierubeiz: Great idea - will make it ENV option.

lib/tasks/create_ocr.rake
```diff
@@ -0,0 +1,12 @@
+namespace :scihist do
+  desc """
+    bundle exec rake scihist:create_or_remove_ocr
+  """
+  task :create_or_remove_ocr => :environment do
+    progress_bar = ProgressBar.create(total: Work.count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
+    Work.find_each(batch_size: 10) do |work|
+      WorkOcrCreatorRemover.new(work, do_now: true).process
+      progress_bar.increment
```
jrochkind, line 9: When we run this as a cron job, we definitely don't want `do_now: true`.  We want the cron job to get out of there as soon as possible, and add any actual work to a bg queue.  

(Why? Because the bg workers can _scale_ and do things in parallel; and because long-running tasks are dangerous and can be interupted by platform restarts and such, a bunch of shorter bg jobs are much preferable). 

I am not convinced we need a `do_now` option at all. But I'm curious to hear use cases!  If we do, the rake task should definitely by default execute with bg jobs though. 

Also I don't think `do_now` is a good name for it. If it remains, maybe the option can be called something involving the word `foreground` (as opposed to` background`, alternately it might be something involving `background` set to `false`)  or `synchronous` (as opposed to asynchronouos), or something like that.

eddierubeiz:
Two more comments from J.:
- make the nightly rake task enqueue the jobs, not run them immediately
- `overwrite_existing_ocr` isn't really needed.

eddierubeiz:
This takes a little over 3 minutes to run in staging, btw (excluding the actual downloading and OCR'sing work done by the resque workers.)

jrochkind:
Thinking about the fact that the AssetHocrCreator might in the future add _two_ artifacts (the hocr and the text-layer PDF)... it is very clear where in this job we'd add the extra check.  (blank both of them out; re-do OCR if either of them is missing). So great! We just will have to remember to do it! 

If you see a use case for force_ocr you can leave it in… but honestly I think we’re more likely going to need something like “re-do OCR for all things created before date X”, when we know we changed how we did things at date X or whatever.

I think we’ll just have to write special code when we need it anyway.

Similarly, I honestly see no use case for doing things in the foreground at all. It would just be too slow, it’s like 10 seconds a page on my MacBook I think. We might want a progress bar if it was doing things in foreground, but rather than flesh this feature out...  But did you have a use-case in mind?

jrochkind:
nice job on the tests

lib/tasks/create_ocr.rake
eddierubeiz: OK - got rid of `do_now`.

app/services/work_ocr_creator_remover.rb
eddierubeiz: Oh yeah, that was a holdover from a previous version. Fixed.

app/services/work_ocr_creator_remover.rb
```diff
@@ -0,0 +1,39 @@
...
+#   arranges for their OCR to be created (either via a job or immediately) or removed,
+#   depending on the work's `ocr_requested`.
+#
+#   Ignores child works completely.
+class WorkOcrCreatorRemover
+  attr_reader :work
+
+  # @param work [Work] Work to look at
+  # can be used to update a progress UI.
+  def initialize(work, ignore_missing_files=false)
```
jrochkind, line 12: can you make this a keyword argument, instead of mysterious positional argument?

compare (current, worse) `WorkOcrCreatorRemover.new(work, true)` to (better) `WorkOcrCreatorRemover.new(work, ignore_missing_files: true)`. 

Thanks!

lib/tasks/create_ocr.rake
```diff
@@ -0,0 +1,13 @@
+namespace :scihist do
+  desc """
+    bundle exec rake scihist:create_or_remove_ocr
+  """
+  task :create_or_remove_ocr => :environment do
+    progress_bar = ProgressBar.create(total: Work.count, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
+    ignore_missing_files = (ENV['ignore_missing_files'] == 'true')
+    Work.find_each(batch_size: 10) do |work|
+      WorkOcrCreatorRemover.new(work, ignore_missing_files:ignore_missing_files).process
+      progress_bar.increment
```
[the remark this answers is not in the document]
eddierubeiz: No, I'd meant to fix that. Thanks for the reminder.

app/services/work_ocr_creator_remover.rb
eddierubeiz: Sure - makes sense.

jrochkind:
thank you!

PR #2121, opened 2023-04-04
Simple version of work-level OCR display.

eddierubeiz:
- A simpler version of the UI in https://github.com/sciencehistory/scihist_digicoll/pull/2115 in response to @jrochkind's comments.
- No work-level OCR tab for now; we can bring it back later if we want.
- To keep things simple, this PR does not (and will not) include the feature of viewing the concatenated OCR (or HOCR) from the entire work.
- Ref #2095.
- Simple test included; could add more, I suppose.


jrochkind:
lovely, thanks!

PR #2127, opened 2023-04-10
Bug fix for ocr component count query

eddierubeiz:
We were checking the key, not the value.
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2095


jrochkind:
I wonder if a query for "IS NOT NULL" would work?  But if this works, that's fine!

Do we want a test though, ideally one that would have failed before but would pass now?

jrochkind (approved):

eddierubeiz:
I added that test - good call.

Issue #2119, opened 2023-03-31
DC Redesign 


apinkney0696:
Please find some sketches for the redesign located in P:\Othmer Library\Digital Collections - Internal Access\Website Updates\2023 Website Redesign Updates
https://files.sciencehistory.org/navigate/folder/d14ff8a8-a0ff-430f-87db-1b1e431d4a9d

These are my suggested looks for the DC which will closely match the new design and branding themes of the main website.  This is a draft of a large-scale eventual goal, open to adjustment and feedback based on what is feasible, of course. Please add your feedback to the comment section of each .pdf in Egnyte to keep them organized. 

In order of importance, here are the changes we should prioritize:
- Add new Main Website header and footer
- Add a thin DC header to separate the new MW header from our page body
- Create a DC footer with About, FAQ, and Contact options
- Update the colors and select fonts to match the new brand guidelines. 

Eventually, we can move on to upgrading pages, such as the homepage to add a new header image/mimic the [support](https://files.sciencehistory.org/navigate/file/5b44f1e4-3e67-4d98-a10c-e96f62d72c24) page; the OHMS page to apply changes from the feedback we received during user testing; and collections pages to also apply user testing feedback.

I understand that it isn't ideal to undergo rebranding efforts while other projects are put on hold, however, it is a priority to adopt the design, branding, and feel of the main website in the DC, even if it takes a few months and several rounds of iteration.

jrochkind:
> • Add a thin DC header to separate the new MW header from our page body

While at first I was partial to this solution, I actually think we can do something that looks good and matches the main site better _without_ adding a "divider" bar.  I'd like to try it out and show it to you.

apinkney0696:
Sure; as long as it's clear that the user is on the digital collections site and there is still a search bar, I'm totally open to other ideas!

apinkney0696:
Feedback from updates on Wednesday, May 3rd. Minor adjustments to conclude phase 1:

(1) make the FAQ, ABOUT, and CONTACT buttons all the same size, center text, and slightly larger than they are currently.
(2) to adjust the section titles on the homepage and the OH collection page so their margins begin at the same point as the header icon/text
(3) add a little bit more white space between the text on the contact/faq/about pages and the footer.

jrochkind:
Thanks @apinkney0696 !

[...]

PR #2128, opened 2023-04-11
Site redesign phase 1 -- New header/footer 

jrochkind:
This ended up being a LOT of code, mostly CSS, for "just" header/footer changes.  More than I expected, I was imagining it would only be like maybe two HTML and two CSS files. 

This consists of enough that we would feel good deploying it, when the main site redesign goes live, to synchronize. But we want to go forward, in another PR, with transitioning to stuff between the header/footer to new fonts/colors/etc too. 

Ref #2119


eddierubeiz (approved):

jrochkind:
Before deploying, we should:

- [ ] check all header/footer links to make sure they match main site

There are at least a couple places subject to change. Still waiting on a link for "Terms of Use", and they are still deciding what link to use for "Podcasts".

PR #2131, opened 2023-04-14
Make test pass with Nokogiri 1.14.3

eddierubeiz:
Ref #2132 . Ref #2130 .
I have a fix and the sketch of an explanation.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -390,7 +390,7 @@ GEM
     net-smtp (0.3.3)
       net-protocol
     nio4r (2.5.8)
-    nokogiri (1.14.2)
+    nokogiri (1.14.3)
       mini_portile2 (~> 2.8.0)
       racc (~> 1.4)
     oai (1.2.1)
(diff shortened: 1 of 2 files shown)


eddierubeiz:
One possible explanation of what's going on here:

1. Our test `spec/components/atom_entry_component_spec.rb` was only passing because of a bug in `libxml2`, which caused the code to ignore namespaces in HTML document fragments.
2. [This bug was fixed](https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.10.4) in `libxml2` version 2.10.4, and the fix was bundled in the latest `nokogiri` version.
3. Namespaces were no longer ignored, so the css selector stopped matching.
4. Parse the output of `render_inline` as XML (which is fine - this is an atom feed after all) and refer to the tag we're looking for (correctly) as `rendered.at_css("summary[type=html]")`, and the test passes again.

NB: in css selectors, apparently, the colon separating namespace prefixes in tag names is escaped as a pipe `|` character.

jrochkind:
Thank you for the theory of what was going on, that makes sense!

jrochkind:
> Don't be fooled though. You do not have to use XPath to get the benefits of namespaces. CSS selectors can be used as well. CSS just uses the pipe symbol to indicate a namespace search.

https://nokogiri.org/tutorials/searching_a_xml_html_document.html#namespaces

Fair enough! I had no idea! I might add a comment saying what's going on to the code here but maybe not required

Issue #2133, opened 2023-04-17
Delete OCR immediately when a work's `ocr_requested` is set to false.


eddierubeiz:
Deleting OCR, unlike adding it, is quick and cheap.
Add some code to the admin work controller that immediately removes OCR for a work if the `ocr_requested` metadata has just been set to `false`.
This implies no change to the existing rake task which creates and deletes OCR en masse.

PR #2135, opened 2023-04-17
Delete OCR immediately from the work controller if requested

eddierubeiz:
#2133

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -74,6 +74,7 @@ class Admin::WorksController < AdminController
   # PATCH/PUT /admin/works/1.json
   def update
     authorize! :update, @work
+    delete_ocr if params['work']['ocr_requested'] == 'false'
     respond_to do |format|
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Oops - thanks for catching that!

app/services/work_ocr_creator_remover.rb
```diff
@@ -34,6 +34,6 @@ def image_assets
       members.
       where(type: 'Asset').
       order(:position).
-      select { |m| m.content_type.start_with?("image/") }
+      select { |m| m&.content_type&.start_with?("image/") }
```
eddierubeiz, line 37: Adding this in here as well - just want it to fail gracefully if for whatever reason an asset doesn't have a content_type defined on it.

eddierubeiz:
Thanks again for catching that dumb error! I think it's better now, but take another look when you have a minute.

jrochkind:
I suppose we could also decide this is too much trouble and we'll just let the cronjob do it's thing?

PR #2138, opened 2023-04-26
Remove OCR in one call to the DB.

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2133.

Works with OCR are largely going to be books with lots of pages in them, so we're treating this fairly carefully.
This is a redo of https://github.com/sciencehistory/scihist_digicoll/pull/2135 .

We're not using a transaction to wrap both the update and the deletes, because:
a) we trust PostGres to handle the query correctly.
b) the deletes are unlikely to fail, as this is now only one extra DB call instead of potentially hundreds (as in 2135)
c) even if the deletes were to fail, the nightly cron would get rid of the stray OCR anyway.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -76,6 +76,19 @@ class Admin::WorksController < AdminController
     authorize! :update, @work
     respond_to do |format|
       if @work.update(work_params)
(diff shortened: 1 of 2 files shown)


jrochkind:
If I were going to quibble I'd say that the code for constructing the SQL query etc should be extracted to it's own method (could be in this controller still, not sure). 

But we've spent enough time on this, it's fine!

Issue #2125, opened 2023-04-10
"Copyright Free" checkbox adjustment


apinkney0696:
When checking the public domain checkbox on the homepage search, this should pull in other classifications that fall within Public Domain (see https://rightsstatements.org/page/1.0/?language=en#collection-nc No Copyright options):

- Public Domain Mark 1.0,
- No Known Copyright,
- No Copyright - United States,
- No Copyright - Other Known Legal Restrictions

jrochkind:
Should we change the name of that checkbox if that's what it's going to search.  

I would suggest that for instance "No Copyright - United States" is _not_ in fact the same thing as Public Domain.  

For instance, Wikimedia Commons will, I believe, take Public Domain items but not "No Known Copyright" items, or "No Copyright - United States" items.  

I am worried we are mis-educating people about what "Public Domain" means, and/or giving people who _do_ know what it means misleading results if we call all of those categories "Public Domain". 

Perhaps "No copyright" or something along those lines?

eddierubeiz:
This involves an indexing change.

PR #2142, opened 2023-05-09
Copyright free checkbox

eddierubeiz:
###  ⚠️ Requires a reindex

Ref #2125
- Updated the labels in the UI.
- I didn't feel comfortable keeping `public_domain` as the Blacklight filter class name, controller params etc. as  when we were actually changing the functionality, so I went ahead and changed the code accordingly. This adds some noise to the PR, but it will prevent confusion in the long run.
- ~~Since we're now matching multiple URIs, I'm just using the [terms query parser](https://solr.apache.org/guide/solr/latest/query-guide/other-parsers.html#terms-query-parser). Easy: `!terms` and a comma-separated list of URLs.~~ Instead adding a `copyright free"`string to the rights facet (see @jrochkind 's commend below).
- Where do we keep the list of what we're calling "copyright free"? Initially hard-coded it in CopyrightFreeFilter, then changed my mind. Instead, I added a `matches_copyright_free_searches` boolean flag next to matching terms in the yml file. The filtering happens in the RightsTerm class, and the resulting list is used in the CopyrightFreeFilter. Might be overkill; I could go either way.
- Beefed up the tests a bit.


jrochkind:
This is making a lot more changes than I expected, I wasn't thinking we'd change the names of all the classes/variables. but if you think we have to that's fine!

But also -- I'm afraid we had a misunderstanding about the specs here. 

I want to change the _indexer_ so anything that has rights assigned that's in the list gets a term _added to the facet-field_ that says "copyright-free".  Now the fact limit has to only check for this one, the same as it did before. Also the "Copyright Free" term will now show up in the sidebar facets, so the user can choose it normally from the sidebar, not _just_ from the special checkbox.  And then it also shouldn't need any special logic for displaying the "constraint" either, it'll just show up as a normal constraint -- it's just an ordinarily facet value, just one that we add at index-time as "derived" from the base values. 

That should make it possible to have a lot simpler code, with less special purpose code that needs to be maintained. What do you think?

I'm sorry, looking at the Issue I see I did not spec this out clearly. I thought we had discussed it, but I need to learn to always make sure there's a written spec, not just count on an assumed understanding from a verbal discussion. My fault!

eddierubeiz:
Good call - makes a lot of sense. Fixed!

app/indexers/work_indexer.rb
```diff
@@ -37,6 +37,9 @@ class WorkIndexer < Kithe::Indexer
     to_field ["text_no_boost_tesim", "medium_facet"], obj_extract("medium")
     to_field ["text_no_boost_tesim", "format_facet"], obj_extract("format"), transform(->(v) { v.titleize })
     to_field ["text_no_boost_tesim", "rights_facet"], obj_extract("rights") # URL id
+
+    to_field ["rights_facet"], obj_extract("rights"), transform(->(v) { "Copyright Free" if RightsTerm.copyright_free_filter_uris().include?(v) }) # human label
```
jrochkind, line 41: As ruby style, we don't generally use parens (`()`) on method calls unless there are arguments. `RightsTerm.copyright_free_filter_uris.include?(v)` is preferable. 

We could probably make this more efficient by combining 39 and 41 into one line, but I don't know if it matters.  And I guess not easily if we _don't_ want to put `copyright free` in the text field.

jrochkind (approved):

eddierubeiz:
Deployed and indexed!

PR #2150, opened 2023-05-16
Phase 2 Redesign of standard image-based work show page

jrochkind:


jrochkind:
One general purpose thing that's going to end up in this PR...   refactoring/simplfiying the "max_readable_length" stuff. 

* Too-long times are hard to read, so we keep paragraph text to a certain line limit
* But the line limit wasn't really long enough, lengthen it; and get rid of a separate "longer line limit", if we lengthen the standard one we can do with just one. And then adjust it if we think it's still not long enough.

Issue #2148, opened 2023-05-15
Turn off old Google Analytics.


eddierubeiz:
We think we have a little over a month before [the old GA](https://analytics.google.com/analytics/web/?authuser=1#/a326869w144740760p168958592/)  stops processing information on July 1, 2023.

- [x] Check that this won't affect any remaining workflows (the new dashboard in particular)
- [x] Check that the old analytics data will still remain accessible
- [x] Remove the old code

We want to be in touch with Clare and Erin about this.

See https://github.com/sciencehistory/scihist_digicoll/pull/2053 .

PR #2152, opened 2023-05-18
Remove Google Analytics Universal

eddierubeiz:
Ref #2148 .
We added GA4 in such a way that this step is now fairly simple.
Either before or after deploying, I'll also remove the `GOOGLE_ANALYTICS_PROPERTY_ID` config var in Heroku.


app/frontend/javascript/custom_google_analytics_4_events.js
```diff
@@ -6,7 +6,7 @@
 $(document).on('click', '*[data-analytics-category]', function(e) {
```
eddierubeiz, line 7: The comment below is wrong. As you might imagine,
JS function `gtag` is not defined in
`app/views/layouts/_google_analytics_universal.html.erb .`
It's defined in 
`app/views/layouts/_google_analytics_4.html.erb`.

jrochkind:
hooray!

eddierubeiz:
Deployed and removed the old config variable.

PR #2156, opened 2023-05-23
Redesign Collection show page

jrochkind:


jrochkind:
Also removes CollectionMetadataComponent -- it was trying to package up some re-usable func for displaying colleciton metadata on different pages (standard page and Oral History splash), but ended up not actually being re-usable, too many differences in how we want to display. 

Replaced with a couple finer-grained re-usable objects (`CollectionCountComponent` and `CollectionOpacUrls` service), but also just inlining stuff.

jrochkind:
I'm not super happy with the way some of this somewhat complicated logic and CSS is organized, but am not sure how to improve it, and this is already taking so long.

jrochkind:
trying to override h2...hn default color led to problems with selector specificity when we wanted other colors. Using bootstraps variable instead works better/simpler, although changes ALL headings, less targetted than before. But this is fine.

jrochkind:
Our "featured item" page just uses the same DOM as "collection show", so that the same stylesheet will apply. 

This isn't really a realistic approach, to keep them in sync like this. It would be better to use common ViewComponents that you re-use....  

...I was going to say, "but we'll keep doing it cause it's worked so far", but now I'm wondering.... nah, I don't need to make this mountain bigger, I am going to keep doing it the "wrong" way, but if it gives us any trouble we should refactor to shared ViewComponents instead of trying to sync the HTML between two templates.

jrochkind:
@eddierubeiz OK, this one is ready for merging when approved!

eddierubeiz (approved):

PR #2168, opened 2023-05-30
Fix redirect to main site search results, for phase 1 redesign

jrochkind:
New website will have new URL pattern for linking to search results

eg

    https://dev-sciencehistorywp.pantheonsite.io/?s=frog%20and%20toad


eddierubeiz:
- Also un-hard-coding the website hostname -- good.
- Took a quick look at https://www.rubydoc.info/gems/addressable/Addressable/URI/CharacterClasses and I the change from `PATH` to `QUERY` makes sense at first glance. Merging.

PR #2169, opened 2023-05-31
Oral History pages -- redesign for phase 2

jrochkind:


jrochkind:
Also includes some refactoring of new-site-matching brand buttons in scihist_custom_buttons.scss

jrochkind:
We are doing more major changes to Oral history "work with audio" pages (with or without OHMS), because I didn't think they worked with new design without it. 

But request pages just need minor tweaks, so that's all they're getting for now. #886  remains a separate outstanding ticket.

eddierubeiz (approved):

Issue #705, opened 2020-04-16
Serve Asset model files from a content delivery network instead of from S3 ?


eddierubeiz:
[**See also** suggested notes for how to handle cloudfront/S3 at #2174]

We suspect offering our assets for download on a CDN might make downloading large files faster and more reliable.

Our recent work with oral histories has made this option more appealing, if not necessarily urgent: the audio files tend to be on the large side.

There have been recent reports of users having really slow download times from S3. This could be attributable to slow internet connections, of course. Drawing reliable conclusions about this sort of thing is a project in its own right.

Possible platforms include Cloudflare and Cloudfront.

**this could really be two issues/two part**

1. Do it for _public_ assets (that are publicly viewable) because that's easier
2. Do it for our access-controlled assets that we only deliver with time-limited URLs (this is harder)

jrochkind:
I thought public buckets would be easy and avoid us having to deal with signed URLs through cloudfront -- but we forgot that even for our public buckets, we use signed S3 URLs in order to set content-disposition headers for downloads. 

So even our "public" buckets were basically the same problem as our private ones, and still required figuring out how to deal with the issue of signed URLs through cloudfront -- but actually even harder, required us to deal with passing the specified content-disposition headers through cloudfront too!

So we've given up on this for now, it's no longer highest priority item, it's moved back to backlog/ready future work.

jrochkind:
we tried it, we didn't get very far -- this is very related to #2174 which is still open, and in backlog for now.

PR #2171, opened 2023-06-01
Use a cloudfront distribution for the derivatives bucket

eddierubeiz:
Ref #705
`host: lookup(:s3_bucket_derivatives_host)` seems to do the trick.


jrochkind:
sweet

PR #2175, opened 2023-06-05
Normalize colors and css var names for Phase 2 Redesign

jrochkind:
We create new SCSS variables for all applicable colors in Brand guide color pallette. We also preserve a separate set of SCSS vars with mnemonic names for the colors we actually use. 

We remove all OLD brand color vars and usage, in most cases replacing them with new brand colors. 

Removed a couple CSS classes that weren't actually applying to anything anymore, that we came across. 

As part of Phase 2 we had a variable called `shi-dark-blue`, but after learning that Brand Guide refers to this as gray (it is a dark gray very slightly blue tinted), we changed variable name to `shi-dark-gray`. 

We continue using a couple colors not from current Brand Guide:

* Our links, we haven't yet decided what to do with them, but for now are keeping the old brand blue color, now in scss var `$shi-blue-text`.

* We need a gray for muted text, but none of the Brand Guide grays suffice, they are too dark or too light, skipping a middle-ground we need. Rather than preserve the exact old brand gray we were using, we take the current brand dark gray, and lighten it in color until it is very close in lightness/darkness to previous brand gray.  This is now in SCSS var `$shi-alt-muted-text`.
  *   (There is some precedent for additional grays -- the gray in the logo "Museum & Library" is not actually in Brand Guide either).  

All commits:

- new brand color SCSS variables, with names from Brand Guide
- remove unused -btn-alt-bg variable
- remove bootstrap -dark-* overrides, we don't use them
- remove old brand color -bright-green scss variable
- remove old -yellow scss color, replace use with new brand color -yellow
- replace old -red scss with new brand color -red
- remove -light-gray, replace with new brand colors mostly -bg-gray-border, some red highlights
- link hover color is brand-black, more consistent with main site and don't use old -dark-blue anymore
- remove styles for .search-heading no longer in DOM
- remove old now longer used -dark-blue, replace with new brand colors, mostly -blackish but sometimes other choices
- remove now-unused -blue -- we're still for now using it in -blue-text for links though
- use brand grey for missing-image fill that was previously not brand color
- remove style that was not being applied -- we don't need different color for facet arrow
- remove old brand-dark-grey -- but keep similar custom non-brand color for muted text
- change css var name, the Brand Guide calls this dark gray not dark blue

### DIFF
--- a/app/frontend/stylesheets/_mixins.scss
+++ b/app/frontend/stylesheets/_mixins.scss
@@ -64,7 +64,7 @@
   line-height: normal;
   letter-spacing: $small-caps-letter-spacing;
   a:not(:focus):not(:hover) {
-    color: $special-label-muted-color;
+    color: $shi-alt-muted-text;
   }
 }
 
--- a/app/frontend/stylesheets/_variables.scss
+++ b/app/frontend/stylesheets/_variables.scss
@@ -110,7 +110,6 @@ $special-label-muted-color: $brand-dark-grey;
 // Why do we need so much spacing for it to look reasonable? Good spacing for 14px and weight 600.
 $small-caps-letter-spacing: 0.035em;
 
-$brand-btn-alt-bg: $brand-bright-green;
 $brand-alt-header-weight: 900;
 
 // adelle-sans has a 600 semi-bold available. 700 is standard bold. 500 seems to be the same as 400=normal.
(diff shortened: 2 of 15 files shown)


jrochkind:
Oh, this also uses the new brand green color for bootstrap "success" theme color.

PR #2200, opened 2023-06-13
oh splash center projects

jrochkind:
phase_2_redesign_base


jrochkind:
@eddierubeiz Thanks for catching that, targetting this at `phase_2_redesign_base` instead of `master`, it's now properly only one commit with 6 lines changed and should be more understandable!

Just centering some content on the OH "home" page!

PR #2213, opened 2023-06-15
Default text-link style to p was catching OH born on this day which we didn't want

jrochkind:
Provide a CSS class to opt-out, but in this case actually making the <a> block level and not inside a <p> gives a bigger link target and works better anyway, so did that.


eddierubeiz:
I think Github ate some of your description - it strips what it considers HTML, I. believe. But I get what you're saying. Merging!

PR #2219, opened 2023-06-20
update ruby version to 3.2.2

jrochkind:


jrochkind:
Passes tests, seems to deploy fine to staging, I guess it's good?

PR #2221, opened 2023-06-20
update Rails to 7.0.5

jrochkind:
bundle update rails

Which also updates all dependencies of rails to latest versions

PR #2223, opened 2023-06-20
Update all rubygems dependencies

### LINKED ISSUE #2165: general dependency updates with `bundle update`
Last time I tried `bundle update` it broke tests. Not sure why. 

Possibly it's becuase latest version of `view_component` breaks our version of Blacklight, maybe freezing `view_component` (until we can update to BL8) will resolve. 

But we should get what dependencies can get updated updated; and identify any that can't to restrict in `Gemfile` to the latest version we can use. 

See #2164

jrochkind:
With `bundle update` to update all dependencie within restritions expressed in Gemfile or transitive dependnecy tree.

WHENEVER you update `vite-rails`, you need to run `bundle exec vite upgrade` or you get errors saying that your JS vite stuff doesn't match the vite-rails gem. The vite upgrade command fixes that by updating your yarn.lock and package.json -- all changes to there in this PR were done by that.

I keep forgetting this, and what it looks like in rspec when it fails is... weird. I think that was the only actual problem caused by updates that I encountered in #2165.

Added comment to README to hopefully make that easier to debug when it happens.

Closes #2165

Supercedes #2221, this includes rails update too. 

Built on top of #2219, so that needs to be merged first.


jrochkind:
Also some additional JS updates:

----

update vite-plugin-sass-glob-import to 2.x, apparently needed for vite 4.0 that vite-rails seems to have updated us to.

Was previously getting warning message:

> warning " > vite-plugin-sass-glob-import@1.4.0" has incorrect peer dependency "vite@^3.0.0".

----


update rollup-plugin-gzip to latest version, avoid spurious error message about peer dependency

Was previously getting the warning:

> warning " > rollup-plugin-gzip@3.0.1" has unmet peer dependency "rollup@>=2.0.0".

Which doesn't make much sense, we were using a rollup >= 2.0.0 already. Might have been a warning/dependency bug. Fixed it simply by running `yarn upgrade rollup-plugin-gzip` to get latest
still same major version rollup-plugin-gzip.


----

Adding vite into the mix -- with a couple plugin as above -- has definitely complicated our JS package.json management routine a bit!

jrochkind:
Oops, we're still getting that second warning....

    warning " > rollup-plugin-gzip@3.1.0" has unmet peer dependency "rollup@>=2.0.0"

That doesn't make any sense to me....

```
$ yarn list --pattern rollup
yarn list v1.22.19
├─ rollup-plugin-gzip@3.1.0
└─ rollup@3.25.1
```

So we are using rollup 3.25.1, which is definitely greater than 2.0.0.  Yarn peer dependency checking here is getting weird....  We also are using a very old version of yarn, which maybe doesn't help?

jrochkind:
Oops, this is failing to deploy to heroku. The "uglifier" JS compressor is complaining about JS `?.` optional chaining operator in source code

~~~
remote:        Uglifier::Error: Unexpected token: punc (.)
remote:        --
remote:         6058
remote:         6059 Blacklight.onLoad(function() {
remote:         6060
remote:         6061   $(".range_limit .profile .range.slider_js").each(function() {
remote:         6062     BlacklightRangeLimit.buildSlider(this);
remote:         6063   });
remote:         6064
remote:         6065   // Support for Blacklight 7 and 8:
remote:           =>   const modalSelector = Blacklight.modal?.modalSelector || Blacklight.Modal.modalSelector
remote:         6067
remote:         6068   $(modalSelector).on('shown.bs.modal', function() {


remote:  !
remote:  !     Precompiling assets failed.
remote:  !
remote:  !

~~~

I can't figure out why this would just have started with this PR though... we did not update blacklight here... AH, we DID update blacklight_range_limit from 8.2.3 to 8.3.0. 

Perhaps we should lock blacklight_range_limit to not update minor versions unless we're careful too -- it's possible 8.3.0 doesn't work with our version of Blacklight 7 too, these things happen. 

BUT...  we're going to have to update to it eventually, and there are still a number of questions:
* Why is this failing on heroku when I can't reproduce locally with `bundle exec rails assets:precompile`?
* recent versions of uglify-js ought to be recognizing `?.` anyway.... maybe the rubygem `uglifier` doesn't use the recent versions or something?
* but maybe we should stop using uglifier anyway, it's maybe not so maintained. Should we replace it with a different JS minimizer?  Some people mention `terser`.

jrochkind:
OK, locking `blacklight_range_limit` to 8.2.x _seems_ to have let us deploy to heroku again. Eventually we'll have to deal with the problem though. 

the `uglifier` ruby gems hasn't been released since 2019. https://rubygems.org/gems/uglifier .  It probably doesn't use a recent version of uglifier-js. 

Perhaps switch to terser?  

https://github.com/ahorek/terser-ruby

Not totally sure if we need a JS minifier at all, but we've always had one... But now that we're using vite... it looks like the minifier is happening on top of vite too, maybe not needed/problematic?

jrochkind:
Hm, on heroku deploy, I'm now getting message:

> remote:  !     Warning: Your slug size (363 MB) exceeds our soft limit (300 MB) which may affect boot time.

Were we getting that on master branch too, or have we significantly increased slug size here?

Master deploying is only `219.7M`, that's a pretty big increase! 

It may be a weird temporary artifact involved with heroku cache and changing so much at once... but I've got to look more into this and figure it out, we don't want our slug size going up so much!

jrochkind:
Yep, locking active-encode to 1.1.1, the version previous to this PR, our heroku slug size is back to 208.1M, a slight reduction even. Great. 

We will do that for now. 

Reported to active-encoe at https://github.com/samvera-labs/active_encode/issues/126, and we'll try to fix it there in a subsequent release too.

jrochkind:
OK, I think I'm ready to merge this one, and not scared to deploy upgrade to ruby 3.2 at the same time as general dependency update!

eddierubeiz:
Thanks for all this detective work. Noted re: `bundle exec vite upgrade`. Approving.

eddierubeiz (approved):

eddierubeiz (approved):

PR #2224, opened 2023-06-20
Update ransack to 4.x

jrochkind:
We use ransack to avoid some boilerplate in our admin dashboards, around sorting and filtering on Collection and Work lists. 

Ransack 4.0 has a backwards breaking change where you NEED to list all attributes you want ransack to operate on in allowlist methods in the model (Work or Collection). 

It's not much well-documented, but see https://github.com/activerecord-hackery/ransack/pull/1400

This is a security improvement, there were some potential holes there. (Mostly not applicable to use, because our own uses of ransack are all behind admin logins, but still better to be a tight ship -- and stay on the latest dependencies generally). 

I really don't like having to add methods to the MODELS. This would ideally have been controller-level. But that's not what ransack chose. In general, ransack seems to be kind of barely maintained (see lack of docs for this change, for instance)... I'd love to move away from it. We actually barely use it, the main convenient thing it does for us involves those SORT headers, where you can click on any header to sort ASC or DESC.  Without ransack it'd be a lot of boilerplate.  Perhaps we should extract the parts that just do that into it's own gem or something. 

But meanwhile, we get everything working for ransack 4.0.  

Unfortunately, if we MISSED an attribute in the allowlist -- the default behavior is just for it to no-op not include it in the search/sort.  But i've tested everything manually pretty thoroughly on admin Work and Collection, which seems via a sourcecode search to be only place we use ransack. I've tried turning on the ransack configuration to RAISE -- only in dev and test -- instead of silently ignoring. But that does not seem to actually work for sorting, our main use case. https://github.com/activerecord-hackery/ransack/issues/1427

Commits:

- update ransack gem to 4.0
- allowlist Work attributes for ransack 4.x
- allowlist Collection attributes for ransack 4.x
- try to have ransack error in dev/test when we missed an allowlist item, in cases where it can do so

----

This PR is built on top of #2223 and that needs to be merged first.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -59,4 +59,15 @@ class Collection < Kithe::Collection
   def build_representative
     self.representative = CollectionThumbAsset.new(title: "collection-thumbnail-placeholder", parent: self)
   end
+
+  # For ransack use, we need to list all attributes we want to use ransack to SEARCH or SORT by.
+  #
+  # We really probably oughta stop using ransack, I hate having this in the model.
+  def self.ransackable_attributes(auth_obj=nil)
+    ["title", "id", "created_at", "updated_at"]
+  end
+  # We don't use ransack with associations, but still have to include this method
+  def self.ransackable_associations(auth_object = nil)
+    []
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -124,7 +124,7 @@ gem "content_disposition", "~> 1.0"
 
 gem 'faster_s3_url', "< 2" # for generating s3 urls faster!
 
-gem "ransack", "~> 3.0"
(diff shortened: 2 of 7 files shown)


app/models/collection.rb
```diff
@@ -59,4 +59,15 @@ def representative_attributes=(attributes)
   def build_representative
     self.representative = CollectionThumbAsset.new(title: "collection-thumbnail-placeholder", parent: self)
   end
+
+  # For ransack use, we need to list all attributes we want to use ransack to SEARCH or SORT by.
+  #
+  # We really probably oughta stop using ransack, I hate having this in the model.
+  def self.ransackable_attributes(auth_obj=nil)
```
eddierubeiz, line 66: Noted!

eddierubeiz (approved):

Issue #2222, opened 2023-06-20
update JS dependencies


jrochkind:
`yarn list X` can be used to see what versions(s)  we are currently using. `yarn info X` can (with confusing output) be used to see what latest version is. 

most significant:

* openseadragon: 4.0.0 to 4.1.0 -- used for image viewer
* video.js: We're still on 7.x, can go to 8.5.0. (also see videojs-seek-buttons). -- used for audio and video players

others we maybe could do?

* @shoppify/draggable: 1.0.0-beta.8 to 1.0.0-beta.12. -- used only for admin screen that lets you re-order members

Assorted other things I can't figure out why we are including them at all (`js-yaml`?), and/or we are intentionally using specific older versions (`fontawesome`, `popper`).

jrochkind:
`@shoppify/draggable`: Looking at the repo, i think this may actually not be very maintained, alas. What's going on with subsequent versions is confusing -- beta.8 is still (intentionally?) marked latest "stable" in npm.  And there's nothing of import to us in CHANGELOG for subsequent releases. We'll prob just leave this alone.

PR #2226, opened 2023-06-20
update video.js from 7.x to 8.x

jrochkind:
ref #2222 

skip ahead/forward by X seconds is functionality now built into 8.x -- we were formerly using plugin `videojs-seek-buttons`, but it's README now recommends:

> However Video.js 8.2.0 + has a built-in seek buttons functionality. Consider using that instead of this plugin.

So we removed that plugin, and switched config over to use built-in functionality. We were previusly skipping forward 30 and back 15. The built-in func only supports skipping by 5,10, or 30 seconds -- so the skip back is now 10.

The built-in functionality shows the seconds you will be skipping on screen though, when before it did not. So that's an improvement, that had been requested by some of our local stakeholders too I recall.

### DIFF
--- a/app/components/audio_player_component.html.erb
+++ b/app/components/audio_player_component.html.erb
@@ -16,23 +16,24 @@
           volumePanel: {
             inline: false
           },
+          skipButtons: {
+            # seconds. only 5, 10, and 30 are supported by video.js
+            forward: 30,
+            backward: 10
+          },
           # we want to customize the ORDER of control elements,
           # while also removing some.  Making it similar
           # to default html5 audio players in popular browsers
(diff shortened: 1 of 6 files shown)


archivistsarah:
10 & 30 are fine with me!

rachellane12:
Looks good @jrochkind! I like that there are numbers now!

sarahschneiderSHI:
Looks good to me also! Thanks!

eddierubeiz (approved):

jrochkind:
GAH, this merged into wrong branch, never made it to master. That could be true of others too....

Issue #2143, opened 2023-05-09
Try using Redis instead of Memcached for Rails cache


jrochkind:
to be a store for rack-attack. 

Can we keep using rack-attack but get away from these annoying periodic dalli/memcached errors? What if we try using a redis instance of a memcached instance? Still provided by a heroku plugin. Cost might be similar?

A different redis than the one we're using for activejob queue, since it should have different settings -- to use as a cache it should be set as a cache to evict old values when it gets full (which the ActiveJob queue should not!)

jrochkind:
OK, I'm realizing that all those annoying Dalli::Ring errors we get.... aren't actually interupting our users at all.  

Rails _does_ just move on from the cache failure, rescuing the exception. 

It just also reports the error in a way that Honeybadger hooks into.  Note that in honeybadger the Dalli::Ring errors have tags `[severity:warning ]`, `[handled:true]`.  

Which is coming from [here](https://github.com/rails/rails/blob/de5eb9ed9869fad99af0296267594e5ae6dff74b/activesupport/lib/active_support/cache/mem_cache_store.rb#L192). 

It's already a silent ignore -- but with error logging. 

Maybe we just want this error to never be reported to honeybadger at all, so it doesn't annoy us?

We could do that just by adding to our [honeybadger before_notify config](https://github.com/sciencehistory/scihist_digicoll/blob/32c396e4ec6c5cd68217e692111af8f7adeea48b/config/initializers/honeybadger.rb#L25) something like:

```ruby
config.before_notify do |notice|
  if notice.exception.class < MyError && notice.tags.include?("severity:warning") && notice.tags.include?("handled:true")
    notice.halt!
  end
end
```

Not sure if we want to just totally ignore these or not.... are we bothered that memcached could be super duper misbehaving and we wouldn't know it?

Alternately, can we get honeybadger to group these better, so we only get ONE honeybadger error instead of multiple....

jrochkind:
OK, I think the number of errors we get is actually pretty low -- like one every few weeks or so, which is FINE, especially for this kind of cache. I don't know if it justifies changing to a whole different data store, the one we have is actually working fine. 

[...]

PR #2229, opened 2023-06-22
Tweak config of mem_cache_store used for rack-attack

jrochkind:
With context from the memcachier heroku docs. We are attempting to reduce/eliminate all the errors we get for memcached being unavailable.

We add a 'pool', recommended there for use with multi-thread puma, which we are using.

We also, while looking at the config, go back on our decision to eliminate retry after error, instead enabling a pretty small 80ms retry delay.

Ref #2143, trying harder with memcache before giving up


eddierubeiz:
Thanks for researching this!

PR #2235, opened 2023-06-27
Move override of Query Constraint to component-based override

jrochkind:
We override the "query constraint" area to add the editable "search within your results" form, which also tells you your current search query in current results. 

We had done this by overriding a Rails view template -- in a way that would break in Blacklight 8. We move it over to Blacklight 8's ViewComponent-based arcchitecture, where we have to configure a custom view component... so it can call ANOTHER custom view component of ours... to provide the override. 

This works now in BL7, and should continue to work in BL8. 

We also turn off our silencing of Blacklight deprecation warnings, since with this and past PR's, we should no longer be generating them, hopefully. 

- remove blacklight deprecation silencing, in preparation for move to blacklight 8
- switch override of query constraint component (to be editable form) to be Blacklight8-compatible component-based override


jrochkind:
I'm putting back in silencing of Blacklight deprecation warnings, because we're still getting one -- I think it's from Blacklight itself??

PR #2242, opened 2023-07-03
Outstanding merges....

jrochkind:
This is a mess, we have existing approved PR's that wound up merged into this branch instead of master....

So re-PR'ing this branch and re-merging. 

All of these have already been approved and we attempted to merge, but we totally messed up our git workflow somehow. 

- update ransack gem to 4.0
- allowlist Work attributes for ransack 4.x
- allowlist Collection attributes for ransack 4.x
- try to have ransack error in dev/test when we missed an allowlist item, in cases where it can do so
- update openseadragon JS to 4.1.0
- update video.js from 7.x to 8.x


jrochkind:
Actually we'll handle this by trying to re-PR each of the PRs. GAHHH.

jrochkind:
Nope, actually we're just going to merge all this.

Issue #2220, opened 2023-06-20
Find broken "related links" to www.sciencehistory.org


jrochkind:
With the main site redesign/new platform -- I know they intended to keep all the URLs from old site working with redirect. And many/most of them are. 

But when testing things out and clicking around, I find plenty that unfortunatley are broken. 

We could write a script to check them all (checking for 404 HTTP response vs 200 is enough to identify the broken ones I've found), and make a list -- for either manually finding the new URL for the thing we tried to link to main site, or just removing the broken links. 

(Communications may or may not want to know about broken links, since they intended to keep them all working. We could ask them if they'd like our list). 

-----

Note also, need to check for links that do not 404, but now redirect to `https://sciencehistory.org/stories/`. See below, any former links to videos may now do this.

jrochkind:
cc @apinkney0696

jrochkind:
Oops, here's at least one example where checking for HTTP error status code wouldn't suffice. :( 

https://digital.sciencehistory.org/works/fq977v64q

Includes a link to:

(Video) Distilled #5: Porcelain Painting: https://www.sciencehistory.org/distillations/video/distilled-5-porcelain-painting

that redirects to a generic "stories" page at https://sciencehistory.org/stories/

It's not getting to the actual video, which... may not be on the site anymore?  Did they remove all the Distillations Videos?

apinkney0696:
It looks like they did remove all of the videos from the main site, but not the YouTube channel - I started replacing the links to go instead to YouTube. If you were able to make a list of all the broken links I would be happy to go through and edit/remove the rest!

jrochkind:
It is unfortunately difficult to find "broken links" in an automated fashion, when they redirect to a good page, as this one did, instead of reporting "no page found". 

But I guess if we know all videos now redirect to https://sciencehistory.org/stories/, we can flag all links that redirect to that one. 

We'll have to make a tool to go through all our links and check them -- including checking for that condition. We just need to decide when to prioritize this work, we suddenly have a whole bunch of issues coming up.

eddierubeiz:
To start out with, here's a [list of broken links](https://github.com/sciencehistory/scihist_digicoll/files/12001003/404s.txt).

eddierubeiz:
And here's a [list of links that aren't broken but don't forward you all the way to the destination](https://github.com/sciencehistory/scihist_digicoll/files/12012895/Non-404.broken.external.links.txt) .

[...]

PR #2245, opened 2023-07-10
[WIP] Two rake tasks to output shell script that can help us find "broken" external links.

eddierubeiz:
This could be accomplished with faraday, but I'm going with a local script consisting of calls to `curl` instead.
Ref #2220

### DIFF
--- a/lib/tasks/reports/find_broken_related_links.rake
+++ b/lib/tasks/reports/find_broken_related_links.rake
@@ -1,0 +1,46 @@
+namespace :scihist do
+  namespace :reports do
+    desc """Output a shell script to find broken links in related links.
+      #
+      # bundle exec rake scihist:reports:find_broken_related_links > check_urls.sh
+      #
+      #./check_urls.sh 2>/dev/null
+    """
+    task :find_broken_related_links => :environment do
+      Kithe::Indexable.index_with(batching: true) do
+        Work.find_each do |work|
+          next if work.related_link.length == 0
+          work.related_link.each do |link|
+            friendlier_id = work.friendlier_id
+            #title = work.title
+            #category = link.category
+            #label = link.label
+            url = link.url
+            metadata =  "#{friendlier_id} #{url}"
+            comm = [
+              # --head:       show document info only
+              # --location:   follows redirects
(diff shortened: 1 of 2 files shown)


jrochkind:
Why calls to curl? 

I would really recommend using ruby.  Not necessarily faraday. The ruby HTTP landscape is a bit disjointed which perhaps makes it seem hard to pick what to use in ruby, but how about maybe http-rb -- which is already in our dependencies, although only indirectly, could be listed in our Gemfile as a top-level dependency too. 

https://github.com/httprb/http

```ruby
response = HTTP.get("https://example.com")
response.status #=> 200

response = HTTP.get("https://www.sciencehistory.org/distillations/video/distilled-5-porcelain-painting")
response.status # => 301
response.headers["Location"] # => "https://sciencehistory.org/distillations/video/distilled-5-porcelain-painting"
```

Okay, that last one is not what I expected... but is the same thing `curl` reports, so that's not the issue. 

(I was expecting a redirect to `/stories` which is what i see happen in the browser... this is a different issue, but if the redirect is really happening in, say, Javascript code or something... oh boy what a pain. If we DO end up needing to actually PARSE html returned, that's even more reason to do this in ruby not do weird things with curl. Separate issue is how we actually identify and detect the bad links that can't be detected with http 404).

jrochkind:
I would like one single rake task that produces a report, not two rake tasks. 

I think there may end up being more than two things we need to check for. 

I am not sure checking for redirect to /stories is actually _right_ (see above), it needs to be tested. 

Part of the task/work is the "analysis" of actually figuring out for sure what to check for.

jrochkind:
AHHH

`https://www.sciencehistory.org/distillations/video/distilled-5-porcelain-painting`

redirects to

`https://sciencehistory.org/distillations/video/distilled-5-porcelain-painting`

which redirects to

`/stories`

Redirects to a relative url!  Seems like it could have also been absolute url. 

I think we're going to need some more complicated logic here, and it's going to be a lot easier with a ruby http client that gives you clear access to parsed response.

eddierubeiz:
I rewrote this without curl in #2273. Closing this PR.

PR #2249, opened 2023-07-12
Extract textonly_pdf from tesseract too

jrochkind:
Tesseract has a `textonly_pdf` feature which gives you a (very small in bytesize) PDF with only an invisible positioned text layer. 

We can combine this with an image layer ourselves to create a full image+text PDF for our scans. We store it as a shrine derivative on Asset, at key `textonly_pdf` 

AssetHocrCreator is now called AssetOcrCreator, as it stores not only hocr but tesseract's `textonly_pdf` on an asset. 

WorkOcrCreatorRemover is updated to take account of the fact that there are two places tesseract OCR data is stored. 

The Admin::WorksController now queues an async WorkOcrCreatorRemoverJob to update OCR state after any change to `ocr_requested?` in _either direction)_. Instead of issuing an SQL — this makes WorkOcrCreatorRemoverJob the single source for knowing about what OCR state must be added or removed. 

- AssetHoctrCreator makes both Hocr and textonly_pdf from tesseract
- WorkOcrCreatorRemover should be more efficient with only conditional maybe_remove when necessary
- WorkOcrCreatorRemover knows about textonly_pdf in addition to hocr attribute, in determining when it needs to run
- use fixed kithe method that is less indirect
- reword comments on WorkOcrCreatorRemover
- remove extraneous condition in maybe_add
- Admin::WorksController now queues an async WorkOcrCreatorRemoverJob to update OCR state
- Rename AssetHocrCreator to AssetOcrCreator, as it now adds two different ocr-related products to Asset


jrochkind:
We had to update to a new version of kithe to get a bugfix in some derivative handling methods. This wound up giving us some other dependency updates too.

PR #2252, opened 2023-07-18
A Service to create a single-page PDF out of an Asset, including text layer from OCR

jrochkind:
This uses a handful of shell-out command lines to do it's thing. Including ocr2pdf, a Python pip package, that we will need to figure out how to accomodate install of in production/test/development environments.

Couldn't find any way to accomplish our pipeline with JPEG2000 images specifically, except for the img2pdf python tool.  The alternative, JPG using ruby Prawn library -- is actually more complicated code too. 

And we think we can get significant byte size savings with JP2.  

So it seemed worth it to cross the "python dependency" barrier. I think we can do it reasonably on heroku (prod), Github Actions (test), and developer MacBooks.  But solving that is part of this.


jrochkind:
oy i've really messed up the commit history, deleting and starting over.

PR #2254, opened 2023-07-18
A Service to create a single-page PDF out of an Asset, including text layer from OCR

jrochkind:
This uses a handful of shell-out command lines to do it's thing. Including ocr2pdf, a Python pip package, that we will need to figure out how to accomodate install of in production/test/development environments.

Couldn't find any way to accomplish our pipeline with JPEG2000 images specifically, except for the img2pdf python tool. The alternative, JPG using ruby Prawn library -- is actually more complicated code too.

And we think we can get significant byte size savings with JP2.

So it seemed worth it to cross the "python dependency" barrier. I think we can do it reasonably on heroku (prod), Github Actions (test), and developer MacBooks. But solving that is part of this.


jrochkind:
OK, we're adding a lot more dependencies. 

To usual Aptfile/homebrew dependencies, we now have `qpdf` and a direct (not just as a dependency of vips) use of `imagemagick`. 

And now we have a python dependency. 

We have created a python `requirements.txt` file. Which gets automatically installed on heroku if we add the `python` buildpack, seems to work great in testing on staging. And works fairly easily on Github Actions CI, as initiated in this PR. 

Still have to think through how to handle the python dependencies on MacOS development... I think you'll just need to get python installed somehow and then run `pip install -r requirements.txt`, no big deal. 

I will do another PR to move our "brew" installs to use a `brew bundle` `Brewfile`, as we have a bunch of them now!

PR #2256, opened 2023-07-18
Re-organizing and documenting how dependencies are installed on MacOS dev

jrochkind:
We use a 'Brewfile' for 'brew bundle' installation of brew dependencies doc'd in repo, instead of relying on manual individual brew installs doc'd in README.

We try to doc using a requirements.txt for installing python dependencies, now that we unavoidably have at least one. The requirements.txt is also used for our heroku and Github Actions CI installs, to get python dependencies there.

This is becoming very convoluted, I defintiely understand why people like docker/container based setups instead. At some point we may want to consider figuring that out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ node_modules
 # https://vitejs.dev/guide/env-and-mode.html#env-files
 *.local
 
+# Brew bundle creates this pretty useless file, it's not a useful lock file like you think
+Brewfile.lock.json
+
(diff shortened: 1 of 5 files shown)


eddierubeiz:
Tried this in dev - worked great for me. Thank you!

eddierubeiz (approved):

eddierubeiz (approved):

Issue #2090, opened 2023-03-01
Make a test script for command-line utilities installed on deploy machines (eg heroku)


jrochkind:
We have a number of unix command line utilities we arrange to have installed on our heroku dynos. 

We don't have any automated testing that this actually worked. 

When we're changing something on heroku; or wanting to check our assumptions; or if we were moving somewhere off heroku -- we just need to test manually that all CLI utilities are installed and working properly. 

We do have some notes for manual testing, here: https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/2088271873/Derivative+generation+on+Heroku#Diagnostics-for-derivative-generation-software

We should have a runnable script that run on an actual live deployment machine, and checks that all of our utiltiies are installed properly, and gives good diagnostics. 

This could be bash, but also could just be a simple ruby script. We're going to have ruby available on anything we're checking. Might be easier and more maintainable. (But can we do it avoiding any gem dependencies?)

Tools include:
* mediainfo
* pdfunite
* vips
* ffmpeg
* ffprobe
* (soon) tesseract

We might make sure they are installed and run properly without error; check `--version` is as expected. For any optional configuration or plugins, check they are there. Etc.

eddierubeiz:
Three cheers for this idea.
You'll want to draw on this: https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/2088271873/Derivative+generation+on+Heroku

PR #2262, opened 2023-07-20
New PDF creation, using new AssetPdfCreator with ocr text

jrochkind:
A new WorkPdfCreator2 (note differnet name) is available, that uses the AssetPdfCreator to create PDFs with embedded OCR text. 

It can be opted into with ENV NEW_WORK_PDF=on, which we use on staging. 

The AssetPdfCreator actually needs heroku-22 stack (for mysterious reasons vips has jp2 support there but not heroku-20), so for this to be fully enabled, we are blocked by #1761 which is blocked by #2090

The cache key used for on-demand-derivatives now takes into account the `textonly_pdf` derivative present, if any -- since PDF generation now takes that as input.  Changes to `textonly_pdf` will also be seen as invalidating ZIP derivatives even though they don't really need to, but that's fine and better than much more complicated code to treat them differently. 

Work PDF creation is _very slow_ using htis new method -- about 10x as slow as it was before.  So we may want to look into ways to speed it up, and/or to use more aggressive (page-level pdf?) caching on ingest. 

----

- sample 30x30.png should not have an alpha channel, so it can work with AssetPdfCreator
- Alternate WorkPdfCreator, that uses the new OCR-aware AssetPdfCreator
- feature flag to enable new ocr-aware work pdfs: ENV['NEW_WORK_PDF'] == 'on'
- WorkPdfCreator2, make sure order of pages is correct per position
- on-demand deriv creators when making cache fingerprint take account of textonly_pdf, since we use that as input for pdf_file deriv now


jrochkind:
OK, the new PDF creator switches from `pdfunite` to `qpdf` for joining pdf pages. 

* We already need `qpdf` in new method; we can stop depending on `poppler-utils` once we replace old method with new
* qpdf in my tests is slightly faster at joining pdfs too
* qpdf has a "--linearize" option that produces PDFs that are in linearized "streaming" order, so viewers can start displaying before they've downloaded the whole PDF, and possibly generally make better memory use.

jrochkind:
OK this is still FAR TOO SLOW, as is. 

I haven't been able to actually speed up the end-to-end process... but we can do something with some additional "caching" perhaps in the form of creating an derivative(s)

But this PR as it is has the new PDF generation routine as opt-in with an ENV variable (feature flag), it won't actually be used in production. So I think it can be merged as is....

However, as it creates new cache key for PDFs, all work-level PDFs will be cache misses, and have to be re-created when requested -- regardless of feature flag, it'll happen for everybody. 

https://github.com/sciencehistory/scihist_digicoll/pull/2262/files#diff-b793b2f53674ee8a1b7b4d25063442d14877632538539decf95a217220481962

I think that's ok?

jrochkind:
Okay changed the "on-demand" derivative cache key so it includes the creator class -- so if we switch between `WorkPdfCreator` and `WorkPdfCreator2` those will have different cache keys. 

There are currently 163 works with cached on-demand derivatives (includes zip as well as PDF), so if we merge and deploy this, that many will get cache busted and have to be re-created. (Even though technically the old cached copy is still good now, we busted the cache key).  I think that's okay.

eddierubeiz (approved):

eddierubeiz (approved):

PR #2269, opened 2023-07-25
Remove webdrivers gem -- may no longer be necessary, and currently causing problems

jrochkind:
It may no longer be neccessary, selenium-webdrivers can handle webdriver install itself? And webdrivers may now be causing a problem?  Very unclear what is going on, trying this.

https://github.com/titusfortner/webdrivers/issues/247

@eddierubeiz You may need to `brew remove webdriver` on your Mac, to remove a very old unnecessary webdriver package that was being ignored before this PR, but for me caused problems after this PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -197,7 +197,7 @@ group :test do
   gem 'capybara', '>= 2.15'
   gem 'selenium-webdriver'
   # Easy installation and use of chromedriver to run system tests with Chrome
-  gem 'webdrivers'
+  #gem 'webdrivers'
   gem 'capybara-screenshot'
   gem 'factory_bot_rails'
   gem "database_cleaner", "~> 2.0"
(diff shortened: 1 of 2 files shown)


jrochkind:
Well, this seems to get us back in business, so merging it. 

On MacOS, you will probably see a warning interupting rspec output, like this:

```
2023-07-25 12:57:33 INFO Selenium [:logger_info] Details on how to use and modify Selenium logger:
  https://selenium.dev/documentation/webdriver/troubleshooting/logging

2023-07-25 12:57:33 WARN Selenium [:selenium_manager] Error getting version of chromedriver 115. Retrying with chromedriver 114 (attempt 1/5)
```

For me, the tests still run and pass though. 

I hope the community figures out general solution, things are weird at present, as detailed in https://github.com/titusfortner/webdrivers/issues/247

PR #2270, opened 2023-07-25
move and reword OCR enabled/requested checkbox

jrochkind:
Underlying attribute remains the same for now, may change later.


jrochkind:
Also with:

> fix order of hint in at least one kind of simple-form bootstrap checkbox
>
> Hint should be BELOW checkbox, and nicely lined up. I think maybe this was a mistake before?

PR #2272, opened 2023-07-27
Fix getting dpi from mediainfo CLI output

jrochkind:
We were getting with a regex -- but didn't realize that sometimes the output can have a decimal and fractions, and other times not! This kind of shows the extra gotchas of our shell out to CLI approaches, sort of a "screen scrape" approach rather than having an actual ruby library. But oh well, fixed that one.


eddierubeiz:
This has a `byebug` in it, but otherwise looks good!

jrochkind:
@eddierubeiz Thanks for noticing and fixing the byebug!

PR #2273, opened 2023-07-28
Report broken or bad "related links"

eddierubeiz:
a) Report related links that are, or redirect to, a broken URL.

b) In addition, if the final URL of the rediect chain ends in one of these endings, we don't like that.
      `bad_endings = ['stories', 'magazine', 'distillations-pod']`

Ref #2220

### DIFF
--- a/lib/tasks/reports/find_broken_related_links.rake
+++ b/lib/tasks/reports/find_broken_related_links.rake
@@ -1,0 +1,58 @@
+namespace :scihist do
+  namespace :reports do
+    desc """
+      # bundle exec rake scihist:reports:find_broken_related_links 
+    """
+    task :find_broken_related_links => :environment do
+      # a) Report related links that are, or redirect to, a broken URL.
+
+      # b) In addition, if the final URL of the rediect chain ends in one of these endings,
+      # we want to report it as a problem:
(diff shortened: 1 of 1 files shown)


eddierubeiz:
This supersedes #2245 which used `curl`.

lib/tasks/reports/find_broken_related_links.rake
```diff
@@ -0,0 +1,58 @@
...
+                # construct the next URL in the redirect chain
+                if response.headers["Location"].start_with? 'http'
+                  urls_followed.append response.headers["Location"]
+                else # relative URL
+                  parsed_url = URI(urls_followed.last)
+                  urls_followed.append(
+                    parsed_url.scheme + "://" + parsed_url.hostname  +
+                    response.headers["Location"]
+                  )
+                end
```
jrochkind, line 33: You can just tell http-rb to follow redirects, instead of doing it manually. Then you just get the terminal response as output. Might make for a lot simpler code?

https://github.com/httprb/http/wiki/Redirects

    HTTP.follow.get(url)

eddierubeiz: That's much better yeah. Done.

PR #2277, opened 2023-08-01
Create a `graphiconly_pdf` derivative

jrochkind:
The derivative is a single-page PDF containing an image (of the asset), the image in JP2 format and downscaled. 

We then use this when we create the multi-page Work PDF, putting it together with possible textonly_pdf with OCR text. WorkPdfCreator2

We also make some improvements to WorkPdfCreator2 while we're at it, like skipping  assets that are not TIFF. 

- empty commit to force deploy
- factor out #create_graphical_pdf method
- AssetPdfCreator just calls out to extrated AssetGraphicOnlyPdfCreator, for now
- create graphiconly_pdf derivative, automatically,for all TIFFs
- Rename AssetPdfCreator, now it simply combines existing derivaties, AssetPdfCombiner
- WorkPdfCreator2 ignore non-image files


eddierubeiz:
Thanks for walking me through this last week. Looks like it's coming together nicely.

PR #2283, opened 2023-08-02
optional configuration to set non-default queue for OCR related jobs

jrochkind:
We will set to 'special_jobs' temporarily during a time period we want to bulk-enable OCR for a lot of already exisiting works

> heroku config:set ACTIVE_JOB_OCR_QUEUE=special_jobs


eddierubeiz:
Clever!

Issue #2140, opened 2023-05-08
Homepage image change for Phase II Redesign


apinkney0696:
Hello! I put together some options for the header images. There are a few variations ranging from no edits to some playful edits. You can find options here: P:\Othmer Library\Digital Collections - Internal Access\Website Updates\Homepage\Header Images\JPG OPTIONS

I also have the Photoshop files for all of these, so I can change them as we'd like. I'm thinking we use the following:

BettyeB.jpg
Header_Option_2a copy.jpg
Header_Option_5 copy.jpg
Header_Option_6 copy.jpg
Header_Option_7 copy.jpg
and Header_Option_8 copy.jpg

although I would love to preview what each of these looks like on the new staging site before making a final decision. Let me know if these are the right dimensions too!

jrochkind:
@apinkney0696 

You're caling these "header images", but theyr'e definitely for hte home page on the top left, correct?  That is not an area I call the "header", but we may use different language which is fine!

Currently in that location, the image has a little caption "Wilson College chemistry club", which then links to the relevant Work. The main site redesign does _not_ generally do this when using images from the collection. 

If you'd like to continue doing that, please supply the Work that each image came from so i can make it have the caption/link.  If not, then not!  I have no real opinion on whether we should or not.

jrochkind:
I am not sure people would know to click on the image without a label; if we want people to be able to get from the image to the source record, to me it's more accessible to provide the label as a hyperlink/caption, the same as we are now.  the image as an unlabelled hyperlink does not seem right to me. 

But it's technically possible either way, sure! 

Are those links in the same order as the images you listed?  If you had a moment, it would make things easier for me and skip the step of corresponding if you just provided a list of the images and the URLs together!

apinkney0696:
Good call - we should keep the caption for accessibility then, but I do think that the image should also be linked!

The image files I have listed are just the ones I'd like to test first, but not necessarily the final options. As I mentioned above, I'd like to preview all the images included in the folder and potentially do some additional editing before deciding on the final options. Would you like the links corresponding to all 11 image options now or would it be easier for you to do the linking after I make the final selections? 

We should probably set up a meeting to go through and look at the image options in real time!

jrochkind:
You had mentioned you wanted to preview them... I had imagined that to allow you to do that, we'd put in a (possibly temporary) feature where if you put like `&hero_image=1` in the URL, it would choose a specific image. From `1` to how many images there are. 

[...]

PR #2285, opened 2023-08-02
Rotating home page hero images

eddierubeiz:
Ref #2140
We show each image to all viewers for a period of ten minutes. Then we move on to the next one, and so on.
- The number of images is arbitrary.
- To test each image without waiting, override the timer with ?hero_image=1, ?hero_image=2, and so on.
- Paths to the images, along with metadata, go in a config file at `config/data/home_page_hero_images.yml`
- Images can be commented out, reordered, or added at will; the presenter doesn't assume a particular number of files.
- Image metadata includes the path to the original files on the P drive in case we want to regenerate hero images from the existing crops of the originals.
- `images/homepage/hero_images/README.md` explains how to use `vips` for that process.
- We parse and validate the YAML file when the presenter class loads, not when the homepage displays. Errors in the configuration file will show up as soon as tests are run or as soon as anyone loads the home page, regardless of which image would normally be chosen.


eddierubeiz:
@apinkney0696  -take a look at https://github.com/sciencehistory/scihist_digicoll/blob/2fdf5e5989ccb6cd6ebcf8dfa58a03a9c27e959b/config/data/home_page_hero_images.yml

For now, just to keep things simple, we just have three basic items of metadata we were using before:

- path to the image
- caption
- ID of the work to link to.

No description, no height or width. (Image dimensions are in the file, but commented out, in case we change our minds later.)

eddierubeiz:
https://staging-digital.sciencehistory.org/?hero_image=1
https://staging-digital.sciencehistory.org/?hero_image=2
https://staging-digital.sciencehistory.org/?hero_image=3
https://staging-digital.sciencehistory.org/?hero_image=4
https://staging-digital.sciencehistory.org/?hero_image=5
https://staging-digital.sciencehistory.org/?hero_image=6

jrochkind:
Nice job with the `ticks` algorithm, very clever and clearly the right solution!

I was pondering whether to say something about those tests in the live method, or to just let it go instead of being picky -- I like your solution of moving them to tests, that makes sense.

app/assets/images/homepage/hero_images/2x/resize_to_2000.sh
```diff
@@ -0,0 +1,9 @@
+max_size=2000
+width=$(vipsheader -f width $1)
+height=$(vipsheader -f height $1)
+size=$((width > height ? width : height))
+factor=$(bc <<< "scale=10; $max_size / $size")
+
+new_name=$(sed 's/\.jpg/_new\.jpg/' <<< "$1")
+
+vips resize $1 $new_name $factor
```
jrochkind, line 9: I don't understand the "scale=10" and all this other arithmetic...

Can't you just tell `vips resize` to resize the original to 2000 pixels wide, and let it do the arithmetic?

Ah, the `resize` command annoyingly won't, but how about just use `vipsthumbnail` instead?  This also lets you convert from an original image format (say tiff) to jpg in one go.... while resizing from a full-size JPG as source is almost certainly fine in my experience, technically it's lossy-to-lossy and not best practice. 

    vipsthumbnail input.tiff --size 2000x65500 -o output.jpg

(just having the "height" be a huge number to tell it to resize width to 5000 while keeping aspect ratio is what works for me, and was necessary in older versions of vips, although might be possible to leave off the height now)

app/assets/images/homepage/hero_images/README.md
```diff
@@ -0,0 +1,23 @@
...
+new_size = 2000 # for 2x
+```
+
+```
+width=$(vipsheader  -f width  $original_giant_file)
+height=$(vipsheader -f height $original_giant_file)
+size=$((width > height ? width : height))
+factor=$(bc <<< "scale=10; $new_size / $size") # (or 1000)
+vips resize $original_giant_file $new_file_name $factor
+```
```
jrochkind, line 25: Seems no need to have the recipe documented AND in a .sh file, pick one or the other instead of two places that need to be kept in sync? 

Also see comments about recipe.

spec/components/home_page_hero_image_component_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  describe "images in Rails.application.assets" do
+    it "all needed keys and images exist" do
+      expected_keys = ["link_title", "original_file", "path_1x", "path_2x", "work_friendlier_id"]
+      all_metadata.each do |metadata|
+        expected_keys.each do |k|
+          expect(metadata[k].class).to eq String
+          expect(Rails.application.assets.find_asset(metadata['path_1x']).class).to eq Sprockets::Asset
+          expect(Rails.application.assets.find_asset(metadata['path_2x']).class).to eq Sprockets::Asset
+        end
```
jrochkind, line 22: I'm not sure about this Rails.applicaiton.assets.find_asset -- is that really public API?  OK, fine!  

I'm a bit worried this will be fragile, but I'm glad it's only in tests now, we'll find out!

I wonder if another way to do this would be getting access to ordinary Rails helper methods in tests, and just doing `image_path(thing)`, and making sure it doens't raise. 

Just musings, I am not requiring/requesting a change here. 

Nice test coverage generally! I like the idea of verifying this file in tests.

spec/components/home_page_hero_image_component_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+    end
+  end
+
+  describe "tick = 0" do
+    before do
+      allow(instance).to receive(:tick).and_return(0)
+    end
+    it "renders placeholder" do
```
jrochkind, line 31: What do you mean by the word "placeholder"?  I think of a placeholder as something used as a default when other content is unavailable. I don't think that's what this is testing for? Is this really "renders legend" or "renders title"? Or am I misunderstanding?

eddierubeiz:
@apinkney0696  - feel free to add any other editorial ideas to this list for future home pages images; I can add them into the instructions whenever.

## Create a new home page hero image:

1) Start with a large TIFF file
  * Cropped as simply as possible from an original TIFF asset
  * Wider than 2000px
  * Aspect ratio: 10 x 7

apinkney0696:
Thanks, Eddie! Here are my non-technical instruction additions...

**Create a new home page hero image:**
1. To choose an image, select images that include colorful subjects or people doing science, especially people from under-represented communities. Image compositions should fill the display window flatteringly at different screen sizes (attempt to avoid cropped heads). 
2. Download your image of choice as a TIFF.
3. Crop the original TIFF asset to the following specifications,

-     2000px or wider
-     Aspect ratio: 10 x 7

4. Save the cropped image as a TIFF to P/Othmer Library/Digital Collection - Internal Access/Website Updates/Homepage/Header Images

PR #2287, opened 2023-08-02
update kithe to 2.10.0

jrochkind:
To get a feature that will give us a rake task making it easy to queue up bg jobs to create new graphiconly_pdf for every existing suitable asset

https://github.com/sciencehistory/kithe/pull/166

> heroku run -r production "rake kithe:create_derivatives -- --derivatives=graphiconly_pdf --lazy --bg=special_jobs"


jrochkind:
@eddierubeiz I'm self-merging and deploying

Issue #2288, opened 2023-08-03
new PDF generation pipeline can't handle images with "alpha" channel


jrochkind:
Our new PDF generation generates JPEG2000 (JP2) thumbs, and uses img2pdf to embed them in PDFs. 

The img2pdf tool can't handle JP2 input with an "alpha" channel for transparency. 

Generally our TIFF sources do _not_ have an alpha channel and are not intended to -- but a handful _do_, which then causes an error.  I think these may have been images edited with photoshop, which added an alpha channel undesirably.  

When we convert from TIFF with alpha channel to JP2, using vips, the alpha remains. Then `img2pdf` complains. 

I guess we need to add another step to remove alpha channel when needed?

## References

question on vips discussion: https://github.com/libvips/libvips/discussions/3589
question on img2pdf tracker: https://gitlab.mister-muffin.de/josch/img2pdf/issues/173

## Example images

Example images with alpha:
* https://digital.sciencehistory.org/admin/asset_files/1v53jx931
* https://digital.sciencehistory.org/admin/asset_files/k643b219p
* https://digital.sciencehistory.org/admin/asset_files/7ql6dco
* https://digital.sciencehistory.org/admin/asset_files/gm80hw18z
* https://digital.sciencehistory.org/admin/asset_files/3r074v88k

Also some that are _collection thumbnails_, I didn't know/forgot that we used TIFF instead of JPG for these but makes sense... also probably examples of Photoshop adding an alpha. (we don't really need the PDF derivative for these, and were actually erroneously trying to create one, we can stop doing that)

* https://digital.sciencehistory.org/admin/asset_files/rlvzehj
* https://digital.sciencehistory.org/admin/asset_files/tu5e374

## Sample exiftool output for one image

For https://digital.sciencehistory.org/admin/asset_files/k643b219p

Note "Alpha Channel Name" field, and "Samples Per Pixel" is `4`, instead of typical `3`. Those seem to be the clues it has an alpha channel. 

Also note "Photoshop" mentioned in some metadata, providing clues this was edited with photoshop, which is not necessarily typical for our workflow. 

```
ExifTool Version Number         : 12.50
File Name                       : museum_wormianum_rv042t91s_16_k643b219p.tiff
Directory                       : /Users/jrochkind/Downloads
File Size                       : 210 MB
File Modification Date/Time     : 2023:08:03 11:57:36-04:00
File Access Date/Time           : 2023:08:03 11:57:38-04:00
[...]

PR #2294, opened 2023-08-07
Use an unreleased img2pdf to include unreleased fix for jp2 with alpha channel

jrochkind:
Temporary, should go back to normal pip install after an img2pdf release happens. Not yet included in current version 0.4.4.

ref #2288 

https://gitlab.mister-muffin.de/josch/img2pdf/issues/173

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,16 @@
 # used for installs. https://www.codementor.io/@inanc/how-to-run-python-and-ruby-on-heroku-with-multiple-buildpacks-kgy6g3b1e
 
 
-img2pdf==0.4.4
+
+# img2pdf==0.4.4
+#
+# We need an img2pdf with a fix for jp2 alpha channels that is not yet included
+# in a release.
+#
+# https://gitlab.mister-muffin.de/josch/img2pdf/issues/173
+# https://gitlab.mister-muffin.de/josch/img2pdf/commit/acc25a49265effbbffa36e053ae2a3aa633eddbf
+#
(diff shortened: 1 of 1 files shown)


jrochkind:
Oh, I can't actually reproduce failure, think it has something to do with Github CI setup, perhaps involving caching, and our unreleased version reporting the same version number as the last released version. :(

jrochkind:
OK, part of why this is failing in CI is becuase Github CI is using an older version of vips that has a bug where it sometimes produces un-boxed jpeg2000 files. https://github.com/libvips/libvips/discussions/3597#discussioncomment-6855938

And img2pdf couldn't handle unboxed jpeg2000 files... however a subsequent still unreleased commit maybe can, thanks to img2pdf author responding to my question about what was going on. 

A tangled web. 

I was hoping there would be an img2pdf 0.4.5 _release_ by now, but it appears it may be some time, so we are stuck trying to use unreleased one. 

I will try to update to LATEST unreleased one, and see if that helps.

jrochkind:
OK, passing CI now with LATEST unreleased img2pdf, and also confirmed on staging I can succesfully create `graphiconly_pdf` for an image with alpha channel that failed on img2pdf 0.4.4!

So okay, I guess we'll merge this, use unrelaesed img2pdf with unreleased bug fixes!

@eddierubeiz I'll give you some time if you want to review this in a minute you have internet access at jury duty, but if I don't hear from you may just go ahead and merge it -- but wanted to make sure you saw it, saw what we were doing, and how (specify git with specific commit in requirements.txt)

Issue #2300, opened 2023-08-07
PDF generation should not SCALE UP images believed to have less than target DPI


jrochkind:
A few images have incorrect TIFF metadata saying they are 72 dpi. 

Since our PDF generation was targetting 150 dpi, it was actually scaling these UP to be around 2x the size to meet the target! Of course, they are already enormous images, making them 2x as high makes them truly enormous -- and actually trips an `img2pdf` pixel limit, which is how I noticed. 

For these specific images, I suppose we could manually add corrected dpi metadata and/or add them to the TIFFs themselves (changing fixity). 

but as a general rule the PDF generation should never SCALE UP, if the advertised DPI (rightly or wrongly) is LESS than the target, just leave it alone, don't scale up. 

Examples that we have since they tripped `img2pdf`s maximum pixel limit:

* https://digital.sciencehistory.org/admin/asset_files/2j62s5380
* https://digital.sciencehistory.org/admin/asset_files/3f462576z
* https://digital.sciencehistory.org/admin/asset_files/fj2362467

jrochkind:
OK, created graphiconly_pdf derivatives for 3 assets mentioned above, after fixing code to NOT scale up. 

It still believes the 72 dpi and does not scale down either, so those are some really large PDF pages.

PR #2301, opened 2023-08-07
AssetGraphicOnlyPdfCreator: if original DPI appears LESS than target dpi, don't scale it UP, just leave it the same size

jrochkind:
Ref #2300

- [ ] After deploying, manually create GraphicOnlyPDF for assets mentioned in #2300

They'll still be REALLY BIG since they have incorrect metadata saying they are 72 dpi so they won't be scaled down -- but at least the won't be SCALED UP, triggering `img2pdf` pixel limit error


jrochkind:
Yeah, no test -- existing tests, which do check for dimensions on normal non-erroneous TIFFs, must pass. 

I don't feel like putting in a test for this weird edge case, would like to just get it done. Tested manually.

eddierubeiz (approved):

PR #2309, opened 2023-08-09
Allow sync_prod_to_staging to optionally not put staging app into maintenance mode

jrochkind:
In our actual work, it turns out, we'd often rather have a few hours of staging being inconsistent with some files missing, then a few hours of staging outage.

This is still by default maintenance mode, you have to choose to do it the way that will be less stable.


eddierubeiz:
Been meaning to do this for ever. Thank you!

PR #2313, opened 2023-08-10
Rake task to add missing graphiconly_pdf assets

jrochkind:
We thought we already did this, but somehow a bunch are still missing on production.  Not sure why, but let's go through again.

When new assets are ingested, they are supposed to get this derivative automatically. (Is that not working?) https://github.com/sciencehistory/scihist_digicoll/blob/master/app/uploaders/asset_uploader.rb#L93-L95

The rake task will enqueue derivative creation jobs, for those missing the deriv, on the `special_jobs` queue. So you'll have to run some special workers to process them.

    rake scihist:data_fixes:add_missing_graphiconly_pdf

- [ ] Ran in production
- [ ] Jobs completed in production

### DIFF
--- a/lib/tasks/data_fixes/add_missing_graphiconly_pdf.rake
+++ b/lib/tasks/data_fixes/add_missing_graphiconly_pdf.rake
@@ -1,0 +1,30 @@
+namespace :scihist do
+  namespace :data_fixes do
+    desc """
+      Enqueues a jobs to special_workers for any image assets missing graphiconly_pdf
+      derivative
+    """
+
+    task :add_missing_graphiconly_pdf => :environment do
+      scope = Asset.where("file_data -> 'metadata' ->> 'mime_type' = 'image/tiff'")
+
+      progress_bar = ProgressBar.create(total: scope.count, format: Kithe::STANDARD_PROGRESS_BAR_FORMAT)
+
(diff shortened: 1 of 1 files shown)


jrochkind:
Oh right -- images with alpha channels at the moment still can't be processed. Waiting for a new release of ocr2pdf, or we could use an un-released commit. 

I wonder if all the ones missing are that? Not sure.

jrochkind:
I did just verify on staging with `master` that, ordinarily, for a TIFF without an alpha channel -- after ingesting a new file, the `graphiconly_pdf` derivative _was_ created, as expected! So that is working, at least normally.

jrochkind:
Verified that at least one of the assets missing `graphiconly_pdf` does _not_ have an alpha channel. https://digital.sciencehistory.org/admin/asset_files/ukdql0k

So yeah, there's definitely cause to run this.

jrochkind:
Enqueued 3195 jobs for missing graphiconly_derivaties, to special_jobs queue

Don't really understand how that many were missing, what happened in my previous attempt to do this! But ok, running jobs now....

jrochkind:
OK, ran again. 

17 failed because of alpha channel, or giant image, or corrupt image, or otherwise weirdness. Don't know why those other 3K were missing before.

Issue #2311, opened 2023-08-10
Create a "suppress OCR" flag for Assets


jrochkind:
To disable OCR on an asset even if the Work has OCR enabled. 

(We decided Work flag is primary, and there does not need to be a way to enable an Asset if a _work_ is disabled. Just to opt-out of a parent Work's flag). 

The main use case is handling some Assets that _error_ on OCR, so they won't be re-tried night after night.  See Assets failing at #2306

If there's an asset that just has bad or useless OCR.... it could still be opted out? Not sure how much it matters. Up to collection managers. 

This value once it exists needs to be accounted for in WorkOcrCreatorRemover to decide whether or not to actually OCR. 

It also should be taken account of in our "dashboard" info at WorkOcrShowComponent -- taken account of in arithmetic for whether OCR state is consistent or not, and maybe also give a count of all Assets with OCR overwritten-disabled, if any?  Also in the asset LIST under a work in admin, where it currently has a flag for if the asset has OCR or not, maybe it needs a flag for whether OCR is suppressed?

And then finally, the `/admin/asset_files` dashboard maybe needs a filter to find Assets with OCR suppressed at the asset-level?

Any other places this should be accounted for?

These can be more than one PR of course, since there are a bunch of separate places this ends up being accounted for!

PR #2317, opened 2023-08-15
Feature branch for the `suppress_ocr` flag.

eddierubeiz:
Ref #2311 

Steps to follow:

- [x] Create the metadata.
- [x] User interface:
  - [x] Admin UI to set and unset the flag
  - [x] Admin UI to leave notes about OCR decisions
  - [x] Admin UI to see which assets have the flag set, and see the notes on the asset
- [x] Change OCR creation process:
  - [x] Don't create asset-level OCR derivs when the flag is set
  - [x] Don't attempt to include asset-level OCR derivs into work-level PDFs, etc. when the flag is set
  - [x] During the nightly run, *remove* asset-level OCR derivs, when the flag is set 
  - [x] Turn on & troubleshoot the nightly OCR sweep in staging.
- [ ] Admin UI to find assets where the flag is set, if necessary
- [ ] Rake task to adjust staging and production data, if necessary
- [ ] `bundle exec rake scihist:data_fixes:turn_off_problematic_ocr_assets` in production
- [ ] Turn on the nightly OCR sweep in production.


eddierubeiz:
This has gotten some extraneous files committed to it (e.g. gemfile.lock), so it's confusing. I'm closing it.

Issue #2278, opened 2023-08-01
Enable PDF even for single-page works


jrochkind:
It was disabled cause I was kind of thinking of the whole point of the PDF as being aggregating multiple pages in one file -- if you only have one file, why not just download a JPG?

But in fact, some people would rather have a PDF for all sorts of reasons -- plus now that we're adding OCR text, it's another reason.

jrochkind:
@jrochkind right now PDF downloads for multi-page works show up in a separate "section"....

![Screen Shot 2023-08-07 at 12 17 03 PM](https://github.com/sciencehistory/scihist_digicoll/assets/149304/e04ee215-5dfd-4a32-8517-151e5666c9ba)


While single-page works lack that section:

![Screen Shot 2023-08-07 at 12 17 25 PM](https://github.com/sciencehistory/scihist_digicoll/assets/149304/d81589b2-c3a3-4616-9b48-3c23772f2d73)


I guess for single-page works, the PDF option should show up in the "Download Selected Image" section that is present?

@eddierubeiz This is another small-ish one you could do if you run out of work -- clearly no worry of that!

apinkney0696:
That makes sense to me.

PR #2331, opened 2023-08-22
Allow PDF download for one-page works

eddierubeiz:
Ref #2278
I tightened up the tests for this component a bit as well.


jrochkind:
nice, thanks!

PR #2334, opened 2023-08-25
`suppress_ocr` #2:  Turn off problematic assets

eddierubeiz:
A rake task to set `suppress_ocr` to false for assets that are really wide or have a lot of text.
Depends on #2333.

- [ ] Ran in production


jrochkind:
deployed and ran in production!

PR #2336, opened 2023-08-25
`suppress_ocr` #4: Display on the work pages

eddierubeiz:
Depends on https://github.com/sciencehistory/scihist_digicoll/pull/2333.
Shows which assets have `ocr_suppressed` on the member list page;
displays the aggregate numbers correctly on the OCR tab.


jrochkind:
tricky with the SQL, but nice. 

I'm going to make it only show the "suppressed OCR" line when non-zero (as this is really an edge case), and also see if I can simplify the has/have with some `pluralize` function, I feel like it's supposed to be able to handle this.

PR #2337, opened 2023-08-25
`suppress_ocr` #5: Actually change `work_ocr_creator_remover.rb` to take the new metadata into account.

eddierubeiz:
Depends on https://github.com/sciencehistory/scihist_digicoll/pull/2333.
Actually takes `suppress_ocr` into account when creating and removing OCR.

- [ ] Re-enable heroku scheduled job in production
- [ ] change the settings in production to use the new PDF creation workflow -- `heroku config:set NEW_WORK_PDF="on" -r production`
- [ ] Unset OCR on special_jobs queue (after clearing queue), `heroku config:unset ACTIVE_JOB_OCR_QUEUE -r production`


app/services/work_ocr_creator_remover.rb
```diff
@@ -54,9 +59,8 @@ def maybe_add(asset)
   def maybe_remove(asset)
     # don't need it if we already don't have hocr or textonly_pdf
     return if !asset.hocr && !asset.file_derivatives[:textonly_pdf]
-
     asset.hocr = nil
-    # this kithe command will save record to, persisting the hocr=nil,
+    # this next kithe command will save record too, persisting the hocr=nil,
```
jrochkind, line 65: thanks

PR #2342, opened 2023-08-29
New collection thumb sizes for redesigned collection show

jrochkind:
- add larger collection thumb sizes for redesigned collection show page
- use new collection thumb size on collection show page

PR #2345, opened 2023-08-30
large collection thumbs, oops 2x size has to be really 2x

jrochkind:
Didn't have the sizes expressed right in #2342

After deploying, need to re-create derivatives:

    CollectionThumbAsset.find_each {|a| Kithe::CreateDerivativesJob.perform_later(a, only: :thumb_collection_show_page_2X) }

- [ ] re-created derivs in production


jrochkind:
@eddierubeiz self-merging!

Issue #2346, opened 2023-08-30
Collection list page load is slow


jrochkind:
Can be 4-5 seconds on back-end on heroku.  

Looking at logs, most of this is taken up by waiting on DB! Specifically:

>  Aug 30 14:54:08 app/web.1 DEBUG: [f9a8a788-d2d8-4a2b-bf46-4dd63c1d2ea1]   Collection Load (4510.6ms)  SELECT "kithe_models".* FROM "kithe_models" WHERE "kithe_models"."type" = $1 AND (published = true) AND NOT (kithe_models.json_attributes @> ('{"department":"Exhibition"}')::jsonb) ORDER BY "kithe_models"."title" ASC  [["type", "Collection"]]

Oh interesting. In order to select all items from kithe_models that are of type Collection, it's scanning through all kithe_model rows I guess?  I still wouldn't expect that to be so slow.  The pg optimizer may be doing something weird, like looking through the jsonb first for every single row, which would be non-ideal. 

We could have an index on kithe_models.type, but you're usually advocated not to have an index on a value with only a few possible values... but in this case at least for the use case of just identifying collections, seems maybe needed?

jrochkind:
Having trouble figuring out what's going on/reproducing.  I can't figure out why this DB fetch is taking 4.5 seconds to return, can't reproduce it trying specific commands in console.

jrochkind:
Here's where the fetch that seems to be slow happens:

https://github.com/sciencehistory/scihist_digicoll/blob/5d007bf46c60919b3eb7ee1bd0e28458f23bd0df/app/controllers/collections_list_controller.rb#L6-L10

Fetching ALL collections (SQL will be from kithe_models where type = `Collection`) is pretty quick 52ms

```
irb(main):075:0> Collection.all.to_a ; nil
DEBUG:   Collection Load (52.1ms)  SELECT "kithe_models".* FROM "kithe_models" WHERE "kithe_models"."type" = $1  [["type", "Collection"]]
```

Adding _additional_ constraints should not slow it down that much, we're just filtering _within_ the very small 95 record result set... but it really does, turns 56ms in 2000ms (in this case):

```
irb(main):078:0> Collection.where("published = true").not_jsonb_contains(department: Collection::DEPARTMENT_EXHIBITION_VALUE).order(:title).includes(:leaf_representative).to_a ; nil
[...]

PR #2347, opened 2023-08-30
limit collection list query to non-exhibitions with SQL that doesn't make postgres pathological

jrochkind:
ref #2346

### DIFF
--- a/app/controllers/collections_list_controller.rb
+++ b/app/controllers/collections_list_controller.rb
@@ -5,7 +5,8 @@ class CollectionsListController < ApplicationController
   def index
     @collections = Collection.
       where("published = true").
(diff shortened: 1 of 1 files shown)


jrochkind:
That fixed it. From 2500-4500ms with most of that in DB,  to 171ms, with 80ms in db. 


> Aug 30 15:36:59 app/web.1 INFO: [5cdbf0dd-4692-40f5-9287-b4a401d7e964] method=GET path=/collections format=html controller=CollectionsListController action=index status=200 allocations=64444 duration=171.17 view=86.23 db=80.47 ua=Chrome-116/Mac-10.15 ip=69.243.27.93

The `NOT` jsonb contains query was confusing the pg query planner somehow into doing something pathological. Oh well, this fixes it.

Issue #2282, opened 2023-08-02
Change [Missing] text to Undated


apinkney0696:
In the date facet on the left menu of search results, works with "undated" date metadata are found by a link that says "[Missing]". I would like to change this text to say "Undated" instead if it's possible with Blacklight.

eddierubeiz:
Here's a clue:
https://github.com/sciencehistory/scihist_digicoll/issues/408#issuecomment-555558656

eddierubeiz:
Did a bit more research.
In the blacklight controller, adding:

```
    config.default_solr_params = {
      # [...]
      "facet.missing": true,
```
will in fact enable a "missing" facet for `format` and `department` facets.
Sure enough, all of these facets use the label configured in `blacklight.en.yml`.

eddierubeiz:
I feel like explicitly setting `"facet.missing": false` , just as a pretext to add some comments.

jrochkind:
I don't understand why we're messing with changing how Blacklight is configured to generate Solr queries -- we don't need to change any Solr queries, right? Just labels?  I wouldn't add configuration we don't need, the easiest code to maintain is code that isn't there.

PR #2348, opened 2023-09-05
[WIP] Change [Missing] text to Undated

eddierubeiz:
Ref: #2282
This is a bit of a hack: the only facet that uses `facet.missing` is the date facet, and we're using this to our advantage.
- Set the label for *all* `facet.missing` labels to "Undated".
- Make the default setting of `"facet.missing": false` explicit in CatalogController.
If at some point we decide to turn `facet.missing` back on, we will need to find some way to make this label apply to just the date facet (presumably using some hack.)

### DIFF
--- a/app/presenters/download_options/image_download_options.rb
+++ b/app/presenters/download_options/image_download_options.rb
@@ -33,18 +33,11 @@ module DownloadOptions
       # We don't use content_type in derivative option subheads,
       # cause it's in the main label. But do use it for original.
 
-      if dl_small = asset.file_derivatives[:download_small]
-        options << DownloadOption.with_formatted_subhead("Small JPG",
-          url: download_derivative_path(asset, :download_small),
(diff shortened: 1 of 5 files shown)


eddierubeiz:
@jrochkind  - curious if you would suggest doing this some other way.

jrochkind:
i agree with you it would be better to find some way to customize label for just this facet. 

Have you found the source code (probably .html.erb?)  that is responsible for rendering this line for blacklight-range-limit?  If so, could you link to it so I can take a look?

eddierubeiz:
https://github.com/projectblacklight/blacklight_range_limit/blob/main/app/components/blacklight_range_limit/range_segments_component.rb seems to be the component involved here.

jrochkind:
Thanks! I don't see the call to the i18n label there.  Have you located the source code line that does that, likely an .html.erb?

Issue #2276, opened 2023-07-31
drop one of our download derivatives?


jrochkind:
We offer a LOT of download options. I suspect more than anyone needs; can we drop one, and simplify, and reduce our storage needs a tiny bit too?

We have:

* small (800 px wide)
* medium (1200 px wide)
* large (2880 px)
* Full sized (actual size)

in cases where the "full size" is less than "large" we drop "large". 

I wonder if we have analytics on what download sizes get used the most? (And how much do any of them get used, is this actually a feature that is seldom used at all?)

I think we can probably drop either "small" or "medium", they aren't too different.  Whichever we keep, we'll probably call "small", to have just "small" and "large"?

jrochkind:
@apinkney0696 Cool! Curious if you looked at actual usage analytics at all for these, is "small" seldom chosen?

If we drop current "small", instead of leaving two options `medium` and `large`, my inclination is to relabel teh current medium as `small`, so we still have `small` and `large`?

apinkney0696:
that's fine. And yes, small is the least used option according to GA.

PR #2350, opened 2023-09-05
Remove and stop using download_small derivatives

jrochkind:
In public UI, call the download_medium `small` -- if there are two options to users, it should be `small` and `large`, not `medium` and `larger`.

But it's harder to rename derivative internally, the name `download_medium` is actually part of its' path on S3. So we'll leave it "medium" internally everywhere, just change it to be labelled `small` in the single part of codebase responsible for presenting to user.

Also a rake task to remove all existing `download_small` derivatives, from S3 and from metadata on each asset. 

    rake scihist:data_fixes:remove_download_small_derivatives

Ref #2276

After deploying, the app will no longer CREATE or USE the `download_small` derivative. But we need to run rake task to remove existing. 

(It's okay if you run rake task when new version isn't fully deployed on heroku -- existing app checks for download_small derivative being present before presenting link to it, so it's okay if it's getting deleted!)

- [ ] Run rake task on production

### DIFF
--- a/app/presenters/download_options/image_download_options.rb
+++ b/app/presenters/download_options/image_download_options.rb
@@ -33,18 +33,11 @@ module DownloadOptions
       # We don't use content_type in derivative option subheads,
       # cause it's in the main label. But do use it for original.
 
-      if dl_small = asset.file_derivatives[:download_small]
-        options << DownloadOption.with_formatted_subhead("Small JPG",
-          url: download_derivative_path(asset, :download_small),
-          analyticsAction: "download_jpg_small",
-          width: dl_small.width,
-          height: dl_small.height,
-          size: dl_small.size
-        )
-      end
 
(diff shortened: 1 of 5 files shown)


jrochkind:
I think the rake task will take 15-20 minutes to run. Which seems okay to do as a single rake task, rather than involving bg jobs. 

Tomorrow I'll run it on staging -- which will mean staging won't have `download_small` derivatives anymore!

jrochkind:
Exact code for safely removing a derivative in a production app, in a concurrency-safe way that shouldn't leave any orphans, was taken from Shrine guide:

https://shrinerb.com/docs/changing-derivatives#removing-derivatives

eddierubeiz (approved):

eddierubeiz:
@jrochkind approved!

Issue #2271, opened 2023-07-25
Technical metadata for TIFFs


jrochkind:
For audio and video, we extract some technical metadata and store in database, where it can be displayed on Asset admin screen.  bitrate, etc. 

We don't currently do that for TIFFs -- beyond pixel width and height. 

But as we started using TIFF `dpi` (which is actually metadata embedded in TIFF) to effect PDF generation, we might want it reflected back on admin screen -- to more easily debug it etc. 

And this brought up, there might be other TIFF metadata we might want.  In general, most of this stuff is stored as metadata inside a TIFF -- it's there in the TIFF originals in Dig Coll anyway, but it can't be easily seen or filtered without downloading TIFF files and extracting it. 

Maybe we want to extract it on ingest as part of characterization? In addition to dpi, perhaps things like camera model, photo creation date, color profile, more?

(If DPI were in db metadata, we could use it from there on PDF generation instead of pulling it out from TIFF each time, which would give us some more flexibility in using derivatives as sources, etc). 

Here is everything `exiftool` knows about in a TIFF. (We don't  currently use/install `exiftool` , but possibly could. Or there are other tools that can get some but maybe not all of this, like `mediainfo` which we do already use). 

From the first page of: https://digital.sciencehistory.org/works/dvoykwl

```
[ExifTool]      ExifTool Version Number         : 12.50
[File]          File Name                       : conversations_on_chemistr_dvoykwl_3_7d2d5ks.tiff
[File]          Directory                       : /Users/jrochkind/Downloads
[File]          File Size                       : 31 MB
[File]          File Modification Date/Time     : 2023:07:25 16:05:54-04:00
[File]          File Access Date/Time           : 2023:07:25 16:05:57-04:00
[File]          File Inode Change Date/Time     : 2023:07:25 16:05:54-04:00
[File]          File Permissions                : -rw-r--r--
[File]          File Type                       : TIFF
[File]          File Type Extension             : tif
[File]          MIME Type                       : image/tiff
[File]          Exif Byte Order                 : Little-endian (Intel, II)
[File]          Current IPTC Digest             : 8f05da416079a49c8dbe23beb254942b
[EXIF]          Image Width                     : 2517
[EXIF]          Image Height                    : 4156
[EXIF]          Bits Per Sample                 : 8 8 8
[EXIF]          Compression                     : Uncompressed
[EXIF]          Photometric Interpretation      : RGB
[EXIF]          Make                            : Phase One
[...]

PR #2358, opened 2023-09-07
derived_metadata _jsonb column, move hocr to it

jrochkind:
Create new column `derived_metadata_jsonb` on kithe_models, and move Asset#hocr attr_json to it. 

Since the `hocr` data is not currently actually used by anything, we don't need to worry about race conditions in the migration, it's fine that it deploys with new location, then we migrate the data, etc. 

Prep for #2271

- [ ] run rake task to migrate existing data: `rake scihist:data_fixes:migrate_hocr_to_derived_metadata_jsonb`

- add derived_metadata_jsonb column
- move Asset#hocr to be defined on derived_metadata_jsonb
- rake task to move Asset#hocr to new location

### DIFF
--- a/app/components/work_show_ocr_component.rb
+++ b/app/components/work_show_ocr_component.rb
@@ -26,7 +26,7 @@ class WorkShowOcrComponent < ApplicationComponent
 
         SELECT
           (
-            json_attributes ->> 'hocr' IS NOT NULL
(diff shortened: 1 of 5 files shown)


eddierubeiz (approved):

eddierubeiz:
Approved! I'll let you do the merging / deploying / Rake task migration on your schedule.

PR #2360, opened 2023-09-08
Change [Missing] text to Undated

eddierubeiz:
Ref: #2282
Subclass BlacklightRangeLimit::FacetItemPresenter so we can have more control over labels.
Unlike a previous attempt, this does not "cheat" by taking advantage of the fact that facets.missing is turned off for all facets *except* the date range facet.


jrochkind:
Very nicely done, I concur this looks like how current Blacklight architecture wants us to do it "correctly". 

Thanks for the test, you never know when a future BL version might break it.

Issue #2296, opened 2023-08-07
Change to different heroku redis provider for ActiveJob support, one with much higher redis connection limit


jrochkind:
We are currently using the [redis provider from heroku itself](https://elements.heroku.com/addons/heroku-redis), but there are VARIOUS third-party ones listed on heroku marketplace. 

The heroku redis provider we are using meters "number of redis connections" as their major price point in different plans. 

The plan we are using in production has an 80 connection limit, and 300MB of memory (we aren't _close_ to using all of it), for $30/month. 

but for reasons we don't entirely understand `resque` workers make very inefficient use of redis connections -- like 2-4 connections _per worker_, or even more.   The 80 connection limit is keeping us from ramping up past ~30 workers (15 dynos with two workers each) even when we have a lot of temporary bulk work to do. 

I'm not sure why and seems like a bug... but we don't have to worry about it if we just switch to a different third-party heroku marketplace redis provider that isn't limiting redis connections like this. There are several that will give you 256-512 connections even in low-cost plans. 

We'd have to look carefully to make sure we have everything else we need, and the plan looks good. 

Some possibilities:

* stackhero, $19/month plan has 500MB of RAM and 500 connections https://elements.heroku.com/addons/ah-redis-stackhero
* rediscloud, $25/month plan has 250MB of RAM and 256 connections https://elements.heroku.com/addons/rediscloud
* memetria for even $10/month has 160MB of RAM and 512 connections, but these cheap plans don't have "high IO performance", whatever that means and not sure how it compares to other providers. https://elements.heroku.com/addons/memetria-redis

jrochkind:
At some point the number of _postgres_ connections we have in our Heroku PG plan will become the limiting factor, instead of redis.  

We still wont' be able to have more workers than pg connections available -- although hopefully they'll use a more reasonable one pg connection per worker, instead of the 2-4+ redis connections they seem to consume! Although I"m not certain of that. 

Our heroku standard-0 postgres plan has a 120 connection limit.

jrochkind:
Oh we won't save much money, cause we were using a cheaper redis on staging before. Close to a wash. 

* Before:
  * production:  $30/month
  * staging: $15/month
  * total: $45.month
* after:
  * production: $19/month
  * staging: $19/month
  * total: $38/month

eddierubeiz:
[...]

PR #2362, opened 2023-09-14
Exif Characterization before ingest

jrochkind:
Ref #2271 

The exiftool hash already includes exiftool version number used to create it, but we also add the exact command line args given to it, since that effects the keys in hash etc. We may want to change our mind about those at some point, record exactly how we got the output hash. In `Kithe:CliArgs` key. 

We also tell exiftool to do what validation it can do -- getting the actual validation warnings out from exiftool in json was a bit tricky! We do those tricks, but then copy them over ot a normalized `Kithe:ExifToolValidationWarnings` key for easier access. https://exiftool.org/forum/index.php?topic=15194.0

Will also involve some back and forth with kithe

* Making some changes in kithe so multiple metadata extraction or before_promotion characterization hooks that need a local file for asset, do NOT result in multiple downloads or copies of the source file, they can re-use. 

* Will likely extract the exiftool characterization routine itself to kithe. It was a bit tricky to get right, and makes some choices about extra metadata etc. 

- [ ] Create for existing files in production
    * `heroku run rake scihist:data_fixes:add_exiftool_result` will create bg jobs in `special_jobs` queue, which will have to be managed, will take several hours even running 10 dynos in parallel
    * We really ought to get to #2296 to allow us to run more in parallel

----------

### ALSO, 

This is really setting the ground work for #2271 -- I did extra work to try to capture exiftools-detectable validation errors, which should be enough to catch those "bad" TIFFs in #2271, and the kithe/shrine lifecycle work done here is also helpful to #2271. 

---------

- add exiftool as package dependency
- system_env_spec for exiftool
- use more realistic tiny version of a page scan, with full metadata
- initial exiftool characterization on ingest
- attempt at getting validation warnings from exiftool
- normalize convenience access for exiftool validation warnings


jrochkind:
moved exiftool characterization tools to kithe, switched to using from there.

jrochkind:
Forgot that we needed a custom heroku buildpack to get `exiftool` installed on heroku. 

https://github.com/fnando/heroku-buildpack-exiftool

Note that it will install the _latest_ exiftool on every build; we could set an ENV variable to fix it to a specific exiftool version instead, but for now I'm leaving it. 

Doc'd at https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/1915748368/Heroku+Operational+Components+Overview#Buildpacks

Issue #2363, opened 2023-09-14
Index OCR


eddierubeiz:
If this is a duplicate issue, please close this. But I can't seem to find one.

PR #2364, opened 2023-09-18
indexing for OCR

eddierubeiz:
Ignores the edge case of works with both OCR and translations / transcriptions (which is a separate problem.)
Ref #2363
- [ ] Requires reindex


eddierubeiz:
Re: edge case above:

```
Asset.where("derived_metadata_jsonb ? 'hocr'").where("json_attributes ? 'transcription'").map {|a| a.english_translation }
=> ["", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""]
```

```
Asset.where("derived_metadata_jsonb ? 'hocr'").where("json_attributes ? 'english_translation'").map {|a| a.english_translation }
irb(main):032:0>
=> ["", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""]
```

jrochkind:
@eddierubeiz  I don't understand what those excerpts you posted are demonstrating, can you explain?

eddierubeiz:
Just confirming that there's nothing to worry about - we do not have to take translations or transcriptions into account when indexing OCR.

app/indexers/work_indexer.rb
```diff
@@ -252,13 +260,29 @@ class WorkIndexer < Kithe::Indexer
...
+    work.members.sort_by { |m| m.position || 0 }.map do |mem|
+      mem.asset? && mem.published?.presence && mem.send(string_property).presence
+    end.compact
+  end
+
+  # @return [Array<String>] an array of human-readable plain-text strings
+  # containing *just* the text inside the body of the HOCR, suitable for the full-text index.
+  def ocr_for_work(work)
+    get_string_from_each_published_member(work, :hocr).map { |hocr| index_ready_ocr(hocr) }
+  end
```
jrochkind, line 279: Is this `ocr_for_work` method unused?

spec/indexers/work_indexer_spec.rb
```diff
@@ -290,6 +290,77 @@
...
+      end
+      describe "text known to be in German" do
+        let(:language_test_work) { german_only_work }
+        it "goes in searchable_fulltext_de" do
+          expect(german).to eq(expected_hocr)
+          expect(english).to be_nil
+          expect(unsure).to be_nil
+        end
+      end
+      describe "text in either English or German or something else" do
```
jrochkind, line 354: name this what it actually is. Is it in "English or German or something else"? (not even sure what that is supposed to mean!)  Looks like it's a bilingual work, so name the block that?

spec/indexers/work_indexer_spec.rb
```diff
@@ -290,6 +290,77 @@
...
+      before do
+        allow(nil_members_work).to receive(:members).and_return(nil)
+      end
+
+      let(:output_hash) { WorkIndexer.new.map_record(language_test_work) }
+      let(:english) { output_hash["searchable_fulltext_en"] }
+      let(:german)  { output_hash["searchable_fulltext_de"] }
+      let(:unsure)  { output_hash["searchable_fulltext_language_agnostic"] }
+      describe "text known to be in English" do
+        let(:language_test_work) { english_only_work }
```
jrochkind, line 339: What is the point of definining a `let (:english_only_work)` above, only to use it once here?  Just define it inline here, less indirection, makes the tests easier to read? 

    let(:language_test_work) { create(:public_work, language: ['English'], members: assets )  }

Same for the others?

app/indexers/work_indexer.rb
eddierubeiz: It is - thanks for catching.

spec/indexers/work_indexer_spec.rb
eddierubeiz: Done!

jrochkind (approved):

PR #2367, opened 2023-09-19
Properly handle an array passed into the facet.page param in the collection show controller.

eddierubeiz:
### Context
Blacklight throws a UnpermittedParameters at this particular line when a bot passes an array into the facet.page parameter. We discovered this two weeks ago (see Ref #2354).

The URL the bot brought up was:
```/collections/olbbze5/facet?facet.page[%24testing]=1&id=oh_institution_facet```

This may or may not be a problem in real life, but let's assume for the purposes of this discussion that we don't want to see the error in HB.

### What this does:
- rescues the error
- returns a `status: :unprocessable_entity` response if it is the case we recognize
    - ( in other words, if param `facet.page` is anything other than a string),
- and reraises if it *is* a string (aka something else is wrong that we might want to be alerted to by a HB error).

### Variants on this error:
The HB error actually contains 5 different variants on this error which the bot tried:

- https://app.honeybadger.io/projects/58989/faults/100003448/01H9N4MD2XZS7A0BXZHVFB4GFN?page=0
  - /collections/szmk3u4/facet?facet.page[%24eq]=2&id=subject_facet
- https://app.honeybadger.io/projects/58989/faults/100003448/01H9N4MBCDXBPTEG217M6TX636?page=0
  - /collections/szmk3u4/facet?facet.page[%24testing]=1&id=subject_facet
- https://app.honeybadger.io/projects/58989/faults/100003448/01H9N4M9WJ89TWY84B8ESA8F02?page=0
  - /collections/szmk3u4/facet?facet.page[]=2&id=subject_facet
- https://app.honeybadger.io/projects/58989/faults/100003448/01H9N3RDHG916G8Q4K2VPCKSM0?page=0
  - /collections/g445cd19z/facet?facet.page[%24eq]=2&id=subject_facet
- https://app.honeybadger.io/projects/58989/faults/100003448/01H9N3RBD4FF8EQZ56ZWR28B09?page=0
  - /collections/g445cd19z/facet?facet.page[%24testing]=1&id=subject_facet
 
 All involve passing an array to the same param, and are addressed by this PR.

### CODE REVIEW

**Comment by jrochkind** [comment]
I'm confused what's going on here. 

Our collection show controller definitely has to allow other params than `:id` and `:collection_id`, this is the controller that shows an individual collection like https://staging-digital.sciencehistory.org/collections/qfih5hl , right?

So it for instance has to allow our search params like `q` in https://staging-digital.sciencehistory.org/collections/qfih5hl?q=cuba

But if the tests are all still passing, maybe it's fine?  Or maybe we just lacked a test?

Can you confirm this doesn't break that?

Normally Rails controllers allow random unrecognized query params which just get ignored. Can you explain why you thought it was the right fix to do otherwise here and raise an error on any recognized query param?

The actual problem from issue seems to be that a "bad" value is being passed as `facet.page[%24testing]=` -- but if you are now stripping _all_ `facet.page`, I think that may actually break a part of the search?  Perhaps the "more facets" pagination?

I think you actually need a check for the specific bad value, not just stripping any unrecognized value. 

Also, I think it should be in CatalogController, where it will get inherited by the CollectionShowController, I think it's actually a problem that could apply to CatalogController too (this can be tested).

File: app/controllers/collection_show_controller.rb [hunk 1]
```diff
@@ -43,7 +43,20 @@ def facet
...
+      # a bot tries to pass an array into the facet.page param.
+      if params["facet.page"].is_a? String
+         # Reraise.
+        raise
+      else
+        # Quietly return an :unprocessable_entity.
+        return render plain: "Error: unpermitted parameters.", status: :unprocessable_entity
+      end
+    end
+
```
jrochkind, line 59: I am not opposed to rescueing this error, but if so we should do it using the Rails controller `rescue_from` feature we talked about earlier, rather than our own begin/rescue/end block. 

I also do not understand the if/else going on there.  While the comment suggests the first branch is about when an Array is being passed into `facet.page`, of course the check `params["facet.page"].is_a? String` will NOT be true if an array was passed into facet.page, will it? 

If the intention is to rescue if and only if `! params["facet.page"].is_a? String` (that's what it does I think?), we could maybe use more comments explaining why -- or if this condition really is somehow met only when it was an Array... I don't understand how that is possible so comments explaining that?

I also think it might be defensible to just rescue all `ActionController::UnpermittedParameters`

spec/controllers/collection_show_controller_spec.rb
```diff
@@ -37,5 +37,15 @@
...
   end
 
-
+  describe "Bad params" do
+    let(:collection) { create(:collection, friendlier_id: "faked") }
+    it "doesn't throw an error when facet.page is an array" do
+      get :facet, params: {
+        id:              "subject_facet",
+        collection_id:   collection.friendlier_id,
+        "facet.page" =>  { goat: 'goat' }
```
jrochkind, line 46: the title says "is an array", the actual test is a _hash_.  Which do you mean to test? Make the label match the thing?  Testing an array might look like `"facet.page" => ["one", "two"]`, no?

app/controllers/collection_show_controller.rb
eddierubeiz: Using `rescue_from` now, and better comments.

app/controllers/collection_show_controller.rb
```diff
@@ -81,4 +84,18 @@ def collection
...
+  # Suppress certain Blacklight errors in the
+  # #facet action; respond instead with :unprocessable_entity.
+  def handle_unpermitted_params
+    # This only applies to #facet.
+    raise if params['action'] != 'facet'
+    # If the facet page param is a string (this is the normal case)
+    # but we are somehow *still* getting an UnpermittedParameters, reraise.
+    raise if params["facet.page"].is_a? String
+    # Suppress if the param is e.g. an array or hash.
+    return render plain: "Error: unpermitted parameters.", status: :unprocessable_entity
```
jrochkind, line 98: Aha, I think I understand what's going on now, thanks, and sorry for misreading it before! `raise if params['action'] != 'facet'` is good. 

Would it be better/safer to `raise unless params["facet.page"].is_a? Array`?  I'm not sure what's in the middle of the two versions, or if it matters, but checking for what we actually mean may avoid having to think about it?

(I also could still go along with just swallowing _all_ of the unpermitted params honestly, but this is prob safer)

eddierubeiz: No problem - this is super confusing. Here's what I see this code doing:

We know that `facet.page` is supposed to contain a string (it's just meant to be the page number as a string).

So the intent in this handler is to suppress Blacklight's `UnpermittedParameters` errors for arrays, hashes, anything passed into that param that *isn't* a string. That should cover the various URLs caught in the HB error- all of them are variations that involve passing non-strings to that parameter.

By the time we get to the `handle_unpermitted_params` handler, we already know we have bad params, of course.

Now, in this context, if `facet.page` *is* a regular string, and we're **still** somehow getting `UnpermittedParameters`  error, I **do** want to reraise, as this isn't the problem we're targeting.

jrochkind (approved):

Issue #2097, opened 2023-03-07
date range limit with open end is ignored


jrochkind:
You used to be able to enter an open-ended-on-one-side range in date facet, by leaving one box blank, in the sidebar facet area. 

Like, enter "1900" in the "start" box but leave end box blank, then click "Apply". To get `1900-`

Or leave start box blank but enter "2010" in end box, then click "apply", to get `-2010`. 

Currently, if you do this, your date range entry is ignored completely. 

I'm not sure if something changed in the blacklight_range_limit plugin, or in our app, or what. It would be nice to be able to do that again.

eddierubeiz:
This broke on July 13th, 2022, when [we upgraded Blacklight and blacklight_range_limit](https://github.com/sciencehistory/scihist_digicoll/commit/e8b9ae0eac83baaed222cb7712193de9dab3f675).

jrochkind:
I wonder if it's a bug in blacklight_range_limit. (Or if some developer thought it was an intentional change?)  blacklight_range_limit doesn't have any developers really routinely taking responsibility for it, and is some messy code, a bug on that end will be annoying to diagnose, but I guess we probably should. And then either monkey-patch it locally, or submit a PR, or both. (since it can be hard to get attention for PR's there).

jrochkind:
Working on fix in blacklight_range_limit at https://github.com/projectblacklight/blacklight_range_limit/pull/278

PR #2368, opened 2023-09-19
Better screening for range limit params

eddierubeiz:
Ref #2355 

We were already screening for nil values, but we should actually be testing that both values are strings.


jrochkind:
I'm not sure if screening out nils here is actually something we want to do, could it be part of the problem at #2097 ?

But it looks like that problem is actually caused by a dependency... but maybe we don't want to make it worse?

Issue #2341, opened 2023-08-29
Permanently change to new PDF creation class


jrochkind:
After it's been in a production a bit, and we're satisfied. 

* Remove `WorkPdfCreator`
* **LEAVE** single existing `WorkPdfCreator2` (with `2` in it, cause classname is part of cache key)
* Remove logic for ENV `NEW_WORK_PDF`, always using `WorkPdfCreator2`

eddierubeiz:
Noted, esp. about not renaming the class!

jrochkind:
Also remove `pdfunite` from codebase and installation code at this time -- nothing will use it anymore after original `WorkPdfCreator` is removed. New one uses `qpdf` instead.

PR #2370, opened 2023-09-20
Permanently change to new PDF creation class

jrochkind:
Ref #2341

- remove WorkPdfCreator class, we now only use new WorkPdfCreator2, don't need option to switch back
- no longer need to install pdfunite, with more to WorkPdfCreator2 logic


eddierubeiz:
Deployed. Also removed the `NEW_WORK_PDF` env var from prod and staging.

Issue #2279, opened 2023-08-02
Add text and/or graphic indicating OCR presence on work


apinkney0696:
We want to indicate when a work has OCR applied. Some text on the record page beneath the large photo? Not sure at the moment what it should say - will want to brainstorm.

jrochkind:
@apinkney0696 actually here is the existing "stub" issue for this!

apinkney0696:
I've done some research about this, and I'm ready to chat about some possibilities!

apinkney0696:
Decided to pull the PDF option out of the dropdown downloads menu and create its own button with the title "Searchable PDF." Jonathan to mock up the page appearance.

jrochkind:
@apinkney0696 This ended up being a bit tricky to implement, so I decided to start with just a graphical mockup, to see if this is the right direction or not -- no problem if you think it is not, feedback welcome!

I did try adding both Zip and PDF links to the main page -- for both "Searchable PDF" and non-text PDF, but trying to label distinguishing. 

Even though you were leaning against it, I did try adding some text about the "searchable" PDF to the page -- there was just kind of a spot in the layout for text, and I thought it helped distinguish Searchable from not better.   Just to see what it looked like, can be changed. 

-----------------

![Screen Shot 2023-09-20 at 11 41 15 AM](https://github.com/sciencehistory/scihist_digicoll/assets/149304/65b68251-6d3c-4375-baa1-9b9fb23008bc)

-----------------

![Screen Shot 2023-09-20 at 11 31 33 AM](https://github.com/sciencehistory/scihist_digicoll/assets/149304/47a6effb-d296-48e0-833d-a09664ac59cc)

-----------------

The intent is that clicking on the PDF on this link will actually open up in the browser (for easier searching) instead of forcing a "download" as in other location -- although you still will sometimes get a progress bar first, if it isn't already cached. 

While not shown in this mock-up, the intention is that the PDF/Zip links _will_ still be in the "Download" popup _only_ for the _main_ image (as well as in the viewer), it will no longer be repeated for each thumbnail.

apinkney0696:
Looks nice! Here's my feedback:

- I think both buttons should say download, not one view and the other download. I don't think it should open in the same window, I don't want a user to have to hit the browser back button to get back to the Digital Collections and then lose the file. User testing with researchers has taught me that new tabs are preferred. Let's have it push a download just like in the downloads menu. There will be a full-text search in the browser eventually.

[...]

PR #2372, opened 2023-09-21
Add full-work download links to initial work page, indicating presence of searchable pdf

jrochkind:
Ref #2279 

This basically just moves the PDF and (where applicable) ZIP download links to the main work page, no longer hidden behind a menu. 

Additionally, on those on-page links, the PDF is labelled "Searchable" if it will have a text layer from OCR. 

The PDF and Zip links remain in the "downloads" menu for the main "hero" and inside the "image viewer" -- but individual page thumb download menus now only have links for the individual image. 

A fairly simple change, but required some convoluted and confusing code changes because the way we handle rendering of download menus/options winds up being convoluted.  It's currently inflexible, assumed too much about how it would be used that when it changes... is leading to even more convoluted code. 

That whole thing might could use a refactor for present usage patterns at some point.  I think the approach would be really separating the component for _presentation_ from the logic for _what to present_.  The "DownloadOption" model is already a step in this direction.  The `DownloadDropdownComponent` could perhaps do nothing but be _given_ DownloadOptions, it would be other code that would determine _which_ DownloadOptions to use, that could more easily be different in different calling contexts. Instead of trying to pass an ever-growing list of parameters into `DownloadDropdownComponent` that interact in odd permutations to try to instruct `DownloadDropdownComponent` in how to construct the DownloadOptions -- just _give_ it the DownloadOptions constructed externally separating concerns. 

But the nature of the call chains and component trees make it a bit complicated to make work. Would take some thinking through. May not be necessary... it works.


jrochkind:
Going to re-use the WorkShowOcrComponent for it's logic on identifying whether a work has OCR or not. 

In order to do that safetly, will add unit tests to WorkShowOcrComponent -- it currently has none -- including of specific methods. 

In order to do that, might need some additional Factories for Assets with OCR etc.

eddierubeiz:
Sounds good. Could you double-check there aren't any tests for that component?

jrochkind:
Awesome, I'll check that one out to at least re-use/extract test setup routines though!

jrochkind:
The logic around these on-demand download links is complicated in a bunch of ways. 

I'm doing my best to keep it DRY while keeping the code from getting too convoluted, but it's definitely a bit messy. I'm just proceeding instead of spending too much time working out ideal architectures. 

It involves some calling out to individual methods in ViewComponents to try to to keep DRY, and maybe extracting some such methods.

jrochkind:
Getting there, but we need to

- [ ] keep "whole work" download links OUT of individual thumb download menus 

-- but they should still be in the one on the viewer, AND in the main "hero" image one. The way we have a tree of components here, makes it hard to get the context to know what to do.

jrochkind:
I'm not loving that the SQL to see if all pages _really_ have OCR, on a 422 page book (granted that's big!) adds 45ms to response time. 

What if we just look at `ocr_requested` instead, and show "Searchable PDF" based on that -- even if the OCR is actually missing (maybe it's in queue) so you might not really get text....  should be an unlikely case, hopefully.

jrochkind:
We initially wanted to make the on-page PDF links open in-browser, but in a new tab (instead of forcing download), but complexities involving browser popup blocker made us punt on that for now. See #2389

We originally thought it would be nice to tag PDF as "Searchable" within the `DownloadDropdownComponent` too, but I just kind of got overwhelmed by it, need to do it efficiently.  A redesign really needs to happen as discussed above. It is still flagged in on-page links not hidden behind menu.

PR #2374, opened 2023-09-21
Refactor DownloadOption to DRY code

jrochkind:
For intended re-use at #2372

- move analytics data attributes inside DownloadOption, requires passing work-id to DownloadOption
- extract creation of on-demand-deriv DownloadOption to DRY

This code is definitely kind of squirrely, and remains so... not sure if we made it easier to understand or harder with this refactor, but we do DRY it for re-use.


eddierubeiz:
Makes sense, and thanks for spelling it out in the tests.

eddierubeiz (approved):

PR #2379, opened 2023-09-26
Do not throw error on extraneous range params

eddierubeiz:
Ref #2381
This is in response to URL

`/collections/gt54kn818?q=biomedicine&range[year_facet_isim][begin]=19899205363&range[year_facet_isim][end]=1989&range_end=2021&range_field=year_facet_isim&range_start=1989`

causing Blacklight Range Limit to throw an `unpermitted params` error.

Note: the diff for this PR might not show that the changed code is in a `if params[:range].present?` block .


jrochkind:
Yeah, I think chasing specific unpermitted params to catch is a losing struggle, when there are an infinite number. 

Possibly we should just rescue from this exception _entirely_ in CatalogController, log it, and return a http error without raising?

But actually, wait. I think there's a setting where you can have _Rails_ ignore these unpermitted params instead of raising, I think. 

In fact, I _thought_ it was the default.. and we set it to `raise` only in development and test, but have it ignore in production.... 

https://github.com/sciencehistory/scihist_digicoll/blob/master/config/environments/test.rb#L59

https://github.com/sciencehistory/scihist_digicoll/blob/master/config/environments/development.rb#L42

We don't have that setting in production... so why is it even raising at all? 

> action_on_unpermitted_parameters - Allow to control the behavior when parameters that are not explicitly permitted are found. The values can be :log to write a message on the logger or :raise to raise [ActionController::UnpermittedParameters](https://api.rubyonrails.org/v4.0.2/classes/ActionController/UnpermittedParameters.html) exception. The default value is :log in test and development environments, false otherwise.

https://api.rubyonrails.org/v7.0.0/classes/ActionController/Parameters.html

So I feel like it should be ignoring these in production already!

We could spend more time understanding what's going on and why this config isn't working this way!

OR, I'm fine totally rescue'ing _all_ UnpermittedParams in CatalogController -- removing the earlier attempt to rescue only targetted subset of them. 

This piece-by-piece for just what happened to have been sent by a bot this time, though, seems a losing battle just adding hard to understand/maintain logic.

eddierubeiz:
Works for me. I'm rescuing all UnpermittedParams in the latest version.

jrochkind:
Great, short and sweet.

Issue #2298, opened 2023-08-07
remove share to twitter option


apinkney0696:
I'm okay with keeping Facebook and Pinterest, but now that Twitter is "X" and generally a bit of a controversial company, let's just remove the option.

PR #2385, opened 2023-09-27
Remove share to twitter

eddierubeiz:
Ref #2298

### DIFF
--- a/app/components/scihist_footer_component.html.erb
+++ b/app/components/scihist_footer_component.html.erb
@@ -106,15 +106,6 @@
                                     loading:"lazy"
                             %>
                         </a>
-                        <a href="https://twitter.com/scihistoryorg" class="shi-social__link" target="_blank">
-                            <%= image_tag "from_main_website/twitter-icon.svg",
-                                    width:"22",
-                                    height:"18",
-                                    class:"shi-social__icon",
-                                    alt:"Twitter",
-                                    loading:"lazy"
-                            %>
-                        </a>
                         <a href="https://www.instagram.com/scihistoryorg/" class="shi-social__link" target="_blank">
                             <%= image_tag "from_main_website/instagram-icon.svg",
                                     width:"19",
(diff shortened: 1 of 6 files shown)


app/components/scihist_footer_component.html.erb
```diff
@@ -106,15 +106,6 @@
...
                         </a>
-                        <a href="https://twitter.com/scihistoryorg" class="shi-social__link" target="_blank">
-                            <%= image_tag "from_main_website/twitter-icon.svg",
-                                    width:"22",
-                                    height:"18",
-                                    class:"shi-social__icon",
-                                    alt:"Twitter",
-                                    loading:"lazy"
-                            %>
-                        </a>
```
jrochkind, line 117: i don't believe the ticket meant to include removing this one?  Double-checking, this link to Institute twitter is still on the main website footer, so I believe we still want to match that here.

app/views/works/_meta_tags.html.erb
```diff
@@ -31,13 +31,6 @@ fb:app_id Facebook ID A Facebook app ID
 fb:pages  Array<Facebook ID>
 fb:profile_id %>
 
-<meta name="twitter:card" content="summary_large_image">
-<meta name="twitter:site" content="@<%= Rails.application.config.twitter_acct %>"/>
-
-<%# not sure twitter really uses these, but maybe it used to? %>
-<meta name="twitter:label1" content="Rights Statement"/>
-<meta name="twitter:data1" content="<%= attributes.rights_statement %>"/>
-
```
jrochkind, line 40: I think we should leave these -- We don't have a twitter share button anymore, but someone, anyone, without our coordination, can still share our page to twitter, and when they do, we should still provide the meta tags to enable twitter to produce a good tweet.

app/views/application/robots.text.erb
```diff
@@ -52,10 +52,6 @@ Disallow: /works/*/transcription$
 <% else %>
 # Non-production, no robots please, although we let twitter in to test twitter integration
 
-# let twitter scrape cards metadata for testing
-User-agent: Twitterbot
-Disallow:
-
```
jrochkind, line 58: I'd leave this in here just in case we still want to test twitter share on staging, it doesn't really hurt.

jrochkind:
This does quite a bit more than is in the ticket, let's just remove the share button as in the ticket?

Issue #2378, opened 2023-09-26
Remove n+1 from download options parent lookup


jrochkind:
When we added the zip link to the individual asset download options, with these lines:

https://github.com/sciencehistory/scihist_digicoll/blob/a36e8f1e085be41c2d75ccc94c54e600735248a9/app/presenters/download_options/image_download_options.rb#L24-L28

We added an example of what can be generally called an "n+1 query problem" to our page -- _every_ ImageDownloadComponent on a page (one per thumbnail) repeats this query, potentially going to the database one time for each thumbnail. 

It turns out Rails saves this from happening with a per-request "query cache", as evidenced by below log lines. Which keeps it from being catastrophic. But it's still sloppy and a bit fragile -- and also leads to all those log lines in dev, which is annoying!

```
  CACHE Kithe::Model Count (0.0ms)  SELECT COUNT(*) FROM "kithe_models" WHERE "kithe_models"."parent_id" = $1  [["parent_id", "9a5614e3-d2a2-4278-86ed-2deede5c586c"]]
  ↳ app/presenters/download_options/image_download_options.rb:24:in `options'
  CACHE Kithe::Model Count (0.0ms)  SELECT COUNT(*) FROM "kithe_models" WHERE "kithe_models"."parent_id" = $1  [["parent_id", "9a5614e3-d2a2-4278-86ed-2deede5c586c"]]
  ↳ app/presenters/download_options/image_download_options.rb:24:in `options'
  CACHE Kithe::Model Count (0.0ms)  SELECT COUNT(*) FROM "kithe_models" WHERE "kithe_models"."parent_id" = $1  [["parent_id", "9a5614e3-d2a2-4278-86ed-2deede5c586c"]]
  ↳ app/presenters/download_options/image_download_options.rb:24:in `options'
```

(repeat once per thumb on page). 

This is actually a general problem that comes up with our component-based architecture for displaying thumbs, I've had this problem before, where it's hard to figure out how to get "state" information down a component tree. 

In this case, I can think of two alternative sensible solutions:

1.  Pass in a "should show zip link" param to the ImageDownloadOptions component -- it's caller, the DownloadDropdownComponent already has the state it needs to decide this. (not preferred, see number 2)
2. Just have DownloadDropdownComponent add the zip link on to what ImageDownloadOptions provides -- DownloadDropdownComponent already knows what it wants to do and has that state, to decide to for instance _not_ show the "whole work" download options int eh first place. 

[...]

PR #2386, opened 2023-09-27
Extract logic on whether to include a PDF to OUTSIDE ImageDownloadOptions

jrochkind:
ImageDownloadOptions just gets a boolean whether to include pdf, the caller has to decide if it wants it or not.

Is this better? Hard to say. All this dropdown logic has become very convoluted. I have not been able to figure out how to make it incontrovertibly better, and am just resigning to this being good enough. What this does do is:

* Resolve the n-query problem in #2378. I still don't totally understand what was even going on there, but this resolve it.
* Set things up for #2372 to be easier, where in some cases the dropdown menu shouldn't include the whole-work-download options at all.

Separating this out from #2372 for easier review.

I am definitely not thrilled with how weird this has all become, but best I can do for now. If you have any other ideas though, let's talk.

### DIFF
--- a/app/components/download_dropdown_component.rb
+++ b/app/components/download_dropdown_component.rb
@@ -113,7 +113,11 @@ class DownloadDropdownComponent < ApplicationComponent
     @asset_download_options ||= if asset&.content_type&.start_with?("audio/")
       DownloadOptions::AudioDownloadOptions.new(asset).options
     else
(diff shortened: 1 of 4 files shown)


eddierubeiz:
This isn't bad at all. Also, thanks for separating the two PRs. Merging.

PR #2387, opened 2023-09-28
keep very lengthy attributes out of Asset#inspect

jrochkind:
To keep displays of Asset in logs and console from being so unmananageable


jrochkind:
The ActiveRecord `filter_attributes` feature's use case was really maybe sensitive info like passwords, but we can also use it to keep some of our enormously lengthy json attributes from making `puts some_asset` in console, or output of an asset in a log, completely unuseable due to crazy lengthy output.

eddierubeiz:
Many thanks.

PR #2388, opened 2023-09-28
Fix padding on cookie banner for narrow screens

eddierubeiz:
Ref #2388.
Removed the `flex-shrink` css property.
Fixes the problem on phone, and doesn't seem to make a difference on other platforms I've been able to test on.

### DIFF
--- a/app/frontend/stylesheets/local/accept_cookies_banner.scss
+++ b/app/frontend/stylesheets/local/accept_cookies_banner.scss
@@ -15,7 +15,6 @@ nav.accept-cookies-banner-nav {
 
   .i-accept-copy {
     flex-basis: 25rem;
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Screenshots from my phone (which is where I actually noticed the bug in the first place).
Iphone 12, IOS 16.7, Safari:

Before
After

jrochkind:
Huh, I probably had SOME reason to think I wanted `flex-shrink:0`, although perhaps I was just in error. 

@eddierubeiz Can you make sure to test at various window sizes (like just drag the browser window through a range)  to make sure the flex-shrink wasn't keeping it from looking bad at some of them somehow?

jrochkind:
thanks!

PR #2389, opened 2023-09-28
Ability for the OnDemandDerivative system to deliver inline links

jrochkind:
For use in #2372, extracted to separate PR

Ackowledging the lack of tests. This stuff wasn't well tested to begin with. It is difficult to test because we aren't great at testing Javascript; and because content-disposition headeres on download links aren't even currently available in test where we use shrine local file adapter instead of S3 adapter in production.

Regrettable, but I coudln't figure out a reasonable way to improve things. I did test manually. I also improved the docs at top of JS code.

### DIFF
--- a/app/controllers/on_demand_derivatives_controller.rb
+++ b/app/controllers/on_demand_derivatives_controller.rb
@@ -6,15 +6,25 @@ class OnDemandDerivativesController < ApplicationController
 
 
   # Returns a JSON hash with status of on-demand derivative, including a URL
-  # if it's available now
+  # if it's available now at `file_url`.
   #
   # GET /works/:id/:derivative_type
   #
-  # returns JSON status info, including URL to download derivative if status success
+  # Will default to a URL with content-disposition attachment from S3, but for
+  # inline client can request:
+  #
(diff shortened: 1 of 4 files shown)


jrochkind:
OK, so the original brief was to have capability to have links for PDFs display in browser window inline, but in a new tab. 

It turns out Chrome (and presuambly other browsers) popup blockers make the new tab part challenging, when combined with our lazily-created on-demand derivatives that can involve a progress bar lasting 10+ seconds after click until delivery. 

We can open the new tab immediatey on click -- but the browser focuses on it, while it remains blank for the 10+ seconds of the progress bar in the original tab you can't even see anymore.  That's not a good or even probably acceptable UX. 

We can't really open the new tab only after progress bar completion, cause of popup blocker logic. 

I suppose we could do something so the progress bar, on completion, said "click here to view PDF", and _that_ would work, since the popup blocker issue is that you can only open up new tab on an actual user click... 

hmm, might try that a bit, or might just punt on this for now.

jrochkind:
yeah, not going to do it for now, but this code can remain as notes to look at if we want to come back to it.

Issue #2312, opened 2023-08-10
Allow manual upload of tesseract products


jrochkind:
We currently use tesseract OCR to produce both an `hocr` file and a `textonly_pdf`. 

There are some Assets for which OCR uses too much RAM on our existing heroku infrastructure to complete. See #2306 

We could look into a way to run these on larger heroku dynos... but it's only a handful of items.  They can be easily  run on a developer Mac laptop. 

What if we want to do that and then attach them in production? We could do it, kind of inconveniently, by writing code interactively by hand in the developer conflict. But that's kind of error prone and a pain. 

So probably if we want o do this, we want a feature that supports it?

Web GUI, something in "Asset" management where you can just upload, via the web, the two files, HOCR and textonly_pdf?  (Once you do so, it should turn off the suppress OCR flag from #2311 if necessary, because this presumably means you want OCR!)

We're considering this approach we're not sure yet.

jrochkind:
Normally our tesseract products can be "idempotently" created -- not quite the right word maybe, but normally we can run our OCR tasks again at any point to re(create) them. 

Are we worried about flagging these as externally created and imported, in some way so it would take some kind of extra step to overwrite and delete them?

The main use case for this, I guess, is just for when our automated tesseract _fails to complete_ because it needs more resources. So maybe it doesn't really matter, you're not going to succesfully overwrite them if you tried, cause it would fail?

Unless we started using the feature for, like, corrected/improved/QA'd OCR files or something. Not really sure. This is a bit of a cumbersome feature, really.

PR #2392, opened 2023-09-29
Form to replace hocr and textonly_pdf with local files

eddierubeiz:
Ref #2312 
If all goes well we won't be using this feature a lot, but I would like it to work as well as possible when we do use it.
For now, I am treating the hocr and textonly_pdf as ephemeral.


app/views/admin/assets/show.html.erb
```diff
@@ -405,5 +405,6 @@
     <% else %>
       <p class="mt-3 mb-3">No OCR for this asset.</p>
     <% end %>
+    <%= render"upload_hocr_and_pdf", asset: @asset unless @asset.suppress_ocr %>
```
jrochkind, line 408: So, the ones we WANT to manually upload tesseract products for.. have `suppress_ocr` set, because we don't want the nightly process trying to OCR and getting a guaranteed error!

So that's the actual use case.... I think this conditional will mean the manual upload won't be available for the actual use case we have?

eddierubeiz: Oh, you're absolutely right! Will remove that.

jrochkind:
Normally our existing automated tesseracct process will create both a hocr and a textonly_pdf.  We don't normally expect just one but not both of these -- while I tried to write all the code I wrote to fail gracefully if it's missing what it needs, it is an unusual circumstance. 

Right now this has two separate forms to upload them both separately, if I am reading it right, making it easy/possible/convenient to upload just one.  Which we don't actually want to happen, it would be a mistake. 

What do you think about, instead, a single form with fields for both hocr and textonly_pdf, so you can upload both at once? And maybe it actually errors if you don't _actually_ upload both at once (and they both pass the limited validation!)?  it actually asks you to and makes you upload both at once, so they're both there?  Since a "good" file in our system needs to have both? 

Thoughts?

app/views/admin/assets/show.html.erb
jrochkind: And then I was thinking before... should the suppress_ocr bit be automatically _unset_ after succesfully uploading both files? (See also below about uploading both files together)

eddierubeiz: Yeah - of course we can try doing that.

jrochkind: OH and then though... how do we get the work re-indexed after a succesful assignment of manual teseract products, so it will show up in our site search?  

Should we just re-index it after succesful assignment?  (Maybe in a bg job?). If someone is uploading manual prdoucts for _multiple_ assets in a work, it'll get reindexed after each one, so multiple times unnecessarily... but that's prob not a big deal and is not an expected use case anyway.

eddierubeiz:
Sure - I can take both forms (and controller methods) and merge them - easy enough.

app/views/admin/assets/show.html.erb
eddierubeiz: Yup - good call. We already reindex the work on any change to any member asset's OCR status; not a problem.

app/controllers/admin/assets_controller.rb
```diff
@@ -180,6 +180,41 @@ def refresh_active_encode_status
...
+      return
+    end
+
+    # validate hocr
+    hocr = params[:hocr].read
+    parsed_hocr = Nokogiri::XML(hocr) { |config| config.strict }
+    unless parsed_hocr.css(".ocr_page").length == 1
+      redirect_to admin_asset_url(@asset), flash: { error: "This HOCR file isn't valid." }
+      return
+    end
```
jrochkind, line 198: I think you have to rescue a `Nokogiri::XML::SyntaxError` here, parallel to how you rescue PDF::Reader::MalformedPDFError below. 

Is this enough code that we should pull it out into a helper object? that then maybe can be refactored to be a bit cleaner. But doesn't need to be TOO much, not trying to add lots more work, just pulling out into a helper object would be good no matter what (and maybe adding like two tests of it? possibly replacing some of the controller tests?)

jrochkind (approved):

jrochkind:
@eddierubeiz I left another comment sorry, but marked approved after you take care of it however you think best.

jrochkind (approved):

PR #2395, opened 2023-10-03
Ingest/Promotion Validation

jrochkind:
During the shrine "promotion" step, we want the ability use logic to cancel promotion based on validation, and leave validation errors stored somewehre. 

Ref #1652

Then we want to catch certain corrupt files by looking at exiftool validation errors. And in some cases unexpected content_types. 

We do this by using the kithe `before_promotion` hooks, and `throw :abort` (a standard convention for ActiveSupport callback chains, which `before_promotion` is). But then we have to store the validation errors too. We store them in shrine metadata under `promotion_validation_errors` key -- just as plain strings for now. 

This was all a bit tricky to do becuase we have to _save_ the asset to store the validation errors, but _without_ triggering another promotion (which would be an infinite loop as it gets aborted again).  Just a bit weird to figure out, fighting a bit with our perhaps over-complex use of shrine via kithe. 

We should consider making this a feature in `kithe` that doesn't have to be implemented app-level, since it's a bit tricky, but for now we'll maybe link to this PR in kithe, and see how it works out. 

We also provide some cover methods and scopes for accessing the errors, and finding and identifying assets with errors.

Then we will need some UI for showing them in our admin screens, and possibly alerts.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -25,6 +25,12 @@ class Asset < Kithe::Asset
 
   set_shrine_uploader(AssetUploader)
 
+  scope :promotion_failed, ->{
+    # jsonb in file_data column, metadata.promotion_validation_errors exists as array,
+    # we consider that promotion failed.
+    where(%Q{file_data @> '{ "metadata": { "promotion_validation_errors": []}}'})
+  }
+
   # We are doing a weird thing with shrine making it use an attr_json attribute instead
   # of a db column. We 1) create the `attr_json`, then 2) in the shrine attachment we tell it
   # column_serializer:nil tells shrine not to serialize the JSON, let
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -335,12 +335,12 @@ class Asset < Kithe::Asset
     end
   end
 
-  def ingest_validation_errors
-    self.file_metadata["ingest_validation_errors"]
+  def promotion_validation_errors
+    self.file_metadata["promotion_validation_errors"]
   end
 
-  def ingest_validation_failed?
-    ingest_validation_errors.present?
+  def promotion_failed?
+    file_attacher.cached? && promotion_validation_errors.present?
   end
 
   def invalidate_corrupt_tiff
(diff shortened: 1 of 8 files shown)


jrochkind:
Hi @apinkney0696 !

We are going to solve "[Catch corrupt original files (graphics and sound) earlier and more cleanly #1652](https://github.com/sciencehistory/scihist_digicoll/issues/1652)" with logic that checks files upon ingest, and can cancel ingest the file appears invalid in some way. 

The next question is how we will display or alert staff (such as mainly you) to these failed ones. 

First step, we will just simply it as a badge next to asset in listings, as well as on the asset detail page. 

-------------
Screen Shot 2023-10-03 at 3 32 45 PM

-------------
Screen Shot 2023-10-03 at 3 32 35 PM

----------------
Screen Shot 2023-10-03 at 3 32 25 PM

-----------

I think that is a necessary part of our UI, but definitely not sufficient -- we know how easy it is to miss things like this!

So next....

Perhaps in some part of the Admin dashboard we warn if there are ever non-zero "ingest failed" assets.  Like a big red block at the top of the `Assets` (and/or `Works`?) dashboard pages?

That idea is based on the assumption that if something fails, we would relatively quickly delete the asset and re-ingest it. If we instead are going to leave it lying around in 'failed' state for a while, then we'd just start ignoring the big red dashboard notice "5 assets failed ingest" or something, so it might not work. 

Either way, on _top_ of that, maybe we need some kind of actual push alert?  Email?  Slack? Both? Every time an asset fails? (So if 50 fail one day, you might get 50 messages)?  

Hmm, what might be easiest here is to provide a _slack_ message (to what channel), approximately hourly, listing assets that failed in the last hour. (Technically, we'd implement that just by logging it, and then having papertrail alert slack when such log messages are found). 

Any thoughts on these and other approaches?

Cc @eddierubeiz

apinkney0696:
The combination of the red alert badge and a push alert (maybe we could make a separate channel for alerts in slack?) seems like a great option. I think if we have those two failsafes and I _still_ manage to miss a corrupt file, then a prominent error message on the admin dashboard could be a good backup option, though I don't think we'll need it. Thanks - everything looks great!

jrochkind:
Created an alert in papertrail for production, to send alerts to #digital-technical if the string `AssetPromotionValidation:` is seen in logs, with 10 minute granularity. 

I'm marking this ready for review -- but will do some more work subsequent work in other tickets, for ease of review.

PR #2403, opened 2023-10-05
Asset promotion validation for no content-type detected

jrochkind:
application/octet-stream is the generic "binary" content type which effectively means no content type detected.

This will also catch 0-byte files, as they can't have a content-type detected.

If it ends up catching files we want -- we'll have to fix our content-type detection, cause even if we want them, we don't want them without being labelled properly with content type!


jrochkind:
We can see we had to copy some weird boilerplate about succesfully saving validation error. 

We really want to figure out how to pull this out into kithe, and have it be supported with better API. But we'll mess with it a bit locally first.

PR #2415, opened 2023-10-11
Update to view_component 3.x

jrochkind:
We finally have a version of Blacklight 7.x that supports view_component 3.x. And will soon have a version of blacklight_range_limit that does. 

So we can actually get on board with recent view_component -- without needing to update to Blacklight 8 yet, which we still have some blockers for. 

Some of our tests made reference to the view_component test controller at `controller`; it's now changed to `vc_test_controller`, and is maybe meant to be internal API.. but there's some test setup we don't know how to do without it, and it still works as long as we use the new name (and view_component docs even _recommend_ it for devise test integration, https://viewcomponent.org/guide/testing.html#rspec-configuration), so it's good enough for now. 

- update view_component to 3.x, now that we're on a version of Blacklight that allows it
- view_component 3.x uses 'vc_test_controller' instead of 'controller' for it's test controller.
- ~temporarily need to use blacklight_range_limit from main git, waiting release supporting view_component 3.x~ (update to blacklight_range_limit 8.4.0)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,9 @@ gem 'scout_apm'
 # version may require an update with yarn to `blacklight-frontend`, has to be
 # checked manually.
 gem "blacklight", "~> 7.34.0"
-gem "blacklight_range_limit", "~> 8.3.0" # version no longer sync'd with blacklight, not sure how we tell what version works with what version of BL
+
+# Temporarily need to use blacklight_range_limit master for view_component 3.x support.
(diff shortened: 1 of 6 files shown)


jrochkind:
just awaiting a release of blacklight_range_limit instead of using it off git `main` branch.

jrochkind:
With blacklight_range_limit 8.4.0 released to support view_component 3.x, we can now update to view_component 3.x!

Issue #2322, opened 2023-08-18
Asset admin page: navigation between adjacent assets in the same work


eddierubeiz:
Let's look into making it easier to browse through the admin asset pages sequentially.

Compare:
- https://digital.sciencehistory.org/works/3n203z173/viewer/f7623c65f
- https://digital.sciencehistory.org/admin/asset_files/f7623c65f

The viewer page has a `30 / 32` and `<` and `>` buttons, which makes it easy to navigate between sequential pages of a book.
The admin asset page doesn't. Should we add these features to it?

## Why we might want to do this
**Use case**: Identifying and correcting small metadata problems (OCR and transcription especially) that affect pages that occur in sequence, in works that have a lot of pages.

Once a book has more than 10 pages or so, _going "up" to the work's asset view, finding the asset you were looking at, shifting your focus to the next or previous asset, and going back "down" to the new asset_ is tedious when you have to do it several times in sequence, and adds cognitive load.

## Why we might not
The page is really crowded already.

jrochkind:
Sort of a pain to do because of how the web works -- currently this page has no idea what previus and next asstes would be. 

But I guess not too hard, we just need to load the whole asset list on the individual asset page, to find the current assets place in teh list, and what next/previous assets are?  Or perhaps cleverly SQL for just the immediately preceding and succeeding assets somehow?  Or another solution?

It can be done if it's important -- has this come up as an actual admin user request that would make their lives easier?

eddierubeiz:
Absolutely did not intend to close this. Unclear how I even did it.

PR #2418, opened 2023-10-13
Retrieve previous and next members in a work

eddierubeiz:
Ref #2322 .

The UI is clunky, but perhaps good enough for a first pass.
I may actually add a second issue for figuring out where to place the links on the page.

Requirements:
- Should work with a combo of assets and works as members
- Doesn't break if there are nil or duplicate positions
- Doesn't break if the positions aren't contiguous integers
- Should be moderately fast
- Should work if the asset or work has no parent

### DIFF
--- a/app/controllers/admin/assets_controller.rb
+++ b/app/controllers/admin/assets_controller.rb
@@ -30,6 +30,9 @@ class Admin::AssetsController < AdminController
 
     @exiftool_result = Kithe::ExiftoolCharacterization.presenter_for(@asset&.exiftool_result)
 
+    previous_and_next = MemberPreviousAndNextGetter.new(@asset).previous_and_next
+    @previous_model, @next_model = previous_and_next[:previous], previous_and_next[:next]
+
     if @asset.stored?
       @checks = @asset.fixity_checks.order('created_at asc')
       @latest_check   = @checks.last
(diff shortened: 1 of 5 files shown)


eddierubeiz:
Featuring my new friends `LAG` and `LEAD`.

jrochkind:
i did not know about lag/lead... I would hae just done two queries `> $current_pos, LIMIT 1`, and `< $current_pos LIMIT 1`. (with ordering ot break ties consistently etc). 

Which I think would have been fine!  But maybe using this fancy thing is fine too? pg-specific? Also fine. As long as there's a test that will fail if it breaks in the future!

eddierubeiz:
(Did a bit of research - turns out `LAG` and `LEAD` are pretty widespread.)

jrochkind:
Very well organized code, with tests, thanks! I like making the helper class, and calling it in the controller. 

Do we really need the _names_ of the previous/next assets on screen? I think not, especially because asset titles aren't very meaningful?

Unless @apinkney0696 has asked for them or wants them, I think simple links  `« PREVIOUS | NEXT »` are best, perhaps greyed-out/not-a-link when you are at first/last?  Similar to our bottom of the page search results pagination. 

Putting extra text on the page can end up just confusing with too many things for the eye to look at, instead of helpful. There is already a lot on this page! 

If you _really_ wanted to, you could make the title a hover popover of some kind, since we do already have it, but I suspect it's not necessary!  As a user, do you want it?

eddierubeiz:
I'll take all these suggestions - thanks!

jrochkind:
And actually, you know what, if we don't need to display the title, taht actually saves you some SQL fetches, does it not? You are doing separate fetches to fetch in the records to get title?

There is probably a way to do that all in one fetch... but if you don't need the titles, and I don't think we do, it's not an issue, don't do those extra fetches!

eddierubeiz:
Good call- stay tuned...

jrochkind:
looking good, thanks

Issue #2353, opened 2023-09-06
Large numbers for year facet leads to SOLR error


eddierubeiz:
Error is:
```
RSolr::Error::Http - 400 Bad Request
Error: Invalid Number: 19679496550 for field year_facet_isim
```

Two examples:

```
/focus/food-science/facet?id=year_facet_isim&range[year_facet_isim][begin]=19679496550&range[year_facet_isim][end]=1967
```

```
/catalog?f[place_facet][]=California--Fullerton&f[rights_facet][]=Copyright%20Free&q=1&range[year_facet_isim][begin]=19679769164&range[year_facet_isim][end]=1967&search_field=all_fields
```

PR #2422, opened 2023-10-23
Don't allow dates after 3000

eddierubeiz:
Ref #2353


jrochkind:
i'm sure solr cxan handle more than 3000, but I guess for dates this is reasonable? Hopefully won't bite us somehow, once we start adding logic like this it always seems to bite us somehow!

PR #2431, opened 2023-11-09
Rails 7.1

jrochkind:
Passing and working on Rails 7.1, but still with deprecation warnings and other config that needs to be updated. 

We'll leave this in draft until that is done in a separate PR, but doing in a separate PR for ease of review. 

- Rails 7.1 requires explicit require of active_job/log_subscriber for some reason
- Rails 7.1,  plus updates to blacklight, qa, and browse_everything necessary for Rails 7.1
- move restricted derivatives to factory and stop using #dup on ActiveRecord
- fix param check for Rails 7.1 where inner hash is also an ActionController::Parameters instead of Hash
- more comments on rack 2.x lock


jrochkind:
Weird

> Your bundle only supports platforms ["arm64-darwin-21"] but your local platform
is x86_64-linux. Add the current platform to the lockfile with
`bundle lock --add-platform x86_64-linux` and try again.

Why have we not encountered that before? But I will do what it says.

jrochkind:
Deployed to staging, basic sanity check that it deploys and displays, didn't QA every feature or anything

jrochkind:
@eddierubeiz I don't want to merge this until I have more things done on Rails 7.1 -- getting rid of deprecations and such -- but this ready for your _approval_ now!

eddierubeiz (approved):

eddierubeiz:
Done. Thanks for refactoring the test code into the factory too.

Issue #2433, opened 2023-11-17
Find published works with 0 public assets


apinkney0696:
There are a handful of items (I'm not sure how many) in the Booth collection, and perhaps in the DC in general, that may have 0 published assets on a published metadata record. If we can identify these, then I can bulk edit to publish the assets or leave for you to publish if it's easy to do on your end.

Please only look for items that have 100% unpublished assets and are not Oral Histories - there are a couple of records in the DC that are only partially public for copyright reasons. We fully digitized and included all assets from these works because their copyright expiration is within the next couple of years and they can be made fully public at that point.

eddierubeiz:
- https://digital.sciencehistory.org/admin/works/i1fumf5 (Letter from Mary A. Post to James Curtis Booth, December 17, 1874)
 - https://digital.sciencehistory.org/admin/works/bakyr0u (Letter from Mary A. Post to James Curtis Booth, June 22, 1879)
 - https://digital.sciencehistory.org/admin/works/byjqznl (Letter from James Curtis Booth to Mr. J. R. Savage)
 - https://digital.sciencehistory.org/admin/works/y324fqg (Letter from Frederick Sylvester & Brother to James Curtis Booth, October 28, 1880)
 - https://digital.sciencehistory.org/admin/works/ljed9mk (Letter from F. C. Yarnall to James Curtis Booth)
 - https://digital.sciencehistory.org/admin/works/zdgfasg (Letter from Samuel W. Levis to James Curtis Booth, April 1883)
 - https://digital.sciencehistory.org/admin/works/us1tcd1 (Letter from Sanderson Smith to James Curtis Booth)
 - https://digital.sciencehistory.org/admin/works/7fmggmu (Letter from Marcus Benjamin to Margaret M. Booth, March 28, 1888)
 - https://digital.sciencehistory.org/admin/works/pq82gi2 (Letter from Emily Winsor to James Curtis Booth)
 - https://digital.sciencehistory.org/admin/works/truandq (Letter from M. Alsberg to James Curtis Booth)
 - https://digital.sciencehistory.org/admin/works/s1hmlbi (Letter from Rev. Bishop Morris to James Curtis Booth)
 - https://digital.sciencehistory.org/admin/works/a1sdtup (Letter from Dr. Charles M. Wetherill to James Curtis Booth, December 13, 1863)
[...]

PR #2434, opened 2023-11-17
Report of published works with no published members

eddierubeiz:
Ref #2433 

```
heroku run "DRY_RUN=true  bundle exec rake scihist:data_fixes:published_works_with_no_published_members"
```


eddierubeiz:
When you have a second @jrochkind  - could you review this? We can then merge it; run it in staging (dry run first); and run it in prod (dry run first).

jrochkind:
I would have written logic as "at least one member that is published" instead of "all members are non-published" -- but if the attribute is truly boolean this should be identical?

But I wonder if the attribute not being truly boolean somehow is part of how it missed ssx4mym?  

Have you figured out why it missed `ssx4mym`, does this code _currently_ catch `ssx4mym`? We would ideally want to know that before merging -- you could `copy_data` `ssx4mym` to your local dev copy to see if it gets caught or not?

jrochkind (approved):

eddierubeiz:
OK, I'm happy with this and have tested it thoroughly in staging.

Issue #2437, opened 2023-11-21
Unreliable OH system test


eddierubeiz:
This system test tends to fail unpredictably. It's quite a waste of time.
rspec ./spec/system/oral_history_with_audio_spec.rb:397 # Oral history with audio display With OHMS synchronized transcript and ToC can display, and search, without errors

PR #2438, opened 2023-11-21
Attempt to fix flaky test

eddierubeiz:
This adds an extra `expect` right after changing tab to OH description.
Technically this shouldn't be necessary, but fixing these flaky systems test is more of an art than a science.
Ref #2437 
I'll run CI tests a bunch of times on this and see if I can get it to fail.

### DIFF
--- a/spec/system/oral_history_with_audio_spec.rb
+++ b/spec/system/oral_history_with_audio_spec.rb
@@ -414,6 +414,7 @@ describe "Oral history with audio display", type: :system, js: true do
       expect(page).to have_selector("*[data-ohms-hitcount='transcript']", text: "43")
 
       click_on "Description"
+      expect(page).to have_css("#ohDescriptionTab[aria-selected='true']")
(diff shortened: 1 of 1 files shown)


eddierubeiz:
See https://thoughtbot.com/blog/write-reliable-asynchronous-integration-tests-with-capybara

jrochkind:
I would actually say technially this _should_ be necessary, and this makes sense -- after you tell capybara to click on something that will change the page, you have to tell it to wait for that change to take effect, before it can interact with the changed page. 

That `expect` is how you do it, this is actually all "science" rather than "art"!

jrochkind:
This is confusing to talk about, and I'm having trouble finding docs on specifically this sort of issue, but here are capybara's docs on it's waiting in general, that extrapolate to this

https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends

jrochkind:
Handled with #2448 instead

PR #2440, opened 2023-11-21
ONLY AFTER WEEKS OF Rails 7.1 in production: Final Rails 7.1 deploy

jrochkind:
fully opting into all Rails 7.1 defaults

Including ones that move various serialization formats (such as cookies) to things NOT backward compat with Rails 7.0, so once this is deployed app cCANNOT seamlessly go back to 7.0. So we deploy this only after several weeks of succesful Rails 7.1 deployment.

ONLY TO BE DONE AFTER AT LEAST A COUPLE WEEKS OF SUCCESFUL RAILS 7.1 LIVE IN PRODUCTION

Final followup to #2439 and #2431


jrochkind:
just remembered this, let's try to deploy it next week!

eddierubeiz (approved):

eddierubeiz:
Yup! Approved it pro forma. We can do Tuesday for sure.

jrochkind:
We should do a staging deploy just for safety, will record here when it's been done.

jrochkind:
We're going for it!

PR #2441, opened 2023-11-22
fix tests for new Rails 7.1 default action_dispatch.show_exceptions = :rescuable

jrochkind:
We can (and must) actually test for 404's now instead of testing for the unrouteable exception -- actually much better.


eddierubeiz (approved):
Great- so much simpler.

Issue #2429, opened 2023-11-06
Add Time of Request to Oral History Requests Page


sarahschneiderSHI:
We would appreciate your help adding the time that an oral history request was made to the section on the Oral History Requests page where there is already the date of the request. This would allow us to quickly and easily see what time the oral history was requested.
<img width="107" alt="image" src="https://github.com/sciencehistory/scihist_digicoll/assets/103284115/f4627c09-5c7e-437c-96e7-099aab107b8f">

PR #2442, opened 2023-11-22
Use time (and date) instead of just date for OH request list

eddierubeiz:
Ref #2429


eddierubeiz:
See https://staging-digital.sciencehistory.org/admin/oral_history_access_requests

eddierubeiz:
Done.

jrochkind:
@eddierubeiz I see you removed the "OH #" column?  This was intentional, and cleared with OH staff that it's not useful to them, they don't want it? 

Did you remove it because we needed the space? With the line wrap happenign in the date/time anyway... I'm not sure it actually takes up any more horizontal space than it did before? But was it in fact better to remove a column?

Or if they just don't use it or don't need it anyway, I guess might as well remove it when we're looking at this?

jrochkind:
Ah, ok!  Approved!

jrochkind (approved):

Issue #2411, opened 2023-10-10
Sweep for corrupt files


apinkney0696:
This ticket is to see if there are any existing corrupt files currently in the repository (at least ones that we currently know how to detect, vis-a-vis #1652)

Issue #2407, opened 2023-10-09
Detect and block accidental JPEG instead of TIFF


jrochkind:
If an Asset is NOT an OH portrait or a collection thumb, do not allow it to be a JPEG (or other image format?), image asset must be a TIFF!

While this came up when doing asset ingest validation at #1652, this particular problem isn't really amenable to that appraoch, since it relies on the characteristics of the parent Work, and other mutable metadata. 

An asset can change it's "role" or it's Work, or it's Work's metadata (including oral hsitory genre) _after_ ingest. So checking these things as Asset ingest isn't really sufficient. Also, it can be a performance problem (n+1 queries) to insist on having the parent Work available at time of Asset save/promotion. 

The better place to check this might be upon Work publishing, which I think was @apinkney0696 's original suggestion in fact.  It may or may not actually fit into ActiveRecord validation approach, due to it's need to consult parent Work though -- again, want ot be careful of performance causing work to be fetched. 

we will need to investigate.

PR #2447, opened 2023-11-28
[WIP] Scan for missing files or metadata; scan for corrupt files based on that metadata.

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2411
Still needs some more adjustments.

### DIFF
--- a/lib/tasks/reports/scan_for_corrupt_files.rake
+++ b/lib/tasks/reports/scan_for_corrupt_files.rake
@@ -1,0 +1,50 @@
+namespace :scihist do
+  namespace :reports do
+    desc """Scan for missing metadata, missing files, incomplete characterization, and corrupt files.
+
+      bundle exec rake scihist:reports:scan_for_absent_files_or_metadata
+      bundle exec rake scihist:reports:scan_for_corrupt_files
+
+      See also:
+        bundle exec rake scihist:data_fixes:add_audio_characterization
+        bundle exec rake scihist:data_fixes:add_exiftool_result
+
+
+    """
+    task :scan_for_absent_files_or_metadata => :environment do
+      progress_bar = ProgressBar.create(total: Asset.count, format: Kithe::STANDARD_PROGRESS_BAR_FORMAT)
+      Kithe::Indexable.index_with(batching: true) do
+        Asset.find_each(batch_size: 10) do |a|
+          if !a.stored?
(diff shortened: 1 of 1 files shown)


eddierubeiz:
The sweep reports no problems in staging (which has been very recently synced).

jrochkind:
Awesome that no existing known problems!

I guess since we theoretically have protections in place to prevent ingest, there's really no reason to try to run this automated regularly, and this completes teh ticket?

jrochkind:
Hmm, @eddierubeiz I'm not sure this is running all the same checks that we instituted for ingest validation? 

I understood the ticket as about taking the same checks we recently added for ingest validation, and running them on existing files -- did you understand it the same way, and do you think you have done that?  Or did you understand the ticket differently?  

I haven't found the PR's yet for that recent ingest valiation work, have you?  The issue linked here links to #1652, would have to look at comments and references in that ticket to track down the specific ingest validations added referenced there.  I'm not sure this PR is actually checking those? But it may be? What do you think?

eddierubeiz:
Yup, we're on the same page.  The relevant PRs that beefed up our on-ingest validation code are:
  - Check files are stored and have a content_type: #2403, which introduces `invalidate_unknown_type`. Note: I'm not calling that method (it's trivial; I just run the code in the rake task.)
 - For tiffs: #2395, which introduces `invalidate_corrupt_tiff`
 - For audio: #2406, which introduces`invalidate_audio_missing_metadata`

So this code is just running the two most importand methods, `invalidate_corrupt_tiff` and `invalidate_audio_missing_metadata`, just as if the asset was in the process of being ingested, and reporting any errors flagged by those methods -- and indeed there are none.

Note that I'm not testing for JPEG assets (this doesn't really fall under the scope of this ticket, but see #2407 )

eddierubeiz:
> I guess since we theoretically have protections in place to prevent ingest, there's really no reason to try to run this automated regularly, and this completes teh ticket?

I think there's no need to run this in a cron job. This is just to reassure us that all assets currently in the collection pass these tests. Future assets will be tested on ingest.

Note: if we make the ingest tests more stringent in the future by adding further requirements into the invalidation methods, we can run this test again to ensure all existing assets meet the new criteria.

jrochkind:
Ah thanks for explaining, great. Approved, alhtough I guess it's already run!

jrochkind (approved):

PR #2452, opened 2023-12-04
Add `barnes` gem (needed to report ruby runtime metrics to heroku)

eddierubeiz:
Ref #2449

This PR adds the `barnes` gem, which reports Ruby runtime metrics to Heroku, where we can monitor them at e.g.

Note: to turn on ruby runtime metrics:
```
heroku labs:enable "runtime-heroku-metrics"
heroku buildpacks:add -i 1 heroku/metrics
```
After these are run and this PR is merged and deployed, you will see three more time series graphs on the `Metrics` tab for the app in Heroku's admin interface. (e.g. https://dashboard.heroku.com/apps/scihist-digicoll-staging/metrics/web), namely
- Ruby: Heap Objects
- Ruby: Puma Pool Usage
- Ruby: Free Memory Slots

https://devcenter.heroku.com/articles/language-runtime-metrics-ruby

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -249,3 +249,8 @@ end
 # to configure Blacklight, for automatic retry
  gem "faraday", "~> 2.0"
  gem "faraday-retry", "~> 2.0"
+
+
+# Barnes reports Ruby runtime metrics to Heroku, where we can monitor them.
+# See https://devcenter.heroku.com/articles/language-runtime-metrics-ruby
+gem "barnes"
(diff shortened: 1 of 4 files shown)


jrochkind:
Hmm. config/puma.rb isn’t actually the file we use on heroku!  Please also add to config/heroku_puma.rb — I’m surprised we got metrics at all this way, wonder if maybe they aren’t reliable!

Perhaps we should add a big comment warning to `config/puma.rb` since we are doing something a bit odd here (maybe we shoudln't be and should just use config/puma.rb on heroku -- but I didn't want to run multiple puma workers in dev and test etc!)

jrochkind:
thank you for url to heroku docs next to barnes before_fork line, love it!

jrochkind (approved):

Issue #2436, opened 2023-11-21
Update to Python from 3.11. 4 to 3.11.6


eddierubeiz:
Just noted the following warning during a deploy to Heroku:

```
remote: -----> No Python version was specified. Using the same version as the last build: python-3.11.4
remote:        To use a different version, see: https://devcenter.heroku.com/articles/python-runtimes
remote:  !
remote:  !     A Python security update is available! Upgrade as soon as possible to: python-3.11.6
remote:  !     See: https://devcenter.heroku.com/articles/python-runtimes
```

jrochkind:
let's ask Heroku a feature request for specifing a range of python, to get the latest patch when it comes out automatically

eddierubeiz:
I finally put in that feature request and it's at https://help.heroku.com/tickets/1337432 . Will report back if and when they respond. In the meantime, moving this to "review".

PR #2456, opened 2023-12-05
Adding a runtime.txt specifying python-3.11.6

eddierubeiz:
Ref #2436.
I followed the instructions at https://devcenter.heroku.com/articles/python-support#supported-runtimes .
This PR just adds a file explicitly stating the python version.
I've tested `rspec system_env_spec ` against the new version of python and everything still works fine. I think this is safe to merge.


eddierubeiz:
Merged a bugfix to `rspec system_env_spec` into this branch just now...

eddierubeiz:
And now  I was able to run `system_env_spec` with python  3.11.16 in staging. Everything works fine.

I can confirm that:
```eddie-m1-laptop:scihist_digicoll erubeiz$ heroku run "python --version && rspec system_env_spec" --app scihist-digicoll-staging
Running python --version && rspec system_env_spec on ⬢ scihist-digicoll-staging... up, run.9257 (Standard-1X)
Python 3.11.6

System Environment
  mediainfo
    is present
    has acceptable version
  imagemagick
    convert
      is present
      has acceptable version
    identify
      is present
      has acceptable version
      runs
  vips
    is present
    has acceptable version
    includes build feature
      pdf load
      jpeg2000 save
      tiff save
      tiff load
      jpeg save
    vipsthumbnail
      is present
      has acceptable deveversion
      can read a pdf
      can write a jp2
  ffmpeg
    is present
    has acceptable version
  ffprobe
    is present
    has acceptable version
    can fetch URLs
  qpdf
    is present
    has acceptable version
  python CLI utilities
    img2pdf
      is present
      can create a pdf from a jp2
  tesseract
    is present
    has acceptable version
    has expected language packs
  exiftool
    is present
    has acceptable version

Finished in 3.21 seconds (files took 0.52569 seconds to load)
32 examples, 0 failures
```

jrochkind:
Is there any way to tell it to use the latest python that's `3.11.x` or even `3.x`? Or does it only support specifiying a specific version?

If the latter... we'll probably need to pay more attention to it than we want to, to keep it up to date and not running on an old version that has un-applied security patches -- the same thing the heroku warning was warning us about in the first place!

Python is really a side-issue for our app, and we don't _really_ want to pay attention to it, or know how to fit it into our workflow to pay attention to it...  I wonder if there's like a Github Action that could tell us when there's a newer version than runtime.txt or something.... or if it's worth asking Heroku support about if there's a way to do this....

jrochkind:
I'm a bit annoyed that we are still responsible for manually monitoring for security patches... I guess hopefully heroku will give us the same warning if we are out of date, as it's giving us now?   Like it will no longer say:

```
remote: -----> No Python version was specified. Using the same version as the last build: python-3.11.4
remote:        To use a different version, see: https://devcenter.heroku.com/articles/python-runtimes
```

But maybe still will say?
```
remote:  !
remote:  !     A Python security update is available! Upgrade as soon as possible to: python-3.11.6
remote:  !     See: https://devcenter.heroku.com/articles/python-runtimes
```

Hey, just so we know and can document here, do you want to try deploying with runtime.txt saying 3.11.5 to staging, to see if we still get that warning at least? (although it's pretty easy to miss in our sea of other heroku output, at least it's something!)

Also, how about figuring out how to file a heroku feature request asking for ability to specify "latest python matching spec" to heroku?

jrochkind:
If you feel like it, still seems to me to register with heroku the feature request for a range. We are paying customers after all.

PR #2458, opened 2023-12-05
Fixes to env spec helper

eddierubeiz:
Ref #2459

Now works (and passes) in staging.


jrochkind:
Awesome thanks!

It's also possible we don't need the `require rspec` at all, since of course we're running this thing with `rspec`... but unclear, and this seems fine! Merging!

jrochkind:
Oh, it's still marked draft! Not merging! But approved!

jrochkind (approved):

Issue #2460, opened 2023-12-05
Oral History Request Dashboard


jrochkind:
Rather than send an email to every user per OH requested with links to files... we will provide an online "dashboard" of their requests, showing multiple requests, including pending and approved. 

Breakout of #1877 

This is meant to make things easier for the significant number of user who make multiple requests. And possibly better also for the majority who make only one request -- or at least no worse. 

Rather than make users create an account, we will deliver "magic links" that login with a token to their email. To avoid friction for especially one-request users -- you do NOT need to auth to make a request, you can make a request un-auth'd -- it'll be attached to the email address you entered, you'll get access to it from "magic link".  

Since access to email provided access to protected files before -- and allowed unauthed users to trigger emails with files and info to be sent to arbitrary email addresses already -- the security model is unchanged. 

For first round at least, the admin dashboard will also remain unchanged -- admin will still approve/reject/etc each request one by one. 

## A wireframe draft of a potential 'dashboard' 

![dash](https://github.com/sciencehistory/scihist_digicoll/assets/149304/73a122c1-cabe-4d2b-9dba-213dea6b3ce7)

## Considerations

- Note we used to keep emails themselves in the database _encrypted_ with application level encryption. This was probably overkill anyway; we won't do so anmore, since we need to associate multiple requests by email address. 
- [ ] Do we need to user-test this to make sure it is intelligible? Including workflow around expired magic links etc. 

## Additional development jrochkind would like to do to polish it off
- [ ]  Consider how long 'logins' should last in browser; likely separate OH login into separate cookie to more easily support this (devise login is in separate cookie)
- [ ]  If you are logged in and make a request, redirect to dashboard _before_ filing out the form!

## Future questions

- [ ] Time limit on approved files?  Can you keep logging in and accessing them forever, or is there a time limit?  See comments on renewals [below](https://github.com/sciencehistory/scihist_digicoll/issues/2460#issuecomment-1864847304), needs more specification on what UX looks like when they expire (renewal?), not yet implemented. 
- [ ] Right now multiple individual emails are sent out when you have access (auto or manual) (and none are sent out just for submitting request for manual).  Do we need **batching** of emails though? (which prob necessarily means there will be a delay to get)
- [ ] Should we provide a link somewhere for the user to "sign in" or go to dashboard, other than finding link in their email? On the Oral History homepage? Would they find it there anyway? (may need user testing)
- [ ] Do admin need to see a user's "dashboard"? 
[...]

PR #2464, opened 2023-12-07
Move OralHistoryAccessRequest patron_email to a separate associated record

jrochkind:
First step of #2460, we make a new associated object Admin::OralHistoryRequesterEmail -- each one can have one or more Admin::OralHistoryRequests, and the requests will no longer hold an email attribute, but instead be linked to that RequesterEmail object.

Except actually the patron_email attribute still exists for existing data; the code can handle it in both places (the method looks in both); after deploy we'll migrate the old data, then remove the db column and code for it, for a nice migration.

In this change everything else is the same, we just move where that data is. For easier review, and step by step implementation. 

it could be deployed anytime, but we'll actually leave it in draft until we get further, to verify this approach is actually useful and working out. 

## After deploy

- [ ] Run rake task to migrate data
  -  `rake scihist:data_fixes:migrate_oh_request_email`
  - Test on staging first?
  - Will actually leave email in old place too. Double-check data make sure it's in new place?
- [ ] Prepare separate PR to remove old db column and associated logic


app/models/admin/oral_history_requester_email.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+# -- we let people request without authenticating, so all we know is the
+# email they specified -- names and other things are attached to individual requests,
+# becuase we don't know they are authorized to overwrite OR SEE any existing info
+# at time of request!
+#
+# But this is the record that logins (really one-time magic link logins)
+# are attached to, that DO confirm the user has access to the email specified!
+class Admin::OralHistoryRequesterEmail < ApplicationRecord
+  self.filter_attributes += [ :email ]
```
eddierubeiz, line 12: Duplicate line!

PR #2466, opened 2023-12-11
Patch in Rails fix for Rails 7.1 memory consumption increase, prior to it's release in presumably Rails 7.1.3

jrochkind:
See https://github.com/rails/rails/pull/50298

And ref #2449


eddierubeiz (approved):

jrochkind:
While our RAM usage has overall been much lower on average since we deployed the changes to reduce allocations -- we did just get some more R14's this morning. 

I'm going to merge and deploy this and see how it effects our average and max/spiked RAM usage!

Issue #2446, opened 2023-11-28
enable heroku ruby runtime metrics, particularly interested in puma pool usage


jrochkind:
https://devcenter.heroku.com/articles/language-runtime-metrics-ruby

As we keep getting RAM errors, this will help us understand our puma usage better, hopefully let us know if, under our actual experienced traffic, we can make do with fewer puma workers to stay under RAM limits -- or if we will need to increase our heroku resources to add more puma workers.

jrochkind:
settings changed in production. Deploying shortly. 

```
jrochkind-shi scihist_digicoll (master ?) $ heroku labs -r production
=== User Features jrochkind@sciencehistory.org
[ ] app-overview                         An overview page for an app
[ ] dashboard-mlab-shutdown-banner       Enable display of mlab shutdown banner
[+] dashboard-pb-websockets              allow user to connect to particleboard for eventbus events via websocket [alpha]
[ ] metrics-beta                         App Metrics Beta Features
[ ] metrics-chart-config                 Set order and visibility preferences for your metrics charts
[ ] metrics-data-lag                     Adds an indicator to your graphs when data is still processing, or not available. Prevents graphs from falling off before data is available, or when a data gap occurs.
[ ] metrics-p99-max-response-times       Modifies the Response Time plot to include P99 and Max latency time series.
[ ] metrics-platform-status-events       Display Heroku platform incidents and scheduled maintenance for your application’s region in the Application Metrics Events chart.
[ ] metrics-request-volume               Modifies the Throughput plot to change units from request per minute to requests per second and also display request volume stratified by status code.
[ ] metrics-response-time-summary        Shows summary average statistics for response time over a given time period.
[ ] metrics-seven-day-lookback           Allows you to see data from the past seven days of data on the same chart
[ ] metrics-three-day-lookback           Allows you to see data from the the past two 24-hour periods
[ ] metrics-three-day-lookback-combined  Allows viewing the past three days of data on the same chart
[+] mfa-required                         Allows rollout of mfa requirement to all users [limited]
[ ] team-internal-routing                Enables access to Internal Routing in Private Spaces for a team

=== App Features ⬢ scihist-digicoll-production
[ ] app-alerting               Allows the creation of alerts for Web Dynos
[ ] build-in-app-dir           Make the build system use /app as the build directory
[ ] go-language-metrics        Display language specific metrics for Go apps
[ ] http-end-to-end-continue   Send 100-continue headers to the backend
[...]

PR #2468, opened 2023-12-12
Remove barnes config from development puma

jrochkind:
Barnes is used for Heroku statistics capturing, from #2446

But the config/puma.rb is onlu used in development, where this is irrelevant AND since we don't actually have cluster mode configured here, it doesn't do anything, and generates a warning:

> Warning: You specified code to run in a `before_fork` block, but Puma is not configured to run in cluster mode (worker count > 0 ), so your `before_fork` block did not run
> Puma starting in single mode...

Remove this non-functional code from config/puma.rb.  It's already also in config/heroku_puma.rb where it's actually used.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -5,15 +5,6 @@
 # and maximum; this matches the default thread size of Active Record.
 #
 
-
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Thanks!

PR #2474, opened 2023-12-18
remove unused db column OralHistoryAccessRequest#status and #notes

jrochkind:
We have a used column 'delivery_status'. Perhaps initially we were considering calling it 'status', and they both wound up in there? 'status' is not used in code, all rows in prod db have nil value for it. We can remove it to avoid confusion.

We never got around to using #notes, I guess either? All nil in db.  We are going to give it a more specific name now that we're about to use it in OH request dashboard redesign


eddierubeiz (approved):
Approving, but I'll let you merge and migrate.

PR #2480, opened 2023-12-21
Holding area for Oral History requests dashboard

jrochkind:
Ref #2460

Holding branch/PR where we will merge individual PR's in one at a time to make code management of this feature easier, without landing things on master until we have more confirmation that this approach is working out. 

## After deploy
from #2464 

- [ ] Run rake task to migrate data
  -  `rake scihist:data_fixes:migrate_oh_request_email`
  - Test on staging first?
  - Will actually leave email addr in old place too. Double-check data make sure it's in new place?
- [ ] Prepare separate PR to remove old db column and associated logic


jrochkind:
Aha, tried to run `scihist:data_fixes:migrate_oh_request_email` on staging for a test, and indeed it didn't work, it's gotten stale. `ActiveRecord::RecordInvalid: Validation failed: Patron email must be blank`

So this is why I tested first, will be working on getting this working again. Let's get this merged!

PR #2485, opened 2024-01-04
Enhancmenets to FileListComponent for use in OH request delivery

jrochkind:
The FileListItemComponent shows an individual file in a list of files. It was formerly mostly only used on the public work pages for _restricted_ Oral Histories, where the public would see any public files -- basically only the "Front Matter" PDF!

So it was _really_ only polished for this use case.  Although admins could see other files there, marked with "Private", we hadn't really polished how these files displayed or worked. 

_________________
Screen Shot 2024-01-04 at 4 58 54 PM
_________________

It was also used on the "Downloads" tab for public Oral Histories -- again only for a PDF file, the public transcript. 

We are going to re-use this component in the re-designed Oral History in-browser request delivery, for delivering both PDFs and Audio files. 

We had to add/polish some functions/features to it, to make it work how we wanted for audio files. 

Notably, we want a generic File Icon used for a thumbnail, instead of just a cross-box. (The latest FontAwesome "free" package had one that was okay -- not perfect, but the best we found that was licensed for free re-use and modification. We deliver it as an inline SVG). 

For FLACs, we also want clicking on the title to load the M4A version directly in the browser (at least for first pass) -- not the FLAC original as previous, it's too big!

And with that in mind, while the component displays original file size ordinarily, which works great for PDFs -- for FLACs, the huge original file size is disconcerting, and they have their choice of FLAC or much smaller M4A -- it's a bit confusing how to display all this information and options, this is just first pass. But for audio and video we now show _duration_ and _not_ file size on initial line. 

_________________
Screen Shot 2024-01-04 at 5 04 07 PM
_________________

Or, without "private" badge, and with full download menu with both derivative options, as we'll use it on OH request page:

_________________
Screen Shot 2024-01-04 at 5 05 41 PM
_________________

## All commits/changes

- make lines in file listing shorter, to fit three next ti thumb icon
- audio assets get default audio file icon thumbnail
- initial basic FileListItemComponent spec
- FileListItemComponent for FLAC should link inline to M4A derivative
- fix  Asset audio factories to have duration_seconds as float, matching actual app behavior
- FileListItemComponent for A/V includes duration instead of file size
- FileListItemComponent has new ability to turn off show_private_badge

### DIFF
--- a/app/components/file_list_item_component.rb
+++ b/app/components/file_list_item_component.rb
@@ -86,8 +86,18 @@ class FileListItemComponent < ApplicationComponent
     if asset.content_type.present?
       details << ScihistDigicoll::Util.humanized_content_type(asset.content_type)
     end
-    if asset.size.present?
-      details << ScihistDigicoll::Util.simple_bytes_to_human_string(asset.size)
+
+    # For A/V include duration, else include original file size
+    # Note that in some cases there may be alternate derivatives -- FLAC size may be
+    # huge, but maybe m4a is what you'll download.
+    if asset.content_type&.start_with?("audio/") || asset.content_type&.start_with?("video/")
+      if asset.file_metadata["duration_seconds"].present?
+        details << helpers.format_ohms_timestamp(asset.file_metadata["duration_seconds"])
+      end
+    else
+      if asset.size.present?
+        details << ScihistDigicoll::Util.simple_bytes_to_human_string(asset.size)
+      end
     end
 
     str = details.join(" — ")
--- a/app/components/file_list_item_component.rb
+++ b/app/components/file_list_item_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class FileListItemComponent < ApplicationComponent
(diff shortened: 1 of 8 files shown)


app/components/file_list_item_component.rb
```diff
@@ -12,11 +12,17 @@ class FileListItemComponent < ApplicationComponent
   # @param download_original_only [Boolean] default false. If true, instead of the
   #   download menu component in the action column, there will be a single download
   #   button to download original.
-  def initialize(member, index:, view_link_attributes: {}, download_original_only: false)
+  #
+  # @param show_private_badge [Boolean] default true. If true, when a private item is
+  # visible to the user, it will be marked with a 'Private' badge, to warn (usually admin)
+  # they are seeing something the public wouldn't see. Main use case for 'false' is
+  # when we we are showing OH by-request items to authorized users.
```
eddierubeiz, line 19: Got it.

eddierubeiz (approved):

Issue #2477, opened 2023-12-20
Delete Parent, Promote Children


apinkney0696:
If possible on your end, please promote the children in the following works to individual records while retaining their metadata. The original parent work may be removed. Thank you!

https://digital.sciencehistory.org/admin/works/e172xgb

https://digital.sciencehistory.org/works/ulbrb9w

https://digital.sciencehistory.org/works/xoe9glc

https://digital.sciencehistory.org/works/vx7n5ou

PR #2505, opened 2024-01-18
Rake task to promote the child works of four specific works.

eddierubeiz:
Ref #2477 .
This unpublishes the parent works rather than delete them. I'll do that manually after I run this.
Tested in staging.
- [ ] run in production
- [ ] delete the 4 parents manually


jrochkind:
Can you put checkboxes in the Issue to check off a) running this in prod, and b) the manual work you plan to do afterwards?

jrochkind (approved):

Issue #2478, opened 2023-12-20
Staff Admin Oral History requests: add filter by status


jrochkind:
The staff dashboard page for reviewing oral history requests should have a filter by status, that especially lets staff look at only "pending" ones (ones requiring a response!)

This was something that was asked for/prioritized in past discussions with @sarahschneiderSHI and @rachellane12 , but accidentally fell out in the wash.

PR #2507, opened 2024-01-19
Filter OH requests by status

eddierubeiz:
Ref #2478.
See also the "add pagination" PR linked below

### DIFF
--- a/app/controllers/admin/oral_history_access_requests_controller.rb
+++ b/app/controllers/admin/oral_history_access_requests_controller.rb
@@ -1,7 +1,6 @@
 # STAFF FACING
 # Public facing actions are in app/controllers/oral_history_access_requests_controller.rb
 require 'csv'
-require "active_support/all"
 
 class Admin::OralHistoryAccessRequestsController < AdminController
   def index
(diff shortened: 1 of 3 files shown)


app/controllers/admin/oral_history_access_requests_controller.rb
```diff
@@ -1,12 +1,28 @@
 # STAFF FACING
 # Public facing actions are in app/controllers/oral_history_access_requests_controller.rb
 require 'csv'
+require "active_support/all"
```
jrochkind, line 4: You should not need to require `active_support/all` here, as we're in a Rails app, it should already be there I think?

Curious why you added this; if it's actually needed, let's documetn why? (And it makes me worry what's going wrong or what I don't understand if it's actually needed!)

app/controllers/admin/oral_history_access_requests_controller.rb
```diff
@@ -1,12 +1,28 @@
...
+
+  def status_filter_options
+    statuses = Admin::OralHistoryAccessRequest.delivery_statuses.keys
+    capital_statuses = statuses.map {|s|  ActiveSupport::Inflector.titleize(s) }
+    items =  ([['Any', 'any']] + capital_statuses.zip(statuses)).to_h
+    helpers.options_for_select(
+      items,
+      params.dig(:query, :status) || ""
+    )
   end
```
jrochkind, line 24: should we bother memo'izing this so it isn't re-calculated every time it's called in display? not sure if it matters.

eddierubeiz: I think that's a good idea.

eddierubeiz:
Also adding some tests for the page itself (heretofore we only had tests for the csv download)

app/views/admin/oral_history_access_requests/index.html.erb
```diff
@@ -55,7 +72,7 @@
           <td class="col-sm-4">
              <%= oh_request.intended_use %>
           </td>
-          <td class="col-sm-1">
+          <td class="col-sm-1 request-list-delivery-status">
```
jrochkind, line 75: You added a class name even though you didn't add any styles targeting it? Doesn't hurt though!

jrochkind (approved):

app/views/admin/oral_history_access_requests/index.html.erb
eddierubeiz: Yeah. I use it in the tests and it looks harmless enough.

PR #2508, opened 2024-01-19
Add pagination to OH requests; remove "three month summary" filter.

eddierubeiz:
Ref #2511.

### DIFF
--- a/app/controllers/admin/oral_history_access_requests_controller.rb
+++ b/app/controllers/admin/oral_history_access_requests_controller.rb
@@ -9,7 +9,7 @@ class Admin::OralHistoryAccessRequestsController < AdminController
     unless status == "any" || status.blank?
       scope = scope.where(delivery_status: status)
     end
-    @oral_history_access_requests = scope.order(created_at: :desc).page(params[:page]).per(100)
+    @oral_history_access_requests = scope.order(created_at: :desc).page(params[:page]).per(300)
 
   end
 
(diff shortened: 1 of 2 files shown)


jrochkind:
When we first implemented, we asked if they wanted pagination, and the staff at the time said no, they would only use this page to look at recent stuff, and if they wanted old stuff they'd use the export. 

What about the new filters seem to you to work best with pagination? I'm not following, sorry!

jrochkind:
I think the use case for the new filters is to be able to focus on the 'pending' ones, and make sure they haven't missed any. 

Ah, if there WAS a "pending" one from earlier than last 3 months, i guess they wouldn't want to miss it!

Is that what you were thinking?

eddierubeiz:
Yeah, that's basically what i was thinking. I might be mistaken, but doing exactly what the issue asks for might lead to some confusion. Let's not worry about it until we merge the other PR - if I am wrong we can ignore this PR.

jrochkind:
Yeah, you're right, it makes sense to me too. just to have to clear it with OH staff to make sure it does to them too. Show it on staging to them? I dunno, I leave to you?

app/controllers/admin/oral_history_access_requests_controller.rb
```diff
@@ -9,7 +9,8 @@ def index
     unless status == "any" || status.blank?
       scope = scope.where(delivery_status: status)
     end
-    @oral_history_access_requests = scope.where('created_at > ?', 3.months.ago).order(created_at: :desc).to_a
+    @oral_history_access_requests = scope.order(created_at: :desc).page(params[:page]).per(100)
```
jrochkind, line 12: We were showing way more than 100 before I think, should we make it 200 or 300 any thoughts?

jrochkind:
omg i love how little code is required. thanks kaminari

Issue #2511, opened 2024-01-23
Remove the "three month summary" filter on the OH page, and add pagination.


eddierubeiz:
If a request is pending for more than 3 months (which is the case for at least one request at the moment, due to requester not answering our follow-up questions), there is currently no way to approve or deny the request.

Also, and more generally,  the "new filter by status" menu is confusing when combined with the 3-month time limit.

I have @sarahschneiderSHI's approval to do this.

jrochkind:
let's add pagination to teh top of the results too, as a standard practice!

PR #2513, opened 2024-01-24
Remove the "three month summary" filter on the OH page, and add pagination (intended for merge into `oh_request_dashboard_holding`).

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2511.
Same code as #2508, but intended for merge into the `oh_request_dashboard_holding` holding branch.

### DIFF
--- a/app/controllers/admin/oral_history_requests_controller.rb
+++ b/app/controllers/admin/oral_history_requests_controller.rb
@@ -9,7 +9,7 @@ class Admin::OralHistoryRequestsController < AdminController
     unless status == "any" || status.blank?
       scope = scope.where(delivery_status: status)
     end
-    @oral_history_requests = scope.where('created_at > ?', 3.months.ago).order(created_at: :desc).to_a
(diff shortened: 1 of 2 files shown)


jrochkind:
Thanks, but we still will need to merge master into the holding branch -- I'm not sure I'm visualizing the workflow where this extra PR helps?  Wait, i guess we merge this in.. .then merge master... then when there's conflict say to just take "our branch" version of everything... not sure if it makes it easier in this case. 

If you have a vision of it, do you want to take care of it?  Otherwise, if doing it myself I'll prob just do the merge without this branch!

eddierubeiz:
Yeah, I can take care of it.

jrochkind:
@eddierubeiz Thanks! Taking care of it means merging `master` into `oh_request_dashboard_holding` and then pushing that to git.  Since you'e already merged this, requiring that next step to be done in a way I'm not totally sure how to do, can you do that today before we forget? Thank you!!

eddierubeiz:
Yup, it's taken care of.

PR #2516, opened 2024-01-25
restore deleted tests; reorganize by method

eddierubeiz:
Restores deleted tests, and reorganizes them by method.


eddierubeiz:
Yeah, as is normal, the more I look at this the more things I want to change about it, but I am going to walk away -- we have other stuff on our plate!

Issue #2476, opened 2023-12-20
Adjust progress options for items in digitization queue


apinkney0696:
Hi! In doing metadata this morning I thought of a small change that might be useful for my predecessor. The digitization queue currently has a drop-down menu listing the following status options:

Awaiting Dig on Cart
Imaging in Progress
Post Production Completed
Batch Metadata Completed
Individual Metadata Completed
Closed (Which hides the queue entry)
Hold
Re Pull Object

From a workflow perspective, I think the following list would be more helpful and accurate:

Awaiting Digitization
Imaging in Progress
Image Export Complete
Metadata In Progress
Closed
Hold

If you'd like me to justify why these changes would be helpful I can, but probably would be easier in person at our next Monday meeting.

jrochkind:
Eliminating "re-pull object" is the only one i'm especially curious about... eliminating that one was intentional? 

I'm a bit nervous about changing these when you won't be here to _use_ them and ensure the changes actually work and ask for more changes if they don't... but if you're sure, I guess?  It's not a difficult change to make. (If any exist using to-be-eliminated statuses, we'd have to map them to one of the new statuses)

jrochkind:
Post Production Completed => Image export complete
Batch Metadata Completed => Metadata in progress
Individual Metadata Completed =>Metadata in progress

PR #2517, opened 2024-01-25
Change list of dig. queue item statuses

eddierubeiz:
Ref #2476
- Changes the status values in the model
- Adds a migration to change the default status (I would love to remove that default status altogether, but I also don't want to modify more code than I have to.)
- Fixes tests
- Adds rake task to do the crosswalk

Note: I didn't bother trying to get the migration to change the actual statuses in the table. None of this code faces the public, and this controller is not in heavy use at the moment. So if a staff member visits that page between deploy and when the rake task is run, they'll see an error.

Steps:
- [ ] Deploy
- [ ] Run rake task

### DIFF
--- a/app/controllers/admin/digitization_queue_items_controller.rb
+++ b/app/controllers/admin/digitization_queue_items_controller.rb
@@ -165,13 +165,6 @@ class Admin::DigitizationQueueItemsController < AdminController
         },
         params.dig(:query, :status) || ""
       )
-
-      # helpers.options_for_select(
-      #   Admin::DigitizationQueueItem::STATUSES.
-      #     find_all {|s| s != "closed" }.
-      #     collect {|s| [s.humanize, s]},
-      #   params.dig(:query, :status)
-      # )
     end
     helper_method :status_filter_options
 
--- a/app/models/admin/digitization_queue_item.rb
+++ b/app/models/admin/digitization_queue_item.rb
@@ -25,8 +25,12 @@ class Admin::DigitizationQueueItem < ApplicationRecord
   validates :collecting_area, inclusion: { in: COLLECTING_AREAS }
 
   STATUSES = %w{
-    awaiting_dig_on_cart imaging_in_process post_production_completed
-    batch_metadata_completed individual_metadata_completed closed hold re_pull_object
+    awaiting_digitization
+    imaging_in_process
+    image_export_completed
+    metadata_in_progress
+    hold
+    closed
    }
   validates :status, inclusion: { in: STATUSES }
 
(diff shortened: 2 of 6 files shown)


jrochkind:
We are first deploying change in allowable statuses, THEN running the data migration rake. So for a brief period there will be things with not allowed statuses in DB. 

What happens to those? Are they still fetchable?  Like, the data migration rake can still work, have you tested in this on staging to ensure this order of operations is okay?  (and if someone does access an admin dashboard in the brief period in which some things may have illegal statuses, what happens?)

eddierubeiz:
You're right - in the interval between deployment and running the rake task, visiting e.g. https://digital.sciencehistory.org/admin/digitization_queue_items/archives will absolutely cause an error. But since a) it's a staff-facing page and b) the interval will only last a minute or two, I'm not worried about it. (See my note above).

jrochkind:
OK, I had a scenario where depending on how 'bad keys' are handled your migration script might actually not work... but rather than go into the details, can you just confirm you've tested this on staging? 

Good with me if you've tested on staging and it worked. 

I still don't really understand all the ins and outs and how we can be sure the bad data will -only- interupt a couple staff pages and not public pages for that interim period, it might be worthwhile to test that on staging too, but your call!

eddierubeiz:
Yup - tested in staging!

jrochkind (approved):

eddierubeiz:
I'd like to deploy / run this soon; if you can approve the PR, I will take care of the rest, @jrochkind.

eddierubeiz:
```
"Changing #782 from post_production_completed to image_export_completed"
"Changing #1118 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1119 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1120 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1121 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1122 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1123 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1124 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1125 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1126 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1127 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1128 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1129 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1130 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1131 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1132 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1133 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1134 from awaiting_dig_on_cart to awaiting_digitization"
"Changing #1135 from post_production_completed to image_export_completed"
```

PR #2522, opened 2024-01-29
Update redis configuration finding to allow stackhero and more flexibility as to where config is found

jrochkind:
Ref #2296 

Temporary testing on staging of this caused some disaster, see https://github.com/sciencehistory/scihist_digicoll/issues/2296#issuecomment-1915149051

- [x] More careful testing, including test failure conditions
- [x] TODO: stop disabling verify if we are not using Heroku redis!

---

- remove old unused persistent_redis_host config
- change how redis URL is found to use REDIS_PROVIDER_ENV redirection, and also default to looking for stackhero provider
- Refuse to use localhost redsi default in production, refuse to boot if no redis

### DIFF
--- a/lib/scihist_digicoll/env.rb
+++ b/lib/scihist_digicoll/env.rb
@@ -222,8 +222,10 @@ module ScihistDigicoll
       end
 
       error_message = "No redis url could be found, cannot provide!\n\n" +
-       "REDIS_PROVIDER_ENV=='#{ENV['REDIS_PROVIDER_ENV']}' ;\n" +
-       "ENV[ ENV['REDIS_PROVIDER_ENV'] ]=='#{ENV[ ENV['REDIS_PROVIDER_ENV'] ]}'\n"
+       "REDIS_PROVIDER_ENV=='#{ENV['REDIS_PROVIDER_ENV']}' ;\n"
+      if ENV['REDIS_PROVIDER_ENV']
+        error_message += "ENV[ ENV['REDIS_PROVIDER_ENV'] ]=='#{ENV[ ENV['REDIS_PROVIDER_ENV'] ]}'\n"
+      end
 
       raise RuntimeError(error_message)
     end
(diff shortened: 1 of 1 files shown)


jrochkind:
We can merge this even if we aren't quite ready for stackhero switch -- the config changes here are good with current setup too, and without ENV change should not change production's current behavior if deployed. 

Once merged, we now have the config to change over to stackhero using ENV settings whenever we want.

jrochkind:
@eddierubeiz oops, I was wrong, there were still a couple things I wanted to improve here, putting back in draft status haha sorry

eddierubeiz:
Yeah, I saw the checkboxes. No worries - just keep me posted.

jrochkind:
OK, we'll disable SSL verification (as heroku redis requires) only when redis url is coming from `REDIS_TLS_URL`  or `REDIS_URL`. 

That is where heroku redis provides it's redis url -- super annoying that this could apply to non-heroku using those same ENV variables (and annoying that heroku uses self-signed certs!!), but better for now than introducing another config ENV var.

jrochkind:
OK, I was trying to raise on boot if redis not found to prevent a deploy -- but then I accidentally found a problem where this created an even worse error, an infinite loop of heroku trying to restart over and over, that could only be fixed by dropping dyno counts to 0!

Keep it simple, we won't raise ourselves, but just log and return `nil`.  

Not sure if  in some cases that might still result in an exception and either infinite heroku restart loop and/or thousands of errors again... but at least let's not make the problem worse ourselves by introducing our own raise known to cause that problem. 

Will continue to do more testing.

jrochkind:
OK, in the curren version if no redis is available, there _will_ still be an infinite loop of frequent errors. 

There seems to be no way to fix this with our configuration alone -- if you give _no_ config to redis/resque, it still defaults to localhost on default port. 

The infinite loop of high-volume errors is caused by some logic in some combo of redis, resque, and resque-pool, that would have to be fixed. Out of scope of this PR. 

Just a good thing to be aware of. 

We can stlll merge this PR, which ought not to change production behavior without config change, but provides for us to make a config change to switch to stackhero.

eddierubeiz (approved):

PR #2524, opened 2024-01-30
Few more things missed in OH class rename

jrochkind:
Including some that created actual errors not caught by existing tests. :( 

- fix bad old method name to renamed route helper admin_oral_history_request_url
- rename association to new name #oral_history_requests
- use correct new param name oral_history_request to preserve existing values on invalid redisplay


jrochkind:
I'm going to take the liberty of self-merging this (just into our holding branch at the moment), trying to get ready for demo tomorrow.

Issue #2494, opened 2024-01-12
Automatically sort CHM and Dist. collections chronologically


apinkney0696:
Due to some donor promotion that's going to happen with these collections, it would be ideal for the Distillations and Chemical Heritage collections chronologically sorted by default, from oldest to newest. 

https://digital.sciencehistory.org/collections/g6go8ar?sort=oldest_date
https://digital.sciencehistory.org/collections/avyv728?sort=oldest_date

The date fields are a little funny for this collection because only the year and season are included in the publication information. This means they appear slightly out of order, even when the "SORT BY OLDEST DATE" setting is applied. I wonder if the technical configuration of these collections should be automatically sorting the titles alphabetically instead. This would accurately put them in chronological order.  

This setting could apply to other collections as well, but I'm not sure which ones at this point. I recommend applying this change to only these two collections for now so that the donor work can move forward swiftly, and then circling back upon my replacement to establish a more thought-out application of this setting to the DC more broadly.

jrochkind:
We have to figure out how to have code figure out _which_ collections to change the sort on, since there's nothing distinguishing these. 

Or I guess we could just hard-code some ID's in, but we don't love that. But if there's a deadline? When you say "due to some donor promotion", is there a particular tight deadline for this request?

We could also just change all collections, or all Library collections, to sort this way. The idea of "recently added" sort was you might be a regular visitor and want to see the new stuff, but that's never really been validated with users, we don't know if it's useful, we could just change all collections to default to publication year oldest first.

jrochkind:
Once @apinkney0696 is sadly no longer working here, who would we check in with about this donor promotion timeline/priority, and also check in with on UX feedback?

I would tag Gabriella, but I cna't find her github account, we should probably add her to the project so it auto-completes if nothing else!

jrochkind:
> This means they appear slightly out of order, even when the "SORT BY OLDEST DATE" setting is applied. I wonder if the technical configuration of these collections should be automatically sorting the titles alphabetically instead. This would accurately put them in chronological order.

I am loathe to add a new "sort by title" feature, because the use cases for it seem so minimal. 

Perhaps we could make sort by publication date sort secondarily by title when publication date is identical, although I don't love it either, maybe it's not too bad.

jrochkind:
ONE way to handle this is we coudl add metadata to the Collection that actually specifies the sort order for that specificc collection -- publication date, order of addition, etc. 

[...]

PR #2542, opened 2024-02-14
Allow per-collection default sort

eddierubeiz:
Ref #2494
Allows you to specify a default sort field. No changes to the admin interface in this PR (that is #2546 ).
TODO:
- [ ] Reindex after deploy

### DIFF
--- a/app/controllers/collection_show_controller.rb
+++ b/app/controllers/collection_show_controller.rb
@@ -69,9 +69,16 @@ class CollectionShowController < CatalogController
 
   private
 
-  # Our custom SearchBuilder needs to know collection id (UUID)
+  # Our custom SearchBuilder needs to know:
+  #   collection id (UUID)
+  #   the default sort order for this collection, if specified.
   def search_service_context
-    super.merge!(collection_id: collection.id)
+    super.merge!(collection_id: collection.id, collection_default_sort_order: collection_default_sort_order)
+  end
+
+  # Some collections define a default sort field. Look up its sort order in blacklight_config and use that.
+  def collection_default_sort_order
+    blacklight_config.sort_fields.dig(collection&.default_sort_field)&.sort
   end
 
   def check_auth
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -124,7 +124,7 @@ class Admin::CollectionsController < AdminController
     # This could be done in a form object or otherwise abstracted, but this is good
     # enough for now.
     def collection_params
-      permitted_attributes = [:title, :description, :department]
+      permitted_attributes = [:title, :description, :department, :default_sort_field]
       permitted_attributes << :published if can?(:publish, @collection || Kithe::Model)
 
       Kithe::Parameters.new(params).
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -400,12 +400,12 @@ class CatalogController < ApplicationController
 
     config.add_sort_field("newest_date") do |field|
       field.label = "newest date"
-      field.sort = "latest_date desc"
+      field.sort = "latest_date desc, title desc"
     end
 
     config.add_sort_field("oldest_date") do |field|
       field.label = "oldest date"
-      field.sort = "earliest_date asc"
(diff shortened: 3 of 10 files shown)


jrochkind:
1. I don't think we need to add new title sorts. I don't see a use case for that.  We may disagree on that, or the spec may have been unclear?  I think for collections we want to allow sort by *publication date*, an already existing sort. 

2. I would like to make sure there's no way to do this with blacklight configuration before adding customizations to the search builder. I feel like there should be.

jrochkind:
There is a blacklight config `default_sort_field` -- I think in the CollectionController we can do an over-ride of the blacklight_config at an instance level (so it's executed per-request, not just at class level fixed), that can look at the current Collection to decide whether/how to override. 

The specs look good, and hopefully can stay the same (or close to it) verifying behavior implemented a different way!

jrochkind:
Oh , if we want to make the current `publication_date` field sort by title secondarily, we may still need to index title. 

But I don't want to add title searches that show up for every search, while we have an edge case where title search kind of "accidentally" does something useful, I think for 99% of cases it does not, and should not be an option cluttering things up. 

Let's talk more?

It is impressive that you did this with so few lines of code, it's not the worst appraoch! But I think using existing blacklight_config will be more future-safe than customizations to search builder.

eddierubeiz:
Consensus:
* sort by publication date, then title;
* no need for the `config.add_sort_field` menu options in `collection_sort_controller.rb`;
* attempt to configure blacklight settings instead, instead of customizing the search builder.

app/controllers/collection_show_controller.rb
```diff
@@ -65,13 +65,14 @@ def facet
     # and we need to make sure collection_id is allowed by BL, don't totally
     # understand this, as of BL 7.25
     config.search_state_fields << :collection_id
+    config.search_state_fields << :collection_default_sort_order
```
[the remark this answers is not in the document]
eddierubeiz: Removed!

app/controllers/collection_show_controller.rb
```diff
@@ -69,9 +69,16 @@ def facet
...
+  #   collection id (UUID)
+  #   the default sort order for this collection, if specified.
   def search_service_context
-    super.merge!(collection_id: collection.id)
+    super.merge!(collection_id: collection.id, collection_default_sort_order: collection_default_sort_order)
+  end
+
+  # Some collections define a default sort field. Look up its sort order in blacklight_config and use that.
+  def collection_default_sort_order
+    blacklight_config.sort_fields.dig(collection&.default_sort_field)&.sort
```
jrochkind, line 81: Can we make sure that if the thing in collection.default_sort_field does not exist at all in the blacklight_config -- this doesn't error, and does give us the same default behavior as if it were blank?  It may already do that! Maybe just add a test?

Other than that, looks great!  Good job figuring out the pretty baroque Blacklight architecture stuff!

eddierubeiz: Oh, the collection model validation rules will actually _prevent_ you from saving a value for `collection.default_sort_field` that's not one of the 2 allowed values, namely:
-  `nil`
- `'oldest_date'`.
If you were to remove the 'oldest_date' sort field from catalog_controller, a whole bunch of tests would fail in addition to this one.
We could make this more sophisticated, of course, but for now crude is good.

jrochkind: I guess those validation rules are maybe in the next PR i haven't looked at yet, not here? But great!

But one of the cases I'm thinking of is where we _change_ the allowed value, and a value _already_ exists in the db that accidentally is no longer allowed. 

Just seems fairly easy and wise to make sure if this happens it's treated as "no value chosen" rather than blowing up!

eddierubeiz: You are correct - the validation rules are *in fact* not in this PR.

jrochkind (approved):

app/controllers/collection_show_controller.rb
eddierubeiz: OK - added the test!

jrochkind:
approved!

eddierubeiz:
I also changed the metadata for the Distillations collection.

Issue #2543, opened 2024-02-15
Email address lookup needs to be case insensitive on the OH sign in page


eddierubeiz:
On the oral_history_session/new page:
Safari changed the user's email address by capitalizing the first letter, upon which the user was told there was no record of a user with that email.

Please make the match case-insensitive.

jrochkind:
REalize this also means we would create duplicate records for varying cases, which I guess we shoudln't. I guess we need to normalize on insert as well as lookup. Perhaps in the DB? Need to do a bit of research on best practices here.

PR #2544, opened 2024-02-15
OH requester email is normalized to downcase and strip whitespace

jrochkind:
Use Rails 7.1 normalization api hooray

Ref #2543


jrochkind:
I just went to console in staging and manually ran:

```
OralHistoryRequester.find_each do |requester|
  requester.email = requester.email.downcase.strip
  requester.save!
end
```

I suppose if anyone is doing more testing with new email addresses between now and when this is merged, it would be possible to create unnormalized ones in the window.

Issue #2545, opened 2024-02-16
Proposed copy changes to the OH interface.


eddierubeiz:
Typo:
`insitute` should read `institute`.

eddierubeiz:
`The files you requested are immediately available, from [...]`
should read
`The files you requested are immediately available from[...]` (no comma)

eddierubeiz:
Run-on sentence:
'Sorry, we have no email on record for #{params[:email]}, check your entry?'
should read e.g.
'Sorry, we have no email on record for #{params[:email]}. Please check your entry.'

PR #2553, opened 2024-02-28
Copy changes to new oh interface

eddierubeiz:
Ref #2545

3 copy changes:

| Before    | After |
| -------- | ------- |
| The files you requested are immediately available, from: `OH_TITLE`  | The files from '`OH_TITLE`' that you requested are immediately available.   |
| Sorry, we have no email on record for `EMAIL`, check your entry? | Sorry, we have no email on record for `EMAIL`. Please check your entry.    |


jrochkind:
Please confirm these have been approved by OH team? As they specified the original copy. Thanks!

eddierubeiz:
Confirmed these copy changes are ready to go. Checked with OH staff.

spec/controllers/oral_history_requests_controller_spec.rb
```diff
@@ -179,7 +179,7 @@
             }
 
             expect(response).to redirect_to(oral_history_requests_path)
-            expect(flash[:success]).to match /The files you requested are immediately available, from: #{Regexp.escape work.title}/
+            expect(flash[:notice]).to match /The files you requested from '#{Regexp.escape work.title}' are immediately available./
```
jrochkind, line 182: OK, the reason I had it the other way first, is the name of the Oral History can be quite long in some cases.  I wanted them to see "are immediately available", which is the actually important content, before getting lost in the really long OH title. 

Your motivation to change it is just to make it a sentance without colon?   I guess if you've considered my point and still prefer it that way, I'll leave it to you, just mark this resolved and go ahead and merge?

eddierubeiz: Yeah, I'm going to argue that the single quotes make it clear that the title is a title, no matter how long. I sincerely believe this is better copy, and I've cleared it with OH staff.

jrochkind: OK, thanks!

It's not that I thought it wasn't clear it was a title, it's that I was worried the reader skimming might never make it to the most important words of the sentence, the whole point of the alert, telling the user that their files are immediately available!

Issue #2519, opened 2024-01-26
`can't find executable rspec` error, even after setting BUNDLE_WITHOUT="" and waiting.


eddierubeiz:
IMPORTANT NOTE: 5 days later, on Tuesday Jan. 30th, I can't reproduce the problem.
Could I have set `BUNDLE_WITHOUT="" ` on staging instead of production? Not impossible, since I was dealing with a lot of stuff on Friday.

Original notes re: this problem:
Just adding this for later investigation. I'm following the instructions athttps://github.com/sciencehistory/scihist_digicoll/tree/master/system_env_spec .
`heroku config:set BUNDLE_WITHOUT="" --app scihist-digicoll-production`

[wait half an hour or so]

Make sure the var is set:
```
$ heroku config --app scihist-digicoll-production | grep WITHOUT
BUNDLE_WITHOUT:
```

We still get:
```
$ heroku run "rspec system_env_spec" --app scihist-digicoll-production
Running rspec system_env_spec on ⬢ scihist-digicoll-production... up, run.6590 (Standard-1X)
/app/vendor/ruby-3.2.2/lib/ruby/3.2.0/bundler/rubygems_integration.rb:308:in `block in replace_bin_path': can't find executable rspec for gem rspec-core. rspec-core is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)
	from /app/vendor/ruby-3.2.2/lib/ruby/3.2.0/bundler/rubygems_integration.rb:347:in `block in replace_bin_path'
	from /app/bin/rspec:29:in `<main>'
```

jrochkind:
I think you LATER wrote:

> I ran all the "is present" tests in https://github.com/sciencehistory/scihist_digicoll/blob/master/system_env_spec/system_env_spec.rb and the only utility we're missing is exiftool.

https://github.com/sciencehistory/scihist_digicoll/issues/2518#issuecomment-1912280897

How'd you manage to do that without this working?

eddierubeiz:
Oh, I just ran the commands from a bash shell.

jrochkind:
> Oh, I just ran the commands from a bash shell.

I'm still confused how THAT could work if the BUNDLE_WITHOUT didn't!  

When we get to this, lets pair on it initially.

eddierubeiz:
I want to test this one more time...

eddierubeiz:
As of today I can't reproduce this in production.

jrochkind:
[...]

PR #2554, opened 2024-02-28
Edit to system_env_spec/README.md

eddierubeiz:
Practically speaking we only use this script during an emergency, but for that same reason, the docs should be as clear as possible.
Ref #2519.


jrochkind:
thank you!

Issue #2498, opened 2024-01-17
Heroku: use nodejs buildpack to specify version of node and yarn


jrochkind:
Similar to what was going on with python in #2436, we're letting heroku pick the version of `node` and `yarn` installed, which some months ago heroku changed to deprecated behavior. 

They want us to use the nodejs buidpack and specify the exact version of node and yarn we want. In this case we SHOULD be able to specify a range of versions, like "latest 10.1.x", which is probably what we want to do. 

But it still does mean we'll have to periodically manually update these version numbers, when we are using a too-old (EOL etc) version. 

But this is what heroku wants, we should use heroku recommended setup. 

We will want to pay attention to how it effects our heroku "slug size".  If it increases it too much, it could be a problem. Heroku max slug size is 500MB and it starts warning you at some number under that. (350?).  Bigger slug size means slower deploy too. We want to avoid even the warning. As of this writing, our slug size is.... 

```
remote: -----> Compressing...
remote:        Done: 256.3M
```

(most of our slug size is node packages! That may not be needed at runtime, just at asset compile time. We could find a way to trim them. Anyway, a separate issue!)

Currently we get these deprecations on deploy.

```
remote: ###### WARNING:
remote:
remote:        Installing a default version (1.22.19) of Yarn
remote:        This version is not pinned and can change over time, causing unexpected failures.
remote:
remote:        Heroku recommends placing the `heroku/nodejs` buildpack in front of the `heroku/ruby`
remote:        buildpack as it offers more comprehensive Node.js support, including the ability to
remote:        customise the Node.js version:
remote:
remote:        https://devcenter.heroku.com/articles/ruby-support#node-js-support
remote:
remote:
remote: ###### WARNING:
remote:
remote:        Installing a default version (20.9.0) of Node.js.
remote:        This version is not pinned and can change over time, causing unexpected failures.
remote:
remote:        Heroku recommends placing the `heroku/nodejs` buildpack in front of
remote:        `heroku/ruby` to install a specific version of node:
remote:
remote:        https://devcenter.heroku.com/articles/ruby-support#node-js-support
remote:
```

jrochkind:
@eddierubeiz this is another straightforward fairly quick one that needs doing, if you wind up with extra time.

PR #2557, opened 2024-03-04
Add yarn and node to package.json

eddierubeiz:
This assumes you've added the `nodejs` buildpack like this:
```
heroku buildpacks:add heroku/nodejs -i 1
```

Ref #2498
- Tested in staging
- Edited confluence wiki documentation


jrochkind:
We have docs on heroku buildpacks  on confluence, please make sure to edit accordiongly!

eddierubeiz:
This raises the slug size from 263 to 270MB. That's fine.

.github/workflows/ci.yml
```diff
@@ -49,11 +49,16 @@ jobs:
             sudo apt-get update
             sudo apt-get -y install libvips-tools ffmpeg mediainfo tesseract-ocr qpdf imagemagick exiftool
 
+        # Note: as of early 2024, 'yarn install' causes GitHub to install node 18.19.1.
+        # Whatever version is installed, make sure it is permitted in file
+        # package.json under key "engines".
```
jrochkind, line 54: Sorry, a nit-pick, but it's not that "yarn install" causes github to install node, I don't think, but that Github Actions VM's have a pre-installed node, that `yarn install` _uses_, that has to match requirements.

package.json
```diff
@@ -29,5 +29,11 @@
   },
   "browserslist": [
     "defaults"
-  ]
-}
+  ],
+  "science_history_institute_comment_about_engines" :
+    "We permit a range of node versions here:18.19 is GitHub CI. 20.9  is for everything else.",
+  "engines": {
+    "node": ">=18.19.0 <20.10.0",
```
jrochkind, line 36: I wonder if it's safe and reliable to go all the way up to any `20.x`, ie less than 21?  To keep getting bugfixes and improvements without backwards breaking changes that will come in next major version? Assuming node is pretty safe on backwards breaking changes not happening unless major version?   It looks to me like the minor versions come pretty often, so I think maybe we do want this, not to require us to manually opt in to 20.11 20.12 etc, what do you think?

eddierubeiz: Yeah, that's exactly my thinking. Based on your description of how we're using Node, I think we can rely on node working reliably for us, for all versions lower than `20.10`.

package.json
```diff
@@ -29,5 +29,11 @@
   },
   "browserslist": [
     "defaults"
-  ]
-}
+  ],
+  "science_history_institute_comment_about_engines" :
+    "We permit a range of node versions here:18.19 is GitHub CI. 20.9  is for everything else.",
```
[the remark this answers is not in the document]
eddierubeiz: Sounds good.

package.json
jrochkind: Why not 20.10 and 20.11 (which are already out), and future versions less than 21?

Let's look at recent releaes history:

20.11.0: Jan 10, 2024
20.10.0:  Nov 22, 2023
20.9.0: Oct 24, 2023
20.8.0: 2023-09-28
20.7.0: 2023-09-18

It looks to me like the second digit (10.9, 10.10., 10.11) releases come out roughly monthly. I am suggesting maybe we we want to allow the app to use the new 20.x version (and have heroku install it) as available automatically, instead of having to change the package.json monthly to allow the new minor release?

So we can get these monthly minor releases with security fixes and improvemetns on heroku automatically as they come out? 

But require a manual change to package.json to upgrade to a `21` in the future, as it may have backwards-breaking changes. (Also 20.x is a "Long Term Support" release and 21 might not be, so we might not want it anyway). 

Right now, you are freezing it at 20.9, even though 20.10 and 20.11 are already out.

eddierubeiz: Sorry, I misunderstood your comments. No problem at all - we now permit everything less than 21.0.0 .

jrochkind:
@eddierubeiz While I did merge... I'm still getting the heroku warnings on pushing master once merged!

```
remote: ###### WARNING:
remote:
remote:        Installing a default version (20.9.0) of Node.js.
remote:        This version is not pinned and can change over time, causing unexpected failures.
remote:
remote:        Heroku recommends placing the `heroku/nodejs` buildpack in front of
remote:        `heroku/ruby` to install a specific version of node:
remote:
remote:        https://devcenter.heroku.com/articles/ruby-support#node-js-support
remote:
remote:
remote: ###### WARNING:
remote:
remote:        Installing a default version (1.22.19) of Yarn
remote:        This version is not pinned and can change over time, causing unexpected failures.
remote:
remote:        Heroku recommends placing the `heroku/nodejs` buildpack in front of the `heroku/ruby`
remote:        buildpack as it offers more comprehensive Node.js support, including the ability to
remote:        customise the Node.js version:
remote:
remote:        https://devcenter.heroku.com/articles/ruby-support#node-js-support
```

Your comments to me in slack suggested you saw something different on deploying this branch!

```
remote:        Resolving node version 20.x...
remote:        Downloading and installing node 20.11.1...
remote:        Using default npm version: 10.2.4
remote:        Resolving yarn version 1.22.x...
remote:        Downloading and installing yarn (1.22.21)
remote:        Installed yarn 1.22.21
```

I think more investigation is required before we close the related issue?

Also....   I realized I missed this in my review, but I'm still not sure what yarn `"yarn": "~1.22.19"` does in npm -- I think we want `"yarn": "^ 1.22.19"` instead?  Can you confirm what the `~` does in NPM version restrictions? 

I believe the carot in `^ 1.22.19` would mean any `1.x` greater than that, which similar to discussion of node, is I think what we want? Googling looks like `~ 1.22.19` would lock us to `1.22.x`, so similar discussion as before with npm. (Note that `~>` in ruby is different!!)

eddierubeiz:
The warning you mentioned is gone - it's just because the deploy happened before we added the buildpack. The buildpack is now installed in production.

eddierubeiz:
I'm moving our discussion about the correct range for `yarn` into issue https://github.com/sciencehistory/scihist_digicoll/issues/2565

jrochkind:
@eddierubeiz Have you/can you document the new buildpack in relevant Confluence page on heroku buildpacks please?

PR #2560, opened 2024-03-05
    When migrating to new OH request schema, erase email addr in old location

jrochkind:
We had tried to leave it in both for safety, but our validation rules don't actually allow that! Which is good, our validation rules make sure it's in the right place.

- handle normalization when ensuring we migrated properly


jrochkind:
* Tested on staging
* Verified that all OralHistoryRequest#patron_email are now nil!
* Verified that all OralHistoryRequest#oral_history_requester are non-nil
* Verified that all OralHistoryRequester#email are non-nill
* Email addresses still show up on OH request dashboard in staging
* Downloaded report of all requests in staging...  email addresses all present and look right! (they aren't all the same email addr or anything)

PR #2562, opened 2024-03-08
Remove workaround from Rails 7 migration

eddierubeiz:
Ref #2563

Rails 7 stopped automatically running `yarn install` at app startup, which led to the problem discussed [here](https://github.com/ElMassimo/vite_ruby/discussions/316).
@jrochkind  introduced a [workaround](https://github.com/sciencehistory/scihist_digicoll/commit/6c5672ce53f1dc1c4c4d38496460b55d1dfb5a78) for this problem.

I was [messing around with vite and node](https://github.com/sciencehistory/scihist_digicoll/issues/2498) this week and I believe this workaround is no longer needed, so I'd like to remove it.

With this PR applied, Vite still works fine in dev, and assets still compile on Heroku.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,8 +142,6 @@ We preferentially use Vite.js (an ES6-style JS bundler, https://vite-ruby.netlif
   * image is in `./app/frontend/images/*`
   * reference using standard CSS `url(@/images/*)`, vite will properly serve/build and translate url. Note the `@/` at the beginning that vite will use to look it up in vite entrypoint directories, which are configured by vite-ruby to include ./app/frontend/images.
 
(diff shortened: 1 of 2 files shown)


jrochkind:
Thanks for the archeological spelunking!  You've confirmed you can still deploy to heroku (say staging) on this PR?

jrochkind:
i guess this may expalin that doubling of deprecation messages we were getting from heroku! And getting rid of it may speed up our deploys, avoiding a double install!

eddierubeiz:
Yes, I've deployed this to staging and everything works fine.

jrochkind:
thank you for noticing and fixing this!

Issue #2352, opened 2023-09-06
Handle null byte injections [not doing this]


eddierubeiz:
Four examples below, courtesy of HB. Will link to original HB errors later.

```
/downloads/orig/image/1yrphmgdpgulaszriylqiipemefmacafkxycjaxjs%00.jpg
```

```
/rights/InC-RUU/1yrphmgdpgulaszriylqiipemefmacafkxycjaxjs%00.jpgkjizcx
```

```
/collections/1yrphmgdpgulaszriylqiipemefmacafkxycjaxjs%00.jpg
```

```
/works/1yrphmgdpgulaszriylqiipemefmacafkxycjaxjs%00.jpg
```

eddierubeiz:
To summarize our decision this morning:

We have a working fix at https://github.com/sciencehistory/scihist_digicoll/pull/2570/files, but:
1) we don't like the fix (we think it's overkill)
2) this problem has only occurred once and we'd rather leave this broken than implement the fix.
I'm moving this to the review column; after we discuss next week we should close this issue.

PR #2567, opened 2024-03-12
Redirect requests with a null byte to /422 error page

eddierubeiz:
Ref #2352
This feels clumsy, but I'm not really sure what else to do.
Adding middleware to catch this problem lower in the stack seems like overkill at first glance.
Other ideas welcome!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,18 @@ class ApplicationController < ActionController::Base
   end
 
 
+  # Bots occasionally feed us URL params containing null bytes.
+  # Since normal use by humans never leads to this error,
+  # there's no point in throwing a distracting 500 error.
+  rescue_from "ArgumentError" do |exception|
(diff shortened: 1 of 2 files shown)


jrochkind:
Hm, yeah, I don't love it either, seems like a pretty heavy hammer for such an edge case problem. 

I see looking at the examples that it really DOES apply to like every controller possible, not a particular specific controller!

This is really a Rails problem.  I think you probably HAVE found the only way to do it though -- but I wonder if other people have run into this and what they've found...

We didn't capture the actual rails error message in the ticket, do you have it now from your reproduction to store for posterity... and to use to google to see if other people have been annoyed by this in Rails and solved it differently?

Alternately, i wonder if we just have HoneyBadger ignore it/not alert us to it; I think we can configure HB to ignore certain things, I wonder if we can get it to ignore very specifically this thing, or at least not alert us about it. What do you think?

PR #2570, opened 2024-03-13
Don't send  null byte errors to honeybadger

eddierubeiz:
#2352
This works great. We can and should reuse this pattern for similar errors:
- not affecting any particular controller
- not caused by human users
- thus not visible to human users

### DIFF
--- a/config/initializers/honeybadger.rb
+++ b/config/initializers/honeybadger.rb
@@ -23,7 +23,11 @@ Honeybadger.configure do |config|
   }.delete_if { |_k, v| v.blank? }
 
   config.before_notify do |notice|
-    notice.halt! if notice.error_message =~ /string contains null byte/
+    
+    # see https://github.com/sciencehistory/scihist_digicoll/issues/2352
+    if notice.error_message =~ /string contains null byte/ && notice.backtrace[0] =~ /postgresql_adapter\.rb/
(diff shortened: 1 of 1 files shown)


jrochkind:
Hm, what I worry about here is whether ANY error message with this string in it is really one we want to totally ignore and have be basically invisible. 

Would it make sense to try to scope it by stack trace to ensure it's really this one we've identified as ignorable?

Also, I wonder if there's any way to have honey badger _record_ it, but not _alert_ us. So it's there if we go looking for it at least.

eddierubeiz:
In theory you should be able to add
a search phrase of the form `-message:"string contains null byte"` to the [settings for the HoneyBadger integration with Slack](https://app.honeybadger.io/projects/58989/channels/1121712/edit).
I wasn't able to get that to work in Staging.

eddierubeiz:
I'm also checking the stack trace now, per your idea of being a bit more specific.

jrochkind (approved):

jrochkind:
OK still makes me nervous (that we might end up totally ignoring an error that DOES matter, and isn't just caused by a bot constructing bad URLs), but I'll approve it!

eddierubeiz:
We decided it's not worth fixing this - the fix is worse than the problem in this particular case.

Issue #2463, opened 2023-12-06
Search within work: evaluate Internet Archive BookReader


eddierubeiz:
We now have a fair number of works with full text. It's time to turn our attention to getting search results bringing the user to a specific page within the work (instead of to the work page.)

Use case: you search for the word "owl", and for each matching work, you get [a set of links to each matching page](https://archive.org/details/birdbookillustra00reedrich/page/225/mode/1up?view=theater&q=owl), instead of a [single link to the work](https://digital.sciencehistory.org/catalog?search_field=all_fields&q=owl).

Based on the [discussion in the wiki](https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/2206203905/OCR+planning+notes#Search-within--Work), evaluate the [Internet Archive BookReader](https://openlibrary.org/dev/docs/bookreader), as a way to provide this functionality.

For now, this will not be part of the digital collections code, but a proof of concept that will stand alone, using a sample of a few pages from the digital collections.

eddierubeiz:
I'm wrapping up work on this. My report is at:
https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/2285371393/Internet+Archive+BookReader

jrochkind:
For the record, some emails from Drini, BookReader developer. 

-----

Hi Jonathan and Eddie,

I'm glad you like the UX! And fantastic work on your archive, that seems
like a great initiative -- and very well designed, too. It seems like a
good use case for BookReader. We are actively maintaining BookReader for
use on [archive.org](http://archive.org/) , so it will be getting regularly updated. But there
might be some bumps when trying to integrate it into third party sites
since we don't do that very often, but I'm happy to help iron out any
issues you discover.

You are correct, the IIIF demo was _very_ out-of-date. I spent some time
and just opened a PR to fix it up. If you spin up a IIIF server, you can
paste a manifest URL here to try it out:
https://deploy-preview-1312--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-iiif
. IIIF is convenient since it's a standard that works with a number of
different viewers, and it's used by a _lot_ of archives and libraries,
so you'd be in good company!

For zoom levels, yep it does support it! It needs the width/height of
each page, and then you can specify the `getPageURI` option to be a
custom function. This function is given the page index and the reduce
factor -- i.e. how much do shrink the image. For Internet Archive, we do
something like:

```js
[...]

PR #2571, opened 2024-03-14
store selected exiftool characterization in normalized metadata location

jrochkind:
We store complete exiftool characterization result in native json format, so we can always refer to it later, if we want more from it or there were bugs in using it. This is good archival practice.

But future exiftool versions may change their results. For anything we want to use computationally, we also want to extract and store in a standardized normalized metadata location. For now, we defintely need dpi for giving to InternetArchive BookReader ref #2463. 

For now only that!

With rake task to extract from existing objects -- quick enough that no bg jobs are needed, just do it foreground. 

**Don't merge until you are ready to deploy and:**

- [ ] `heroku run rake scihist:data_fixes:extract_normalized_exiftool`

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -359,9 +359,29 @@ class Asset < Kithe::Asset
     Rails.logger.info("Asset Destroyed: #{info_str}")
   end
 
+  # Run exiftool and write results to attributes in memory, does not save record
   def store_exiftool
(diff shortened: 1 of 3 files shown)


jrochkind:
Tested on staging

eddierubeiz (approved):
Looks good!

PR #2575, opened 2024-03-14
WIP Book Reader initial integration prototype

jrochkind:
Ref #2463


jrochkind:
We decided not to proceed with IA BookReader for now. Instead, going with #2587

PR #2578, opened 2024-03-18
Substitute zero if delivery_status_changed_at is nil

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2577
Fixes the problem in staging. Just want to fix the bug; no tests for this.


eddierubeiz:
I'm merging this myself.

jrochkind:
thanks!

Issue #2568, opened 2024-03-12
Oral history request backend cleanup: remove unused data column


jrochkind:
After #2840

* Remove unused `patron_email` column from OralHistoryRequest. 
* Remove validation related to it
* make validation _require_  an associated OralHistoryRequester record
* Optionally if feasible, in DB make the foreign key for the associated OralHistoryRequester also required

This can be done independently of the feature flag for new request style -- the new data layout is now used either way! So we do not need to wait until we are sure we won't be changing back, it can be done starting now.

PR #2581, opened 2024-03-22
Remove unused patron email column

eddierubeiz:
Ref #2568 
- [x] Remove unused patron_email column from OralHistoryRequest.
- [x] Remove validation related to it
- [x] Make validation require an associated OralHistoryRequester record
- [x] in DB make the foreign key for the associated OralHistoryRequester also required

### DIFF
--- a/spec/controllers/oral_history_requests_controller_spec.rb
+++ b/spec/controllers/oral_history_requests_controller_spec.rb
@@ -143,40 +143,6 @@ describe OralHistoryRequestsController, type: :controller do
       }
     end
 
-    describe "with old email functionality" do
-      describe "automatic delivery" do
-        let(:work) { create(:oral_history_work, :available_by_request, available_by_request_mode: :automatic)}
-
-        it "emails files" do
-          expect {
-            post :create, params: full_create_params
-          }.to have_enqueued_job(ActionMailer::MailDeliveryJob).with { |class_name, action|
-              expect(class_name).to eq "OralHistoryDeliveryMailer"
-              expect(action).to eq "oral_history_delivery_email"
-          }
-
-          expect(response).to redirect_to(work_path(work.friendlier_id))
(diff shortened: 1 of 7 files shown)


eddierubeiz:
Note: I'm keeping `OralHistoryRequest#requester_email` for now, just for the sake of keeping the PR small. It's not doing any harm.

spec/controllers/oral_history_requests_controller_spec.rb
```diff
@@ -143,40 +143,6 @@
...
-              expect(class_name).to eq "OralHistoryRequestNotificationMailer"
-              expect(action).to eq "notification_email"
-          }
-
-          expect(response).to redirect_to(work_path(work.friendlier_id))
-          expect(flash[:success]).to match /Your request will be reviewed/
-        end
-      end
-    end
-
```
jrochkind, line 179: So if I remember right, the old flag functionality SHOULD still work even without this column.  Removing the column ought not to remove ability to switch the feature flag back and forth at this point. So ideally these tests might remain?

Did these tests fail? Is it possibly just a test setup problem (trying to set it up with the email in the old column), and if you just set up the data in the new way the tests as written will still pass?

eddierubeiz: I am not sure why these got mistakenly removed. Glad you caught this!

jrochkind:
The other thing we _could_ do is change the database schema to make column correspoinding to `OralHistoryRequest.oral_history_requester_id` required non-nillable.  

What do you think, is it worth it? Your call, other than that approved!

eddierubeiz:
That migration is in here.
`change_column_null :oral_history_access_requests, :oral_history_requester_email_id, false`.

jrochkind:
Oh nice, thank you!

Issue #2569, opened 2024-03-12
oral history request backend: remove old request UX


jrochkind:
After #2840

When we are sure we won't be going back, remove logic for OLD way of doing requests, remove `FEATURE_NEW_OH_REQUEST_EMAILS` feature flag. 

We could change the default for FEATURE_NEW_OH_REQUEST_EMAILS to true earlier, I suppose.

PR #2584, opened 2024-04-02
Remove old oral history request code

eddierubeiz:
We have @Gabz73 and @sarahschneiderSHI 's blessing to remove the old code.
Merging this PR means there's no going back to the old system.
Ref #2569

### DIFF
--- a/spec/controllers/admin/oral_history_requests_controller_spec.rb
+++ b/spec/controllers/admin/oral_history_requests_controller_spec.rb
@@ -67,6 +67,10 @@ RSpec.describe Admin::OralHistoryRequestsController, :logged_in_user, type: :con
     describe "#respond", queue_adapter: :inline do
      let(:message) { "custom message from staff" }
      let(:oral_history_access_request) { create(:oral_history_request, delivery_status: "pending") }
+      before do
+        allow(ScihistDigicoll::Env).to receive(:lookup).and_call_original
+        allow(ScihistDigicoll::Env).to receive(:lookup).with("feature_new_oh_request_emails").and_return(true)
+      end
 
       it "can approve" do
         post :respond, params: {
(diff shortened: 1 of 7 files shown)


spec/controllers/admin/oral_history_requests_controller_spec.rb
```diff
@@ -104,56 +107,11 @@
         expect(oral_history_access_request.delivery_status).to eq "rejected"
 
         last_email = ActionMailer::Base.deliveries.last
-        expect(last_email.subject).to eq "Science History Institute: Your request"
+        expect(last_email.subject).to eq "Science History Institute Oral History Request: #{oral_history_access_request.work.title}"
         expect(last_email.from).to eq [ScihistDigicoll::Env.lookup(:oral_history_email_address)]
+        expect(last_email.body).to match /Unfortunately we could <b>not<\/b> approve your request/
         expect(last_email.body).to include(message)
       end
-
```
jrochkind, line 111: It looks to me like maybe you accidentally removed some useful specs here for new (now only) func, rather than just moved them around? Can you confirm?  Or I can checkout the code and spend more time with it confirmiming myself!

I am just suspicious of how many lines are removed in the diff, without corresponding lines being added. I added lots of new tests for the new func that we DO want to preserve, although we don't need to label it as "new functionality" anymore of course!

eddierubeiz: Will absolutely look into this - thanks for checking.

jrochkind: I looked it over, it's actually all good!

jrochkind (approved):

jrochkind:
nice work!

eddierubeiz:
Reopening until we actually remove FEATURE_NEW_OH_REQUEST_EMAILS in production and staging.

Issue #2587, opened 2024-04-04
search within book for OCR text: add feature to existing custom Viewer


jrochkind:
See https://sciencehistory.atlassian.net/l/cp/p3BPKJBU

Related to #2463

PR #2588, opened 2024-04-04
search within book for OCR text: add feature to existing custom Viewer

jrochkind:
ref #2587

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -38,12 +38,14 @@ class WorksController < ApplicationController
 
   def viewer_search
     if params[:q].blank?
-      render status: 422, plain: "Error status 422: Missing required `q` parameter."
-    else
-      render json: HocrSearcher.new(@work, query: params[:q],
-          show_unpublished: can?(:read, Kithe::Model)
-      ).results_for_osd_viewer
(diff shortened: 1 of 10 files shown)


jrochkind:
@eddierubeiz While there are fairly good tests for the search function back-end, there are no tests included for the actual JS feature. We don't currently have any tests for the JS viewer, so I'm not crossing that bridge now -- but what do you think? We haven't really figured out great approaches to testing complex JS like this, but it could be done if we want to go down that path. Maybe a separate issue for us to file (and prob ignore to be honest!)

Otherwise, this is ready for review!  It's behind a feature flag, so we an merge it in now and keep this PR from becoming even more monstrous and hard to review -- and later decide when we want to actually reveal the feature to users.

eddierubeiz:
The more complicated our custom viewer becomes, and the more time the digital collections is likely to be an active website, the more useful a set of JS tests is likely to be. I feel like we are getting to the level of complexity and maturity where it's worth giving it some thought. But I'd like to frame it as a separate, long-term project involving a big chunk of initial research. 

In other words, I still think we should try and get the viable project out the door, if possible, within a few weeks -- and that should not include any front-end tests.

So yeah,  I think we put a separate issue in the "someday-maybe" column for now.

eddierubeiz (approved):

PR #2594, opened 2024-04-16
Viewer search area can be shown/hidden by user

jrochkind:
#2592 must be merged first, this is on top of that. #2592 must be merged first, this is on top of that.


jrochkind:
Note, I researched accessibility a bit and tried to do some things to improve accessibility. 

* Make sure when the drawer is hidden, it has CSS `visible: false` instead of just being off-screen, where it would still be treated as present by screen-readers etc

* Make sure after closing and opening drawer, "focus" is restored to a reasonable convenient element. 


This was a useful resource: https://knowbility.org/blog/2020/accessible-slide-menus

eddierubeiz:
Resolved trivial conflict; will merge after that.

PR #2607, opened 2024-04-29
Viewer, next/prevous through search results

jrochkind:
Includes some refactors of data structures in the JS, because we need to keep more data on hand now, such as what page-index each result corresponds to, so we can find "next result after current page being viewed". 

* When you have a "current selected result", next/prev takes you to next or previous
* When you manually change pages, it removes teh "current selected result". 
* If you have no "currently selected result", next/prev finds the next/prev search result starting from the page you are currently looking at

-----

- refactor viewer search results to separate object
- refactor viewer search results to use js data structures not store in DOM
- viewer search results, most basic next/previous func
- fix finding first search result
- viewer hide searchNav on no results


jrochkind:
@eddierubeiz This is ready for review, although we may not want ot merge it until I have improvements to highlighting done too. But I'm going to do them in other PRs -- this one is already hard enough to review, sorry!

app/frontend/javascript/scihist_viewer/viewer_page_info.js
```diff
@@ -0,0 +1,73 @@
...
+      idToIndex[value.memberId] = index;
+    });
+    self._idToIndex = idToIndex;
+  }
+
+  // @param index [Integer] 0-based index into ordered list, return
+  // the object with page info for that index. Object is as returned by server.
+  //
+  // it's a kind of complicated object not entirely documented sorry, but looks like eg:
+  //
```
eddierubeiz, line 23: (It does make sense though.)

eddierubeiz (approved):
Went through the code, tested in dev (and it's been in staging, too, for a bit). I approve this PR.

PR #2609, opened 2024-05-01
aggressively rate limit with rack attack to 30 requests per minute per IP

jrochkind:
one every other second, averaged over a minute. A human seems unlikely to get that? Googlebot might and we don’t want to harm our SEO, but what can you do.

Trying some things to deal with the apparent bot traffic we are getting which is running up our S3 costs -- although this only applies to our app not S3, the bots are getting the S3 urls from our app, and this is an easy thing we can try in an urgent situation.


jrochkind:
@eddierubeiz self-merging

PR #2611, opened 2024-05-01
Viewer, highlight current search result in results list

jrochkind:
And scroll it into view if it wasn't already. 

Also change method we use to scroll things into view -- to scroll to CENTER of list, instead of top or bottom. We can do this with new browser API that wasn't widely avail when we first wrote it. Simplifies need to track if we're going next or prev for top/bottom, also I think center is prob better UI anyhow. 

- viewer, highlight currently selected search result in results list
- viewer, factor out generic scrollElementIntoView for re-use
- viewer, scroll selected search result item into view
- viewer scroll into viewer always scrolls so element is centered now

### DIFF
--- a/app/frontend/javascript/scihist_viewer.js
+++ b/app/frontend/javascript/scihist_viewer.js
@@ -136,15 +136,19 @@ ScihistImageViewer.prototype.show = function(id) {
   });
 };
 
+// Scroll given element into view only if it's not already in full view. Unlike built-into
+// browser function which always scrolls so it's at the top, even if it already was in view.
+//
 // position can be 'start', 'end'
-ScihistImageViewer.prototype.scrollSelectedIntoView = function(position) {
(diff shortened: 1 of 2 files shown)


jrochkind:
So there are times when the UI scrolls a list to a new point based -- scrolling the thumbnail to the current one, and now scrolling the search result list. 

There is a newer browser feature wehre you can have the browser do a "smooth" visible scroll instead of just abruptly jumping -- we hadn't been using it before, since I think it wasn't universallly available when code was first written! It looks *really nice* and makes the UX very pleasant. 

BUT, now that we're doing it in TWO palces, sometimes simultaneously (pressing the search result 'next' button might require a scroll to a new thumbnail AND a new search result that were both off screen) -- it turns out there is a bug in Chrome where it can't do them both simultaneously in a "smooth" fashion, and doesn't scroll one  of them at all!

There are ways to write more manual code to work around this, which I would like to do -- but not in this PR, it turns out, I won't hold this PR for it!  But maybe in this general project before we leave it.

PR #2612, opened 2024-05-02
DISABLE_DOWNLOADS flag

jrochkind:
Meant to temporarily halt the S3 asset accesses by bots that are costing us lots of money, until we can figure out a better solution.

Initial implementation removes ALL *image* download options (derivative and original) from download menu.

It does NOT actually disable the download controller, if someone already has a URL.

It also does not disable PDF or Audio downloads from OH-specific pages. (That is somewhat more confusing to do, since the links are kind of integrated all over the place).

This can be adjusted in either direction based on whether it works to stop our problematic access, and whehter it interferes with staff work too much, or what.

- [ ] heroku `heroku config:set DISABLE_DOWNLOADS=true -r production`


jrochkind:
@eddierubeiz self-merged

Issue #2561, opened 2024-03-07
Single DC queue


Gabz73:
This is open to discussion, and you may have already discussed this in the past and decided against it:  is it possible to have a view of teh DC queue that isn't sorted into Collecting Areas?

eddierubeiz:
Yeah, absolutely. We can discuss on Monday.

jrochkind:
We have never discussed it before, that was not what previous iterations of staff wanted!

It should be possible, if we allocate some time to do it! 

Would you also still want it separated by collection areas? I'm wondering if your current workflow just doesn't really separate by collection areas anymore, like it was designed for.

Gabz73:
It is helpful to see it by collecting area as well, so Jahna would like to keep that too.

__________________________________

Gabriela Zoller   (she, her, hers)
Director of Library Services
Othmer Library of Chemical History
t. +1.215.873.8289

Science History Institute
Chemistry • Engineering • Life Sciences
315 Chestnut Street • Philadelphia, PA 19106 • U.S.A.
We tell the stories behind the science.
sciencehistory.org<https://www.sciencehistory.org>

PR #2616, opened 2024-05-16
Simplify digitization request list

eddierubeiz:
Ref #2561.
User interface works for @Gabz73 and @JAuerbach26.
I added extra controller tests, too.


jrochkind:
oh wow this is a lot of files changed. I'm going to actually check it out to play with it to review

jrochkind:
When we last talked to @Gabz73, my understanding was they still wanted the current view by collecting area but ALSO wanted a view that listed everything regardless of collecting area.  As recorded in [this comment](https://github.com/sciencehistory/scihist_digicoll/issues/2561#issuecomment-1984568882)

So I had imagined we'd still have the home screen by a list of collecting areas that you could click through, to support that previous workflow where staff was working focused on a specific collecting area. 

But in this PR, that has been removed, and you only get one "main" list of queue items. You can choose from a popup menu to restrict by collecting area, a little bit different UI. 

@Gabz73 can you confirm this UI is what you wanted, you wanted to get rid of that list by collecting area as initial screen, it's no longer best for current workflow?

When we **are** looking at the list by all collecting areas -- the initial list screen does _not_ have "collecting area" on it, you don't know what collecting area an item is in unless you click through. Is this okay?

This is more change to UI than I had anticipated, just want to make sure these changes were intentional responses to new workflow needs?  Do we want to like user test it with digitization staff or anything?

Also, I note that there's still a "Rights & Reproduction" checkbox in digitization item metadata -- I no longer remmeber what this does or how these even show up -- this is still used, and has been tested to ensure it still works as desired with these changes? @eddierubeiz @Gabz73

app/models/admin/digitization_queue_item.rb
```diff
@@ -23,6 +23,7 @@ def validate_each(record, attribute, value)
   COLLECTING_AREAS = %w{archives rare_books modern_library museum}
 
   validates :collecting_area, inclusion: { in: COLLECTING_AREAS }
+  validates :collecting_area, presence: true
```
jrochkind, line 26: why this change, no longer making sure it is in the allow list of allowed collections?

Gabz73:
I may not have expressed it well in the original request, but this is what we want:  you can see the entire queue, by order of deadline, or you can filter the queue to a particular collecting area. When you are in the "entire queue" view, you can tell what collection something belongs to by virtue of the parenthetical qualifier. So, yes, can confirm that we don't need the initial screen that shows the list by collecting area. 



>Do we want to like user test it with digitization staff or anything?

**This is a good reminder that Sarah Newhouse is also adding items to the DC. I spoke with her, and showed her the draft queue page, which she had no issues with. Jahna has already seen it.  That's all the dig staff!**



>Also, I note that there's still a "Rights & Reproduction" checkbox in digitization item metadata -- I no longer remember what this does or how these even show up -- this is still used, and has been tested to ensure it still works as desired with these changes? @eddierubeiz @Gabz73

It is still used, and yes it works.

jrochkind:
Thanks @Gabz73 !

app/models/admin/digitization_queue_item.rb
eddierubeiz: This was a mistake - thanks for catching it!

Issue #2605, opened 2024-04-29
Edit social media logos


eddierubeiz:
From Clare Hirai:

You may have noticed, but just in case, I wanted to call your attention to the update we made to the social logos in the main site footer. Could you follow suit with Digital Collections when you have a chance?
 
![image](https://github.com/sciencehistory/scihist_digicoll/assets/94723/18afa67b-f22a-4c10-ab91-3b093f1d3fd7)

jrochkind:
Hi Claire, thanks, Eddie created an issue for us here: https://github.com/sciencehistory/scihist_digicoll/issues/2605

I don’t know if you use github? If you do and have a github account, let me know what your account is, and I’ll tag you on the issue and in the future on issues related to you!

Let us know if you consider this urgent, otherwise we’ll get to it at our next convenient time! Thanks!

Also – before we were using .svg graphics, I’m not sure if that’s a customization I just did on my own or if the main website had SVG too.  Using svg can provide sharper graphics on high-resolution mobile screens while also being smaller file size, if you wanted to find/use SVG alternatives instead!

Jonathan


__________________________________

Jonathan Rochkind   (he, him, his or they, them, theirs)
Software Developer/Technical Lead
Othmer Library of Chemical History
t. +1.215.873.8239

Science History Institute
Chemistry • Engineering • Life Sciences
315 Chestnut Street • Philadelphia, PA 19106 • U.S.A.
We tell the stories behind the science.
sciencehistory.org<https://www.sciencehistory.org>
From: Eddie Rubeiz ***@***.***>
Date: Monday, April 29, 2024 at 9:34 AM
To: sciencehistory/scihist_digicoll ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [sciencehistory/scihist_digicoll] Edit social media logos (Issue #2605)

From Clare Hirai:

You may have noticed, but just in case, I wanted to call your attention to the update we made to the social logos in the main site footer. Could you follow suit with Digital Collections when you have a chance?

image.png (view on web)<https://github.com/sciencehistory/scihist_digicoll/assets/94723/18afa67b-f22a-4c10-ab91-3b093f1d3fd7>

jrochkind:
Also: 

Hi again,
 
I clicked on the Staff & Fellows link in the digital collections footer and it seems to be broken. Could you fix if it’s not just me?
 
https://www.sciencehistory.org/about/staff-fellows/

PR #2621, opened 2024-05-20
new social media icons; fix broken link to main website staff page

eddierubeiz:
Ref #2605 

- Switched order to instagram; facebook; twitter/x; youtube, to match main site
- Found two SVGs for the two new icons.
  - The Twitter/X one is from [Twitter/X's site](https://about.x.com/en/who-we-are/brand-).
  - The Facebook one is from www.svgrepo.com
    - [licensing terms look fine to me](https://www.svgrepo.com/page/licensing).
    - If anyone can find a better SVG that matches what we have on the main site, great.
    - I changed the fill on the Facebook from black to white.
- Fixed broken "Staff and Fellows" link in the footer to point to https://www.sciencehistory.org/about/staff-fellows/


jrochkind:
thanks!

PR #2622, opened 2024-05-21
Get Gemfile.lock up to date

eddierubeiz:
I let it get out of date, resulting in an error on deploy.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -843,7 +843,7 @@ DEPENDENCIES
   webmock (~> 3.5)
 
 RUBY VERSION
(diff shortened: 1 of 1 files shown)


jrochkind:
I believe this is fixing my mistake in #2618, updated .ruby-version without changing and commiting new `Gemfile.lock`. 

I think that's a new thing bundler requires you to do as of fairly recent, that I didn't know/forgot about!

I believe you enacted this change @eddierubeiz by running `bundle update --ruby`, and then committing resulting `Gemfile.lock` change? 

Thank you!

PR #2623, opened 2024-05-21
Fix broken link to staff and fellows.

eddierubeiz:
Ref #2621  - I didn't get this right the first time.


jrochkind:
ok thanks for noticing and fixing.  i actualy tried url this time, I think it is working.

jrochkind (approved):

PR #2629, opened 2024-05-23
viewer, when navigating to search result, pan viewer to make sure it's on screen if necessary

jrochkind:
Comes AFTER #2628, merge that first


eddierubeiz:
Will merge as soon as tests pass.

Issue #2410, opened 2023-10-10
Prevent staff from publishing work with corrupt file


apinkney0696:
Per #1652, prevent publishing a work with an asset falling into some basic categories: 

> When TIFF is detected as corrupt, for some actual kinds of corrupt TIFF we have encountered (can't promise it will get all corruption), https://github.com/sciencehistory/scihist_digicoll/pull/2395
> When our system could not determine the file type of the file at all (probably means it's corrupt, also catches zero-byte empty files). https://github.com/sciencehistory/scihist_digicoll/pull/2403
> For audio files which our system can't detect a duration, bitrate, or sample rate: https://github.com/sciencehistory/scihist_digicoll/pull/2406

A red banner pop-up, the same way it prevents one from publishing a work with insufficient metadata seems appropriate to me.

PR #2632, opened 2024-05-24
Refuse to publish a work unless all members have valid (stored) files

eddierubeiz:
Ref #2410.

Includes a simple test.

Note: to check the files, I use
`asset.promotion_failed?`
rather than
`!asset.stored?`.

This is because:
- published works are required to have at least one published representative
- our `:published` work factory trait thus has:
```members {  [ build(:asset_with_faked_file, published: true) ] }```
- `:asset_with_faked_file` uses files that aren't stored
- so requiring published works to have stored files on their assets would mean we would have to stop using `:asset_with_faked_file` in our tests featuring the `trait :published`.

### CODE REVIEW

**Comment by jrochkind** [comment]
OK, I'm confused whats' going on here....

`with_options` was new to me... this is the same thing as a `if: :published?` on the end?  I would find that less confusing. 

`member_files: true` also confused me -- is this saying to use the validator called `MemberFilesValidator` translating `member_files` to `MemberFiles[Validator]`? OK.... 

Still trying to figure out what this is actually doing...   it's loading all members in every time a work save happens on a published work... I'm worried about the performance implications. 

It means that if we try to do a bulk change on works, say iterate through every work to change some metadata -- now for every single published work, it will have to load in all members to check them -- which is actually a pretty expensive operation, and will seriously slow down such an operation. 

So that worries me... 

Some ideas....

We could check not all things that are `published` but only if `published` has _changed_ from false to true.... 

You might do that with a less magic more manual ` validate do |work|` block, that I think would be more readable than using the custom validator and Rails magic. 

**BUT** for such an expensive thing I'm _still_ not sure about doing this as a validation taht gets executed on _every applicable save_ (unless you disable all validations!)

I wonder if it would be better to do this only in actual UI actions that publish work? I guess we might have TWO of those, the individual work controller and the bulk work controller?  (anything else I'm missing?) What do you think about writing some re-usable logic that only gets executed there?

**Comment by eddierubeiz** [comment]
Good suggestions- thanks!
I got rid of `with_options` (although we do use it elsewhere in the model); you're right, no need to use that here.
I also rewrote the code so it only adds one more call to the DB. It's cheap now , so we can treat it like any other validation.
```
  Kithe::Model Pluck (0.7ms)  SELECT kithe_models.friendlier_id,kithe_models.file_data -> 'metadata' -> 'promotion_validation_errors' FROM "kithe_models" WHERE "kithe_models"."parent_id" = $1  [["parent_id", "abc_123"]]
```
Making these changes was easy with the test already written.

**Comment by jrochkind** [comment]
Hey thanks. 

It's actually still one SQL query per Work, right?  It was that before too -- although it's a cheaper one now. 

But one SQL query per published work is still not "like any other validation" --most validations do not trigger a trip to the DB.  And is still pretty expensive. 

It would be interesting to actually compare.  Want to try doing a bulk edit to say, 1000 or 5000 published works on staging, with and without this validation, and see the time difference?

I suspect it will be significant enough to sway you or at least make you think! 

Right now I think it's doing it for anything that is published, whether newly published or not -- meaning the valication can also _stop you from editing a published work_ that was already in a bad state, not just stop you from publishing it.  That also seems like a potential headache. 

While one optimization would only be doing the check if the work has had it's `published` state _change_ from `false` to `true` -- i'm still not feeling great about this, it just seems like a foot gun. This is the sort of thing that makes people not like ActiveRecord callbacks, not even validations.  I think a validation that triggers an SQL query is probably something to avoid unless you really can't. 

I'm still thinking this should be code executed in a controller, not model-level validation code.

**Comment by eddierubeiz** [comment]
I tried validating all the works in the Beckman collection:
```
Benchmark.measure { Collection.find_by_friendlier_id('wh246s128').contains.each { |w| puts "invalid" unless  w.valid? } && nil }
```
Before: 2.4 sec. After: 3.8 sec.
Point taken - I will move this over to the controller update method, yeah.

jrochkind:
Cool. Only trick is that there are at least two places we need to do it, also the bulk update controller?

So still re-usable code somewhere... it probably would even be possible to report the errors as validation errors... just a validation that isn't normally run... I'm not sure.

eddierubeiz:
I moved the code to the controller, and now we're only hitting the DB if a work (or works) are actually being published. And it's one query, even if more than work is being published at once.

PR #2641, opened 2024-05-30
update gem dependencies

jrochkind:
update our ruby gem dependencies.  Partially motivated by preparing for Rails 7.2 release in next couple weeks. 

Includes explicitly opting into new rack 3, as our dependency tree can now handle it -- but wont' necessarily resolve the correct way unless we explicitly list it. 

And includes a couple major version bumps of dependencies. 

- update to rack 3, ensuring sinatra indirect dep is newer 4 that can use it
- 'bundle update' all dependencies
- update alba dependency from 2.x to 3.x
- update 'browser' gem dependency from 5.x to 6.x
- remove rack config workaround that in rack 3.x is a deprecated no-op


jrochkind:
oops heroku doens't have ruby 3.3.2 yet! Have to wait for it.

jrochkind:
@eddierubeiz let's do a dependency update, now that we have succesfully deployed ruby 3.3.2!

PR #2644, opened 2024-05-30
update to latest ruby 3.3.2 with latest bundler

jrochkind:
Redo https://github.com/sciencehistory/scihist_digicoll/pull/2639 (reverted in sciencehistory/scihist_digicoll#2642)

Once heroku supports ruby 3.3.2!


jrochkind:
heroku has it now, so re-applying! Already reviewed

Issue #2284, opened 2023-08-02
Export CSV from Cart


apinkney0696:
Add an "export CSV" option for items in the cart.

Metadata should include:

- Title
- URL
- External ID
- Department
- Creator
- Date
- Place
- Genre
- Series Arrangement
- Physical Location
- Collection

However, if it's possible, I think it might be ideal if the export included all or most of the metadata fields. I can then remove data based on my needs once I download the CSV export. Here are some examples of instances I've requested metadata exports before, for reference:  
Ticket #   (Data Requested)
#1951 (Title, Date, Place, External ID)
#1927 (Title, Creator, Place, Date, Genre, External ID, Department)
#1563 (Title, URL, Department, Genre, Creator)
#1449 (Title, URL, Date, Description)

An issue is how to structure the data in the CSV. Jonathan has a great summary of this problem in ticket #1449:
 
> "Our data for a Work doesn't always "obviously" map to a flat list of individual values though. For one obvious example, let's say if they wanted individual asset filenames or sizes. A given work might have 1 asset, or 2, or 10. How do you map the filenames and sizes of multiple children to a single line for a "Work" in a CSV? You can come up with various schemes if you want, whether they'll be workable or not I don't know.
> 
> Or, just staying within Work metadata, for instance, we model "inscriptions" as a "location" and "text". There can be 0 to multiple inscriptions, each of which has a "location" and a "text". How do you map this to CSV? Again, i could come up with some hacky way, but there isn't just one true way....
> 
> Or even subjects -- a work can have many subjects, do you just put them all in a single CSV value, seperated by... maybe not a comma, that gets confusing, because CSV already uses commas to separate values! By semi-colons? And you count on no subject ever containing a semi-colon itself? I think that's a safe assumption in the case of FAST/LCSH.
> 
> This is just a list of examples. In general, determining "how do I represent this all in a CSV" is a non-trivial task, that could take, oh a week or two, to try to go through and get "all" the metadata, and if we just try to come up with something we think is "reasonable"...

I think separating multiple values in one cell by semi-colons, as Eddie did in the export in #1927 would work well. Let's talk through any other info you need and a timeline.

jrochkind:
"All or most metadata fields" in a CSV gets pretty hairy. What do you think @eddierubeiz ?

I suppose we could imagine a feature that let you SPECIFY fields to include, but that is much more complicated to implement. 

[...]

PR #2651, opened 2024-06-07
CSV report from cart

eddierubeiz:
Ref #2284.
UI is approved by @Gabz73.
Works as advertised - includes some tests and happy to add more.
I'd like the serializer to be easy to subclass and call from a rake task or the command line, so we're not reinventing the wheel anytime we need a CSV report based on a particular set of works.


jrochkind:
Looks pretty well done, thanks!

Issue #2652, opened 2024-06-10
Report of textual works without OCR


Gabz73:
Can I have a report of works for which
Language = French or German or Spanish 
Department = Library or Archives or Museum
Format = text
OCR enabled = NO

Report should show:
work ID (what do we call this - I mean the "of8lva6" in "https://digital.sciencehistory.org/works/of8lva6")
title
date/s
desc
genre/s

eddierubeiz:
(We call that the `friendlier_id`).

Gabz73:
Awww!


__________________________________

Gabriela Zoller   (she, her, hers)
Director of Library Services
Othmer Library of Chemical History
t. +1.215.873.8289

Science History Institute
Chemistry • Engineering • Life Sciences
315 Chestnut Street • Philadelphia, PA 19106 • U.S.A.
We tell the stories behind the science.
sciencehistory.org<https://www.sciencehistory.org>

eddierubeiz:
Just for historical interest:
https://github.com/sciencehistory/scihist_digicoll/issues/2080

eddierubeiz:
```
scope = Work.
        where("json_attributes -> 'department' ?| array[:depts  ]", depts:   ['Archives', 'Museum', 'Library'] ).
        where("json_attributes -> 'language'   ?| array[:langs  ]", langs:   ['English', 'French', 'Spanish', 'German'] ).
        where("json_attributes -> 'format'     ?| array[:formats]", formats: ['text'] ).
        where("(json_attributes -> 'ocr_requested')::boolean = false" )

report = WorkCartSerializer.new(scope, columns: [:title, :url, :date, :description, :genre]).to_a

puts CSV.generate { |csv| report.each { |line| csv << line }}
```

eddierubeiz:
With that code in a file named `export.rb`,
```
cat export.rb | heroku run console --app=scihist-digicoll-production --no-tty > prod_out.csv
```

eddierubeiz:
Here's the resulting file. Happy to adjust, of course.
[June_12th_report.xlsx.gz](https://github.com/user-attachments/files/15809593/June_12th_report.xlsx.gz)

eddierubeiz:
Here it is. Let me know if you need anything adjusted.
[june_12_report.xlsx](https://github.com/user-attachments/files/15809621/june_12_report.xlsx)

eddierubeiz:
I created OCR for these in production today.

PR #2663, opened 2024-06-24
Add OCR to some more works

eddierubeiz:
This is a whittled-down list of the works we listed in #2652.
This adds the jobs to the `special_jobs` queue - doesn't actually add the OCR . That way we can run the OCR creation jobs on separate dynos.


eddierubeiz:
Awaiting final version of the list from @JAuerbach26  :)

PR #2667, opened 2024-07-02
Support Cloudfront distros in front of select S3 buckets

jrochkind:
New ENV varibables corresponding to each S3 bucket, name of S3 bucket configuration with `_HOST` appended. If set, there is cloudfront in front of this bucket, and the value is the cloudfront distro domain hostname. 

When ENV indicates a cloudfront distro in front of an S3 bucket, our code for configuring the [shrine](https://shrinerb.com/) storage for the bucket will take that into account, and configure such that it will use Cloudfront URLs to specified hostname. 

To do that, we create our own custom subclass of the Shrine S3 storage class, that knows how to generate CloudFront urls properly, including Cloudfront-signed urls when necessary, and passing through the response-content-distribution (et al) headers we have configured out CloudFront distros to pass through to S3 to request certain response headers. 

Accompanying terraform PR at https://github.com/sciencehistory/terraform_scihist_digicoll/pull/81 has the infrastructure changes to actually set up the cloudfront distros (and provide values for new ENV variables). That PR will also have some additional work checkboxes on it for documentation changes, roll-out plan, etc. 

The intent is that *this* PR should leave things functioning as-is if relevant ENV vars are NOT set, so can be merged when ready to be waiting for actual roll-out of CloudFront. 

- add Shrine storage for CloudfrontS3Storage, for generating URLs for our public and restricted cloudfront distros in front of buckets
- env can create cloudfront-fronted S3 buckets, that will generate cloudfront urls properly, including signed urls, and response-content-disposition etc query params

**Deployment checklist on terraform PR at https://github.com/sciencehistory/terraform_scihist_digicoll/pull/81**


jrochkind:
This should be deployable with no terraform changes and no change to behavior as long as new `_HOST` config is **not** set, behavior should remain the same. So this is the first step. 

**Full deployment checklist on terraform PR at https://github.com/sciencehistory/terraform_scihist_digicoll/pull/81 **

jrochkind:
So, the Cache policies we are using, suggested for S3, such as the "CachingOptimized", will by default cache for 24 hours, but respect origin cache-control headers where present. 

Some of our content we uploaded to S3 with metadata saying 1-year cache-control headers, so CloudFront will cache for a year. Great, this is what we want! Our content is immutable at a given url (we put fingerprints etc in urls), and can be cached forever. 

But other of our S3 content we accidentally did not upload with metadata for cache-control headers. It would be a pain and take a while and have some expense to try to set that on all headers now. 

So we will use custom Cache policies that _default_ to a 1 year TTL in CloudFront cache. 

As long as we only use this on top of buckets where we always use immutable-content urls (urls with digests, unique ids, etc in them), we're good.

jrochkind:
Nevemrind, we are NOT going to change the default TTL. 

a) fewer than I thought of our Shrine storages were failing to set S3 cache-control metadata -- the big ones are, the only ones missing are on_demand_derivatives (temporary anyway), and combined_audio (would be nice, will fix for the future, may think of other fixes)

b) Changing the defualt TTL increases complexity of what is already a complex system. It would result in for instance 404's being cached for a year too!

jrochkind:
S3 objects can be uploaded with metadata that will set the cache-control response headers they are returned with. 

Cache-control response headers will be used by CloudFront to determine cache times. 

We have added this metadata with far-future cache-control headers for on-demand derivatives and combined-audio-derivatives, missing before. These are public files with fingerprinted urls, so they are publicaly cacheable forever, and we can get them to maybe cache longer by doing this. 

We are not currently going and setting metadata and all existing objects. The existing objects may still be cached for cloudfront default TTL, and delivered to end-clients without cache-control headers. May not be ideal, but also may not matter very much for these unusual storages (combined audio is requested only via RANGE requests anyhow; Cloudfront default is 7 days anyhow).

lib/scihist_digicoll/shrine_storage/cloudfront_s3_storage.rb
```diff
@@ -0,0 +1,125 @@
...
+module ScihistDigicoll
+  module ShrineStorage
+    class CloudfrontS3Storage < ::Shrine::Storage::S3
+      attr_reader :host, :public_mode, :cloudfront_signer, :key_pair_id
+
+      QUERY_PARAMS_PROXIED = {
+        response_cache_control: "response-cache-control",
+        response_content_disposition: "response-content-disposition",
+        response_content_encoding: "response-content-encoding",
+        response_content_language: "response-content-langauge",
```
eddierubeiz, line 28: Typo

jrochkind: thanks for catching!

Issue #2664, opened 2024-06-27
Oral history page: "Copy" button on the table of contents page sometimes doesn't work


eddierubeiz:
### Setup
  - Start here: https://digital.sciencehistory.org/works/dqn3jcv#tab=ohToc
- Type "Kenya" into the search box under the play bar.
- Click the "Copy" button

### Expected behavior
- "https://digital.sciencehistory.org/works/dqn3jcv#t=0&tab=ohToc" should be copied into the clipboard
- A tooltip appears with "Copied to Clipboard"

### Actual behavior
- Nothing happens.

Reported by @sarahschneiderSHI

PR #2668, opened 2024-07-09
Fix js bug in oh toc page

eddierubeiz:
Ref #2664
In OH searches, don't replace the `html` of the entire section; instead, replace the innerHTML of particular subsections marked by CSS as being safe.
Adds to the system tests so this won't break again.

### DIFF
--- a/app/components/oral_history/ohms_index_component.html.erb
+++ b/app/components/oral_history/ohms_index_component.html.erb
@@ -48,21 +48,21 @@
           </div>
 
           <% if index_point.synopsis.present? %>
-            <p>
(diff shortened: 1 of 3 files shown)


jrochkind:
nice, thanks!

I don't completely follow, but haven't had a whole brain avail to grok it, ok! Do you think it need any comments explaining anything?

jrochkind (approved):

eddierubeiz:
The code was wiping out the DOM event associated with the "copy to clipboard" button. I'm just being a little more conservative about where we add the red "match" highlights, so now the button works again even if there's a match in that section.

PR #2671, opened 2024-07-11
Fix flaky oh system test

eddierubeiz:
We've had this problem before: a click triggers a bunch of slow javascript, and the next `expect` in the system test fails.... part of the time.

I borrowed [a recipe](https://github.com/sciencehistory/scihist_digicoll/pull/2448) from @jrochkind which, coincidentally, appears in the same test.

This fixes the problem and barely slows the test down.


eddierubeiz:
I mistakenly merged another branch with the same name into this PR. Let's just try again.

PR #2675, opened 2024-07-16
cart: select / deselect all works on a page

eddierubeiz:
- Cleaned up JS code a bit.
- Includes controller test and a system test.
- Doesn't work on mobile, but neither does the other cart functionality, so eh. We're not going to be doing mass metadata updates from a phone browser.
- As of this PR, the database is hit once for every work added.


spec/system/batch_edit_spec.rb
```diff
@@ -33,6 +33,45 @@
...
+    # Admin works page
+
+    visit admin_works_path
+
+
+    # Check all items
+    find("#check-or-uncheck-all-works").check
+
+    # Remove this line at your own risk.
+    stall =  [work0, work1, work2, work3,].map {|w| find("#cartToggle-#{w.friendlier_id}").checked?}
```
eddierubeiz, line 46: Workaround for flaky behavior; this will have to do for now.

app/views/admin/works/index.html.erb
```diff
@@ -20,7 +20,7 @@
 
   <div class="admin-filters">
     <div class="admin-filter">
-      <label for="q_published">Published</label>
+      <label for="q_published_eq">Published</label>
```
eddierubeiz, line 23: This label was incorrect.

app/views/admin/works/index.html.erb
```diff
@@ -89,7 +89,10 @@
       <th class="datestamp"><%= sort_link(@q, :created_at, "Created", page: nil) %></th>
       <th class="datestamp"><%= sort_link(@q, :updated_at, "Last Modified", page: nil) %></th>
       <th class="action"></th>
-      <th>Cart</th>
+
+      <% if can?(:update, Kithe::Model) %>
+        <th>Cart</th>
+      <% end %>
```
jrochkind, line 95: 1. What makes this the right privilge to check for adding/removing someting from your personal cart? And why do we need a privilege to check, is there anyone who has access to this page who does NOT have access to a Cart?

spec/system/batch_edit_spec.rb
jrochkind: this actually seems reasonable i guess

jrochkind:
OK, so this looks like it'll work....

but it's a fairly lot of JS, and maintaining this kind of JS isn't our favorite. 

I'm still wondering...  what are the downsides of just doing an actual page submit/reload instead of leaving the same page loaded and having JS disable the form and then change state of all checkboxes?  We think much worse UX?

app/views/admin/works/index.html.erb
eddierubeiz: Oh, I logged in as one of the "staff viewer" users and noticed  the cart column is entirely empty for them. But this might belong in another PR.

jrochkind:
OK I'm just merging and deploying then! Enough bikeshedding! Thanks for this!

jrochkind:
Man, looking at how much code was in the PR when I merged it... I'm still feeling suspicious! We can always change it later if it gives us trouble?

eddierubeiz:
Absolutely - we can either refactor it or redo it in a less front-endy way if we feel like it's more trouble than it's worth.

PR #2677, opened 2024-07-16
Changing name of vite config from .ts (typescript) to .mjs (javascript module) avoids vite deprecation message.

jrochkind:
Vite deprecation message looks like:

```
remote:        Build with Vite complete: /tmp/build_93418466/public/vite
remote:        The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
```

And started in vite 5.x.

Relation to .ts typescript and simple solution suggested from https://github.com/ElMassimo/vite_ruby/issues/431

Seems to work great. test deploy to staging succesful


jrochkind:
Better way to do it at #2679 it turns out

Issue #2683, opened 2024-07-17
July 17th outage due to misbehaving SearchStax Solr


jrochkind:
We had a ~4 hour time window this morning in which Digital Collections was largely unavailable. 

The short summary is that SearchStaxh hosted Solr was being incredibly slow and/or timing out; this was resulting in our puma worker threads spending a lot of time waiting for Solr; which caused running out of puma worker threads, and large queue. 

Restarting and even reconfiguring SearchStax solr did not eliminate bad behavior. But creating a _new_ SearchStax solr instance, indexing, and switching production over to use it, seems to have given us a Solr performing as we expect, and resolving the problem. 

More complete notes below.

jrochkind:
Operational notebook kept while diagnosing and solving the problem....



* Solr was overloaded somehow and not responding
  * Solr admin showed spiking load average (7 to 8) and heap memory usage
  * Searchstax health checks were showing not healthy with "pulse agent is down"
  * Heroku metrics were consistent with puma workers being exhausted despite not having heavy traffic or heroku CPU usage — which app responses timing out at 30 seconds — which is consistent on puma workers being busy waiting for solr. (Our app is supposed to have a timeout waiting on solr shorter than 30  seconds, so not sure what was going on there! Consider trying to improve this….)

* searchstax did not report any general system outages or problems on status page https://status.searchstax.com/

* Restarting searchstax solr… problem quickly re-exhibited same way. 

* So we're going to try putting app into maintenance mode so the solr should not be utilied, then actually deleting and re-creating the whole solr "collection" in case it has become corrupt in some way. 

* put app in maintenance mode so there should be no traffic on solr server

* Delete solr "collection" entirely….
  * delete collection from admin
    

* create new collection configured for us, heroku production bash then 
	RAILS_ENV=production ./bin/rake scihist:solr_cloud:create_collection
  * Ran into problem where this was not actually working.  "SolrConfigsetUpdater::SolrError: The configuration scihist_digicoll_baee394 already exists in zookeeper (SolrConfigsetUpdater::SolrError)". 
  * Have to create collection with existing "configuration", which could do manually in production like this:
        * `updater = SolrConfigsetUpdater.configured`
        * `updater.create_collection(configset_name: updater.configset_digest_name)`
  * TODO need to fix the rake task to handle this situation. 




* OK, now reindex, from production heroku bash
[...]

PR #2686, opened 2024-07-17
scihist:solr_cloud:create_collection rake task can handle an already existing configset, and use that

jrochkind:
Normally doesn't happen, but ran into that state when trying delete and re-create a collection in #2683

So this makes this task more resilient to that condition, so it does what it needs even when we're in that state trying to deal with a problem!

It's okay to assume an existing configset with same name is the same configset, because we put a digest code into config set name. If you need more precise things, like manualy deleting the config set instead of re-using it, you just gotta open a console and use SolrConfigsetUpdater helper directly.

### DIFF
--- a/lib/tasks/solr_cloud.rake
+++ b/lib/tasks/solr_cloud.rake
@@ -12,7 +12,15 @@ namespace :scihist do
     task :create_collection => :environment do
       updater = SolrConfigsetUpdater.configured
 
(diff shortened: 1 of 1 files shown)


jrochkind:
manually tested on staging in various configuration states, seems to work well

eddierubeiz (approved):

eddierubeiz (approved):

PR #2687, opened 2024-07-18
Replace lazysizes.js with browser lazy loading, and browser aspect-ratio

jrochkind:
We previously used lazysizes.js to lazily load images -- but since then, lazy loading is widely supported in browsers, and we can just use that. 

Similarly, we previously used a hack involving a wrapper DIV and padding to *preserve proper space* on page for not-yet-loaded image (with lazy loaded image being main but not necessarily only use case). But now browsers support an `aspect-ratio` CSS that can be used for that purpose, that we can add as an inline style. 

So we do all that in the places we were using it, simplifying JS, CSS, and DOM. 

We also add lazy loading to a few random spots on the home page, cause it's so easy now. 

- replace lazysizes.js with native lazyloading in ThumbComponent
- ThumbComponent uses native aspect-ratio CSS to reserve space instead of wrapper padding hack
- fix comment we dont' use lazysizes.js here anymore
- viewer uses native lazy loading instead of lazysizes.js
- remove spinner icon from non-loaded image; it doesn't look great, and if the user is seeing it with native lazy loading the image prob is broken not loading
- image placeholder style for new native lazyloading instead of lazysizes.js
- stop including lazysizes.js in pages, we don't use it anymore
- remove lazysizes.js from package.json, we don't use it anymore
- add some browser lazy loading to some front page thumbs


jrochkind:
This is failing very mysteriously on push to heroku staging. 

```
remote:        info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
remote:        error Command "vite" not found.
remote:
remote:        pid 4968 exit 1
remote:        Build with Vite failed! ❌
remote:
remote:  !
remote:  !     Precompiling assets failed.
```

That doesn't make much sense. The only JS change we made was removing lazisizes from package... why would that trigger failirue to find `vite` command line that was fine before?

Something related to heroku caching? 

I may post a question on a vite-ruby forum.

jrochkind:
OK I think I know what's up, here's my hypothesis:

Indeed, `vite` was in a heroku cache. When I changed package.json, it busted the cache, and when the deploy process re-installed JS, it was not there. 

What changed between the last time we installed JS (and got it in the cache) and now?  We started using the NPM buildpack.  #2498   Which installs our JS for us, possibly different than the ruby buildpack did. 

Vite is listed as a `devDependency`.  

https://github.com/sciencehistory/scihist_digicoll/blob/53a1f0cd346f7456f547f37d9c5cf0852a419a1b/package.json#L24. 

Maybe the old build installed devDependencies but new one did not?  The nodejs buildpack _says_ it installs devDependencies, but says:

> After running the installation and [build steps](https://devcenter.heroku.com/articles/nodejs-support#heroku-specific-build-steps), Heroku strips out the packages declared under devDependencies before deploying the application.

It's possible heroku doesn't really install devDependencies, [someone on StackOverflow does suggest that may be the case](https://stackoverflow.com/a/69935179/307106)

I think it's more likely that the point the nodejs buildpack strips out devDependencies is _before_ the asset precompile runs, when we still need them. 

And indeed if we look at the deploy console, we can see `remote: -----> Pruning devDependencies` appear _before_ `rake assets:precompile` 

**OK, so that's it**. 

So... how to solve?  It would be awesome to still get devDependencies pruned, is there any way to rearrange things so vite assets precompile happens before they get pruned? Unclear.  But we probably weren't getting devDependencies pruned with just ruby buildpack anyway. 

Possilby we don't need *any installed JS dependencies* after we get assets precompiled, after that point only the precompiled assets are used for web, so maybe we could use this buildpack to prune *everything*: https://elements.heroku.com/buildpacks/leoafarias/heroku-buildpack-node-modules-cleanup

But we may still be using some dependencies at runtime for various command-line things we run for media processing, hmm. 

Ha, look, indeed [vite-ruby did warn about this](https://vite-ruby.netlify.app/guide/deployment.html#using-heroku)

> When precompiling assets in Heroku, it's better to [skip pruning](https://devcenter.heroku.com/articles/nodejs-support#skip-pruning) of [dev dependencies](https://vite-ruby.netlify.app/guide/deployment.html#development-dependencies-%F0%9F%94%97) by setting:

>        heroku config:set NPM_CONFIG_INCLUDE='dev' YARN_PRODUCTION=false
>         # or NPM_CONFIG_PRODUCTION=false in versions of npm < 7
>
> That will ensure that [vite](https://vitejs.dev/) and [vite-plugin-ruby](https://github.com/ElMassimo/vite_ruby/tree/main/vite-plugin-ruby) are available, along with other build tools.

OK, for now will do that, maybe in another PR, after testing.

jrochkind:
Hm, even thiough I know see in deploy console:

```
remote: -----> Pruning devDependencies
remote:        Skipping because YARN_PRODUCTION is 'false'
remote:
```

I'm still getting an error on deploy, `error Command "vite" not found.`!

I'm going to have to keep experimenting, perhaps with a new dummy app.

jrochkind:
OK added vite to "dependencies" instead of "devDependencies", and pruned heroku build cache... still having `error Command "vite" not found.`, what??

jrochkind:
Can't deploy until we fix the heroku problem (which I guess I should make a separate ticket for), but ready for review any time @eddierubeiz

eddierubeiz (approved):

Issue #2627, opened 2024-05-22
add finding aid links to all archives department works without one


archivistsarah:
### TODO
- [x] Remove now-obsolete finding aids using the rake task at https://github.com/sciencehistory/scihist_digicoll/pull/2726 . The command is `bundle exec rake scihist:data_fixes:remove_extra_finding_aids` .

### Description
Currently, our metadata creation practice is that all works/items belonging to the archives department contain a publicly accessible link from the work in the digital collections to the public finding aid in ArchivesSpace for the collection the work belongs to.

But because we launched the Digital Collections well before the ArchivesSpace public user interface was in use, only archives DC works created _after_ the ASpace PUI was in use contain that link. Approximately 5000 archives works created before the ASpace PUI launch are missing this link and archives staff would like for it to be retroactively added to their metadata. 

Links to finding aids are added to metadata using the Related Link fields:
Category = Finding aid
URL = [something that looks like] https://archives.sciencehistory.org/repositories/3/resources/236

It appears on the front end of the Digital Collections as a visually distinct button ("collection guide") in the Institutional Location section:
![institutional location fields DC work](https://github.com/sciencehistory/scihist_digicoll/assets/25084902/fdf62920-63c3-4bee-bac5-cb729779985b)

Other info:
All archival collection-level works in the DC _should_ already contain a link to their finding aid in the collection's related link field. Is it possible to have works belonging to these collections automatically inherit this link? As a one-time batch update?

All archival works in the DC _should_ be from a fully processed archival collection that _should_ have a public finding aid in the ArchivesSpace PUI. It is possible, however, that archival items were digitized from collections that do not yet have public records in ArchivesSpace (this shouldn't be happening, but just trying to cover my bases here).

I can generate a CSV containing the accession number (in the DC metadata this is stored in External ID: Accession No.), title, and public URL for each archives collection with a public finding aid and/or only the collections that have digitized content in the DC. 

Also, ASpace has an API and OAI-PMH functionality that are options if we need them for this or other projects.

jrochkind:
Let's talk about if we want to just display link from Collection a Work is part of (possibly changing workflow going forward?), or if we do want to continue adding this duplicate link to every Work, and bulk do that. 

Initially at next weekly meeting?

archivistsarah:
Sure, I can pop into the weekly meeting.

jrochkind:
We decided to display the Finding Aid link from linked Collection, so it does not need to be entered in individual Work. 

[...]

PR #2691, opened 2024-07-23
Include parents and collections when looking for finding aids

eddierubeiz:
Part of Ref #2627 

See in particular https://github.com/sciencehistory/scihist_digicoll/issues/2627#issuecomment-2274186057 which spells out the metadata guidelines for which works have and don't have finding aid metadata directly attached to them.

### DIFF
--- a/app/components/work_show_info_component.rb
+++ b/app/components/work_show_info_component.rb
@@ -60,12 +60,18 @@ class WorkShowInfoComponent < ApplicationComponent
     @links_to_opac
   end
 
-
-  # We look these up via the collection and its related links.
+  # We look these up on the work itself, the work's parent, and the collections they're in.
+  # This would be simpler if we didn't have to support child works, but eh.
   def links_to_finding_aids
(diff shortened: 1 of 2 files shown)


jrochkind:
@eddierubeiz You write 'rather than', but the spec I was anticipating was to look in both places. Does that seem workable?

eddierubeiz:
Missed your comment the other day, but yeah, you're of course right. I'm also looking at parent works (and their collections) in this draft version of the PR.
Once Sarah adds links to the child works mentioned in the [report](https://github.com/user-attachments/files/16366645/Archival.works.in.the.digital.collection.with.no.finding.aid.xlsx), I can amend the PR to ignore parent works before we merge.

eddierubeiz:
Note: I'm replacing a call to method `RelatedLinkFilter#finding_aid_related_links` and replacing it with  a couple lines of inline code. It makes more sense now that we're listing the finding aids for more than one model instance.

jrochkind:
Nice, love how little this took!

PR #2698, opened 2024-08-06
patch Questioning Authority gem to add sort param to OCLC AssignFAST API calls

jrochkind:


jrochkind:
Want to submit a PR to qa gem itself and link here too.

jrochkind:
failure?? What, ok have to fix that, sorry not actually ready

jrochkind:
Doh failure is just an unreliable test. :(  I think it'll pass on re-run. 

>  1) Oral history with audio display With OHMS synchronized transcript and ToC can display, and search, without errors
     Failure/Error: expect(page).to have_content(%r{Table of Contents — 1 / 7}i)

If so, we gotta either fix that or disable it.

eddierubeiz (approved):

jrochkind:
Submitted upstream at https://github.com/samvera/questioning_authority/pull/390

Issue #2492, opened 2024-01-11
Support PDF files in Digital Collections


apinkney0696:
@jrochkind I'll let you note the technical to-do's below. 

Our first PDFs in the DC should be the Best Of Distillations magazine files for 2019 and 2023.
The two PDF files can be found here: https://files.sciencehistory.org/navigate/folder/e3c82a7b-cb79-490a-a262-2ac294475a83

Their complete metadata records, awaiting assets, are here:
https://digital.sciencehistory.org/works/h3bv3vt
https://digital.sciencehistory.org/works/hvlp3f4

jrochkind:
Awesome. Those PDFs are about 50MB each. Not sure if we might want to investigate making an "access copy" with lower resolution images for a smaller download, either as the main thing we ingest or an alternate download?

jrochkind:
And those copies also have a "page" being the two-page spread, which I guess we'll leave that way?

I am curious the provenance of these PDFs to understand what's going on.

apinkney0696:
I believe these were the PDFs shared with Clay by the designers from their InDesign files. We would have to pay them about $800 for access to the original InDesign files if I remember correctly. 

I would wait to see what Michelle and/or Clay respond to the email about whether they even think we should put the large file in the DC at all.

jrochkind:
Cool cool, I don't see any need to have the original InDesign!

the PDFs we create ourselves from scanned images use 150dpi JP2s for around 500kb per page on average -- where each page is really nothing but a photograph, since we create them from photographs. But just to have a sense of what size we decided for access copy previously. 

I haven't fully investigated these PDFs to see if they are whole-image pages, or (hopefully!) real born-digital PDFs with just text and the images are only used for image elements. I haven't looked at what formats the images are in.  But they weigh in at around 5MB per page, so a lot larger than what we are otherwise using for access copies.  

One thing to consider them would be downsampling the images to something more similar size/quality to what we are otherwise using for access copies.  I think the main reason to do this would be convenience of our end users, more managable files.  It could be done manually using Adobe Acrobat, for instance.

Gabz73:
Hi, following up on our earlier meeting with some thoughts. In my files I have an email from Annabel to Michelle and Clay, cc'ing Jonathan and me, inquiring whether their desire is that the "DC preserve the original files or only act as an access point". To my knowledge, they did not respond. 

[...]

PR #2702, opened 2024-08-08
Create image assets for each page of a PDF work

jrochkind:
PdfToPageImages can create an Asset representing a specific page of a PDF. 

This is meant to be used on "born digital" type PDFs that have embedded text and usually vector graphics -- if it's all raster graphics anyway, not sure if this is really useful.  Main use case is our "Best of Distillations" PDFs -- this is part of continuing #2492 (not done yet). 

* Extract a rendered image as a JPG, at 300 dpi, using `vips` CLI
* Extract the text with poppler `pdftotext` CLI, then convert it to hocr
* Make an Asset model that has the image and hocr, with a new role `extracted_pdf_page` to keep track of it
* role is set to `extracted_pdf_page`, and new metadata is embeddded with the page index from the PDF. This allows us to identify whether an Asset has already been created for a given page, for, say, idempotently-lazy creation.  
  * `position` is also set to page num to order the pages appropriately, and `title` is also set with the page number so alphabetical sort of title will be be proper order! Order of course is easily changed on purpose or accidentally by various processes, so isn't reliable metadata identifying page index of original pdf. 

Also store the sha512 fingerprint and asset uuid PK of original PDF Asset, for bookkeeping, if we ever need to retrace the steps of the auto creation of these, or verify them!

# overwriting and lazy creation

Since we can identify whether an Asset already exists extracted for a given page index, we can have page creation code take account of that, to, depending on argument:
1. do not create new one if existing one already exists, just abort. (Or standard bg job will prob do this, for efficient idempotence)
2. UPDATE existing one with new data and metadata, leaving Asset object there with existing friendlier_id. (If, say, generation routines have been improved and we want to re-gen -- leaving bookmarked links and ids the same, hooray)
3. Don't bother checking for duplicate, just insert new dup (honestly don't know a use case, maybe it shoudln't have been included)


jrochkind:
Saw a couple things I want to tweak....

* Remote no longer relevant TODO comment
* clean up tempfile for extract jpeg when using create asset

Still can be reviewed,but making draft to give me a chance to clean those up next week!

eddierubeiz (approved):

PR #2703, opened 2024-08-13
Block accidental JPEGs from publication

eddierubeiz:
Ref #2407 .

All image assets in the digital collections should be TIFFs except if they are portraits or collection thumbnails.

Otherwise stated:
**IF** you're publishing, or bulk publishing, a work or set of works,
**AND** one of their members is an asset with an image file attached,
**AND** the asset is not a tiff,
**AND** the asset is not a portrait,
**THEN** we you may not publish the work.

I'm putting this code in an existing method, `works_with_members_with_invalid_files`, modifying the method slightly.

Note that our commonly-used factory call, `build(:asset_with_faked_file)` actually uses a `.png` as its default file, and thus breaks the new rule. Turns out this isn't such a big deal; only a few tests need to be adjusted.

### DIFF
--- a/spec/controllers/admin/works_controller_spec.rb
+++ b/spec/controllers/admin/works_controller_spec.rb
@@ -122,7 +122,16 @@ RSpec.describe Admin::WorksController, :logged_in_user, type: :controller, queue
   end
 
   context "#batch_publish_toggle", logged_in_user: :admin do
-    let(:representative) {  build(:asset_with_faked_file, published: true)}
+    let(:representative) {
+      create(
+        :asset,
+        :inline_promoted_file,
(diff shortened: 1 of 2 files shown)


spec/controllers/admin/works_controller_spec.rb
```diff
@@ -122,19 +122,18 @@
   end
 
   context "#batch_publish_toggle", logged_in_user: :admin do
```
eddierubeiz, line 124: Without the :tiff trait, the asset actually ends up with a png file, which it is now illegal to publish.

jrochkind:
nice this was implementable so cleanly!

Issue #2521, opened 2024-01-26
Add a "dismissed" (or equivalent) OH request status


eddierubeiz:
@sarahschneiderSHI would like to consider adding a "request dismissed" option (label to be determined).
Use cases:
- Someone submits the same request twice in a row/at the same time (happens on occasion).
- Other miscellaneous cases

PR #2704, opened 2024-08-14
Add a "dismissed" status for OH requests

eddierubeiz:
Ref #2521
@sarahschneiderSHI has approved the user interface, but open to any advice.


jrochkind:
Stakeholders definitely want "dismissed" to send an email to patron?

eddierubeiz:
They do not want "dismissed" to send an email to patron, no.

jrochkind:
The admind OralHistoryRequestsController is getting a bit squirrely... I wonder if instead of having all those conditionals with sort of false-economy DRY, it would be better to just have separate methods `respond_to_approve`, `respond_to_reject`, `respond_to_dismiss`. 

But just a thought, not important enough to spend time doing at present, it works!

jrochkind:
Re: Where the button is placed -- Staff before was concerned about buttons being clearly separated to avoid accidentally hitting the wrong one.... ah, but I see you put a "are you sure" prompt on dismiss, cool!

I feel like i still might move the button either to the right, or put a bigger margin on top?  I think it's too close to the most commonly used "approve" button?  What do you think?  

Other than this, it's approved!

jrochkind:
I'd also keep the prompt from stretching across teh whole screen -- it's too wide to be readable on wide screens, limit it's width.

eddierubeiz:
Moved the button down a bit and made the text narrower, as you suggested.

Also decided to move some of the branching code into private methods; I think this makes things a lot more readable.

jrochkind:
nice!

jrochkind (approved):

Issue #2559, opened 2024-03-05
Consider changing color of oral history requests sign-out banner


sarahschneiderSHI:
It would be nice for this banner (that appears after signing out of oral history requests) to be dark green to match the other banners:
![image (3)](https://github.com/sciencehistory/scihist_digicoll/assets/103284115/d723bdc7-a3c1-4f04-8f8d-b77d394ad724)

Thanks,
Sarah

P.S. I cannot mark things with the tag oral history, so please feel free to do so.

jrochkind:
Technically that's showing up with color theme labelled "info", whereas other ones are with color theme labelled "success". Thus the difference in color. 

But we dont' use too many of these alerts user-facing, and haven't been too intentional with how we use them, this is kind of just what we inherited from bootstrap.

PR #2706, opened 2024-08-16
Use success flash message on OH session logout

eddierubeiz:
Ref #2559


eddierubeiz:
I don't have strong feelings about this either way, but if this is what we want, this will make the message green. :)

eddierubeiz:
Still need to fix test:
``` 
OralHistorySessionsController#destroy signs out
Failure/Error: expect(flash[:notice]).to include("signed out")
expected nil to include "signed out", but it does not respond to `include?`
# ./spec/controllers/oral_history_sessions_controller_spec.rb:80:in `block (3 levels) in <main>'
```

Issue #2606, opened 2024-04-29
Track how many people use "search inside the text" feature


Gabz73:
I would like us to be able to track how many users use the "search inside the text" feature from the work's page, from the image viewer.I would also like to know what term they are searching.

 I don't know if this is something we can achieve via Google Analytics, so that I could then capture it as part of the monthly stats?  If it isn't achievable via GA, I would like a way that I could easily access the monthly #s without having to bother the tech team.

eddierubeiz:
We decided this morning not to report the search phrase to GA (this feature is in PR #2715, which is closed for now) until we have a chance to test it in staging. Instead, let's report just the *fact* that someone searched inside the work.

PR #2707, opened 2024-08-16
Report search-inside-the-book event to google analytics

eddierubeiz:
Ref #2606.
- Refactored the code in `custom_google_analytics_4_events.js` so that it exports a function we can call from the search code. Seemed like the DRYest thing to do, although I'm open to other ideas.
- We're using the `event_value` parameter, which we don't use anywhere else in the digital collections, to report the query string to GA. (This is a bit of an experiment.)
  - 💡In case we decide to use `event_value` somewhere else later, I tweaked the `custom_google_analytics_4_events.js` to also look for a `data-analytics-value` on the decorated link, and send it to GA along with the other metadata. Does not affect any existing reporting, as `getAttribute` will just return `null` in existing code.

### DIFF
--- a/app/frontend/javascript/custom_google_analytics_4_events.js
+++ b/app/frontend/javascript/custom_google_analytics_4_events.js
@@ -1,28 +1,67 @@
 // Send a custom Google Analytics event with data-attributes on a link.
 // Details about migrating to the GA4 event tracker are here:
 // https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs#measure_events_with_the_default_tracker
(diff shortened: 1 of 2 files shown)


jrochkind:
Hm, before we were just adding some data-* attributes on every hyperlink, and had some code that caught click events with those data events, and then reported them to ga. 

I was anticipating doing the same thing, but expanding our code to look for form submits not just hyperlinks? 

I'm not totally following what's going on here. Maybe we should talk about it?

eddierubeiz:
Closing in favor of #2715 . Cleaner, more maintainable.

PR #2708, opened 2024-08-19
update device_detector gem

jrochkind:
For making sense of user-agent strings. When updating all deps to prepare for Rails 7.2, I noticed that updating device_detector from 1.1.2 to 1.1.3 caused our test to fail... it's for weird reasons not even worth going into, particular string we were using for 'bad' user agent was triggering a new detector, doens't really matter, just fixed to green.


jrochkind:
@eddierubeiz self-merging this straightforward one

Issue #2712, opened 2024-08-21
List asset_ids of problem assets when files are blocked from publication due to e.g. zero-length files.


eddierubeiz:
The error message does not include the list of asset_ids that are causing the problem.

PR #2713, opened 2024-08-21
Include asset_ids in error message when preventing work publication due to incorrect or corrupt files

eddierubeiz:
Ref #2712
Does not list asset_ids when you're publishing more than one work via the cart form.


jrochkind:
My use case wasn't the cart from, it was just clicking "publish (and publish all child works)" on the Work admin page. Sorry for ot filing the ticket and being clear on that. 

Does this apply to both? Or is that a separate issue?

jrochkind:
In an ideal world the bad asset might be marked in red or something... but with multi-page pagination that might be non-trivial... and this is totally good enough first step, for something that we think prob wont' happen hardly ever!

jrochkind (approved):

PR #2715, opened 2024-08-22
Report search_inside_the_book to Google Analytics

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2606.
- We're using the `event_value` parameter, which we don't use anywhere else in the digital collections, to report the query string to GA. (This is a bit of an experiment.)
- 💡Since the query string can change at any time, the reporting function looks up its current value using a selector that we provide in attribute `data-analytics-value-css`.
- Does not affect any existing reporting, as getAttribute will just return null if  `data-analytics-value-css` isn't specified.


app/views/scihist_image_viewer/_scihist_viewer_search_area.html.erb
```diff
@@ -27,7 +26,20 @@
 
 
       <div class="input-group-append">
-        <button type="submit" class="btn btn-brand-dark" title="Search" id="search-submit-header">
+        <%#
+          data-analytics-label is set to @work&.friendlier_id instead of @work.friendlier_id .
+          This is because spec/components/work_image_show_component_spec.rb renders this partial,
+          via app/components/work_image_show_component.html.erb, with a nil value for @work.
```
jrochkind, line 32: Should we just fix the spec to pass in a work instead of allowing the mistake in real code to not have a work, if a work is always required to get good analytics among other things?

app/frontend/javascript/custom_google_analytics_4_events.js
```diff
@@ -8,21 +8,39 @@ $(document).on('click', '*[data-analytics-category]', function(e) {
...
 
+    // In certain cases we want to set event_value to the contents of a text input - one that
+    // might be different every time the link or button is clicked.
+
+    // Example use case: we want to set event_value to the query string in search-within-the-work search form.
+
+    // If we find a selector in data-analytics-value-css, we can look up the current value of the text input 
+    // using that selector, and then pass that string to GA as event_value.
+    var elementToLookUp =  e.currentTarget.getAttribute("data-analytics-value-css");
+    var eventValue = elementToLookUp ?  $(elementToLookUp)[0].value.replace(/[^a-zA-Z 0-9]+/g, '') : null;
```
jrochkind, line 23: I guess we do want to capture the query ideally!  Although this wasn't part of the spec (are there privacy implications? Should we check with @Gabz73 ?)

Using a field pointing to a CSS selector is clever, I guess it works!

I feel like there was a reason I didn't use the GA `value` field before, like it was unsuitable for such things.. but also there's a whole new GA version we're using since then, let me look up the docs of GA "value" just to verify... I sort of feel like `value` had  to be an integer at least in old GA....

I'm having trouble finding docs. And the new GA is very different than the old GA!

But I'm wondering if we'd want to use a different mechanism in GA to capture queries, if we wanted to capture queries. Rather than metadata on a "event", there's a whole different mechanism I think...  but I'm having trouble finding that too, GA docs are terrible!

I wonder if we watn to bail on tracking queries unless that's been specifically asked for by stakeholders?  Alternatively, have you actually verified this works to capture search terms *in a way that can be usably accessed in GA console*??

jrochkind: It actually looks to me like maybe we can put _arbitrary_ metadata on the event in GA4 now, when before we were limited? So that is positive for this working, altough I still want to see it verified in GA Search Console that this new value can be accessed usefully, before committing code to record it!

And if we can call the metadata lable anything we want, should we actually call it "search_terms" or something?

Not totally sure about any of this!

https://developers.google.com/tag-platform/gtagjs/reference#event

jrochkind:
@eddierubeiz this has failing test

eddierubeiz:
Yup - hopefully it will pass once we have 2723 merged into it.

eddierubeiz:
Yup, it's passing now.

app/frontend/javascript/custom_google_analytics_4_events.js
```diff
@@ -8,21 +8,44 @@ $(document).on('click', '*[data-analytics-category]', function(e) {
...
-      // param eventName
+      // Only used to send a search phrase (or, more frequently, null).
+      'event_value': eventValue
+    }
+
+    // Based on https://support.google.com/analytics/answer/13675006?hl=en, let's try a more user-friendly format as well:
+    if (eventValue) {
+      data_to_send['search_phrase'] = eventValue;
+      data_to_send['friendlier_id'] = e.currentTarget.getAttribute("data-analytics-label");
+    }
```
jrochkind, line 40: OK, I know I was sort of suggesting this, but i was suggesting changing the `data-` attribute to something specific like "search phrase" too. 

It actually doesn't work to just assume that `data-analytics-value` is ALWAYS a search phrase, like undocumented API that when we say "send a value" we *really* mean "send a search phrase", and will record it in our analytics as such. 

And same for assuming that `data-analytics-label` *always* means "friendlier-id".  If that were true, we should just make the html data attribute (which is up to us), `data-analytics-friendlier-id`, right?

Are you following me? We need to think of those `data-` attributes as the "API" to our "record in GA" routine, and the API needs to be transparent and legible.  

We can talk about this more next week?

jrochkind:
In addition to our back-and-forth about what the data-attribute API shoudl be and how we should record it in GA4 -- i dont' want to merge this until we have actually _verified_ that it's getting recorded in GA in a way that we can usefullly access in reports and consoles etc. And that whoever is accessing GA reports knows how to get to. This stuff isn't actually straightforward.  

If we write code taht's trying to do something, but isnt' actually doing it in a useful way and none of the GA-reporting staff uses, and we don't realize it, it will just confuse us down the line. 

(We also still have the option of NOT recording thigns that our GA-reporting-staff didn't actually ask for?  Then we don't need to do the work to make sure it's being recorded in a way they can access usefully?)

It's possible you have already done this testing, but we just need to docuent it's been done here, for traceabilty and record-keeping!

eddierubeiz:
We've decided we don't need to record the search phrase in Google Analytics after all, at least for now.

Issue #2717, opened 2024-08-23
Bug in JS code that submits click metadata to Google Analytics


eddierubeiz:
In `app/frontend/javascript/custom_google_analytics_4_events.js` we attempt to refer to the link the user clicked with `e.target`.

In cases where the user clicks a _child_ of the link (e.g. a `span` or `i` or `img` element), the code attempts to look up the Google Analytics metadata on `e.target`, but it's not there, so we submit null metadata.

I can reliably reproduce this bug in dev by zooming in and clicking on the grey text listing the size of the downloads here:
![Screenshot 2024-08-23 at 10 05 39 AM](https://github.com/user-attachments/assets/b00dd5cb-7d68-4402-9ad7-1f57936d2e5c)

PR #2718, opened 2024-08-23
Fixing google analytics bug

eddierubeiz:
Ref #2717
Using `event.currentTarget` rather than `this` although https://api.jquery.com/event.currentTarget/ says they are equivalent!
At any rate, fixes the bug.


jrochkind:
thanks!

Issue #2722, opened 2024-08-26
Allow references to the work being searched in `app/views/scihist_image_viewer/_scihist_viewer_search_area.html.erb`


eddierubeiz:
Background:
`app/components/work_image_show_component.html.erb`
calls
`app/views/scihist_image_viewer/_scihist_viewer_modal.html.erb`
which in turn calls
`app/views/scihist_image_viewer/_scihist_viewer_search_area.html.erb`

Problem:
In #2715, we are going to need a reference to the work in `_scihist_viewer_search_area.html.erb`, 
but `_scihist_viewer_modal.html.erb` is not passing it in as an argument.

PR #2723, opened 2024-08-26
Pass `_scihist_viewer_search_area.html.erb` a reference to the work being shown.

eddierubeiz:
Ref #2722
app/views/scihist_image_viewer/_scihist_viewer_search_area.html.erb needs a reference to the work, so pass it in as an argument.
We need this so we can report search-within-the-work events to Google Analytics.
See #2715


jrochkind:
Huh, looks like it was good we tracked this down instead of editing the code to allow null and skip work reporting, yeah??

jrochkind:
So in recent rails like we have, we actually could edit the partial with a "magic comment" saying it requires the local. maybe makes sense here?

https://masilotti.com/safer-rails-partials-with-strict-locals/

    <%# locals: (work:) %>

jrochkind (approved):

PR #2726, opened 2024-08-27
Remove extra finding aids

### LINKED ISSUE #2627: add finding aid links to all archives department works without one

eddierubeiz:
Last task before we close #2627 .

### DIFF
--- a/app/services/extra_finding_aid_remover.rb
+++ b/app/services/extra_finding_aid_remover.rb
@@ -1,0 +1,51 @@
+class ExtraFindingAidRemover
+  attr_reader :work
+
+  def initialize(work, dry_run:false)
+    @work = work
+    @dry_run = dry_run
+    @notes = []
+  end
+
+  def process
+    return unless current_related_links.present?
+    return unless current_related_links.map { |rl| rl.category }.include? 'finding_aid'
+    return if current_related_links == new_related_links
+
+    @work.update!(related_link: new_related_links) unless @dry_run
+
+    @notes << "#{ @work.friendlier_id.ljust(10) } before: #{current_related_links.to_a}"
+    @notes << "#{ @work.friendlier_id.ljust(10) } after:  #{new_related_links.to_a}"
+  end
+
+
+  def notes
+    @notes
+  end
+
+  private
+
+  def current_related_links
+    @current_related_links ||= Set.new(@work.related_link)
+  end
+
+  def new_related_links
+    @new_related_links ||= Set.new(
+      current_related_links.reject do |rl|
+        rl.category == "finding_aid" &&
+        finding_aid_urls_from_other_sources.include?(rl.url)
+      end
+    )
+  end
+
+  def other_models_to_check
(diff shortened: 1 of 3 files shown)


jrochkind:
Looks good! This is only going to be used by the data_fixes we probably wont' ever run again?

At first I wondered if we could keep it in data_fixes somehow... but that migth make it hard to test....

Can you at least put a comment at top of file and test saying they are only for use by data_fixes task mention it by name, link to this PR, and say in comment these files can be deleted once data fix task is run?

And add a checkbox to either this PR or the Issue (forget which we do, or both!) for running the data fixes task, with exact copy-and-paste command line to run? 

Thanks!

eddierubeiz:
Added comments - good call.
We'll remove the test, but let's keep the service class and rake task around. I think it's very likely we will end up with more of these stray finding-aid links in the future, so it'll be nice to have this tool to clean them up every couple of years.

jrochkind:
Don't remove the test if you're leaving the code it tests! Ok then.

jrochkind (approved):

eddierubeiz:
Yeah, I do think we will want to rerun this rake task. Somebody is going to add a duplicate finding-aid link somewhere.

eddierubeiz:
[Results of running this in production](https://github.com/user-attachments/files/16873991/production_results_from_removing_extra_finding_aids.txt)

Issue #2721, opened 2024-08-26
flakey Rspec test for displaying OH ToC


jrochkind:
This test fails flakily, such that re-running tests with no code change can cause it to pass again. 

It's doing it enough to be annoying. 

```
  1) Oral history with audio display With OHMS synchronized transcript and ToC can display, and search, without errors
     Failure/Error: expect(page).to have_content(%r{Table of Contents — 1 / 7}i)
```

./spec/system/oral_history_with_audio_spec.rb:429:

PR #2728, opened 2024-08-28
Fix flaky Github CI test

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2721 .
The corrected test runs reliably in CI; I ran it fifty times and it didn't fail.


jrochkind:
it annoys my sensibilities to have to do this hacky thing, but I don't understand any other possible solution!

I think it just needs to wait for the Javascript event handlers to actually be attached to handle the click! Not sure what can be done about that. 

Thanks for figuring this out! Will be super nice not to have it failing!

Issue #2626, opened 2024-05-22
Fail more quietly when an on-ingest checksum validation fails in certain cases.


eddierubeiz:
**Scenario**:

- User ingests an asset
- A number of characterization and ingest jobs get enqueued
- The one that takes the most time is the checksum calculation
- User deletes the asset right after ingesting it

**Error**:
We get an error of the form:
```
NoMethodError: undefined method `id' for an instance of SingleAssetCheckerJob Rails.logger.error("#{job.class.name} (#{job.id}): 
Cancelling job due to ActiveJob::DeserializationError: #{error.message}") ^^^
```

Interestingly, the error is due to the SingleAssetCheckerJob not having a method `id`.
Let's rewrite this line so that it doesn't actually cause a 500. It doesn't want to! It's just trying to quietly write to the logger.
`single_asset_checker_job.rb  16  `

This is an infrequent bug -- shows up about once a month -- but it would be nice to address it.

https://app.honeybadger.io/projects/58989/faults/107139971 is the relevant HB link.

PR #2730, opened 2024-08-29
Deal better with logging to rails log after ActiveJob::DeserializationError

eddierubeiz:
This took two tries, but it's actually a simple problem - a `ApplicationJob` does not have an `id` but it does have a `job_id`. (Putting `id` in the error message was throwing a NoMethodError.)
Ref #2626


jrochkind:
Thanks!

Not having the job id in the error line is unfortunate, and won't allow the error line to tell us much, or be findable when we're looking for what hapepend with a given job id. 

If we're going ot do this, maybe at least log "No job id available" or something, instead of just a blank space more easily missed?

eddierubeiz:
Good call - done!

PR #2727, opened 2024-08-27
Change ocr_requested boolean attr_json to text_extraction_mode

jrochkind:
Part of #2492

text_extraction_mode can be `ocr` (former ocr_requested true), `pdf_extraction` (for extracting text from born-digital pdf), or nil (for no text extraction).

- [ ] Run data migration rake task `heroku run rake scihist:data_fixes:ocr_requested_migration -r production`

We will leave this in draft until we have verified this appraoch is convenient for implementation of PDF extraction

This also changes admin UI on work form for setting OCR, so need to run past staff.

PR #2732, opened 2024-08-29
Procedures and metadata for expanding a PDF into asset-per-page in a Work

jrochkind:
Part of #2492 . Built on TOP of #2727, merge that one first -- once we're ready. 

This PR is kind of sprawling, sorry!  Hard to keep it all separate at some point, this is kind of the big sprawling one. I will try to lay it out:

- add Asset role work_source_pdf
  * To be used for the Asset holding the PDF that will have pages extracted
  * update hint for "role" entry on Asset, in a kind fo hacky way that uses html with a <details> collapsible

- Button on Asset Admin page, for a PDF asset only, that will like "Setup work from this PDF source"
  * It will set the "work_source_pdf" role on this asset; and set the Work to `extract_text_mode` `pdf_extract`
  * And enqueue bg jobs, one-per-page, to make an Asset for the page, with graphic and hocr, and role set to "extracted_pdf_page"
  * Some validation also included, will refuse to do it if it seems like the work is set up in a conflicting way

- WorkImageShowComponent won't use 'work_source_pdf' Assets in it's list or as rep poster/thumbnail
   * That "work_source_pdf" PDF Asset?  should NOT be listed in the list of thumbnails on the public work page, as it's not actually a page and that was designed for pages. So suppress it from there, and from being shown as a "representative", which should also be set to a page Asset. 

- include exiftool pdf page_count in standard metadata
   * We already had a system for "raising up" exiftool stuff to easily accessible metadata, previously used for dpi. We need a pdf page count a lot iwth this code, so do the same with that. 

- there was a compoennt WorkShowOcrComponent, which was the "OCR" tab on Work admin page
  * It was meant to kind of veirfy everything was present for OCR, but our new PDF-extracted things were confusing it. 
  * So we expand it to be able to verify things are basically as expected for OCR *and* pdf-extracted, and rename it `WorkAdminTextExtractInfoComponent`
  * The tab on the work page now could say "OCR" or "PDF Text", it's a bit confusing. We'll have to demo it and try it out. 

- change admin text_extraction mode input to radio buttons instead of select 
  * Oops, that was meant to be in #2727, but my git got confused, and it's challenging to disentable now


jrochkind:
Sorry @eddierubeiz , this is the big one!  This is the next review task, then there will be more (smaller, easier to review I hope!) on top of this one!

eddierubeiz (approved):

PR #2735, opened 2024-09-03
report search_inside_book to Google Analytics

eddierubeiz:
Ref #2606 
We're not yet sure how to report the search phrase data to Google Analytics, so for now let's just report the fact that the search happened.

### DIFF
--- a/app/components/work_image_show_component.html.erb
+++ b/app/components/work_image_show_component.html.erb
@@ -39,12 +39,17 @@
                 </label>
               </div>
 
-              <%= search_field_tag :q, '', class: "q form-control",
+              <%= search_field_tag :q, '', class: "q form-control search-inside-query-field",
                   id: "search-inside-q"
               %>
 
               <div class="input-group-append">
-                <button type="submit" class="btn btn-brand-main" aria-label="Submit search" title="Search" id="search-submit-header">
+                <%# This GA action is also reported from app/views/scihist_image_viewer/_scihist_viewer_search_area.html.erb %>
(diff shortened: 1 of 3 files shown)


jrochkind:
Love the simplicity, I like that we've split this into two PRs!  But I suspect we do need some additional JS code after all -- if you enter text in the search bar but then just hit "enter" instead of actually pressing the button, it still submits the form -- and I think _won't_ trigger the JS here?

(I think that's probably a more common way to submit the search)

I think what might be a better approach -- if we put the data- attributes on the form itself, and modify the JS code to ALSO listen for form submit events, I _think_ it will catch form submissions regardless of whether they were triggered which way. (Should be tested!)

eddierubeiz:
Hitting "enter" in the form does actually cause the same code to run.

This is because the browser actually triggers a click event on the button _on your behalf_. See e.g. [the HTML5 spec](https://html.spec.whatwg.org/#implicit-submission).

`If the user agent supports letting the user submit a form implicitly (for example, on some platforms hitting the "enter" key while a text control is focused implicitly submits the form), then doing so for a form, whose default button has activation behavior and is not disabled, must cause the user agent to fire a click event at that default button.`

I interpret this as:

If the browser supports letting the user submit a form implicitly [e.g. by hitting enter] [...], then doing so for a form [...] must cause the browser to fire a click event at [the] default [submit] button [of that form].

jrochkind:
Oh nice, ok then!

If it were me, I'd test this in a few browsers just to make sure!

eddierubeiz:
Tested!

eddierubeiz:
Going to self-merge this - it's actually trivial in this PR

PR #2738, opened 2024-09-03
forgot to batch solr updates on ocr_requested data migration, needed for performance to actually execute

jrochkind:


jrochkind:
Gah, also use our custom util `find_each` to try to prevent memory leak; we were getting killed by heroku for using too much memory trying to run the rake task!

jrochkind:
self-merging! Trying to get production shaped up

PR #2743, opened 2024-09-05
Override Rails IRBConsole so console prompt in staging says staging

jrochkind:
even though our staging uses "production" env, we want it to show up differently in console, so prompt can serve it's cautionary function to distinguish production

Maybe. Asked question in Rails forum at https://github.com/rails/rails/discussions/52809


jrochkind:
There's a better way to do this suggested in Rails forum, let me try that.

jrochkind:
Actually I don't think this is appreciably worse than the way suggested in rails forum in the end. Let's try this.

Issue #2450, opened 2023-11-30
stop using ransack


jrochkind:
it's a pain, we don't really need it. 

The one place where it's being convnenient is in generating our clickable headers on columns in search results, to change sort order. 

So that would be a bit annoying to replace BUT

1. We can replace ALL OTHER users of ransack right now
2. We really CAN replace the sort stuff too we jjst have to write some annoying code.

PR #2744, opened 2024-09-11
Remove Ransack from collection model; use new component instead for table header links.

eddierubeiz:
Ref #2450

**Background**
Ransack did two things for us:
- It took care of *some* aspects of searching, filtering and sorting.
- It automatically kept track of the state of sorted tables and made it easy to create sortable column header links.

This PR separates these two tasks. The first is done in the controller; the second, inside a new component class.

More specifically:

**1) Creating the list of collections to show:**
Searching, filtering, sorting and pagination happen entirely in the `#index` method, using sane and legible code.

**2) Table header links:**
- Default values for params are set in a new `index_params` method, which returns a permitted strongparams object.
- This strongparams object is used to create a factory object named `:link_maker`.
- `:link_maker` is instantiated in a `before_action` call, so it doesn't crowd the `#index` method.
- `:link_maker` stores and encapsulates the current state of the sorted table.
- In the template, `:link_maker` is then used to generate several `SortedTableHeaderLinkComponent` objects, one per sortable column, which are then rendered to display the actual links.

**Tradeoffs**
Since we intend to use this component in at least three (and likely more) admin controllers, this PR aims to simplify the model, controller and view, at the expense of slightly complicating the component.

- The component, even with its internal class, is actually pretty simple considering what it does, and certainly much easier to understand than Ransack.
- Likewise, If you _ignore_ the component and its test, and *just* look at the modifications this PR makes to the model, controller and view: the improvement is obvious:
   - No mention of Ransack in the model
   - For sorting and filtering, we're using tried-and-true ActiveRecord patterns instead of magical and nested param names
   - Dev works the same way as prod.
   - Search now works in dev (see note below)
   - Sorting and filtering are decoupled from generating links
   - The code in the template is short and self-explanatory

**Note**: The `#index` method was not using permitted params, so we created an `index_params` method to return sanitized params that `#index` can use. These are then handed directly to the component code, along with keys that tell the component which keys important for what it's trying to do.

**Note**: Ransack search is _currently_ broken in dev on the master branch.
This is one of the many reasons we want to get rid of it, but it's easy to forget about.
(Try going to `http://localhost:3000/admin/works` in dev on `master` and searching for `Goat`. You get an 
`ArgumentError` in `Admin::WorksController#index` with error message `Invalid search term q`
unless you comment out 
	`c.ignore_unknown_conditions = Rails.env.production?`
in `ransack.rb`.

**What this means**
If you are testing this PR in dev, and note that the search is broken on the admin works page, that's not because this PR broke it. It's because it is broken in `master` and this PR does not modify the admin works page.


app/controllers/admin/collections_controller.rb
```diff
@@ -8,14 +8,12 @@ def index
...
+    scope = Collection
+    params[:sort_field] = "title" unless ['title', 'created_at', 'updated_at'].include? params[:sort_field]
+    params[:sort_order] = "asc"   unless ['asc', 'desc'].include?                       params[:sort_order]
 
-    # NOTE WELL: To use ransack, all attributes we want ransack to search or sort
-    # on NEED TO be listed in Colletion.ransackable_attributes and/or Collection.ransackable_associations
-    @q = Collection.ransack(params[:q]).tap do |ransack|
-      ransack.sorts = 'title asc' if ransack.sorts.empty?
-    end
+    scope = scope.order(Arel.sql("#{params[:sort_field]} #{params[:sort_order]}"))
```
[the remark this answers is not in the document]
jrochkind: that might not make sense though... at the moment it's just the controller that knows these things and passes them to component... maybe just two constants in the controller? Would be an even easier way of making it so there's only one place that knows we use `sort_field` and `sort_order`.

app/controllers/admin/collections_controller.rb
```diff
@@ -30,9 +26,23 @@ def index
       @department =  params[:department]
     end
 
+    scope = scope.order(Arel.sql("#{params[:sort_field]} #{params[:sort_order]}"))
```
jrochkind, line 29: What if the `link_maker` handles this, so the knowledge about what the param keys are is in ONE place?

    scope = link_maker.add_order(scope)

app/controllers/admin/collections_controller.rb
```diff
@@ -30,9 +26,23 @@ def index
...
+  # an object that can *return* a component.
+  # This code does NOT perform any actual sorting or filtering.
+  def link_maker
+    @link_maker ||= SortedTableHeaderLinkComponent.link_maker(
+      params: index_params,
+      table_sort_field_key: :sort_field,
+      table_sort_order_key: :sort_order,
+    )
+  end
+  helper_method :link_maker
```
jrochkind, line 44: This is specifically creating _sort_ links, so let's rename it to make that clear. `sort_link_maker`.  Or with it's possible repsonsibility for adding order to scope too, maybe `sort_link_manager` and class could also be `LinkManager` instead of maker?  Just thoughts.

app/components/sorted_table_header_link_component.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def call
+    link_to sort_column_title, sort_column_path
+  end
+
+  def sort_column_path
+    url_for ( {
+        sort_field:   @sort_field,
+        sort_order:   (reversed_sort_order if @sort_field == @table_sort_field)
+      }.merge!(@other_params)
```
jrochkind, line 65: `sort_field` and `sort_order` as param keys are hard-coded here. Can we make them arguments to the initializer instead, so one page could call it a different thing like `search_order` and `search_direction` or whatever ?  You can still have them have defaults, optional args.

jrochkind: Also, use rails `reverse_merge!` or change the order of what's merged into what, so if @other_params has conflicting `sort_field` and `sort_order` keys, our new intentional ones are not overwritten.

jrochkind:
This is a great approach, good work!

Some comments!

app/controllers/admin/collections_controller.rb
```diff
@@ -8,14 +9,9 @@ def index
...
-
-    # NOTE WELL: To use ransack, all attributes we want ransack to search or sort
-    # on NEED TO be listed in Colletion.ransackable_attributes and/or Collection.ransackable_associations
-    @q = Collection.ransack(params[:q]).tap do |ransack|
-      ransack.sorts = 'title asc' if ransack.sorts.empty?
-    end
-
-    scope = @q.result
+    # Searching, filtering, sorting and pagination.
+    params.delete("button")
```
jrochkind, line 13: I don't think you need to delete this from params for any reason? just an unnecessary line?

app/components/sorted_table_header_link_component.rb
jrochkind: The line `(reversed_sort_order if @sort_field == @table_sort_field)` is confusing to me, although I understand what the intention is. 

Your'e leaving out `sort_order` value UNLESS it's reversed, otherwise it's blank? Is this advisable/what ransack did?  If it does have a blank value and that's intentional, let's make sure it doesn't have an unneeded key in the url either with a blank value?

I think it might be more straightforward to put something in the constructor or even the link_maker that decides what the direction shoudl be and passes is as an arg or stores it in a variable called `@direction` or `@order_value` or something, that's present whether or not it's reversed?  I'm just finding this logic confusing, although it very well may be correct!

app/controllers/admin/collections_controller.rb
jrochkind: what do you think, nah?

app/components/sorted_table_header_link_component.rb
```diff
@@ -0,0 +1,139 @@
...
+    if @sort_field == @table_sort_field
+      # IF the column whose title we want to display is the *same*
+      # as the column the table is currently sorted by,
+      # THEN, IF the user clicks on this title,
+      #  we reverse the direction of the sort.
+      url_param_hash[@table_sort_order_key] = reversed_sort_order
+    else
+      # For columns OTHER than the one the table is currently sorted by,
+      # clicking the title of the column will sort the table by that column,
+      # in an arbitrary order.
```
jrochkind, line 85: I _think_ I understand, but what do you mean by "arbitrary"? WE only have two kinds, asc and desc, don't we know for sure which one we are doing? With "asc" being the default?

app/controllers/admin/collections_controller.rb
```diff
@@ -118,6 +120,18 @@ def set_collection
       @collection = Collection.find_by_friendlier_id!(params[:id])
     end
 
+    # Params method just for #index.
+    # Also sets default sort column and order.
+    # This does not actually perform any sorting or filtering.
+    def index_params
+      @index_params ||= Kithe::Parameters.new(params).permit(
```
jrochkind, line 127: You don't need to use Kithe::Parameters here, let's just use ordinary rails strong params. https://api.rubyonrails.org/classes/ActionController/StrongParameters.html

Which means this can just be `params.permit...`

app/controllers/admin/collections_controller.rb
jrochkind: Oh this is also a security problem, passing `params` directly to `Arel.sql` like that opens up to SQL injection. Either need to use proper routinet o escape, or I think maybe you can use this to do it for you? `order(params[:sort_field] => params[:sort_order])`

jrochkind (approved):

Issue #2685, opened 2024-07-17
update searchstax solr to latest 8.x


jrochkind:
We are on 8.6.3, SearchStax has the latest 8.11.2. 

Solr 9.x is also out and offered by SearchStax, but we're not ready for that yet (may or may not break Blacklight version we are on)

But let's go to latest 8.x for sure. 

May require creating new SearchStax instance and deleting old one? Not sure if SearchStax lets you update in place somehow. Needs investigation.

PR #2759, opened 2024-09-30
Solr 8.11.2

eddierubeiz:
So the tests pass against 8.11, and it seems to work fine in staging.
I suppose it's fine to merge this, @jrochkind - let me know if you can think of any reason not to :)
Ref #2685

### DIFF
--- a/.solr_wrapper.yml
+++ b/.solr_wrapper.yml
@@ -22,4 +22,4 @@ collection:
   persist: <%= Rails.env.test? ? "false" : "true" %>
   dir: solr/config
   name: scihist_digicoll_<%= Rails.env %>
-version: 8.6.3
+version: 8.11.2


eddierubeiz:
Tests pass; no problems using 8.11 in staging.

jrochkind:
How do we coordinate the production deploy... if we merge this, does it effect production? Maybe not, this only effects development and CI maybe? Are we still concerned about having this in develpoment before we have it in production?

Where are you documenting your production rollout process, and when it's complete?

eddierubeiz:
Based on my experience with staging, it's OK to have an `8.11` version of `solr_yml.rb` file in production even though the searchstax deployment production is pointing to is still `8.6`.
So the checklist would go like this (thanks to @jrochkind for suggestions)
 - [ ] Merge and deploy this PR to production
 - [ ] Override current `SOLR_URL` and deploy configset to new deployment
 - [ ] Override current `SOLR_URL` and index to new deployment
 - [ ] Change `SOLR_URL` permanently
 - [ ] Index a second time

jrochkind (approved):

Issue #2762, opened 2024-10-10
Sort  not working on Works page in Admin


Gabz73:
The "Created" and "Last Modified" sorts on the Works page in the Admin dashboard )https://digital.sciencehistory.org/admin/works) are not working: when you click on the header, the sort doesn't change.

eddierubeiz:
I very likely broke this - stay tuned.

jrochkind:
I guess add a test that would fail on the broken code and pass on the fix, please, @eddierubeiz !

eddierubeiz:
Yeah. I'm considering it high priority.

PR #2763, opened 2024-10-10
Fix sort on admin work page.

eddierubeiz:
Ref #2762
Classic blunder - should have used [order!](https://apidock.com/rails/v7.1.3.2/ActiveRecord/QueryMethods/order%21) instead of [order](https://apidock.com/rails/ActiveRecord/QueryMethods/order).


jrochkind:
Sorry if this is obvious and i'm being annoying, but important to have actually verified the test failed before fix, when adding a regression test. Sometimes you think you've written a good test, but it turns out it was green before fix too!

eddierubeiz:
Yup! The test fails with `order` and passes with `order!`.

PR #2769, opened 2024-10-25
Microsoft SSO ( keywords Entra / Azure / OmniAuth / Oauth )

eddierubeiz:
Ref #2564
We want to make sure we get @Gabz73 's blessing before merging it.
General documentation is [in the wiki](https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/1915748368/Heroku+Operational+Components+Overview#Microsoft-SSO); feel free to append or emend.

### Changes to routes
Because we may need to redirect users to the logout page, we replaced the old `delete 'logout'` route with a `get 'logout'`.
If Microsoft SSO is turned on:
 - we override Devise's `passwords` controller to make it unusable (see below)
 - we implicitly define some Microsoft SSO routes by editing `devise_for :users`
 - we turn of the `get 'login'` and `post 'login'` routes, which displayed and processed the login form, respectively
   - Note: `user_entra_id_omniauth_authorize_path` requires a `POST` request for [security reasons](https://github.com/cookpad/omniauth-rails_csrf_prot), so we won't be pointing `/login` to it in this PR.
 - the new logout will go to `auth#sso_logout`, which logs the user out and then sends them to the Microsoft logout page.

### Changes to `env.rb` and`devise`
- A feature switch, `:log_in_using_microsoft_sso`, turns the entire feature on and off;
  - If you set it to `true`, you will also need values set for these three new `env` values which are needed to configure the new auth provider in the `devise.rb` initializer:
    - `:microsoft_sso_client_id` identifies the app to Microsoft SSO;
    - `:microsoft_sso_client_secret` authenticates the app to Microsoft SSO;
    - `:microsoft_sso_tenant_id` identifies the Microsoft SSO directory the app wants to check (namely the Institute one. This ID is the same for dev, staging and prod.);
- We are not making any changes to the User class other than adding the above auth provider.

### Controllers
 - A new controller, `controllers/auth_controller.rb`
  - handles the interaction with entra (via two non-private methods: `entra_id` and `logout`.
    - `entra_id` is the callback method where users land after being successfully authenticated 
- We subclass `Devise::PasswordsController` to make it unusable if `ScihistDigicoll::Env.lookup(:log_in_using_microsoft_sso)` is true.
- We add a method to `ApplicationController` so Devise knows what to do with users who just logged in. (since you can now log in from any page in the app, we want you to go back to wherever you were before after successfully logging in).

### Other changes
- We add a login / logout button on the footer.
- Two new gems, `omniauth-entra-id` and `omniauth-rails_csrf_protection`;
- We change some comment verbiage in `scihist.rake`
- A new test file `spec/requests/auth_controller_spec.rb` which tests the new auth controller. It includes a bunch of tests that used to live in `spec/system/login_spec.rb` but were slow and flaky, and adds more.
- Another new test file, `spec/requests/passwords_spec.rb`, contains some extra smoke tests, just to make sure password workflows are still functioning when `:log_in_using_microsoft_sso` is _not_ set .
It should help guard against regression bugs that could creep into our *old* auth workflows while we are using the new ones. If and when we need to turn off Microsoft log in temporarily in the future for whatever reason, we have a bit more confidence that it will it works the way it's supposed to.
- The only thing left in `login_spec.rb` is a test to show that post-login redirect works.

### Notes on logout
Currently, clicking on any of the logout buttons:
- logs you out of the app
- if SSO is turned on, subsequently redirects you to the Microsoft SSO page.
- From the SSO page, you are supposed to get redirected to the root path of the app. It's Microsoft, so your mileage may vary.


app/controllers/application_controller.rb
```diff
@@ -18,13 +18,7 @@ class ApplicationController < ActionController::Base
...
   rescue_from "AccessGranted::AccessDenied" do |exception|
-    redirect_path = if current_user.present?
-      root_path
-    else
-      new_user_session_path
-    end
-
-    redirect_to redirect_path, alert: "You don't have permission to access requested page."
+    redirect_to root_path, alert: "You don't have permission to access that page."
   end
```
jrochkind, line 22: Before, If they tried to access a page when they weren't logged in that would require a login, it would redirect them to a login screen. 

I think this was useful. The idea was you might bookmark a partiuclar admin page you need to use frequenty or whatever. 

I guess it was changed here because we can't actually easily redirect to login anymore?  I guess we could have a landing page with a login button on it (`/login` could go there too), but I guess it's not _that_ useful unless after logging in you actually return to where you were trying to go in the first place, which is kind of a pain to keep track of through a process that stops on a "login button landing page". 

I dunno, what do you think, worth thinking about, or it's fine like above?  Ask Gabriela?

app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,63 @@
+class AuthController < Devise::OmniauthCallbacksController
```
jrochkind, line 1: Can you add a comment at the top explaining what this is for, and mentioning the ENV var that controls it?

config/initializers/devise.rb
```diff
@@ -260,10 +260,50 @@
...
+
+      # "Client" or "Application" is what Microsoft calls a website that uses Microsoft SSO.
+      # This will have a different value for dev; staging; and production.
+      client_id:     ScihistDigicoll::Env.lookup(:microsoft_sso_client_id),
+
+      # This is effectively a password - a shared secret.
+      client_secret: ScihistDigicoll::Env.lookup(:microsoft_sso_client_secret),
+
+    }
+  )
```
jrochkind, line 306: Shoudl/could the config.omniauth block also be inside the guard, only if `ScihistDigicoll::Env.lookup(:log_in_using_microsoft_sso)`?

(Along with the `ready_to_configure_microsoft_sso` definition too why not).

config/initializers/devise.rb
```diff
@@ -260,10 +260,50 @@
...
-  # Add a new OmniAuth provider. Check the wiki for more information on setting
-  # up on your models and hooks.
-  # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
+
+  # MICROSOFT SINGLE SIGN ON / SSO / ENTRA / AZURE / OAUTH
+  # More details about how we use Microsoft SSO (aka Entra, aka Azure) are in the wiki.
+  #
+  # See also https://github.com/sciencehistory/scihist_digicoll/pull/2769
+  # See also config/initializers/devise.rb
+  # See also the wiki
```
jrochkind, line 269: Add wiki url?

config/routes.rb
```diff
@@ -53,11 +53,24 @@ def self.matches?(request)
...
+  if ScihistDigicoll::Env.lookup(:log_in_using_microsoft_sso)
+    devise_scope :user do
+      get 'logout', to:'auth#sso_logout'
+    end
+  else
+    devise_scope :user do
+      get  'login',  to: 'devise/sessions#new',     as: :new_user_session
+      post 'login',  to: 'devise/sessions#create',  as: :user_session
+      get  'logout', to: 'devise/sessions#destroy'
+    end
```
jrochkind, line 73: love it

spec/requests/auth_controller_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+
+  context "using Microsoft SSO to log in" do
+
+    # An authenticated email. This email address belongs to a person who has gotten authenticated.
+    let(:incoming_email) { 'the_user@sciencehistory.org' }
+  
+    before do
+      allow(ScihistDigicoll::Env).to receive(:lookup).and_call_original
+      allow(ScihistDigicoll::Env).to receive(:lookup).with(:log_in_using_microsoft_sso).and_return(true)
+      Rails.application.reload_routes!
```
jrochkind, line 19: Thsi will end up doing the reload_routes before every example -- is it worth seeing how much this slows down our tests? Maybe not, as it's a pain to do it any way other than this reliably!

jrochkind:
Looks pretty great overall, thanks!

A few comments, some of them are questions for discusisons, I'm not sure if they need to be done or not just bringing it up to see what you think. 

Can you link to wiki page to make it easier for me to find and review? Thanks!

app/controllers/application_controller.rb
eddierubeiz: We did discuss this today briefly and decided to leave this as is and make changes, if any, in separate PRs after this has been in production for a bit.

config/initializers/devise.rb
eddierubeiz: Done!

config/initializers/devise.rb
```diff
@@ -260,10 +260,50 @@
...
+  # See also https://portal.azure.com/
+  # See also our password store.
```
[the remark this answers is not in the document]
eddierubeiz: Oh, I meant 1Password. I clarified the comment.

spec/requests/auth_controller_spec.rb
eddierubeiz: It's fast enough - roughly 10 ms every time you reload the routes. I did add a comment to that effect.

config/initializers/devise.rb
eddierubeiz: I moved the `ready_to_configure_microsoft_sso` inside the guard (good idea!) but, unfortunately you can't reload your devise configuration on the fly. [The docs](https://github.com/heartcombo/devise) say:
"You should restart your application after changing Devise's configuration options (this includes stopping spring). Otherwise, you will run into strange errors, for example, users being unable to login and route helpers being undefined.
In practice, this means we have to have the `config.omniauth` block outside the guard.

jrochkind: No problem, it seems okay to configure that then -- but leave a comment explaining why it needs to be configured (so it's there for testing!)

app/controllers/auth_controller.rb
eddierubeiz: Done!

eddierubeiz:
Wiki notes are at:
https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/1915748368/Heroku+Operational+Components+Overview#Microsoft-SSO .

jrochkind:
Wiki looks good I think, but what do you think about putting it on it's own wiki page, and link to it from the heroku page?

It's kind of lengthy content, that could get lengthier if we need more notes, and isn't really about our heroku deployment. 

I also wonder if you want to note that Entra registrations are tied to hostnames, so if we add/change a hostname they need to be edited? 

Also -- can you add back to README the rake tasks for creating users with passwords, that we'll want in development environments?

eddierubeiz:
- Moved the wiki docs to its own file in the wiki, and linked to that; makes sense.
- I added a shortened version of the docs in the wiki to a new README file:
https://github.com/sciencehistory/scihist_digicoll/blob/7db236d8011cc41879c2083f2ea01f119ae37b5b/config/initializers/MICROSOFT_SSO_README.md
- Added your note about Entra registrations to the wiki.

jrochkind (approved):

PR #2776, opened 2024-11-07
update rspec-rails to 7.x and bundle update rspec-rails and dependencies

jrochkind:


jrochkind:
But temporarily avoid updating to rackup 2.2.0 to avoid for now having to figure out some dependency issues we're not looking to deal with right now. 

https://github.com/rack/rackup/issues/26

PR #2794, opened 2024-12-03
Remove bootstrap gem from gemfile

jrochkind:
We now use bootstrap 5 via npm with vite, not from the rubygem. The rubygem is unneeded.

Remove duplicate sprockets/sprockets-rails in gemfile, only sprockets-rails, higher in gemfile. Having them both with sprockets first was effecting load order in a weird way that prevented startup, in the absence of sassc-rails which had been a bootstrap gem indirect dependency.


jrochkind:
After merging and updating locally, you may need to run `rm -rf tmp/cache`.

jrochkind:
approved by eddie verbally

PR #2795, opened 2024-12-03
fix featured topic controller #facet action for blacklight 8

jrochkind:
Parallel to #2791 for collection controller, must do the same for featured topic controller. No longer need #facet action method override, which had become out of date in BL8."

### DIFF
--- a/app/controllers/featured_topic_controller.rb
+++ b/app/controllers/featured_topic_controller.rb
@@ -2,27 +2,6 @@
 class FeaturedTopicController < CatalogController
   before_action :set_featured_topic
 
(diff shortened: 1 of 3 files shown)


jrochkind:
Ugh we have a new flakey test, that was the one I added for #2791!

```
  1) CatalogController general smoke test with lots of features loads
     Failure/Error: expect(page).to have_selector("h1.modal-title", text: "Subject")
       expected to find css "h1.modal-title" but there were no matches
```

For this high-priority fix, I'm just going to hit rerun job in Github CI.

jrochkind:
okay self merging to fix bug.

Issue #2789, opened 2024-11-27
Eliminate special code to work around bad range limit data


jrochkind:
Some of it may no longer be needed in latest blacklight_range_limit; others, if identified, we can add logic to blacklight_range_limit so it's no longer needed. Should be handled up stream.

PR #2797, opened 2024-12-04
remove local blacklight_range_limit bad-param catching to rely on upstream in plugin

jrochkind:
After we upstreamed bad-param protection in blacklight_range_limit, we can remove our custom local protections. 

Ref #2789 

The goal is to prevent any URL given to the app from resulting in an uncaught exception being raised -- bad input should be handled with an intentional recovery or error message, not an uncaught exception, so we don't get an alert from our exception monitoring service. 

In some cases the upstream protection results in a different result after this PR than before (changed test expectations), but the important thing is it still doesn't result in an exception. Almost all of these cases are bots trying weird URLs, they aren't going to work, we just don't want an exception. 

* what before was a 4xx (bad input error) might now be a 200 where bad input is just ignored. (or vice versa). 
* in controller tests, sometimes we test that an `ActionController::BadRequest` is raised, but that will be caught by Rails in production and turned into an http 400, it won't be an uncaught exception. 
* In some cases the bad input will result in StrongParams "unpermitted" -- in production, we just ignore these, but in dev we do raise on them to catch bugs. For those tests, we temporarily make it match production where they are ignored, so we can test properly. 

---

This will remain in draft until we can at least depend on `blacklight_range_limit` *main* branch, instead of an unmerged PR.  (wait for https://github.com/projectblacklight/blacklight_range_limit/pull/305 to merge). 

But it's a beta version anyway, we're okay referring to unreleased main until next beta or release maybe?

---

- update to blacklight_range_limit 9.0.0.beta2
- temporarily update to not yet merged PR in blacklight_range_limit
- remove all local blacklight_range_limit bad-param catching to rely on upstream in plugin

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -67,7 +67,7 @@ gem 'scout_apm'
 # version may require an update with yarn to `blacklight-frontend`, has to be
 # checked manually.
 gem "blacklight", "~> 8.7.0"
-gem "blacklight_range_limit", "9.0.0.beta2", github: "projectblacklight/blacklight_range_limit", branch: "bad_range_limit_params" # version no longer sync'd with blacklight, not sure how we tell what version works with what version of BL
+gem "blacklight_range_limit", "9.0.0.beta2"
 
 # for some code to deal with transcoding video, via AWS MediaConvert
 # Lower than 1.2.1 had far too big gem builds! https://github.com/samvera-labs/active_encode/issues/126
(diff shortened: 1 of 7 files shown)


jrochkind:
@eddierubeiz you approved an earlier version of this. I just self-reviewed the diff to be sure it seemed good, and am going to merge! Hooray.

PR #2799, opened 2024-12-04
upgrade blacklight to just released 8.7.0

jrochkind:
This release is compat with Rails 8, so starts setting us up for that upgrade.

Did not upgrade other related dependencies because some are causing problems, just strictly BL.

Edited Gemfile to require BL 8.7.0, then ran `bundle install`. Sync'd package.json to matching blacklight-frontend.


jrochkind:
approved by eddie verbally

PR #2803, opened 2024-12-05
try archiving capybara failure screenshots

jrochkind:
Experiment to get github to store the screenshots taken after failed browser tests by capybara-screenshot gem. 

Have to make sure it doesn't make the github CI pipeline report success on a failure, since we had to get an archive step to run even after failure, need to make sure it doesn't flip job to success!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,13 @@ jobs:
           run: |
             bundle exec rspec
 
+        - name: Archive capybara failure screenshots
+          uses: actions/upload-artifact@v4
+          if: failure()
+          with:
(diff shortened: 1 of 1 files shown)


jrochkind:
Success!  Created an intentnionally failing capybara test. 

On a capybara test failure, you can see a screenshot url/filename reported in github CI log:

Screenshot 2024-12-05 at 1 38 30 PM

Then in the "Archive capybara failure screenshots" step, there is a clickable URL reported. Clicking on it will download a zip file that includes all screenshot filenames mentioned in any failing tests. 

Screenshot 2024-12-05 at 1 37 56 PM

PR #2805, opened 2024-12-05
Don't call `role_portrait?` on child works

eddierubeiz:
Ref #2804
This is a guard against a weird edge case in which a child work is mistakenly created in an oral history.
Not adding any tests here.


app/components/oral_history/downloads_list_component.rb
```diff
@@ -33,6 +33,7 @@ def all_members
```
eddierubeiz, line 33: This change is a mistake - I'll fix it before I make this `ready for review`.

app/components/work_file_list_show_component.rb
```diff
@@ -30,7 +30,7 @@ def member_list_for_display
       end
 
       # omit "portrait" role
-      members = members.find_all {|m| ! m.role_portrait? }
+      members = members.find_all {|m| !(m.asset? && m.role_portrait?) }
```
jrochkind, line 33: A slightly more fluent way to ssay this might be `!(m.respond_to?(:role_portrait?) && m.role_portrait?)

OR, rails `try` can do this for you -- just `!m.try(:role_portrait?)`, if there is no method it will return `nil` which is false, `!nil` is true, good. 

Same for below.

jrochkind:
Thanks for this! Minor style comment, then approve. 

I think I agree we don't need a test here, but it's an option to consider?

Then feel free to merge!

eddierubeiz:
I definitely considered adding tests; if we want to, we can add a request spec in which we add a child work to an oral history and visit all the OH pages to make sure we don't get any errors. But the scenario is unlikely to reoccur, so I'm not that worried about it.

PR #2806, opened 2024-12-05
try to make copy to clipboard test more reliable

jrochkind:
In some cases the 'copy to clipboard' button is ending up off-screen when capybara tries to click it. I think this is because bootstrap accordion is doing some crazy animation when we automatically open/close the different elements. We can try to work around it by actually scrolling the browser to right position?

a) This may be an actual problem for users of course that we should try to solve, I have kind of noticed it

b) I'm not sure if our previous attempts to improve reliability with the rescue right here are necessary working or not? But leaving them in too!

### DIFF
--- a/app/frontend/javascript/audio/helpers/ohms_player_helpers.js
+++ b/app/frontend/javascript/audio/helpers/ohms_player_helpers.js
@@ -21,6 +21,7 @@ As we interact with Bootstrap 4 Javascript in here, it does include some JQuery 
 to be available. We try to avoid JQuery except for that.
  */
 
+import * as bootstrap from 'bootstrap';
 
 // Does not switch to tab, assumes ToC tab is open.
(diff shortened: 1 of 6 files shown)


jrochkind:
OK, still got a failure of the very test we are trying to solve. 

```
     Failure/Error: copy_button.click
         (Session info: chrome=131.0.6778.85)
```

And the screenshot reveals that it really is off screen, hidden by our fixed nav. 

Screenshot 2024-12-05 at 1 38 30 PM

Try to scroll in a rescue too? Maybe it was still in process of moving when we scrolled?

(Can/should we TURN OFF bootstrap animations in tests?)

jrochkind:
Same result, same screenshot.  

Wow, the screenshot feature at least let's me really feel my failure here!

jrochkind:
OK failed experiment, we're gonna start over elsewhere.

PR #2810, opened 2024-12-11
Switch to use new bootstrap 5 non-jQuery API

jrochkind:
In Bootstrap 4, the JS API for interacting with tabs and collapsible elements was a jQuery plugin. 

In Bootstrap 5, the jQuery plugin _ought_ to still be supported if you have jQuery installed. But we're going to switch to the new straight-JS API here. This is a head start on likely future switch to not use jQuery, but the main reason we are doing it is in hopes that it will help with some of our super mysterious flakey JS that seems to be around code for showing tabs and collapsibles.

### DIFF
--- a/app/frontend/javascript/audio/helpers/ohms_player_helpers.js
+++ b/app/frontend/javascript/audio/helpers/ohms_player_helpers.js
@@ -21,6 +21,7 @@ As we interact with Bootstrap 4 Javascript in here, it does include some JQuery 
 to be available. We try to avoid JQuery except for that.
  */
 
+import * as bootstrap from 'bootstrap';
 
 // Does not switch to tab, assumes ToC tab is open.
 export function gotoTocSegmentAtTimecode(timeinSeconds) {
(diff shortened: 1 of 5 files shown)


jrochkind:
Definitely hasn't totally solved it, we just got one of the more common ones:

````
     Failure/Error: page.find(copy_to_clipboard).click

     # ./spec/system/oral_history_with_audio_spec.rb:423:in `block (3 levels) in <main>'
```

And now we have a screenshot image....

![failures_r_spec_example_groups_oral_history_with_audio_display_with_ohms_synchronized_transcript_and_to_c_can_display__and_search__without_errors_516](https://github.com/user-attachments/assets/cf69f3f8-f320-4003-b91a-19a0b6172630)

OK the was not scrolled far enough to actually put the "copy to clipboard" link on screen. Hmm.

eddierubeiz (approved):

jrochkind:
OK, you actually have to click "Share link" to open up the section with "copy to cliipboard". 

Our capybara script didn't include that!  And mostly worked anyway, because the OHMS Search code that opens up a ToC section with a result was sloppy, and could open up too much or the wrong thing....  still not sure why test was then flakey instead of always failing, but trying to make everything more precise.

jrochkind:
OK I think this was not related to the problem and does not fix it, but as long as we're here we might as well make the change.

eddierubeiz (approved):

PR #2816, opened 2024-12-12
Upgrade to Rails 8

jrochkind:
- rails 8, with updated samvera dependencies that support rails 8
- Rails8 was pickier about that bad formatted date, we want to set it to nil, we should say that
- patch devise for Rails 8
- update to Rails 8.0 framework defaults
- Remove require that is no longer required in Rails 8 to patch IRB how we like

upstream merged but unreleased devise patch to work with Rails 8: https://github.com/heartcombo/devise/pull/5728

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ ruby "~> #{File.read(File.join(__dir__ , '.ruby-version')).chomp.split('.').slic
 gem 'lockbox'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 7.2.0'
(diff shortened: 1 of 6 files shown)


jrochkind:
This looks like a small PR, but it's because of lots of work finding problems and separating solutions into previous PR's that already went in, that would work in both Rails 7 and 8.

jrochkind:
Note a patch to Devise to make it work with Rails 8 before they release a version with the fix!

jrochkind:
@eddierubeiz I'm not going to merge until after the new year, but avail for review whenever!

eddierubeiz (approved):

PR #2817, opened 2024-12-12
Remove yarn resolution to bootstrap4 no longer needed or wanted because we are using bootstrap4

jrochkind:
This may have caused both bootstrap4 and 5 to be loaded in our app? Lucky it didn't cause terrible problems if it didn't!

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,6 @@
     "vite-plugin-ruby": "^5.1.0",
     "vite-plugin-sass-glob-import": "^3.0.2"
   },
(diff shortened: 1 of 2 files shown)


jrochkind:
oops prob got to commit changed yarn.lock after `yarn install` too.

jrochkind:
had to do a bit of manual editing of yarn.lock to get rid of the "extra" bootstrap, annoying.

eddierubeiz (approved):

PR #2818, opened 2024-12-12
On the collection show page, allow users to sort search results by box and folder.

eddierubeiz:
We can change the sort later if we want, but I'd like to have it in place so we can start using it and see how we like it.

- This sort is only available on "search within a collection" search results.
- We use the title of the work as the tiebreaker.
- For now, if a box or folder doesn't have a number in its label, we show it at the end of the search results.
-  Works within the same non-integer box or folder are sorted together.
- Works with non-numeric titles go at the end, but still stick together collated by same box/folder, sorted alphabetically by title.

### DIFF
--- a/Aptfile
+++ b/Aptfile
@@ -14,6 +14,9 @@ libglib2.0-0
 libglib2.0-dev
 libpoppler-glib8
 
+# to get vips CLI tools we need
+libvips-tools
+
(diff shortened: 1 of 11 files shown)


eddierubeiz:
I think this works the way you'd like. I added a couple tests to make absolutely sure we are on the same page.

PR #2819, opened 2024-12-12
Avoid gap in viewer footer for child works with very large windows

jrochkind:
This CSS has gotten a bit hack on top of hack, but good enough for now.

## Before
Screenshot 2024-12-12 at 2 56 59 PM

## After

Screenshot 2024-12-12 at 4 20 56 PM


jrochkind:
Oy as long as we're at it, that all uppercase title of a child work seems wrong to me, undid that too.

PR #2822, opened 2024-12-16
Show box and folder metadata on search results page.

eddierubeiz:
This PR subclasses `WorkComponent` (which is itself a subclass of `BaseComponent`) with a new subclass: `SearchWithinCollectionWorkComponent`.

* The subclass is only used in `CollectionShowController`.
* The subclass defines a new method, `box_and_folder`
   * `box_and_folder` returns `nil` unless the model is a Work and has `department` set to 'Archives'.
   * Otherwise, `box_and_folder` returns a string listing the box and folder.
     * On missing or unusual box and folder metadata, the component displays what it can without errors.


jrochkind:
Looks good. Can we find a way to do this only on Collection search result pages, not every search result? (Or maybe you already have one?). Not totally sure how to appraoch this, may need to figure out how to change architecture to support. 

Also make sure the line doesn't show up if there is no box/folder values (including just empty strings).

eddierubeiz:
This only shows on Collection search results, since the subclass is only used in `CollectionShowController`.

eddierubeiz:
Will include tests that prove the above assertions!

eddierubeiz:
Added tests.

jrochkind:
Nicely done. 

I don't LOVE sub-classing WorkComponent, as sub-classes can be fragile ("prefer composition over inheritance"). 

In an ideal world, I'd like to add this to WorkComponent, as a configuration option in initializer on WorkComponent (`show_box_and_folder` or something like that), and somehow have CollectionController (and its sub-classes) set up the WorkComponent with that config. 

I think that could possibly result in ultimately simpler code... this is a bit spaghetti?  But that might also be just spagetti too. 

I guess we have tests that should catch us if we break?  So maybe we'll just go with this for now!

You did do a great job of figuring out how to fit this into existing architecture, which wasn't _too_ bad, so I guess we should count our blessings!

Approved if you feel good with moving forward even after these comments!

jrochkind (approved):

eddierubeiz:
That's a good idea and I don't believe it's a huge change. Hang on, let me see if I can do that...

eddierubeiz:
Sorry - that approach turns out to be tricky because `CollectionShowController` doesn't actually _instantiate_ any `WorkComponent` objects, it just defines a _class_ to use as a view component given a type of `model` that will later be displayed displayed. I think we _do_ have to subclass `WorkComponent` here.

I could be wrong of course, in which case I might refactor this later; we're not in a rush.

PR #2824, opened 2024-12-16
fix method call for race condition recovery in OnDemandDerivativeCreator

jrochkind:
Ref #2821


jrochkind:
oops, I guess we hadn't quite fixed this yet! I dragged it to "closed" on Monday. 

Thanks for noticing and approving, we have now!

PR #2838, opened 2025-01-07
Bot Detection with Turnstile

jrochkind:
Making everything configurable and architecting in a way that should be not too rough to extract into a gem if desired.  All of the actual logic is really just in the BotDetectController, including methods that are called in other places to "install" it in routes, and init rate limit tracking with rack-attack. 

See more in wiki at https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/2645098498/Cloudflare+Turnstile+bot+detection

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,6 +94,8 @@ gem "resque-pool"
 gem "resque-heroku-signals" # gah, weirdly needed for graceful shutdown on heroku. https://github.com/resque/resque#heroku
 
 
+gem "http", "~> 5.2" # for http client access
+
 # using memcached for Rails.cache in production, requires dalli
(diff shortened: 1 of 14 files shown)


jrochkind:
A challenge success means the checkmark goes green and you get automatically redirected to your destination -- with the challenge page removed from your browser history (hooray). It usually takes maybe 5 seconds?

A challenge failure can look like one of these two things, depending on at what point it failed. Most users shouldn't see challenge failures, unless they are identified as bots somehow, or something has gone wrong (perhaps a bug). 

Screenshot 2025-01-13 at 3 01 28 PM
Screenshot 2025-01-13 at 3 09 57 PM

eddierubeiz (approved):

PR #2839, opened 2025-01-10
Box and folder: use multivalued field; leave tokenization to solr. REQUIRES REINDEX.

eddierubeiz:
I was using a single-valued field for box and folder, and doing the tokenization myself. Instead, let's use a multivalued field and leave the tokenization to SOLR. All the tests still pass.

@jrochkind pointed out the problem - thanks!

- [ ] merge
- [ ] deploy
- [ ] reindex


eddierubeiz:
Oops - this was actually more changes than we needed. Rescinding this PR :)

Issue #2658, opened 2024-06-17
Broaden the scope of the ID search in the admin pages


eddierubeiz:
On these pages:
- https://digital.sciencehistory.org/admin
- https://digital.sciencehistory.org/admin/collections
filtering on an ID should match on any values of `external_id`: bib record ID, accession number, etc.

jrochkind:
@eddierubeiz Is this one actually done? We never put it in review and reported it, I don't think

eddierubeiz:
Yes, this is done - you had a small suggestion which I implemented. Glad you noticed it!

PR #2844, opened 2025-01-14
Admin collection list page: allow users to filter collections by arbitrary external_id

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2658

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -12,22 +12,19 @@ class Admin::CollectionsController < AdminController
     # Searching, filtering, sorting and pagination.
     scope = Collection
     if index_params[:title_or_id].present?
-      sanitized_search_phrase = Collection.sanitize_sql_like(params[:title_or_id])
-      # This expression will match any element of the external_id jsonb array,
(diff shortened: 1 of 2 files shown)


app/controllers/admin/collections_controller.rb
```diff
@@ -12,11 +12,22 @@ def index
...
       scope = scope.where(id: index_params[:title_or_id]
       ).or(
-        Collection.where(friendlier_id: index_params[:title_or_id])
+        scope.where(friendlier_id: index_params[:title_or_id])
+      ).or(
+        scope.where("title ilike ?", "%" + sanitized_search_phrase + "%")
       ).or(
-        Collection.where("title ilike ?", "%" + Collection.sanitize_sql_like(index_params[:title_or_id]) + "%")
+        scope.where(matches_external_id)
       )
```
jrochkind, line 31: I _think_ this actually should be `Collection.where` instead of `scope.where`?  They may be equivalent right now (or not), but as we add things onto scope, they could creep into the `or` clauses, when each or clauses meant to be independent and start from `Collection` I think?

I could be missing something!

eddierubeiz: You're right - thanks for catching this!

PR #2849, opened 2025-01-16
log of throttles should not log bot detection too

jrochkind:
In the event subscription meant to log our throttle monitoring, look at match_name to make sure we're not triggered by our new bot detection stuff. That has it's own logs elsewhere, these logs are meant only for overall traffic monitoring


jrochkind:
a real error, my bad! Thank you tests!

PR #2854, opened 2025-01-17
More robust system spec for Turnstile

eddierubeiz:
EDITED - all this PR does is wait up to 4 seconds for Turnstile to finish.

### DIFF
--- a/spec/system/bot_limiting_system_spec.rb
+++ b/spec/system/bot_limiting_system_spec.rb
@@ -82,11 +82,7 @@ describe "Turnstile bot limiting", js:true do
       expect(shows_turnstile).to be true
 
       # which eventually will redirect back to search.
-      begin
-        expect(page).to have_content turnstile_success_re
-      rescue RSpec::Expectations::ExpectationNotMetError
-        expect(page).to have_content turnstile_success_re
-      end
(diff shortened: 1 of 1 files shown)


jrochkind:
Not totally following what this is doing why; is this doing anything different than just adding longer explicit `wait` to the first version, without the abstraction?

eddierubeiz:
The only substantial difference from just waiting longer is that we catch the exception and try the `expect` twice if needed.

jrochkind:
I don't think trying the expect twice at 50% time has any difference whatsoever to trying it once at 100% time. 

Perhaps we just need to bump up the wait time for the last expect?  Although it's still not totally clear why or if there's something to fix to make it faster!

eddierubeiz:
Looks like you're right. I ran my test again with drastically simpler code, *just* to make sure, and the only significant change the PR makes is lengthening the time we wait for Turnstile. The rest was just my own bias - tough to admit but the numbers don't lie. :)

jrochkind:
Oh I had it misnamed copy-paste "cart and batch edit"?!?! Well THAT was confusing me too about what was failing! Thanks for catching and fixing that too.

jrochkind:
A rebase to one commit wouldn't hurt for things like this in the future, but I'm gonna merge anyway right now!

PR #2857, opened 2025-01-21
bot detect stores remote_ip to avoid session replay attack on bot pass

jrochkind:
improvement to #2838


jrochkind:
Tested in staging to be extra careful, seems good.

PR #2860, opened 2025-01-22
reduce number of searches that can be done per/ip before bot challenge to 2

jrochkind:
We are having capacity issues that we haven't quite figured out, let's see if this helps.


jrochkind:
prob ought to make it ENV configurable...

Issue #2491, opened 2024-01-11
Add Neville Catalog PDFs to Digital Collections


apinkney0696:
Here is the background information on the Neville Catalog PDFs:

The Neville Catalog project was completed before Jim started. The highest-quality PDF files of the catalog are located here:

P:\Othmer Library\Neville Collection\1Neville Catalog, Vol. 1
P:\Othmer Library\Neville Collection\2Neville Catalog, Vol.2

These are the files that were used to create the print books. There are several PDF files that make up each Volume. There are merged versions of the Volumes floating around the P drive (P:\Othmer Library\Neville Collection, Neville Vol 1 Combined.pdf or Neville Vol 2 Combined.pdf). I assume that ingesting the merged files to the DC will be easier, but I don't know how much file quality was lost in their merge.

In the physical books, there are two inserts with color images. These inserts are not in the PDFs. Jim doesn't feel strongly about making sure these are included in the DC, so I think noting their existence in the print copies in the metadata is a sufficient way to handle their absence. 

I've created a Digitization Queue item for the Volumes here: https://digital.sciencehistory.org/admin/digitization_queue_items/1135

jrochkind:
This is super helpful, thanks @apinkney0696 !

To be clear, if we have the high-quality PDFs from which the books were printed -- there's probably no reason to spend time photographing the printed books, we can get what we need from the PDFs instead! 

So the existence of the digitzation queue item probably shouldn't be taken to mean we are intending someone to photograph all hte pages!  (Maybe the two inserts, if we don't have digital originals for them anymore?)

We can instead spend time looking at the various PDFs and merged files etc, analyzing what's there, and create new merged files if needed, etc.

jrochkind:
And also to be clear for the record of this ticket -- we will need to do some UI/feature work to make PDFs other than Oral Histories representable in a useful way in Digital Collections! It shouldn't be too hard, I think we have the pieces in place, but will need to assemble them in a new way.

jrochkind:
The borderless white PDFs look strange to me for all page thumbnails, although worst for the blown-up large title page, yeah. 

We might want to add a border (perhaps grey instead of black) only for pages extracted from PDFs using that new system. We could add them directly to thumbnails at the point we create the image captures -- I think we are doing that for OralHistory PDF thumbs at present. Or we could try to add them in CSS, but trickier to only apply to certain "extracted PDF" pages that way.

jrochkind:
Our initial stop gap:

[...]

PR #2864, opened 2025-01-23
Add a CSS brand light grey border around all thumbnails on work page

eddierubeiz:
Ref #2491

"Try putting a CSS brand light grey border around all thumbnails on work page, see how that looks for Neville and for everything else."

This looks OK, but it would be nice if we can do it without modifying the displayed size of the thumbnail image.

### DIFF
--- a/app/frontend/stylesheets/local/search_results.scss
+++ b/app/frontend/stylesheets/local/search_results.scss
@@ -206,7 +206,6 @@
       margin-right: 28px;
       flex-shrink: 0;
       text-align: right;
-      border: 1px solid $table-border-color;
 
       // Keep really high ones from taking up too much space
       max-height: $standard-thumb-width * 2.2;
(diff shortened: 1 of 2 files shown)


jrochkind:
Looks like we still need it on search results page, yeah?

Screenshot 2025-01-28 at 1 23 11 PM

I'm also curious to try it with just 1px instead of 2. I _can_ see it on ordinary non-neville pages, it makes things a bit messy. I wonder if 1px would still do what we need for neville-type, but be even more invisible otherwise?

eddierubeiz:
Sure, why not! Done.

eddierubeiz:
Oh, sorry about that. Fixed (and deployed in staging).

eddierubeiz:
Can we do this without reducing the displayed size of the image?

Looking at how this PR changes the display size of the image in Chrome developer tools:

- **without** the border, the image size is `208x269`
- **with** the border, in staging, the image size is `206x267`.

(The aspect ratio doesn’t change, nor does the “intrinsic size” of the image at 416x538).

PR #2865, opened 2025-01-23
OHMS new-style XML WebVTT parsing and transcript display

jrochkind:
New OHMS transcripts have (a certain subset of) WebVTT in a `vtt_transcript` tag 

1. New class to parse WebVTT https://www.w3.org/TR/webvtt1/
2. OhmsXML class can recognize presence of either legacy or new vtt transcript, and parse either one
3. Indexer can get indexable transcript_text from either one to index to solr

There's more coming for actual display of VTT transcripts, but this is (more than) enough for first PR.

### DIFF
--- a/app/models/oral_history_content/ohms_xml/vtt_transcript.rb
+++ b/app/models/oral_history_content/ohms_xml/vtt_transcript.rb
@@ -1,0 +1,109 @@
+require 'webvtt'
+
+class OralHistoryContent
+  class OhmsXml
+
+    # Can parse WebVTT content -- or at least the subset delivered by the OHMS
+    # new-style 2025 OHMS transcript with <vtt_transcript> element in xml.
+    # https://www.w3.org/TR/webvtt1/
+    #
+    # Uses the `webvtt` gem for initial parsing, but that gem is basic and
+    # not very maintained, so we need some massaging and post-processing
+    # to get what we need.
+    #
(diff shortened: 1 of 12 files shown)


app/models/oral_history_content/ohms_xml/vtt_transcript.rb
```diff
@@ -0,0 +1,132 @@
...
+            ($1.to_i * 60) + $2.to_i + $3.to_f
+          end
+        end
+
+        def paragraphs
+          @paragraphs ||= begin
+            # This tricky regex using both positive lookahead and negative lookahead
+            # will split into voice tags, taking into account that some text might not
+            # be in a voice tag, and that voice tag does not have to ber closed when it's the whole cue
+            (text || -"").split(/(?=\<v[ .])|(?<=\<\/v>)/).collect do |voice_span|
```
eddierubeiz, line 95: -"" looks like a typo to me!

eddierubeiz (changes requested):
Looks great - thanks for the tests. Found what I think is a typo.

app/models/oral_history_content/ohms_xml/vtt_transcript.rb
eddierubeiz: Ah, just shorthand for "".freeze .

eddierubeiz (approved):

Issue #905, opened 2020-10-28
For books with many pages, don't show all pages to the user at first.


eddierubeiz:
In our efforts to see if our app could work on Heroku, one persistent obstacle has been making sure the works that have over a hundred pages don't load too slowly.
While this a very small proportion of our overall works, these works tend to be high-prestige books, [Ramelli](https://digital.sciencehistory.org/works/4b29b614k) being the prime example. At the moment, on Heroku, these works take quite a long time to load compared to on our current setup.
This ticket covers the work of experimenting with changes to the user interface so when you first pull up a book like Ramelli, you do not see all the members (pages) in the browser immediately. This will make the page load much faster.
This is a complex question with accessibility implications.

jrochkind:
My measurements showed that 80 would be a nice break -- 80 children still displays relatively quickly, and there are a pretty small proportion of works that haev more than 80 children. (Although I can't seem to find my notes on the details there at the moment). 

In addition to work show page, another slow pages is the JSON data that powers the viewer, at a `viewer_images_info` route. This currently returns JSON for all children, and can also be fairly slow -- while the user might not notice as it happens in the bg, it can still be bogging down our app, with other requests waiting behind. It would be of benefit to split that up so the front-end JS requests the `viewer_images_info` JSON data in batches of 100 -- the user wouldn't be waiting any longer, but it would distribute between load balanced workers better/more evenly, instead of clogging up a request queue.

jrochkind:
The new Rails "turbo frames" JS library may be useful here. Here is a good overview: https://mailchi.mp/railsspeed/turbo-frames-drive-and-migrating-a-legacy-app?e=e9606cf04b

PR #2867, opened 2025-01-24
[WIP] Do not show all members at first

eddierubeiz:
Ref #2491 
The specification for this PR is:
_"Cap at say 400 page thumbs on work page, with a message "more pages available, not displayed", maybe even with a count of how many."_ (taken from [this comment](https://github.com/sciencehistory/scihist_digicoll/issues/2491#issuecomment-2468582828).) 
Do not merge, do not review. Contains known bugs!

### DIFF
--- a/app/components/work_image_show_component.html.erb
+++ b/app/components/work_image_show_component.html.erb
@@ -74,6 +74,13 @@
         <%= render MemberImageComponent.new(member_for_thumb.member, lazy: (index > 5), image_label: member_for_thumb.image_label) %>
       </div>
   <% end %>
+
+  <% if more_members_to_show? %>
+    <div><%= link_to "Show #{hidden_viewable_members_count} more #{pluralize(hidden_viewable_members_count, 'item')} ...", show_all: true %></div>
+  <% else %>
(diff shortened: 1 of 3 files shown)


eddierubeiz:
Pausing work on this for a bit while I investigate using Turbo to lazy-load the thumbnails instead, per [another comment](https://github.com/sciencehistory/scihist_digicoll/issues/905#issuecomment-1180480718).

jrochkind:
@eddierubeiz I don't think we have turbo/hotwire installed right now, it might be a whole additional overhead to figure that out. 

I'd just proceed with it without if you run into any trouble.  But nothing wrong with exploring turbo/hotwire, it is something we want to look at eventually. 

Note that the actual *image*  media is already lazy-loaded using HTML attributes.  What we want to lazy load is the list of  tags themselves etc.  I think we can do this wihtout too much trouble even without turbo -- the first step is just a link that you click that replace itself with more thumbnails after clikcing using JS/AJAX -- it'll be easy to turn that automatic after that.  But no problem exploring turbo if you feel called!

PR #2873, opened 2025-01-29
Display OHMS new VTT transcripts

jrochkind:
Handling OHMS VTT annotations will be a future PR. 

(This one may need more specs, considering)


jrochkind:
OK, I was thinking of adding system specs for instance "search and highlight term in transcript" for new-style transcript -- but realized we don't actually have that for old-style transcript at the moment either. 

I think at the moment I'll let it pass, instead of trying to add a challenging/slow/likely-to-be-flakey system test for both.

eddierubeiz (approved):

Issue #2878, opened 2025-02-04
Enhance custom OHMS viewer to handle new-style OHMS VTT transcripts


jrochkind:
As doc'd at https://coda.aviaryplatform.com/7-transcript-module-172 , example VTT transcripts under '7.2. Formatting OHMS transcripts'

jrochkind:
Done for now compatible with new transcript format and does what the default OHMS viewer does with them -- we should check in again with actual OH dept workflow on their timeline.

PR #2879, opened 2025-02-04
OHMS new VTT transcripts support footnotes

jrochkind:
The OHMS editor lets you put some limited HTML in footnotes/annotations, and the sample OHMS VTT transcripts they provided had HTML in footnotes/annotations. So I wrote this feature to support limited html, and wrote specs ensuring that. (Which actually complicated things quite q bit!)

But, when I was nearly done with this, i noticed that current OHMS doesn't actually support this feature!  When I try to actually do this in OHMS Studio, while the editor lets me add HTML, the preview strips it out. And the exported OHMS xml has it stripped out too!  

So we can't currently take advantage of it; but I completed and left in the feature to handle html in annotations/footnotes -- in case it's a bug that OHMS can't currently do them, and they'll fix it? Shouldn't hurt to leave it in. 

Ref #2878

- extract OHMS vtt annotations/footnotes
- move OHMS vtt html scrubbing from model to view_component
- render OHMS new-style vtt transcript annotation references
- style links included in ohms popover
- VttTranscriptComponent outputs footnotes at bottom
- fix spec now that whitespace is normalized again

### DIFF
--- a/app/components/oral_history/footnote_reference_component.html.erb
+++ b/app/components/oral_history/footnote_reference_component.html.erb
@@ -1,6 +1,7 @@
 <a  class="footnote"
     data-toggle="ohms-reference"
     data-bs-content="<%= footnote_text %>"
+    <% if footnote_text.html_safe? %> data-bs-html="true" <% end %>
     data-bs-custom-class="ohms-footnote-popover"
     data-bs-trigger="hover focus"
     aria-label="footnote <%= number %>"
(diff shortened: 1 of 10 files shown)


jrochkind:
Note the HTML sanitizing to strip out any but allowed tags and attributes -- I consider this always necessary when displaying HTML that's been saved as a string, even from a "trusted" source.

jrochkind:
OK there are some idiosyncracies with how OH dept is formatting their transcripts for input into OHMS and how OHMS handles some choices -- but they are between them and OHMS, not between OHMS and us. 

Our viewer handles everything similar or better to how default OHMS viewer does. 

So we can def merge this now (all the code paths only effect new-style transcripts), and then work more with OH at the point they are ready to work on really deciding what their formatting and workflow will be, with regard to things OHMS Studio will do that we have no control over, as well as things we have control over on our end.

eddierubeiz (approved):

PR #2883, opened 2025-02-06
white image edge detection, leads to images displayed with border

jrochkind:
Replacement for #2864, ref #2491

This was pretty easy? 

But it's a bit precious....   on the plus side, if it does break, it's probably just what people less picky would have had anyway? Took care of "recent items" too. 

On the negative side... it actually will make our test suite slower, cause any test doing a "real" ingest with metadata extraction will run this now...  our architecture doesn't give us a great way to turn it off in tests (and would we want to? not sure). 

- white edge detection with ImageMagick
- fix description to be more clear
- detect white edges and store in metadata on ingest
- iamge border around white-edge-detected images in ThumbComponent, with styling in three context

## After merge/deploy

Run rake task in production to detect white edges on Neville Catalog

- [ ] `heroku run rake scihist:data_fixes:detect_white_edge WORK_FRIENDLIER_ID=qbtvy9p -r production`


jrochkind:
oh we don't have `magick` on github? That's weird. Need to confirm we have it on heroku, and that it's included in our system env tests. 

Maybe we're still using deprecated `convert`? We can do that.

jrochkind:
OK, there's a known failure mode here, where an ALL white exclusively white image will not be detected as white-edge, because `--trim` doesn't trim on it. 

For now we're just going to postpone dealing with that -- we can manually set the metadata value on certain images if we need to.

jrochkind:
OK I'm gonna do border in viewer in a separate PR, so this is ready.

jrochkind:
OK, all white images now detected as white-edged too.

eddierubeiz (approved):

PR #2884, opened 2025-02-10
Fix yarn.lock to match change at 7cbaaf4

jrochkind:
Commit 7cbaaf4 in PR #2881

Forgot to commit yarn.lock with package.json change, that's a disaster, and too bad our CI allows it!


jrochkind:
@eddierubeiz self-merging to get repo in consistent state!

PR #2888, opened 2025-02-11
Tweak hard rate limits to restrict less

jrochkind:
- exempt 315 Chestnut IPs from hard rate limits
- restore hard rate limit to 180/min with logged track at 90, what it was before bot troubles. See 39fde487b13e

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -29,7 +29,7 @@ end
 #
 # May 1 2024: Limiting much more extensively to 30 req per minute -- one per every two seconds
 # averaging over a minute -- after bot  attacks costing us money from s3.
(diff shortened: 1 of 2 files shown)


eddierubeiz:
Hm -- I know the code runs only once at config time, but I still feel a bit nervous about the placement of that lookup `ScihistDigicoll::Env.lookup(:main_office_ips)`. Can you reassure me that the list of IPs isn't getting looked up on every request?

eddierubeiz (approved):

jrochkind:
Hm, good thought. It IS getting looked up on every request. 

i think it's probably fast enough that it doesn't matter, but I'll extract it anyway into only looked up on load, just in case, why not.

PR #2890, opened 2025-02-13
[WIP] lazy load work thumbnails, proof of concept

eddierubeiz:
Do not review, do not merge. A quick reorganization of the code based on Jonathan's suggestions.
Tests fail, there are bugs, and it's probably inefficient, but it does work.

### DIFF
--- a/app/frontend/javascript/lazy_member_images.js
+++ b/app/frontend/javascript/lazy_member_images.js
@@ -1,0 +1,50 @@
+document.querySelector(".lazy-member-images-link")?.addEventListener("click", getMoreMemberImages);
+
+
+function getMoreMemberImages(event) {
+  var friendlierId = getFriendlierId();
+  if (friendlierId !== null) {
+    var url = "http://localhost:3000/works/" + friendlierId + "/lazy_member_images?" +
+      "start_index=" + document.querySelector(".lazy-member-images-link").dataset.startIndex +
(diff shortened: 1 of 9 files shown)


app/frontend/javascript/lazy_member_images.js
```diff
@@ -0,0 +1,97 @@
...
+
+class LazyMemberImages {
+
+  // this link won't be necessary; we'll just call the method automatically
+  constructor() {
+     document.querySelector(".lazy-member-images-link")?.addEventListener("click", this.#getMoreMemberImages.bind(this));
+  }
+
+  // Retrieve the images and insert them
+  #getMoreMemberImages(event) {
```
jrochkind, line 24: I've never seen starting a js method with a `#` before.  Did you get this convention from somewhere, or does it mean something in JS I am not aware of?

jrochkind:
Hey Eddie, can you merge master into here? (Don't rebase, since I have another branch on top of this!)

That would make my branch diff less confusing on top of this, cause I merged master into it.

jrochkind:
I think the failure might be another race condition (also in `master`) that I have fixed in another PR. :(

jrochkind:
OK I think if we merge #2921 into master, then master into here again, we'll have green.

jrochkind:
Doing it!

PR #2897, opened 2025-02-20
Protect original downloads with turnstile bot challenge in modal

jrochkind:
Since originals download usually results in a download directly to device with no browser page change -- our other generic method of bot protecting by REDIRECTING to a bot challenge and then back to originally desired page -- does not work well. 

Instead, we want to put up a modal with turnstile bot challenge, and if it passes, remove the modal, and navigate to download URL. 

This is a pretty hacky implementation, that still has some rough edges. But maybe enough for now?

- probably ought to re-use more of the bot challenge page stuff, which we are trying to extract into gem. instead we do it again, differently -- also we need to bot check downloads on FIRST try, even allowing 2 free ones was potentially no good. 

- not very DRY or well organized in general

- If close modal in media res, need to cancel turnstile redirect to download

- need to make work in viewer for original downloads, right now it won't and sometimes you'll get blocked from download. 

- Right now, you get challenge on EVERY download, we really want to exempt you if you've already passed bot challenge recently, hopefully sharing mechanism and code from other bot challenge page redirect stuff (but race condition is a mess).  

-----

- client-side turnstile for orig downloads, in modal
- server-side for turnstile protection of orig downloads
- Do NOT block downloads using bot-challenge-page, we need to do it with modal instead
- exempt PDF originals from bot challenge protection
- fix to work for multiple links on page, with delegation
- config variable for turnstile protection of download links, requires actually generating data attr for protection on link
- controller specs for turnstile download orig protection
- test auto add of data-turnstile-protected=true on download option
- system spec for turnstile-protected orig downloads

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -38,6 +38,8 @@ class DownloadsController < ApplicationController
   before_action :set_asset
   before_action :set_derivative, only: :derivative
 
+  before_action :turnstile_protect, only: :original
+
 
   #GET /downloads/:asset_id
   def original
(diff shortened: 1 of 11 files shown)


jrochkind:
Nevermind, this got too messy, let's try much much simpler implementation -- with some UX trade-offs -- at #2900

PR #2900, opened 2025-02-20
Put a bot challenge in front of orig downloads

jrochkind:
On FIRST access -- and with original downloads opening in a separate tab, to provide a somewhat better UX... although it makes it a bit weird flash of tab when there is no challenge; and kind of confusing stuck looking at passed challenge when there is;  we can demo and think more about it, but seems good enough for now maybe, with a much simpler implementation than last try at #2897

- give BotDetectController.bot_detection_enforce_filter an 'immediate' option to ignore rate limit and apply bot challenge on first access
- use bot_detection_enforce_filter on DownloadsController#original with immediate option
- DownloadDropdown does target=_blank on original downloads, to accomodate bot challenge redirect
- remove Downloads from BotDetect standard config, we only use it with immediate filter

### DIFF
--- a/app/components/download_dropdown_component.rb
+++ b/app/components/download_dropdown_component.rb
@@ -239,10 +239,13 @@ class DownloadDropdownComponent < ApplicationComponent
     ])
 
     if download_option.url.present?
+      # download_original gets target=_blank becuase it's protected by bot protection,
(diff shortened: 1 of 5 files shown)


jrochkind:
@eddierubeiz I think I will self-merge this to get downloads back. 

But when you are back from lunch etc, I would love you to give it a once-over, questions welcome.

Issue #2910, opened 2025-02-28
CI test failure: `bot_limiting_system_spec.rb`


eddierubeiz:
As of 2/28/25,  `./spec/system/bot_limiting_system_spec.rb:57` is still failing in master with:

```
Selenium::WebDriver::Error::UnknownError:
       unknown error: unhandled inspector error: {"code":-[32](https://github.com/sciencehistory/scihist_digicoll/actions/runs/13590353338/job/37994694578#step:11:33)000,"message":"Node with given id does not belong to the document"}
```

Example test run:
https://github.com/sciencehistory/scihist_digicoll/actions/runs/13590353338/job/37994694578

jrochkind:
Gah, maybe not taken care of, we got this failure on master  still even after that merge. 

https://github.com/sciencehistory/scihist_digicoll/actions/runs/13661703449

PR #2912, opened 2025-03-03
replace local code with code extracted to BotChallengePage gem

jrochkind:
For turnstile protection of some things


jrochkind:
Huh, removed most of the tests, but the one that remains is now failing reliably, but can't get a reproduction locally!

>   1) Turnstile bot limiting succesful challenge smoke tests
     Failure/Error: expect(page).to have_content(/traffic control/i)
       expected to find text matching /traffic control/i in "MUSEUM HOURS\nDIGITAL 

We're trying to confirm it briefly redirects to the challenge page in a system test... could ti be redirecting back too quickly and missing it? Hard to say.

jrochkind:
OK have run three times all green, I think it's ready @eddierubeiz !  And I believe will make #2910  no longer relevant

eddierubeiz (approved):

PR #2920, opened 2025-03-12
Capybara specs should wait for page to be returned before checking for database change

jrochkind:
Otherwise, if the app is slower than capybara, when we're checking for a db change, maybe it hasn't been done yet. We actually have to wait for the app to return the next page to ensure it finished the action before checking for DB change.

Not really sure why this has started becoming a problem recently, hopefully we haven't slowed down our app. But these are corrections to do things actually reliably correct regardless.

### DIFF
--- a/spec/system/digitization_queue_spec.rb
+++ b/spec/system/digitization_queue_spec.rb
@@ -21,8 +21,8 @@ RSpec.describe "Digitization Queue", logged_in_user: :editor, type: :system, js:
     fill_in "Location", with: "Some location"
 
     click_on "Create Digitization queue item"
-    # return to listing page, now click on item we just made
-
+    # return to listing page, make sure we wait for it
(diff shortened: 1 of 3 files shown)


jrochkind:
OK this one is now reliably failing too -- in master apparently -- for no clear reason, investigating.  Not sure why we recently had so many system specs start failing for race conditions that hadn't before. Maybe new chrome/selenium/capybara versions.

jrochkind:
OK that was a variation of the same thing -- we clicked "sign out" but then the back-end app was slow at actually signing us out before we moved on to trying to test for it. Why'd it start happenign now? Who knows.

jrochkind:
oK I'm gonna self-merge to keep going, I just added additioanl capybara expectations that should already have been true, to force it to wait long enough. 

But @eddierubeiz  you might be curious to see what's up.

PR #2921, opened 2025-03-17
More Capyabara wait conditions

jrochkind:
More times Capybara was checking for some state before the code to execute it had completed. Have to put in stuff so it waits for a front-end UX signal indicating the processing has completed.

Similar to #2920 but these were mostly not reproducing on Github CI, although they were easily reproducible on my macbook.

Still not sure why this whole category started getting worse easily, what changed. Hopefully it wasn't our app getting slower.


jrochkind:
OK, some of the mystery has been solved, other people seem to be running into the problem too -- does NOT seem to be changes to our app, but changes to capybara, selenium, and/or chrome.  

See https://github.com/teamcapybara/capybara/issues/2800

Issue #2659, opened 2024-06-17
On all blacklight search results page, allow the user to add all works in the results to the user's cart.


eddierubeiz:
This should work efficiently, even for results returning thousands of items.
Collections in the search results can be ignored for the purposes of this PR.

eddierubeiz:
See also ref #2660, which discusses exporting large numbers of works to CSV.

Gabz73:
Adding GZ notes: add some / all results to cart from front-end search?

eddierubeiz:
Consider this ticket as a use case for this ticket: https://github.com/sciencehistory/scihist_digicoll/issues/2840

PR #2927, opened 2025-03-20
Add all the works returned by a blacklight search to the user's cart

eddierubeiz:
ref #2659.

We add a button to the search results page that goes to a new controller and method:

- All the works returned by the search (up to 10 million of them) will be added to the logged-in user's cart;
- Sort is preserved;
- Collections returned by the search are ignored, as are works already in the user's cart;
- We have a request test that asserts that the friendlier_ids are exactly the same as the ordinary search results. Not sure if we need more tests.
- The user is then forwarded back to the same results page.

For now, you can't do this on the collection search page, featured topics page, etc.

Does not load the actual items from the database; does not get any data from SOLR except the ids; does not instantiate the works at any point. Therefore, this is quite fast.


eddierubeiz:
Example: compare

Paginated results:

/catalog?search_field=all_fields&range%5Byear_facet_isim%5D%5Bbegin%5D=1950&range%5Byear_facet_isim%5D%5Bend%5D=2023&f%5Bsubject_facet%5D%5B%5D=Employees&f%5Bsubject_facet%5D%5B%5D=Industries&f%5Bcreator_facet%5D%5B%5D=Dow+Chemical+Company&f%5Bformat_facet%5D%5B%5D=Image&f%5Bdepartment_facet%5D%5B%5D=Archives&q=Midland

Just the IDs:

/all_search_result_ids?search_field=all_fields&range%5Byear_facet_isim%5D%5Bbegin%5D=1950&range%5Byear_facet_isim%5D%5Bend%5D=2023&f%5Bsubject_facet%5D%5B%5D=Employees&f%5Bsubject_facet%5D%5B%5D=Industries&f%5Bcreator_facet%5D%5B%5D=Dow+Chemical+Company&f%5Bformat_facet%5D%5B%5D=Image&f%5Bdepartment_facet%5D%5B%5D=Archives&q=Midland

jrochkind:
Nice start!

In our actual implementation, I think we'll definitely want to NOT send the IDs back to the front-end. There could of course be thousands of them -- there's no need to take that trip sending them to the frontend and then having the frontend send em BACK again with a "save to cart" directive. 

Instead the action `all_search_result_ids` can be changed to actually `save_search_to_cart` or something -- do the same fetch you are doing already (the hard part) then instead of returning hte IDs in json, actually do the work in the Rails action to save them to cart (a back-end DB operation) and then either just redirect the front-end to reload, or tell the front-end how many things are in cart now for front-end to update it's cart count without a reload. 

(the former just reloading the page seems easier to me and sufficient -- so maybe don't even make it a JS fetch, make it an actual form submit to the backend, which returns a redirect)

app/controllers/all_search_result_ids_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    config.search_state_fields << :all_search_result_ids
+  end
+
+
+  # A CartItem doesn't contain a lot of data, so this goes fast, even for thousands of ids
+  def add_to_cart
+    conn = ActiveRecord::Base.connection
+    CartItem.transaction do
+      all_ids_from_search.each_slice(500) do |ids|
+        conn.execute add_friendlier_ids_to_cart_sql(ids)
```
jrochkind, line 16: I still think there oughta be a way to do this with ActiveRecord insert_all and/or upsert_all that would be just as performant, and somewhat neater, without having to maintain our own SQL. 

But there's really nothing _wrong_ with doing it this way, and you have done it very tidily!  Not sure if you experimented with insert_all/upsert_all.    I wonder if we should try it -- are your tests good enough that if they pass we are confident? :). 

But also really nothing wrong with this way.

app/controllers/catalog_controller.rb
```diff
@@ -216,6 +216,22 @@ def active_sort_fields
...
+    # app/models/search_builder/all_search_result_ids_builder.rb
+    # app/controllers/all_search_result_ids_controller.rb
+    #
+    # Don't show this tool on e.g. the collection results page or the featured topics page.
+    config.add_results_collection_tool(
+      :all_search_results_add_to_cart,
+      opts = {
+        if: proc { |controller| controller.can?(:update, Kithe::Model) && controller.class.name == 'CatalogController' }
+      }
+    )
```
jrochkind, line 231: Oh my, this is a part of Blacklight I have literally _never_ seen before! Good job figuring it out, that was some real code spelunking! 

I hope it doens't end up fragile towards future BL upgrades... I suppose we will just have to deal with it if it does.

app/views/catalog/_all_search_results_add_to_cart.html.erb
```diff
@@ -0,0 +1,3 @@
+<div id="add-results-to-cart" class="add-results-to-cart">
+  <%= link_to 'ADD RESULTS TO CART', request.params.merge(action: "add_to_cart", controller: 'all_search_result_ids'), class:"btn btn-outline-secondary" %>
+</div>
```
jrochkind, line 3: Should we use a ViewComponent instead of a partial?  Or wait, does it NEED to be a partial to work with the Blacklight API you found? Woah. 

We better have really good tests for this stuff, I don't trust Blacklight at all!

jrochkind:
Well done! Tricky stuff. Just a few things I'm wondering about.

app/views/catalog/_all_search_results_add_to_cart.html.erb
eddierubeiz: I believe add_results_collection_tool will only take a symbol that is assumed to be the name of a partial. I did a bit of digging around but was unable to get it to accept a partial.

jrochkind:
nicely done!

PR #2934, opened 2025-03-25
capybara system spec for lazy load batched members on work image page

jrochkind:
Currently having trouble with the test, let's see if a separate PR about capybara can fix it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,10 +83,9 @@ jobs:
           uses: actions/upload-artifact@v4
           if: failure()
           with:
-            name: dist-without-markdown
+            name: screenshots
             path: tmp/capybara/*.png
             if-no-files-found: ignore
-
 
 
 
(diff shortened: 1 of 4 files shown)


jrochkind:
I cannot figure out why this now has reliable *additional*  test failures, that I can't reproduce on my macbook!

jrochkind:
OK something with trying to resize the browser window is really messing up capybara/chrome on github actions -- even though it works fine on my macbook?

Not only is the test we added failing, but four more tests are failing too, like it messes up Chrome for the future tests even though they are supposed to be isolated?

Screenshots from all failed tests are identical and very mysterious, this tiny slice of... horizontal scrollbar? 

failures_r_spec_example_groups_public_work_show_page_work_with_complete_metadata_lazy_load_of_batches_of_members_loads_in_remaining_images_only_on_scroll_965

jrochkind:
hooray got it working! I think my window size was just too crazy small.

PR #2935, opened 2025-03-25
Try adding some chrome options to avoid recent capybara/selenium/chrome race conditions

jrochkind:
Such as `Node with given id does not belong to the document`

These suggestions from: https://github.com/teamcapybara/capybara/issues/2800#issuecomment-2750363862

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -51,9 +51,41 @@ RSpec.configure do |config|
   # instead of true.
   config.use_transactional_fixtures = true
 
+
+  Capybara.register_driver :local_selenium_chrome_headless do |app|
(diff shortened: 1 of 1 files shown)


jrochkind:
Sadly still got a `unknown error: unhandled inspector error: {"code":-32000,"message":"Node with given id does not belong to the document"}` in CI on first run. :( 

It's possible it makes them somewhat less likely, I am not getting them locally on my macbook anymore with these changes, 4 runs so far.

jrochkind:
OK, well, I can't reproduce that initial failure, either here or on my MacBook. 

So seems worth merging, perhaps it's still possible to get an error but hopefully less likely, at any rate it doesn't seem to hurt.

PR #2941, opened 2025-03-31
Capybara patch for 'node id does not belong to the document' race condition

jrochkind:


jrochkind:
OK, there is some indication that future/recent versions of Chrome may fix this, but not sure when Github Actions will end up with them. 

I don't _think_ this will hurt in any case; if problem is still happening this week, let's just merge it.

jrochkind:
OK still getting failures on github actions CI.  I think this will eliminate or reduce them. Review please Eddie?

PR #2943, opened 2025-04-03
OH's with no requestable files get special message

jrochkind:
If an OH work has request mode set to OFF _and_ has no published available files:

* It gets a special message about being sealed
* And it is not included in the 'immediate' availability facet, or any other 'availability' facet

Additionally, logged in staff now sees things as an ordinary user would in public view, they can go to admin view to see non-published stuff. This makes  implementation easier and is honestly less confusing for staff.


jrochkind:
On-screen language approved by stakeholders.

eddierubeiz (approved):

PR #2948, opened 2025-04-08
reduce free requests before turnstile to 1, the bots are aggressive

jrochkind:


jrochkind:
self-merging to deploy, just changed a `2` to a `1`, and a comment, self-checked the diff.

PR #2949, opened 2025-04-09
Tool to arithemetically alter OH transcript timestamps to create single multi-recording sequence

### LINKED ISSUE #2939: Do not allow the ingest of multipage and multilayer TIFFs.
For these TIFFs, the output of the `identify` command will contain one line per image. Here are three examples:

```
CHV11N1_029.tif[0] TIFF 5404x7001 5404x7001+0+0 8-bit sRGB 108.305MiB 0.000u 0:00.003
CHV11N1_029.tif[1] TIFF 123x160 123x160+0+0 8-bit sRGB 0.000u 0:00.000
```

```
CHV11N1_030.tif[0] TIFF 5413x6893 5413x6893+0+0 8-bit sRGB 106.814MiB 0.000u 0:00.000
CHV11N1_030.tif[1] TIFF 126x160 126x160+0+0 8-bit sRGB 0.000u 0:00.000
```

```
Silver_001.tif[0] TIFF 5019x5765 5019x5765+0+0 16-bit sRGB 386.342MiB 0.220u 0:00.225
Silver_001.tif[1] TIFF 5019x5765 5019x5765+0+0 16-bit sRGB 386.342MiB 0.000u 0:00.000
```

See also very broad ticket https://github.com/sciencehistory/scihist_digicoll/issues/1652  which covers similar concerns.

jrochkind:
SequenceOhTimestamps: A new service obj which can take a Docx transcript with timecodes and a hash of multi-file start times (same format we were already storing them for combined audio derivative), and re-number the timestamps to be correct sequence for combined file, instead of starting over at 0 on every file change. 

There are new shrine attachments on `oral_history_content` sideare for an for INPUT FILE  needing to be transformed (`input_docx_transcript`), and an OUTPUT transformed file produced (`output_sequenced_docx_transcript`). We do keep both of these around. 

There is some admin UI in the "Oral History" tab. 
* You can upload an input file, which will be attached to oral_history_content, and then a bg job will be enqueued to transform and store the output
* If output is present, you can download it -- for you to use with OHMS Studio. 
* If any errors happened when trying to transform, we'll store them in shrine metadata for the attached input file, and display them in admin UI

Closes #2939


jrochkind:
Reviewed and approved by stakeholders!

eddierubeiz (approved):

PR #2957, opened 2025-04-23
Stop logging bot-challenged requests, but do save them in DB

jrochkind:
Goal is to stay within our papertrail logs. We can't control how many bots hit us, so how many challenges we issue. But if we're keeping neough out with our challenges, then if we stop logging the challenges themselves -- our remaining traffic to log should be reasonable. 

One way this isn't entirely true though is that HEROKU still logs all requests as heroku.router logs. But at least we can avoid double logging them. 

If we want to see them, they are all still in the db -- there is no admin UI at present for this, devs can just go to db or whatever. In fact, we're logging MORE in the db than we log in request log, including complete user-agent and complete http headers.

After merge and deploy, we need to add cleanup Heroku scheduled job:

    bin/rails runner 'BotChallengedRequest.trim_old

- [ ] Added in staging
- [ ] Added in prod

- move lograge config into application.rb for any env testing easier
- add model BotChallengedRequests for logging them
- save BotChallengedRequests to db
- do not log bot-challenged requests, try to keep within our papertrail log plan limits. We are writing them to db with BotChallengedReqeusts
- method to trim old BotChallengedRequest

PR #2959, opened 2025-04-23
fix suppression of bot challenge logging

jrochkind:
We were using wrong rack env value and also blocking anyone that had already passed the challenge

Fix mistake from #2957


jrochkind:
gonna self-merge this guy

PR #2960, opened 2025-04-23
5 second timeout for qa autocomplete

eddierubeiz:
It does not look like we have to mess with the QA gem. [The jquery autocomplete](https://github.com/devbridge/jQuery-Autocomplete) plugin allows you to configure an `ajaxSettings` which allows the underlying [jquery ajax settings](https://api.jquery.com/jquery.ajax/#jQuery-ajax-settings)  to be modified. These last allow a timeout option.

### DIFF
--- a/app/frontend/javascript/admin/qa_autocomplete.js
+++ b/app/frontend/javascript/admin/qa_autocomplete.js
@@ -83,6 +83,10 @@ function addAutocomplete(element) {
       showNoSuggestionNotice: true,
       serviceUrl: qa_search_url,
 
+      // see https://github.com/devbridge/jQuery-Autocomplete
+      // see https://api.jquery.com/jquery.ajax/#jQuery-ajax-settings
+      ajaxSettings: { timeout: 5000 },
+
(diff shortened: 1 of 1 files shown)


jrochkind:
Ooh perhaps we should PR to make this a default in QA?  Not sure.

jrochkind:
Wait, nope!

We're not talking about a jquery AJAX request in that ticket. 

Which would not clog up our servers anyway -- the problem to be solved is heroku timeout from waiting on our _rails app_ on the server side to return, because the rails app is wiating on OCLC api.  

We're talking about the back-end ruby code talking to OCLC API. 

So this isn't it!

eddierubeiz:
OK -- I made a patch for now. I might spend some time making something a bit more useful that we can share with QA upstream. But this works.

eddierubeiz:
We won't be needing this PR as we've been able to fix the problem upstream.

eddierubeiz:
No longer need this patch; we fixed the problem upstream.

Issue #2684, opened 2024-07-17
Investigate and optimize solr timeouts


jrochkind:
We *intend* to set a timeout on our app's connection to solr, to keep it from waiting indefinitely on a non-responsive solr.  Becuase that happens sometime, and we don't want to tie up all our web workers waiting on non-responsive solr. 

We intend to have it wait only 4 seconds for solr before giving up:

https://github.com/sciencehistory/scihist_digicoll/blob/53a1f0cd346f7456f547f37d9c5cf0852a419a1b/config/blacklight.yml#L10

Also, it is intended to retry twice, waiting only a fraction of a second between retries. 

https://github.com/sciencehistory/scihist_digicoll/blob/53a1f0cd346f7456f547f37d9c5cf0852a419a1b/app/models/scihist/blacklight_solr_repository.rb#L40-L45

This means that if a Solr is being consistently unresponsive (or taking 10 seconds, or jsut more than 4, to respond), the app ought to wait:  4 seconds + ~1 second delay + 4 seconds + ~1 second delay + 4 seconds before giving up. Or around 14 seconds. 

Which may already be actually far too long to tie up a worker on unresponsive solr? Maybe we want only ONE retry. (we have a retry in because in some cases SearchStax solr was being weird and a retry worked).  Maybe less than 4 second timeout?  (Although solr can be unpredictably slow). 

But additionally, it's not clear to me that it really _is_ working like this. 

In times of Solr unresponsiveness, we are having heroku tell us lot sof responses are timing out at _30 seconds_, heroku's own timeout. Suggesting they were being held up waiting for solr in sum mucb longer than the 14 seconds we expect too. 

So it's possible the configuration isn't working like we want anyway -- it was a bit convoluted to try to set it up. 

**investigate how solr timeout/retry is _actually_ working, with _actual_ max time spent waiting for solr, then consider if we want to change it to keep app more reliable even when solr is not**

eddierubeiz:
**Relevant lit**
 - (Various stackoverflow articles)
 - https://devcenter.heroku.com/articles/request-timeout
 - https://devcenter.heroku.com/articles/h12-request-timeout-in-ruby-mri
 - https://devcenter.heroku.com/articles/error-codes#h12-request-timeout
 - https://scoutapm.com/blog/response-time-analysis

**Tools**
 - HoneyBadger's [list of recent timeout errors](https://app.honeybadger.io/projects/58989/faults?q=TimeoutError);
[...]

PR #2961, opened 2025-04-24
Drop solr timeout to 2s, with one retry after 300 ms wait

jrochkind:
So maximum end-to-end time should be around 2000ms + 300ms + 2000ms == 4.3s

Previously maximum wait time was:

4000ms + 300ms + 4000ms + 600ms + 4000ms == 12.9 seconds.

That's far too long for a single puma thread to be held up waiting on solr, and definitely caused problems when solr was having a brief outage with all our puma threads in use.

This may still be enough time to caues problems, depending on our traffic volume. We will see.

We are having ~4 times a day where Solr is unresponsive for 8-10s, very frustrating, have to figure that one out.

Screenshot 2025-04-24 at 10 22 29 AM

Note bumps where all puma workers are in use coincide with max request time being 13.3 seconds -- strongly suggesting it's workers waiting on solr for max timeout period -- yes, it's weird how *regular* these are (but they do differ day to day by 30-60 minutes, they aren't at exactly the same time every day or anything), and we haven't yet gotten to the bottom of it. Solving that underlying issue would also be good!

Some references:

* https://github.com/sciencehistory/scihist_digicoll/issues/2684
* https://github.com/sciencehistory/scihist_digicoll/issues/2684#issuecomment-2486426237

### DIFF
--- a/app/models/scihist/blacklight_solr_repository.rb
+++ b/app/models/scihist/blacklight_solr_repository.rb
@@ -41,8 +41,8 @@ module Scihist
           interval: (zero_interval_retry ? 0 : 0.300),
           # exponential backoff 2 means: 1) 0.300; 2) .600; 3) 1.2; 4) 2.4
           backoff_factor: 2,
-          # But we only allow the first two before giving up.
-          max: 2,
+          # But we only actually only ONE retry at present, so it should be 300ms
+          max: 1,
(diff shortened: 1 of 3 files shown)


eddierubeiz (approved):
Enthusiastically approve.

PR #2962, opened 2025-04-28
DZI manifest file location is now tracked by a Shrine uploaded file

jrochkind:
Stored in an attr_json, an implementation we previously were using and verified.

Before this change, the DZI file location was assumed based on current asset fingerprint checksum. That meant that it automatically changed when the asset changed -- good. But we had no way to move it to a new location _without_ changing the asset fingerprint. And since we are delivering them with cache-forever headers, when we have an improvement to the generation algorithm and want to regenerate them -- we need to move them to a new location.

Persisting current location in the db lets us change the location on demand, although of course we have to generate them at a different location, but we can change the logic for picking a location -- and the old location can still be remembered in db, to work until it's replaced, and to be deleted.

### DIFF
--- a/app/jobs/delete_dzi_job.rb
+++ b/app/jobs/delete_dzi_job.rb
@@ -1,5 +1,5 @@
 class DeleteDziJob < ApplicationJob
-  def perform(dzi_file_id)
-    DziFiles.delete(dzi_file_id)
+  def perform(dzi_file_id, storage_key)
+    DziFiles.delete(dzi_file_id, storage_key: storage_key)
   end
 end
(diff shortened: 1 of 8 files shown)


jrochkind:
Before we merge, we need a rake task to back-fill the UploadedFile metadata into the db, which we can do before we have merged this!

jrochkind:
Good catch @eddierubeiz thanks!

Committed. Too hard to check in staging cause staging DZI's are so inconsistent, but added a checkbox to check in prod after deploying, and I'll make it top priority to fix if it does not work.  Currently running DZI report on prod pre-merge/deploy to get a baseline reading.

PR #2964, opened 2025-04-30
Change Asset#dzi_file to Asset#dzi_package to be more clear

jrochkind:
It handles multiple files, and not to be confused with our new `dzi_manifest_file` shrine UploadedFile


jrochkind:
@eddierubeiz Now that the new DZI storage stuff is merged and safe -- this is a PR that just changes some method and variable names to be IMO more clear! Straight refactor.

app/models/asset.rb
```diff
@@ -159,12 +160,15 @@ class Asset < Kithe::Asset
...
-    @dzi_file ||= DziFiles.new(self)
+  #     asset.dzi_package.url # url to manifest file
+  #     asset.dzi_package.exists?
+  #     asset.dzi_package.create # normally handled by automatic lifecycle hooks
+  #     asset.dzi_package.delete # normally handled by automatic lifecycle hooks
+  #
+  # See also dzi_manifest_file which points to the single .dzi file -- this object
+  # manages a whole directory.
+  def dzi_package
+    @dzi_files ||= DziFiles.new(self)
```
eddierubeiz, line 171: Why not memoize with the same variable name:
`@dzi_package ||= DziFiles.new(self)`

jrochkind: good catch, just something I didn't notice since tests passed without it, will do.

eddierubeiz (approved):

PR #2968, opened 2025-04-30
Rake task to fill in legacy DZI manifest location in json_attributes location

jrochkind:
Where #2962 will find it.

So we can merge and run this FIRST, to prep database and minimize any unfound DZI's.

- [x] heroku run rake scihist:data_fixes:fill_in_legacy_dzi_manifest_location -r staging
  -  on staging this will result in many that point to non-actually-exsitng DZI since we don't sync all DZI media  -- but that's exactly what will happen after syncing prod to staging after implementing this in prod, so that's an appropriate thing! 
- [ ] heroku run rake scihist:data_fixes:fill_in_legacy_dzi_manifest_location -r production


jrochkind:
@eddierubeiz Ready for review/approval!

jrochkind:
@eddierubeiz I'm going to merge and deploy to stay on track, but please do review!

PR #2970, opened 2025-05-06
update aws-sdk-s3 and dependencies to latest versions

jrochkind:
they weren't really too far behind, prob wont' make that much difference, but just checking all boxes

**update** -- but then there was a breakage! With our hacky using internal API mocking of the S3 client for testing our orphan checker. Found debugged and fixed, okay, glad to get that out of the way.


jrochkind:
Well, this was some unexpected failures, possibly backwards breaking changes within an AWS SDK minor release?

jrochkind:
OK, we were doing some probably too-clever stuff to mock the S3 client, and the new version of the S3 SDK gems broke it. 

I fixed it. But this is maybe fragile. Not sure if there's a better way to do orphan checker tests, when we don't have a real S3 in test (eg on CI).

eddierubeiz (approved):

PR #2978, opened 2025-05-12
update rubygem dependencies

jrochkind:
with simple `bundle update` then commit Gemfile.lock

Keep things up to date, tests pass, good practice to get bugfixes etc.


jrochkind:
Also update rspec-rails to latest 8.0 release with Gemfile change. rspec-rails 8.0 recommended for Rails 8.0 which we are using.

PR #2980, opened 2025-05-15
Video tag uses VTT captions if available

jrochkind:
On top of #2979, supply those VTT, if they exist, to `<video>` tag for video.js to offer em as a caption option. 

How simple is that?

Only does something if vtt tracks are present, so in production won't be visible change at present.

### DIFF
--- a/app/components/work_video_show_component.html.erb
+++ b/app/components/work_video_show_component.html.erb
@@ -47,6 +47,9 @@
             <%= tag "source", src: video_src_url, type: video_asset.content_type %>
           <% end %>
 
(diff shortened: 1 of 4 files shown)


eddierubeiz (approved):

jrochkind:
oops working great on MacOS Chrome, but NOT on MacOS Firefox or Safari. 

Does not hurt to have it merged, but more investigation required.

jrochkind:
Prior comment in error, it is working on eg Firefox MacOS. But does need more investigation for oddities or breakages on more browsers, there may be idiosyncracies.

PR #2983, opened 2025-05-21
patch QA to use https://fast.oclc.org

### LINKED ISSUE #2982: `Could not parse response as JSON` after retrieving FAST subjects from fast.oclc.org
In production, we've been getting errors that look like this for roughly the past week:

```
Processing by Qa::TermsController#search as TEXT
  Parameters: {"q"=>"quantum", "vocab"=>"assign_fast", "subauthority"=>"all"}
Retrieving json for url: http://fast.oclc.org/searchfast/fastsuggest?&query=quantum&queryIndex=suggestall&queryReturn=suggestall%2Cidroot%2Cauth%2Ctype&suggest=autoSubject&rows=20&sort=usage+desc
Could not parse response as JSON. Request url: http://fast.oclc.org/searchfast/fastsuggest?&query=quantum&queryIndex=suggestall&queryReturn=suggestall%2Cidroot%2Cauth%2Ctype&suggest=autoSubject&rows=20&sort=usage+desc
```

Notes:
- The responses from `fast.oclc.org` (e.g. [the one above](http://fast.oclc.org/searchfast/fastsuggest?&query=quantum&queryIndex=suggestall&queryReturn=suggestall%2Cidroot%2Cauth%2Ctype&suggest=autoSubject&rows=20&sort=usage+desc) ) are in fact valid json, so we should consider the possibility that QA is erroneously reporting the JSON responses as invalid.
- QA been largely unmaintained since the beginning of 2025.
- Not all responses sent to Qa::TermsController#search have resulted in the above error.

Note: this is unrelated to a separate ticket currently in the works (ref #2770).

Thanks to @JAuerbach26 for helping diagnose this.

eddierubeiz:
Fixes #2982 .
The URL is hardcoded so we can't just change a setting.
It's likely this has been broken for a couple weeks.

### DIFF
--- a/config/initializers/qa_url_patch.rb
+++ b/config/initializers/qa_url_patch.rb
@@ -1,0 +1,25 @@
+# http://fast.oclc.org/ is now forwarding to https://fast.oclc.org/, (https) which breaks QA.
+# The URL is hard-coded, we need to submit a PR to change that.
+# We are currently unable to get the tests on Questioning Authority to pass, so our PRs can't be approved or merged.
+#
+# For now we're just using a patch.
+# If, in a later version, Questioning Authority can read the URL from a settings file, we can remove this patch.
+#
+# see https://github.com/sciencehistory/scihist_digicoll/issues/2982
+
+SanePatch.patch('qa', '5.14.0') do
+  module Qa::Authorities
+    class AssignFast::GenericAuthority < Base
+      def build_query_url(q)
+        escaped_query = clean_query_string q
+        index = AssignFast.index_for_authority(subauthority)
+        return_data = "#{index}%2Cidroot%2Cauth%2Ctype"
+        num_rows = 20 # max allowed by the API
+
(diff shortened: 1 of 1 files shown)


jrochkind:
yep, we DEF need to PR this upstream.

jrochkind:
Ideally when I do this i include a url linking to the source of the original method you are overriding, at the version used in the current gem release. I'll go look that up at least to document here, but also approve this.

jrochkind (approved):

jrochkind:
Eg:

# Overridden at: https://github.com/samvera/questioning_authority/blob/296085f94c52ed5f617daac3273febc35f9649c6/lib/qa/authorities/assign_fast/generic_authority.rb#L44-L53

jrochkind:
Also in patch upstream, it prob ought to follow redirects!

eddierubeiz:
Thanks for looking up original method URL!

Issue #2770, opened 2024-10-26
Make sure requests to OCLC for FAST time out


jrochkind:
Apparently the OCLC endpoint is timing out -- our app seems to wait 30 seconds for a response, which holds up the puma workers, and eventually at 30 seconds gets reported as an error by heroku. 

The connection should have a timeout set to, say, 5s, so it gives up without waiting forever. 

This may require making a change to the questioning_authority gem. 

```
Oct 25 19:12:41 scihist-digicoll-production heroku/router at=error code=H12 desc="Request timeout" method=GET path="/authorities/search/assign_fast/all?q=United%20States.%20Patent%20%20Office" host=[digital.sciencehistory.org](http://digital.sciencehistory.org/) request_id=a1506fd1-187b-4196-8fb4-291657a9c145 fwd="173.161.142.5" dyno=web.1 connect=0ms service=30001ms status=503 bytes=0 protocol=https
```

eddierubeiz:
See also my notes at https://github.com/sciencehistory/scihist_digicoll/issues/2684 .

PR #2990, opened 2025-06-03
Update rubygems dependencies; remove patch.

eddierubeiz:
I ran `bundle update` and committed the new `Gemfile.lock`.
We solved several bugs in the QA gem lately, including
- https://github.com/sciencehistory/scihist_digicoll/issues/2982
- https://github.com/sciencehistory/scihist_digicoll/issues/2770

We included all of these fixes in the latest QA release, which is referenced in this PR.
Because we now have the latest code, we can now remove a patch we created last month, which this PR also does.


jrochkind:
Normally I would separate the PR's and just update qa in this one, not every gem in the Gemfile.  But I guess this is fine?

jrochkind:
With removing the patch, do we not need a config file config'ing the timeout in qa?

eddierubeiz:
No need for a config file, no, although we could add one if we wanted to override the default, which is 5 seconds.

eddierubeiz:
In practice I do think this is fine, although there would be absolutely no harm in making this two PRs - one to update all the gems and a second one to remove the patch; admittedly, that would be nicer!

jrochkind:
5 seconds seems way long, how about 2?

PR #2991, opened 2025-06-03
Update qa and remove patch

eddierubeiz:
I ran `bundle update --conservative qa` and committed the new Gemfile.lock.
We solved several bugs in the QA gem lately, including

- https://github.com/sciencehistory/scihist_digicoll/issues/2982
- https://github.com/sciencehistory/scihist_digicoll/issues/2770

We included all of these fixes in the latest QA release, which is referenced in this PR.
Because we now have the latest code, we can now remove a patch we created last month.


jrochkind:
I think 2s timeout might be better -- after dealing with solr timeouts and effect on our stability, I now have even less generous attitudes towards timeouts -- but good enough for now!

jrochkind (approved):

eddierubeiz:
Cool - that's a one-line change now because it's in a configuration file. But I feel like that's a different PR.

PR #2992, opened 2025-06-04
Set the oclc assign_fast timeout to two seconds

eddierubeiz:
This is now just a configuration file change.
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2770.
I tested this one more time; we will in fact using 2 seconds as the timeout after this PR.


jrochkind:
love it thanks

PR #2995, opened 2025-06-04
Video.js custom theme to make control bar bigger and with the components and order we want

jrochkind:
- make video.js volume vertical instead of inline bar, much less confusing
- custom video.js theme making controls bigger, while keeping subtitles out of way
- customize video.js controlbar to show elapsed time, and move volume to right side
- hide elapsed time tooltip, restyle mouse hover tooltip
- get menus out of way of progress bar

### DIFF
--- a/app/components/work_video_show_component.html.erb
+++ b/app/components/work_video_show_component.html.erb
@@ -22,7 +22,8 @@
 
           <%= content_tag("video",
                   id: "work-video-player",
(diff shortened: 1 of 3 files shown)


jrochkind:
draft cause I realized I want to tweak the small screen responsive stuff a bit more.

jrochkind:
and ready again

jrochkind:
omg nevermind, needs more investigation

Issue #2998, opened 2025-06-06
More-like-this SOLR queries should filter out unpublished works.


eddierubeiz:
**Assumption:**
We never, ever want to display unpublished works as part of our "Related works".

**Current situation:**
We are currently retrieving *all* similar works from SOLR (published or unpublished) and later on using a DB query to filter out unpublished works.

**Why we don't like it**
 - We are instantiating works that SOLR _knows_ to be unpublished, just to throw them out immediately once the database tells us that that are unpublished.
 - We are soon going to be _caching_ the results of these SOLR queries (ref #2832 ). We don't want to cache information we won't use.

**What we really want:**
Instead, let's ask SOLR to filter out unpublished similar works, and then ALSO double-check with the DB that all the works the helper returns are indeed published.

eddierubeiz:
This is done.

PR #2999, opened 2025-06-06
More-like-this solr queries should filter out unpublished works.

eddierubeiz:
Ref #2998


jrochkind:
Love it. Surprised the value is `1`, not `true`.  Do we need/want a test that this is working right, and is succesfully filtering things out?

eddierubeiz:
I changed it to true, although 1 also does work (see e.g. [this other code](https://github.com/sciencehistory/scihist_digicoll/blob/master/app/models/search_builder/access_control_filter.rb)). I also added an explicit test for that part of the code -- good thinking.

PR #3001, opened 2025-06-06
[WIP] Rename redis queue

eddierubeiz:
We will soon have not one but two redis instances:
- our existing one (used for the rescue active_job queue)
- a new one for the general purpose Rails cache, which we'll use for
  - rack_attack (which we will move over from memcached)
   - general-purpose Rails caching (including, presumably, low-level caching for related items searches)

To avoid confusion, we are going to need to distinguish the methods / env vars we use to distinguish them (such as `REDIS_PROVIDER_ENV`)
Before starting work on the new redis instance, let's rename the current one to make it clear that it's used for the active job queue.


jrochkind:
works for me! Thanks!

PR #3004, opened 2025-06-10
Swap out old unmaintained webvtt gem for more maintained one

jrochkind:
Old and unmaintained (sorry, peer librarian jaron): https://github.com/jronallo/webvtt

Somewhat more maintained and with more functionality: https://github.com/opencoconut/webvtt-ruby

Ended up being very easy to swap out -- we are using only basic functionality, and were already using a wrapper class to cover for lack of functionalkty in @jaron/webvtt, so made it even more trivial.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -189,7 +189,7 @@ gem "warning", "~> 1.2" # managing ruby warning output
 
 gem "rack-attack", "~> 6.6" # throttling excessive requests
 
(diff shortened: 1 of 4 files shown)


jrochkind:
While our test coverage is good, want to test this on staging with several  legacy and new style OHMS transcripts, as well as videos, before merging.

eddierubeiz (approved):

Issue #3015, opened 2025-06-26
Remove memcached code and resources.


eddierubeiz:
In late June 2025, we stopped using memcached for our cache.
Once the new settings have been live in production for a few weeks (say August),  we should remove the memcached code in config/environments/production.rb .
This code is clearly labeled `BEGIN OBSOLETE MEMCACHED CODE` and `END OBSOLETE MEMCACHED CODE`.
See: https://github.com/sciencehistory/scihist_digicoll/pull/2994
Once that's done, we can also turn off the memcached addon (`memcachier-adjacent-27905`)

PR #3017, opened 2025-06-30
Remove memcached, dalli and references to it

eddierubeiz:
The new ephemeral redis is doing fine as a cache in production, so we're going to remove the old memcache code.
Ref #3015


eddierubeiz:
I'm self-merging this code so I can remove memcache in production.

eddierubeiz:
Removed in staging and prod. Also removed the ENV variables that referred to it.

Issue #2647, opened 2024-05-31
Resize "hero" images on main page


jrochkind:
We are using very large JPGs, without srcset. 

The actual images look even larger than 2x the actual screen size taken on a fairly large screen. 

This is a lot more bytes than is needed, slowing down page loading (especially on slower networks) -- and probably slowing down Google's metrics on our page responsiveness, leading ot less Google SEO than we could have. 

We should make a series of resizes for each one, and use `srcset` to let the browser pick the appropriate one for actual size displayed (and network speed if browsers do that, but sadly I don't think they do)

jrochkind:
Also convert to .webp and maybe run through other optimizer tools to make smaller. As largest image on the page, it will help our Google SEO to make these images load faster. 

(Could consider doing that with our single-work large thumbs in the future!)

PR #3019, opened 2025-07-01
Homepage hero images: use webp at four different sizes.

eddierubeiz:
Ref #2647 

# Background
Status quo ante is well documented at https://github.com/sciencehistory/scihist_digicoll/tree/master/app/assets/images/homepage/hero_images#readme .
We were using `srcset` before, with widths 1x and 2x for 1000px and 2000px width respectively.

# New images
- We're now using webp instead of jpg
- We're now providing four options via `srcset`: 250w, 500w, 1000w and 2000w to represent respectively 250px, 500px, 1000px and 2000px of width
- The file size of the webp images is quite small.
- Recipe for producing the images is documented in the readme; it's a good idea to keep the readme current, as we will certainly want to add new images in the future and we want to keep everything consistent.


eddierubeiz:
Good comments - thanks!

_I'm confused by the naming convention of 1X ... 8X._
_Previously when dealing with multiple sizes, I used the "X" designation to refer to pixel density, the 1X was basically the "full size" image, and the 2X was for a "double density" display, for "retina" screens that displayed double "actual" pixels for each "css pixel" --- commonly called a "2X display"._
_I don't think that's what the X's mean here. I think the 1X is actually significantly smaller than "full size", whatever that is?_
_I realize you refernece these in a lot of places and it might require a lot of copy and replace, but woudl it be clearer if you named these after actual physical pixel widths, instead of 1X...8X? I'm not sure what the X's mean, or what size 1X is supposed to be._

**Fixed.** The paths on the filesystem and the method labels are changed, although the same data is sent to the browser.

I also wonder if it would be easier to predict filenames based on a base name and those pixel size designations, instead of having to list them all in a config file?

Absolutely open to doing this; will create another ticket for it.

_And lastly, do we have a wiki page explaining how to prepare these home page images that needs to be updated?_

No, that information is in the readme I mentioned above, which is edited as part of this PR.

_Love using webp! And the srcset seems fairly straightforward? Have you tested in a browser to make sure the right images are being chosen? How about try to verify that "double resolution" is chosen on a MacBook bulit-in "retina" display?_

Yes, the srcset is straightforward, and we've been using it for a couple years now. All we're doing is adding two smaller images for the browser to use if it wants. My mac (retina)continues to use the 2000px display which is great.

PR #3020, opened 2025-07-02
fix js typos for 'container'

jrochkind:


jrochkind:
Will self-marge @eddierubeiz , thanks!

Issue #2852, opened 2025-01-17
Possible n+1 query in Admin::OralHistoryRequestsController#index


eddierubeiz:
If I'm reading the analysis from Scout APM correctly, we could dramatically speed up the `/admin/oral_history_requests page` by loading more associations in the [index method of the admin pages' list of oral histories](https://github.com/sciencehistory/scihist_digicoll/blob/master/app/controllers/admin/oral_history_requests_controller.rb).

Not urgent, but a good idea to look into this.

jrochkind:
Sweet, good catch! We should def investigate

PR #3033, opened 2025-07-08
Fix n+1 query in list of OH request

eddierubeiz:
Ref #2852 
Used `strict_loading` to figure out what `includes` we needed. Fixed.


eddierubeiz:
Tested one last time in staging; deploying to production now.

PR #3037, opened 2025-07-15
solr_escape box_id and folder_id

eddierubeiz:
Ref #2936

I'm not writing a test for this but I will spend a bit of time looking for other similar bugs.


eddierubeiz:
Again - I'm self-merging and deploying to prod, as this is a bug.

eddierubeiz:
Went through the search_builder directory. No sign of any other search builder classes in which we would need to escape user-entered strings before sending to SOLR.

Issue #3038, opened 2025-07-16
Better escaping for admin works page external_id filtering


eddierubeiz:
In ref #3036 I added a bandaid fix for the external_id filtering form.

It's not the best solution, as it would not match any external_ids that contained single or double quotes (admittedly... this is an unlikely scenario).

I think with a bit of cleverness, it's possible to do a bit better and actually match a theoretical work with an external_id containing single or double quotes.

PR #3039, opened 2025-07-16
Properly escape title_or_id to match any category of external_id (works)

eddierubeiz:
Ref #3038

Much better fix for band-aid fix ref #3038 , which which fixed the bug but dodged the real problem by simply removing escape characters from the search string.

The key was to figure out
 - exactly what to pass to the ActiveRecord escaping;
 - exactly what to pass to `to_json'.

### DIFF
--- a/app/controllers/admin/works_controller.rb
+++ b/app/controllers/admin/works_controller.rb
@@ -710,7 +710,8 @@ class Admin::WorksController < AdminController
       if params[:title_or_id].present?
         match_ilike = Work.sanitize_sql_like params[:title_or_id]
         match_uuid = Work.type_for_attribute(:id).cast(params[:title_or_id])
-        match_any_external_id = "[{ \"value\" : #{params[:title_or_id].to_json} }]"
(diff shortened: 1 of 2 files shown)


jrochkind:
Oh man this does look hairy!

I'm having trouble understanding what changed, since formatting/refactoring was done as well as fix...  but okay...

     match_any_external_id = "[{ \"value\" : #{params[:title_or_id].to_json} }]"

That still looks weird.....   I'm curious if this would work?

    match_any_external_id = [{ "value" => params[:title_or_id] }].to_json

Like can we let to_json take care of ALL of it, and not have to do that manual writing a json string at all?

jrochkind:
With multiple substititions with `?` gets confusing ot keep track... there is some way to use named keyword params instead for substitution, but maybe not necessary here if you think still fairly readable.

eddierubeiz:
The other challenge was figuring out exactly to replace the `?` with in the expression  `"json_attributes -> 'external_id' @> ?::jsonb"` .

eddierubeiz:
Used named params - even better. Thanks!

PR #3044, opened 2025-07-17
Set specific window size for headless chrome in specs

jrochkind:
In some cases the window was too small, such that eg facet sidebar would collapse into hidden disclosure, which would cause some of our tests to fail. I don't know what determines default window size or why it seemed to change unpredictably -- for instance, upgrading some dependencies that shoudln't be related, in a different PR, seemed to trigger this. But it's confusing when it does get triggered, and seems better to just explicitly set the window size so it's always the same.

Still small (there may be test benefits to a small window, making sure things work at small window size?), but big enough to NOT have collapsed sidebar etc.


eddierubeiz (approved):
That's awesome.

PR #3047, opened 2025-07-17
update from ruby 3.3 to latest ruby 3.4

jrochkind:
- update to latest ruby 3.4.5
- new ruby serializes hashes with more spaces, so test on serialization has to be fixed
- update bundler version to 2.6.9, version that comes with ruby 3.4.5. Found no way to do this but manual editing of gemfile.lock


jrochkind:
and actually let me fix that test failure a better way.

Issue #3041, opened 2025-07-16
Inches to cm converter in  the work metadata entry form


eddierubeiz:
Let's put a little inches to cm converter in the metadata entry form, so person entering can use it and convert and enter in CM.

This was part of ref #2656, but I never actually did it.

I'm creating a new issue for it as it's really a separate (though small) project.

PR #3053, opened 2025-07-21
Extent converter for work metadata form

eddierubeiz:
Ref #3041

Using a partial instead of a component, as the template is pretty simple.

It's important that users understand this is not a field to enter work metadata, but just a convenience calculator. I think the UI makes it clear, but I'm open to suggestions.

Other feedback welcome!


jrochkind:
A screenshot of what it looks like is helpful when PR'ing UI changes, but I'll load it up and take a look this time!

jrochkind:
also can add some labels/legends for accessibility (we coudl have staff taht needs accessibilty, might as well do it when we can)

PR #3058, opened 2025-07-22
Restore % in the ILIKE query on the Works and Collections filter in Admin

eddierubeiz:
I must have removed the % in my last PR, which was silly. But our tests didn't catch the problem, because they only tested matching against the full title, not a substring of the title.
Mea maxima culpa!


eddierubeiz:
This PR has too many commits in it - let's try this again.

Issue #3063, opened 2025-07-23
try to upgrade to solr 9


jrochkind:
may give us problems with solr_wrapper, jrochkind knows more.

jrochkind:
Trying ot see if there were solr_wrapper problems, I ran into *other* problems, like, yeah, some of the components we are using may be no longer avail in solr 9, not sure. 

```
ERROR: Error from server at http://127.0.0.1:8987/solr/admin/cores?action=CREATE&name=scihist_digicoll_development&instanceDir=scihist_digicoll_development&wt=javabin&version=2: Error CREATEing SolrCore 'scihist_digicoll_development': Unable to create core [scihist_digicoll_development] Caused by: solr.ICUFoldingFilterFactory
```

jrochkind:
Also might as well REMOVE the lib config that are ignored in recent solr:

https://github.com/projectblacklight/blacklight/pull/3497/files

Whatever we have listed there can give you a hint of what you need to add to ENV though.

PR #3064, opened 2025-07-24
Solr 9.8.1

eddierubeiz:
Ref #3063

# Changes
- New syntax for `solr.ICUNormalizer2FilterFactory` ( see [latest solr filter docs](https://solr.apache.org/guide/solr/latest/indexing-guide/filters.html) )
- Add module `analysis-extras` ( fixes `org.apache.solr.common.SolrException [...] Unable to create core [...] Caused by: solr.ICUTokenizerFactory`; [source](https://github.com/projectblacklight/blacklight/pull/3696) - thanks @jrochkind )
- Add module `extraction` (fixes `java.lang.ClassNotFoundException: org.apache.solr.handler.extraction.ExtractingRequestHandler`; [source](https://stackoverflow.com/questions/16340730/solr-extraction-extractingrequesthandler-classnotfoundexception) )
- Set `SOLR_SECURITY_MANAGER_ENABLED` explicitly to `'false'`. (Solr [changed the default value to "true"](https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.htmlfixes) in Solr 9, but the security manager is actually deprecated: _"In practice, however, the permission scheme [of the security manager] is so complex that the Security Manager has always been disabled by default, and its use is exceedingly rare."_ ([source](https://openjdk.org/jeps/486)). We explicitly set it to false to avoid `java.lang.Error: A command line option has attempted to allow or enable the Security Manager. Enabling a Security Manager is not supported`). 
- Changed the paths to the lib directories in`solrconfig.xml`.


eddierubeiz:
Solr Cell is only used for indexing files (like PDFs and Word docs), which to my knowledge we don't do, or intend to do
https://solr.apache.org/guide/solr/latest/indexing-guide/indexing-with-tika.html
I think it's safe to not include it.

jrochkind:
DOH. If there is no way to include "analysis-extras" on SearchStax Solr 9.8, that could be a real barrier. 

We may have to file a support ticket with SearchStax, explaining that we need to include jars that ARE included in solr distro but are not by default active, that in previous Solr we did this thorugh solrconfig.xml but this is no longer possible in Solr 9, listing the ways it can be done in Solr 9 (JVP arg, ENV variable), asking if there is any way to do that?

If not, then I guess we're back to looking for other options. 

We could also temporairly try upgrade to Solr 9.6, which may be the last versoin in which old way of <lib> in config still works, possibly.  We could check if Solr 9.6 gives us the GC we want too, hopefully.

eddierubeiz:
Let's examine our options tomorrow!

jrochkind:
Looks like last Solr 9.7 should probably still let us use `<lib>` in solrconfig.xml, so that's a possible target, confirm it gives us the JVM version and GC we want too!

> `<lib/>` directives have been made into an "opt-in" feature in Solr 9.8 and must be explicitly enabled if desired (using a system property or environment variable like -Dsolr.config.lib.enabled=true)

https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html

eddierubeiz:
So on SearchStax, we are **opted in** to using `<lib/>` directives (whether we like it or not) via the java flag `-Dsolr.config.lib.enabled=true`. I've verified this by looking at the solr admin pages on searchstax on the 9.8.1 deploy. 

As a workaround, I thought that leaving the four `<lib/>` directives in solrconfig.xml would thus allow the JVM to pick up the jar files we need, but so far it has not been able to instantiate the classes we need.

PR #3067, opened 2025-07-29
Create an ENV switch to skip the `scihist:solr_cloud:sync_configset` release task on heroku

eddierubeiz:
Ref #3063 
@jrochkind  noted this afternoon:
`we could certainly just add a env SUPPRESS_SOLR_CONFIG_PUSH or something.`
Great idea. Helpful in Catch-22 situations when you cannot release the next version of SOLR to Heroku, because the release phase involves pushing the config to SearchStax.... but there's a version mismatch. ♻️


jrochkind (approved):

jrochkind:
looks like a good idea. Did it end up helping?

eddierubeiz:
@jrochkind it was your idea! And yes, it helps a lot.

PR #3070, opened 2025-08-07
Update to ViewComponent 4 including blacklight updates necessary

jrochkind:
- view_component 4 escapes quotes better, which is fine
- view_component 4 initializer more strict with arguments.
- ViewComponent 4 requires explicit `helpers.` object call for `simple_form_for`
- view_component 4, including update blacklight and blacklight_range_limit to releases that allow view_component 4


eddierubeiz (approved):

jrochkind:
I'm holding off on this one for a second becuase there is some churn in Blacklight 8 around ViewComponent support, possibly a bug that may have to make it revert. Now I can't find the comment in Blacklight repo, but I'll look.

Issue #3072, opened 2025-08-08
Request less metadata from SOLR


eddierubeiz:
The other day @jrochkind made a suggestion:

_"I wonder what fl our solr queries are being sent with — what solr fields are returned in response — we kind of only need highlighted fields and id. i wonder if we coudl save bandwidth and even improve performance by cutting some."_

See also tip # 9 from https://opensolr.com/faq/view/opensolr-wiki-q-a/32/how-can-i-save-transfer-bandwidth-

Assuming dev solr is on port 8987, compare these two example json searches, loosely based on a real search:

[...]

PR #3073, opened 2025-08-08
add fl=id to default blacklight params

eddierubeiz:
Ref #3072


eddierubeiz:
I'm having trouble believing the tests pass after such a drastic change.

We'll need to double-check that this is indeed *not* fetching the extra fields from SOLR, and definitively rule out the possibility that Solr or Blacklight is simply ignoring the `fl=id`, and the PR effectively changes nothing. I've done only basic tests to determine that this is indeed as advertised, and we should definitely examine a low-level trace of info sent to and from SOLR to confirm I'm not deluding myself.

eddierubeiz:
In dev, in `RSolr::Response#initialize_solr_response`, I compared the value of `@response.to_s.length` for identical searches on this branch and on `master`.

On master:
	4,056,125 chars
On this branch:
	13,582 chars

So roughly three hundred times less data to fetch from SOLR. We should investigate this further.

jrochkind:
Nice, thanks for the clear write-up and investigation in the Issue!

I wonder what the `fl` was before we added this, maybe `*`, literally every stored field?  Curious where the default was coming from, but no matter.

PR #3080, opened 2025-08-13
Main search form, do not send hidden "search_field" input

jrochkind:
This will keep the url cleaner, without unnecessary `search_field=all_fields` in it. This was probably copied from Blacklight but is not necessary.

* We don't use multiple search fields
* Even if we did, this would be the default and not necessary
* It keeps BL logic from recognizing a blank query as an 'unfiltered' query, so is complicating our efforts to allow unfiltered queries through bot challenge


eddierubeiz:
Great. Thank you!

Issue #3075, opened 2025-08-11
Simple phrase triggers `Blacklight::Exceptions::RepositoryTimeout`


eddierubeiz:
In both prod and staging, phrases like
- `Relief of Persons Deprived of the Use of Their Reason` 
   - This is perhaps due to presence of the item `Conversations on Chemistry: In Which the Elements of That Science are Familiarly Explained and Illustrated by experiments and plates`, which searching all by itself is sufficient to trigger timeout on a freshly loaded index. This is a 422 page book, whose total OCR text is around 760KB
- `Periodic table of elements in an irregular spiral`
- `Radioactivity and the parts of an atom`

reliably cause a repository timeout.

(Our current BlacklightSolrRepository retries twice at 300ms and 600ms; after that it throws an error).

The underlying call to SOLR seems fine, and does not cause any SOLR errors.

It's just slow - typically over two seconds.

```
[...]

PR #3082, opened 2025-08-13
[WIP] Add stopwords to fulltext field types `text_de` and `text_en`.

eddierubeiz:
Ref #3075 

SOLR was taking a long time to calculate phrase highlighting for search phrases that mixed common words and stopwords -- just the phrase highlighting took 2129ms for the example phrase `relief of person and`, which triggered a 500 error.

Adding a stopword filter to field types  `text_de` and `text_en` means that three of the four fields we use to produce search highlights, namely `searchable_fulltext_en`, `searchable_fulltext_de` and `description_text4_tesim` now use stopwords.

(`searchable_fulltext_language_agnostic` still doesn't use stopwords, but it contains very few documents.)

Phrase highlighting calculations for the same search phrase now take only 71ms on average.

### DIFF
--- a/solr/config/lang/stopwords_de.txt
+++ b/solr/config/lang/stopwords_de.txt
@@ -1,0 +1,294 @@
+ | From svn.tartarus.org/snowball/trunk/website/algorithms/german/stop.txt
+ | This file is distributed under the BSD License.
+ | See http://snowball.tartarus.org/license.php
+ | Also see http://www.opensource.org/licenses/bsd-license.html
+ |  - Encoding was converted to UTF-8.
+ |  - This notice was added.
+ |
+ | NOTE: To use this file with StopFilterFactory, you must specify format="snowball"
+
+ | A German stop word list. Comments begin with vertical bar. Each stop
+ | word is at the start of a line.
+
+ | The number of forms in this list is reduced significantly by passing it
+ | through the German stemmer.
+
+
+aber           |  but
+
+alle           |  all
+allem
+allen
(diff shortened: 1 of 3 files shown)


jrochkind:
Your descirption says `phrase highlighting`, which made me think maybe phrase search was going on, like a query in double quotes -- but that's not true right, the problematic queries were not double-quoted phrases, nothing actually "phrase" going on here?

So. Having different stopwords for different fields that are all included in our initial search produces very weird results at the query level. It's confusing. 

Looks like i wrote it up in 2010!!  https://bibwild.wordpress.com/2010/04/14/solr-stop-wordsdismax-gotcha/

And another variation in 2011: https://bibwild.wordpress.com/2011/06/15/more-dismax-gotchas-varying-field-analysis-and-mm/

AND here is where we ran into it locally and it was super confusing for us: https://github.com/sciencehistory/scihist_digicoll/issues/1883

Looks like there is a solr `mm.autoRelax` that may work around. https://solr.apache.org/guide/6_6/the-extended-dismax-query-parser.html#TheExtendedDisMaxQueryParser-Themm.autoRelaxParameter

This solution on it's own is not acceptable, needs more investigation.  Either reproduce problem and confirm mm.autoRelax solves it. Or perhaps figure out a way to remove stopwords only from highlighting query, and different ones for different fields -- should be possible. 

Also seems worth asking some solr experts if this performance hit is to be expected, and if there's anything that can be done other than removing stopwords.

Issue #3088, opened 2025-08-21
Improve PageSpeed Insights, acutal experienced front-end performance


jrochkind:
Improviing our PageSpeed insights performance score  (https://pagespeed.web.dev/) will:

* Improve our user's experience, resulting in faster perceived page load
* Improve our Google SEO, as Google uses this for ranking

We have been chipping away at this for some time, but some remaining things that are tacklable:

- [ ] The standard way of loading adobe typekit fonts is something google really doesn't like, leading to blocking of page load until fonts are loaded by browser. `Requests are blocking the page's initial render`. A bit tricky to solve, may resolve flash of unstyled text, may involve attention to fallback fonts, but there are ways. 

- [ ] use "preconnect" link to certain origins. https://developer.chrome.com/docs/lighthouse/performance/uses-rel-preconnect/?utm_source=lighthouse&utm_medium=devtools . For us, typekit font ones, and maybe also our derivatives bucket. 

- [ ] "largest contentful paint" (LCP) is our large "hero"/poster image -- giving it a `fetchpriority=high` to make it show up quicker will help our score, although it's frankly unclear that it matters much to the user, I don't think it will hurt much. 

- [ ] Deliver content gzip'ed when user agent asks for it -- will reduce transfer time a lot, PageSpeed is asking for it. Some additional CPU to compute, but we're doing great for CPU since solving various problems and bot challenge, I think it'll work out great.  Can do with Rack::Deflater middleware. 

- [ ] Consider switching our thumbnails from JPG to WebP, is something PageSpeed asks for, WebP is smaller for same quality -- there may be some very old browsers taht can't handle it, may annoy users? Needs some consideration.

PR #3089, opened 2025-08-21
Add link rel=preconnect for performance for certain additional origin domains

jrochkind:
Does seem to imrpove our PageSpeed score, makes sense.  Ref #3088

 https://developer.chrome.com/docs/lighthouse/performance/uses-rel-preconnect/ 

- link rel=preconnect on typekit font domains to try to improve perf of custom adobe font use, per lighthouse report. Typekit default load is kind of a performance problem, but this at least helps as a first step. 
- link rel=preconnect cloudfront hostname for derivatives (thumanils) if present for performance, since we'll be loading lots of thumbs from there later in page, this does seem to ehlp. 

Actually this by itself has pretty limited effect on PageSpeed scores, but it does check off one thing PageSpeed is telling us to do, and i don't think should hurt. May have synergy with other changes.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,6 +28,11 @@
     <!-- Internet Explorer use the highest version available -->
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
+    <!-- try to improve performance of typekit fonts noted as perf hit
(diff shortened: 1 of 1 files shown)


eddierubeiz (approved):

eddierubeiz (approved):
I thought I'd approved this already - oops!

PR #3094, opened 2025-08-25
try to restore Github Actions CI incremental view output in color

jrochkind:


jrochkind:
nope doens't work. grr.

PR #3099, opened 2025-08-28
Fix small template bug in admin asset files page

eddierubeiz:
Ref #3098 
Small bug, small fix.


eddierubeiz:
Self-merging as this is a bug and the fix appears to work in staging.

jrochkind:
Thanks! Must have broken with the ViewComponent 4 upgrade, not caught by a test.

PR #3106, opened 2025-09-02
Separate two worker dyno pools, one worker per dyno

### LINKED ISSUE #3105: Move our background jobs to one worker per heroku dyno
Right now our background job system runs TWO concurrent redis workers on a single heroku `standard-2X` dyno. 

It has been _often_ triggering heroku `Error R14 (Memory quota exceeded)` -- two workers on a single standard-2X, depending on the nature of the work being done, is too much RAM. 

We actually already several types of queues, including a `on_demand_derivatives`.  We have too-complicated logic to try to make sure taht ingest work piled up in the queue never totally starves out `on_demand_derivatives` work. This is hard to explain, which shows it's too complicated. 

We're going to switch to having _two heroku worker dyno types_, a standard worker and an `on_demand_derivatives_worker`, _just_ for that queue. 

* _Both_ will have only _one_ worker per dyno -- still running on standard-2X -- giving them a lot more RAM headspace. 
* _Both_ will have _one_ worker running always on standby -- before the one single kind of worker had _two_ workers on standby, so our standby cost is the same. 
* Both will need to be managed by hirefire separately, and will autoscale them more straightforwardly knowing one worker per job, needing hirefire config to know which queue goes to which job type.

jrochkind:
Closes #3105

- separate heroku dyno types for our separate queue types
- move existing special worker pool config to subdir for neatnes

Create a new dyno type in Procfile, that will be a worker reading from a new resque pool config, that has it reading only from the (pre-existing) `on_demand_derivatives` queue. 

Remove the `on_demand_derivatives` queue from the general worker. 

Each dyno type is intended to only run _one_ resque worker process. Hirefire and heroku env config must be adopted ot match. 

## After merge and deploy

- [x] (actually this is already this way!) heroku env, `heroku config:set ON_DEMAND_JOB_WORKER_COUNT=1 REGULAR_JOB_WORKER_COUNT=1 -r production`
- hirefire `worker` manager
  - [ ] `Minimum Dyno Quantity` from 2 to `1`
  - [ ] `maximum dyno quantity` from 12 to 20
  - [ ] `ratio` from 2 to 1
  - [ ] `Notification quantity` from 2 to `
 - [ ] ENABLE `on_demand_derivatives_worker` manager
 - [ ] Create some on-demand-deriv work, and some ingest work, and watch dynos scale appropriately!
 - [ ] Ensure wiki page is updated https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/1915748368/Heroku+Operational+Components+Overview


eddierubeiz (approved):

jrochkind:
`worker` hirefire manager config, before and after this PR. (there does seem to be a hirefire terraform provider, that we coudl move to if we wanted. Now that we have an additional worker manager, it may be getting complex enough to consider). 

## before
worker-before


## after

worker-after

PR #3110, opened 2025-09-03
Set `hl.weightMatches` to `false`

eddierubeiz:
Ref #3075
Produces a noticeable improvement in highlighting speed.
Suggested by Trey Pendragon on the Code4Lib #solr channel.
Tested in staging, a bit.


eddierubeiz:
Yes- the words are highlighted individually with this setting - each gets its own `<em>` tag.

PR #3117, opened 2025-09-08
Rack prefers :unprocessable_content arg to :unprocessable_entity

jrochkind:
This started producing deprecation warnings when I tried upgrading other dependencies. Not sure exactly why the deprecation warnings just started showing up as this seems to have been deprecated since rack 3.1, which we were already on: https://github.com/rack/rack/commit/64ad26e3381da2ce1853638a2c4ea241c2ad3729. See also: https://github.com/rack/rack/pull/2137

But Rails seems to be dealing with it too: https://github.com/rails/rails/blob/c6272da25fa823d26f8469ffef3a9dc7a1225d2c/actionpack/lib/action_dispatch/http/response.rb#L51

Easy enough for us to change, all tests still pass (verified that with an unrecognized symbol they fail), let's do this first before further dependency upgrades.

### DIFF
--- a/app/controllers/admin/assets_controller.rb
+++ b/app/controllers/admin/assets_controller.rb
@@ -65,7 +65,7 @@ class Admin::AssetsController < AdminController
         format.json { render :show, status: :ok, location: @asset }
       else
         format.html { render :edit }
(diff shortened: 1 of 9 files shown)


eddierubeiz:
Thanks!

Issue #3115, opened 2025-09-05
Investigate document cache settings for SOLR


eddierubeiz:
Re: #3075 
Advice from Eric Hatcher, from #solr channel on Code4Lib Slack: _Another thing to consider: document cache - that might help keep stored fields available quicker._
https://solr.apache.org/guide/solr/latest/configuration-guide/caches-warming.html#document-cache

jrochkind:
Awesome. When you say "the cache" -- there are actually several different kinds of caches we might want to use eventually, none of which seem configured at the moment.

PR #3124, opened 2025-09-16
Document cache for solr

eddierubeiz:
Ref #3115


eddierubeiz:
Add a ram maximum setting as well before merging this.

eddierubeiz:
We decided to lower the number of docs to 256, at least for the time being; heap memory usage was on the high side with 512 documents in the cache.

See #3163

Issue #3076, opened 2025-08-12
Consider functionality to add or remove works from a digitization queue item.


eddierubeiz:
We're going to have larger and larger DQ items (50 + items) in the future. Jahna would like to talk about possibly adding a way to EDIT the list of works associated with a particular DQ item. Here's a good example:

https://digital.sciencehistory.org/admin/digitization_queue_items/1431

This DQ item has 16 works so far, but by the end of the week it could have 50.

Jahna uses this workflow to automatically add metadata to all works generated from a DQ item, and keep track of deadlines, etc., so this would be really helpful.

eddierubeiz:
One way we could do this would be to add two buttons to the DQ item show page:
- "Add all attached works to my cart"
- "Set attached works to the contents of my cart"

jrochkind:
Admin can definitely add and remove things from the digitization queue already. But need more convenient ways to do it, I think is the ask?

eddierubeiz:
Oh, I need to know how to do this! I couldn't figure it out.

PR #3126, opened 2025-09-17
Edit works attached to dq item

eddierubeiz:
Ref #3076
This PR proposes a way to edit the list of works associated with a digitization queue item:

* Clear your cart
* Export the works from the DQ item to the cart
* Add or delete works from the cart until it contains the works you want on the DQ
* Replace the works in the DQ with the contents of your cart.

The UI is clunky for now.

### DIFF
--- a/app/controllers/admin/digitization_queue_items_controller.rb
+++ b/app/controllers/admin/digitization_queue_items_controller.rb
@@ -1,6 +1,6 @@
 class Admin::DigitizationQueueItemsController < AdminController
   before_action :set_admin_digitization_queue_item,
-    only: [:show, :edit, :update, :destroy, :add_comment, :delete_comment, :destroy]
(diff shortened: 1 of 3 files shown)


jrochkind:
The "attach from cart" button shows up for me as `Attach ...` which is kind of confusing?

Screenshot 2025-09-29 at 2 30 12 PM

Could say "Attach items from Cart" or something?  Does this location of the button work for @JAuerbach26 ?

I like the "de-attach this work" confirm, but prob should include the title of the work to extra avoid any mistake? `De-attach work 'This is the title'?`. ?  And use a font-awesome icon instead of unique svg? 

Thoughts?

eddierubeiz:
Good suggestions, thanks! Included title in the confirmation message, and also fixed the inline SVG; using a font-awesome icon instead.

PR #3127, opened 2025-09-17
Add docValues to Solr id and model pk fields

### LINKED ISSUE #3120: "Add search IDs to cart" causes SOLR timeout, even with only ~300 search results to add.
https://digital.sciencehistory.org/all_search_result_ids/add_to_cart?q=slide+rule just caused a SOLR timeout. I really want this to be as efficient as we can possibly make it.

I want to:

* request as few fields as possible from SOLR (it should really be just IDs, if I'm remembering correctly)
* explicitly turn off highlighting and faceting, just in case.
* anything else?

The code is actually pretty simple and lives in `AllSearchResultIdsBuilder`.

## Backtrace

    line 38 of [PROJECT_ROOT]/app/controllers/all_search_result_ids_controller.rb: AllSearchResultIdsController#all_ids_from_search
    line 13 of [PROJECT_ROOT]/app/controllers/all_search_result_ids_controller.rb: block in AllSearchResultIdsController#add_to_cart
    line 12 of [PROJECT_ROOT]/app/controllers/all_search_result_ids_controller.rb: AllSearchResultIdsController#add_to_cart

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/58989/faults/123503740)

jrochkind:
Fixes #3120

**don't merge until you are ready to do the downtime with cleanup outlined below**

it turns out that our large full text stored fields really add a performance penalty to fetching documents -- *even if* your solr `fl` field list doesn't include the large stored fields.  This isn't necessarily a big deal for our normal UX searches with 25 records a page -- but in "add search to cart" function, we were trying to pull thousands of docs out of solr. 

It turns out that if *all* your fields in the `fl` fieldlist have docValues so Solr can use this instead of the actual "stored field" disk representation -- you can avoid the penalty. 

The relevant query here has only ONE field in `fl`, `model_pk_ssi` -- so just have to make sure it has docValues, and we're good!  As long as we're at it, give our pk `id` field docValues too, because we might want ot use it in similar ways (it actually holds our `friendlier_id`). 

Thanks to @mgibney for the hint!

## Deploy

When changing docValues you need ot completely erase and rebuild your solr index, you can't do it with an in-place reindex. 

There are ways to do this iwthout downtime by juggling multiple solr collections/cores in and out, but we'll just do it with 10-15 minutes of downtime instead and avoid the headache. 

1. [ ] maintenance mode on
2. [ ] merge and deploy to heroku with these changes
3. [ ] delete "collection" in solr admin dashboard
4. [ ] `heroku run -r production rake scihist:solr_cloud:create_collection scihist:solr:reindex`
    -  (indexing is going slow these days, this full reindex could take up to 7-8 minutes)
5. [ ] maintenance mode off

### DIFF
--- a/app/models/search_builder/all_search_result_ids_builder.rb
+++ b/app/models/search_builder/all_search_result_ids_builder.rb
@@ -13,13 +13,16 @@ class SearchBuilder
 
       # This method runs on all searches, but we exit early unless we actually want this.
       return unless scope.context[:all_search_result_ids] == 'true'
-  
+
       solr_parameters.delete_if do |k, v|
         k.start_with?('facet')   || k.end_with?("facet.limit") ||
-        k.start_with?("hl")      || 
-        k.start_with?("stats")  
+        k.start_with?("hl")      ||
+        k.start_with?("stats")
       end
 
+      # It's important for performance that ALL fields in `fl` (in this case just one)
+      # have `docValues=true` in solr schema, to avoid major stored field performance hit
+      # in this query that could have thousands of results.
       solr_parameters.merge!({fl:"model_pk_ssi", rows: '10000000', hl: 'false', facet: 'false'})
     end
   end
--- a/solr/config/schema.xml
+++ b/solr/config/schema.xml
@@ -245,8 +245,7 @@
     -->
     <field name="_version_" type="long" indexed="true" stored="true"/>
 
-
-    <field name="id" type="string" stored="true" indexed="true" multiValued="false" required="true"/>
+    <field name="id" type="string" stored="true" docValues="true" useDocValuesAsStored="true" indexed="true" multiValued="false" required="true"/>
     <field name="timestamp" type="pdate" indexed="true" stored="true" default="NOW" multiValued="false"/>
 
     <!-- this was dynamic typed *_ssi, but we override specifically to add docValues, so we can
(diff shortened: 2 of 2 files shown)


jrochkind:
OK trying to do it without downtime instead, wiht a temporary instance

`$temp_solr_url` means solr url, with username/password credentials, pointing to NEW temporary solr we create, `scihist-digicol-temp` in Searchstax

`$prod_solr_url` means solr url, with username/password credentials, pointing to actual long-term prod solr, in searchstax `scihist_digicoll_production_solr_9_8_1`

- [ ] Create temporary solr instance
- [ ] Index to it:
    -  `heroku run SOLR_URL=$temp_solr_url rake scihist:solr:reindex`
- [ ] note and record current production url, `heroku config:get SOLR_URL -r production`
- [ ] point production at temp solr `heroku config:set SOLR_URL=$temp_solr_url -r production`
    - [ ] Make sure we've waited long enough to take effect (how to tell?)
- [ ] Merge this PR to master, pull locally, on branch master
- [ ]  delete "collection" in solr admin dashboard for production
- [ ] rebuild and index to solr production:
   - `heroku run SOLR_URL=$production_solr_url` -r production rake scihist:solr_cloud:create_collection scihist:solr:reindex
   - (indexing is going slow these days, this full reindex could take up to 7-8 minutes)
- [ ] Point production back to solr production:
  - `heroku config:set SOLR_URL=$prod_solr_url`
- [ ] Once change goes through (how to tell?) delete temp solr

PR #3128, opened 2025-09-26
Add a sizes attribute to the homepage hero image tag

eddierubeiz:
Ref #2647

This appears to work nicely, but i had to hardcode the breakpoint value in the template. Not a huge deal, as long as it's well documented.

### DIFF
--- a/app/components/home_page_hero_image_component.html.erb
+++ b/app/components/home_page_hero_image_component.html.erb
@@ -6,6 +6,13 @@
       [path_1000, "1000w"], # scaled to 1000 width
       [path_2000, "2000w"], # scaled to 2000 width
     ],
+
+
+    # This max-width: 992 value is the 'lg' breakpoint in Bootstrap.
+    # See also homepage.scss.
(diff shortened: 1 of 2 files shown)


jrochkind:
thanks for comments on hard-coded constants. Might be some clever nicer way to do this, but not obvious, this is good enough for now!

jrochkind:
I forgot I wanted to test this one and just merged it. And now trying to test it on prod, I realize I'm confused about _how_ to test it!

Did you test @eddierubeiz, find a way to confirm what image is being loaded by chrome?  I want us to check, but realize I'm not entirely sure how.

eddierubeiz:
Here's what I tried: for various window sizes, you can hover over the image tag in the `elements` tab of the Chrome `DevTools` window and observe the size of the file and its dimensions. It does appear to work.

jrochkind:
OK I was trying to hover over but got confused, it looked like it was just showing me *all* the different SRCs on hover depending on where I was hovering... for now, I'm just gonna go with it I think, thanks!

PR #3131, opened 2025-09-29
Revert "Add docValues to Solr id and model pk fields"

eddierubeiz:
Reverts sciencehistory/scihist_digicoll#3127 .

This is in response to https://github.com/sciencehistory/scihist_digicoll/issues/3130, which I believe was indirectly caused by #3127 .


jrochkind:
Looks unnecessary, we deleted and rebuilt index instead. 

Somehow my attempt to rebuild index with no downtime went awry, I got confused about something, and we ended up with an inconsistent invalid index -- not totally coming up with an expalnation of how this could have happened, but it def appears fixed now.  🧿

Issue #2787, opened 2024-11-19
Fixity report is too slow


eddierubeiz:
The [fixity report](https://digital.sciencehistory.org/admin/fixity_report) is taking way too long. See https://github.com/sciencehistory/scihist_digicoll/issues/2684 for details.

I'm sure we can find a way to speed it up, perhaps by showing less information.

jrochkind:
We also should see if any staff are actually using this report; if not, and they are not interested in doing so, we maybe want to disable it instead of investing further in it?

eddierubeiz:
We could do something like the [storage](https://digital.sciencehistory.org/admin/storage_report) and [orphan](https://digital.sciencehistory.org/admin/orphan_report) reports - calculated nightly.

PR #3159, opened 2025-10-16
Simplify fixity report template by assigning @stale_in_days once in the controller

eddierubeiz:
One of several small refactors that were unrelated to my main quest in ref #2787


jrochkind:
It's not clear to me why this simplifies anything.  If we wanted to change the value, we had one place to change it before, and one place to change it now. 

There might be something I'm missing the change or not considering! Can you say what do you see as the simpfliication of putting the constnat in an iVar instead of referring to the constant directly?

PR #3160, opened 2025-10-16
Tying up loose ends: refactors to the fixity rake task

eddierubeiz:
I'm tying up a couple of loose ends after working on  ref #2787 . These are small refactors that should make things easier to read, manage, and debug.

- Rename `need_checks_assets_relation` to `overdue_assets`
- Move the method to `app/lib/scihist_digicoll/assets_needing_fixity_checks.rb` where it will be easier to find
- Removing the `STALE_IN_DAYS` constant from the report class;  bad practice to assign a constant to another constant.
- Moving both fixity check rake tasks into the `check_fixity` scope
- Create a new task, `prepare_report`, to prepare a report every night.
- Create a new task :nightly in the shared scope that runs first `check_fixity`, then `complete_overdue`, and then `prepare_report`. This moves a bit of infrastructure out of Heroku and into the code.

# Note
- [x] When this is deployed, the fixity check scheduled rake task in Heroku Scheduler can be simplified
## from
`$ bundle exec rake scihist:check_fixity scihist:check_fixity:complete_overdue && curl -s CHECKIN_URL`
## to
`$ bundle exec rake bin/rake scihist:check_fixity:nightly && curl -s CHECKIN_URL`

### DIFF
--- a/lib/tasks/check_fixity.rake
+++ b/lib/tasks/check_fixity.rake
@@ -4,11 +4,11 @@ namespace :scihist do
     desc """
     These three tasks should be completed nightly by the Heroku scheduler.
     * 1) Check a fixed number of assets;
-    * 2) Check any 'overdue' works that have not been checked in a certain number of days;
+    * 2) Check any 'overdue' works that have not been checked in a certain number of days (under normal circumstances, very few);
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Updated the nightly rake task.

PR #3164, opened 2025-10-21
[WIP] Copy half (possibly all) the items in the old cache to the new one when a new one is created.

eddierubeiz:
From watching the document cache stats in production, it appears that the cache is emptied several times a day, and typically doesn't have time to fill up before it disappears.

The cache is getting regenerated every time we create a new IndexSearcher, which is every time we save a work or collection.

If the cache's lifespan isn't long enough for anything to be evicted, it's arguably not all that useful. _(later update: we were able to get up to ~7000 evictions overnight, and close to a 15% hit rate, but the cache was wiped when I saved an work in production.).._

I'd like to see if setting `solr.documentCache.autowarmCount` to 128 or even 256 (copy half (or all) the documents in the old cache into the new one when it's created) will result in a longer effective lifespan for at least some of the documents in the cache.

See: https://solr.apache.org/guide/solr/latest/configuration-guide/caches-warming.html


eddierubeiz:
Based on more testing in dev, this is a really bad idea. We're not doing it. This PR will result in some really creepy SOLR behavior.

PR #3186, opened 2025-11-04
Refactor: editors can in fact publish

eddierubeiz:
As often happens, tidying up the test made a few small problems a lot easier to see. This PR actually fixes them.

In particular I noticed that the test gave the impression that editors cannot publish works. They can, so now that test is under "editor". The only thing an admin can do that an editor can't, basically, is delete things.

More broadly:
- each context associated with a user type has the appropriate `logged_in_user` argument
- each "can" test is at its lowest permission level:
  - things that you can do as an editor (but not as an admin) are listed under "editor".
- each "can't" test is at its highest permission level:
  - things that you _can't_ do as an editor (but can as an admin) are listed under "editor".

```
RSpec.describe Admin::WorksController, type: :controller, queue_adapter: :test
```

**Not logged in**
```
  context "not-logged-in user", logged_in_user: false
    it "redirects to login"
```

**Logged in**
```
  context "logged-in user", :logged_in_user
    it "shows page"
    context "#demote_to_asset"
      context "work not suitable"
        context "because it has no parent"
          it "rejects"
        context "because it has multiple assets"
          it "rejects"
    context "sorting, filtering and pagination"
      it "sorts correctly by title"
      it "filters by partial title"
      it "uses default sort with only include_child_works in the params"
      it "can filter on work format using work_format param"
      it "can filter on work format using work_format param"
      context "sql escaping"
        it "matches title correctly despite quotes"
      context "sql escaping"
        it "matches external_id correctly despite quotes"
      context "filtering by external_id"
        it "can filter on external id, regardless of the category of the ID"
    context "logged-in user cannot"
      it "can not publish"
```

**Editor**
```
  context "editor user", logged_in_user: :editor
    context "#update"
      it "trims leading and trailing spaces"
      it "does NOT trim from specifically excluded fields"
      it "queues a WorkOcrCreatorRemoverJob when ocr_requested is turned off"
      it "does not delete OCR if the save operation fails"
    context "#reorder_members"
      it "can reorder by arbitrary order"
      it "can reorder alphabetically"
    describe "publishing and unpublishing"
      describe "publishing"
        context "work has assets with invalid files"
          it "refuses to publish"
        it "can publish, and publishes children"
        it "does not change unpublished children unless requested"
      context "work missing required fields for publication"
        it "can not publish, displaying proper error and work form"
        describe "child work missing required fields"
          it "can not publish, displaing proper error for child work"
      context "published work"
        it "can unpublish, unpublishes children"
        it "does not change published children unless requested"
    context "editor user cannot"
      it "can not delete"
```

**Admin**
```
  context "admin user", logged_in_user: :admin
    it "can delete, and deletes children"
```

### DIFF
--- a/spec/controllers/admin/works_controller_spec.rb
+++ b/spec/controllers/admin/works_controller_spec.rb
@@ -238,6 +238,114 @@ RSpec.describe Admin::WorksController, type: :controller, queue_adapter: :test d
       end
     end
 
+    describe "publishing and unpublishing" do
+      # works that have the necessary metadata to be published, but aren't actually published yet
+      let(:work_child) { build(:work, :published, published: false) }
+      let(:asset_child) { build(:asset_with_faked_file, :tiff, published: false) }
+      let(:work) do
+        create(:work, :published, published: false, published_at: nil, members: [asset_child, work_child])
+      end
+
+      describe "publishing" do
+        around do |example|
+          freeze_time do
+            example.run
+          end
+        end
+
+        context "work has assets with invalid files" do
+          let(:corrupt_tiff_path) { Rails.root + "spec/test_support/images/corrupt_bad.tiff" }
+          let(:bad_asset) {create(:asset, :inline_promoted_file, file: File.open(corrupt_tiff_path))}
+          let(:good_asset) {create(:asset, :inline_promoted_file) }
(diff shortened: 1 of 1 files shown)


jrochkind (approved):

eddierubeiz:
Yeah - thanks. Very gratifying to tidy up this test - been meaning to do this since the pandemic, not kidding.

Issue #723, opened 2020-04-23
Add github sha to backups


sanfordd:
When we make changes to the database schema, it can cause breaking issues with our postgres backups. We should start tagging the backups to a specific github sha note so we can restore a backup to the version of the app it was made under.

jrochkind:
And really we're not totally sure this is sufficient to cause the problem, it's just one suggested solution and might be a component but nto totally sufficient for solution. 

Larger problem is work out how we handle schema changes (which happen regularly) and backups, do some tests involving schema changes, etc. 

Also consider introducing a routine to regularly restore from older backups (instead of restoring from the most recent one as we do in our staging sync), to test our process better and find out about problems like this from practice instead of relying on us to just thin of them.

sanfordd:
Good points there, I do think testing both older in-AWS versions and possibly getting a workflow made for recovering from our on-site backups would be worth doing.

PR #3188, opened 2025-11-05
Store sha hash in the backup file and in the s3 metadata

eddierubeiz:
Ref #723 

Assumes you've run 
```
heroku buildpacks:add https://github.com/ianpurvis/heroku-buildpack-version
```
This is tricky to get right the first time but I've tested the changes enough in staging.

Deploy sequence
- [x] Add  buildpack in staging and production
- [x] Deploy
- [x] Leave to run overnight
- [x] Double-check no errors and sha hash in backup file
- [x] Updated [buildpack documentation](https://sciencehistory.atlassian.net/wiki/spaces/HDC/pages/1915748368/Heroku+Operational+Components+Overview#Buildpacks)

### DIFF
--- a/lib/tasks/copy_database_to_s3.rake
+++ b/lib/tasks/copy_database_to_s3.rake
@@ -43,7 +43,6 @@ namespace :scihist do
     dump_command = [
       'echo', "\"-- GIT SHA:\"",                 '&&',
       'echo', "\"-- #{ENV['SOURCE_VERSION']}\"", '&&',
-
       # --clean means "include DROP commands at the top of the file."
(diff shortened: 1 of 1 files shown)


jrochkind:
I think the git sha of the current deploy may be available in a heroku variable, does that help?

Googling says it may be available as ENV `SOURCE_VERSION` at *build* step, but not normally available after build step... but there may be a buildpack you can install to make it available? (Alternately I suppose you could have something run at build step to try to preserve it somewhere).

eddierubeiz:
Interesting advice - will look into it - thanks!

PR #3205, opened 2025-11-20
remove gem activerecord-postgres_enum, not needed since Rails 7

jrochkind:
Gem for recording postgres enum types in schema.rb.  not needed since Rails 7.  That it was still there was resulting in double inclusion of the enum type in a dumped schema.rb, which we were assiduously avoiding committing, but it was annoying. Looks like this fixes it.

Slightly differnet form of statement in schema.rb, fixed now. Slightly different best practice in migration that uses it, fixed there too.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -158,9 +158,6 @@ gem 'sitemap_generator', '~> 6.0' # google sitemap generation
 
 gem 'sane_patch', '< 2.0' # time-limited monkey patches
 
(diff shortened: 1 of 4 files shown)


jrochkind:
Yep that was a feature supported by the gem that is not supported by Rails -- and when I look it up, it seems like a really bad idea we didn't actually want!  (not sure if we have it in db now, I am choosing not to worry about it...)

PR #3227, opened 2025-12-16
Attempt to simplify and shorten claude prompt

jrochkind:
Based on review, experience, and advice from Claude itself. 

The "self check" section is probably doing nothing. Too much redundancy.   Got rid of extensive "more chunks" prediction, we'll have to try again focusing on it, but leaving in something basic for now. 

Not trying to improve anything in what's produced, just a step to try to make it simpler/shorter without making it worse. 

Hard to figure out what it changes! Or if it's better or worse!  I think some of these are a bit better and some of these a bit worse, it seems mostly about the same, important to simplify.


jrochkind:
Did have a few problems with footnotes not matching up, so adding back direction: "Inline footnotes must correspond exactly to the footnotes included."

PR #3238, opened 2025-12-22
Abilty to set up non-OHMS OH chunks for vector search for AI chat

jrochkind:
We can't link to these very well, at present just to Work page. But still can form answers on them. 

There are no UI changes in here, just the ability to create "chunks" for PDFs that are NOT OHMS, based for now on the `searchable_transcript_source` attribute that was already present, with text we were using for searchability in Solr.   The app still doesn't automatically chunk OHs, just has the capability to do so manually. 

* Extract "Paragraph" model object to be consisted abstraction used regardless of transcript type
* Create a `OralHistory::PlainTextParagraphSplitter` that can split our plain text transcripts into paragraphs
* Generalize our TranscriptChunker to be able to chunk both legacy OHMS and plain text

Note:

* In plain text we can cite to "paragraph number", which is pretty useless for a human. We don't have a way to identify PDF page number. 

* Citation links to non-OHMS PDFs are at presnt just to work page.  (Could be to PDF page? Or an HTML version we haven't made yet?)

* We'll have to think about what UI we want here and how to get it.

### DIFF
--- a/app/services/oral_history/plain_text_paragraph_splitter.rb
+++ b/app/services/oral_history/plain_text_paragraph_splitter.rb
@@ -27,7 +27,9 @@ module OralHistory
       current_speaker_name = nil
       paragraph_index = 0
 
-      things = trim_transcript(plain_text).split(/(?:\r?\n\s*){2,}/).collect do |raw_paragraph|
+      # some transcripts have paragraphs split only by 2 `\r` -- like 90s MacOS?  Not sure
+      # where this comes from but okay. So two (or more) \r or two \n or two \r\n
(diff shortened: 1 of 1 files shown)


jrochkind:
Need to fix bug with OralHistoryContents: 470 , 131, 577, 127

```
app/services/oral_history/transcript_chunker.rb:226:in 'OralHistory::TranscriptChunker#build_chunk_record': undefined method 'speaker_name' for nil (NoMethodError)

      speakers = [list_of_paragraphs.first.speaker_name || list_of_paragraphs.first.assumed_speaker_name] +
````

Why is it creating a zero-paragraph chunk??

eddierubeiz (approved):

PR #3250, opened 2026-01-07
More clear labelling of chunks and paragraphs when providing as evidence to LLM

jrochkind:
To try to encourage more accurate and specific referents and citations. 

Every chunk will now have an (eg) "ORAL HISTORY ID: OH0627" metadata line. 

Every paragraph will begin with "[OH0627|P32] SPEAKER:" -- additional labelling of OH id and paragraph number on every paragraph; and every paragraph has a speaker label if we can provide one, not just assumed from prior paragraph.  

- move chunk formatting for LLM into ERB template
- add OH numbers to chunks metadata provided to LLM
- in chunks to LLM add OH and Paragraph number before every paragraph, to try to improve correctness and specificity of citations
- When creating Chunk records, make sure every paragraph in text begins with a speaker label, assuming from previous paragraph if needed

------

~After merge, need to re-create all chunks on staging (which will cost us maybe $20), to have properly labelled speaker paragraphs stored in chunk records~

Bah, nevermind -- right now, deleting/regenerating chunks will cause all bookmarked questions to disappear, we don't want to do that without thinking and ideally a solution. 

So this PR will include new chunk generation code, to ensure all paragraphs have a speaker label, but the chunks on staging won't necessarily reflect it yet. Any new chunks we make in future will. Other improvements in this PR will be immediate.

### DIFF
--- a/app/models/oral_history_content/paragraph.rb
+++ b/app/models/oral_history_content/paragraph.rb
@@ -42,10 +42,14 @@ class OralHistoryContent
     # If a paragraph does not begin with a `SPEAKER:` label (usually cause same as last
     # one), add it -- for LLM, helpful if every paragraph begins, no assumptions.
     def text_with_forced_speaker_label
-      if text =~ OralHistoryContent::OhmsXml::LegacyTranscript::OHMS_SPEAKER_LABEL_RE
-        text
+      # if text doesn't already start with speaker, and we HAVE a speaker to add,
+      # AND the text doesn't start with "[" which is usually used for labels like [END OF TAPE],
+      # then add a speaker.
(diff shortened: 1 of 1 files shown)


jrochkind:
~After merge, need to re-create all chunks on staging (which will cost us maybe $20), to have properly labelled speaker paragraphs stored in chunk records~

Bah, nevermind -- right now, deleting/regenerating chunks will cause all bookmarked questions to disappear, we don't want to do that without thinking and ideally a solution. 

So this PR will include new chunk generation code, to ensure all paragraphs have a speaker label, but the chunks on staging won't necessarily reflect it yet. Any new chunks we make in future will. Other improvements in this PR will be immediate.

Issue #3232, opened 2025-12-18
user generated/selected thumbnails for video works


archivistsarah:
It would be nice if staff users could override or replace the thumbnail and poster frame images that are currently being autogenerated upon ingest of video assets

Possible workflows:
a) after or upon ingest, select a frame from an mp4 to use as a thumbnail and poster frame image, using an interface or tool in the DC
or 
b) create and ingest a jpeg/png thumbnail image outside of the DC and add it to a video work as an additional asset, selecting it as a thumbnail from the dropdown in the metadata form

[disclaimer that these are just off the top of my head]

 Ideally the new thumbnail would replace both the auto-generated thumbnail that appears on a list/search results page to the left of the metadata _and_ replace the poster frame image that appears in the video player on the work's public page, before the viewer starts playing the file.

This came up when ingesting video that has color bars (with spoken audio) for the first minute of footage: https://digital.sciencehistory.org/works/vwrcmdp 
It would be nice if I could replace the current thumbnail/poster frame of the color bars with one more representative of the video and meaningful to a public viewer. There are other examples of the thumbnail not matching the content of the file, most commonly when the thumbnail is the person introducing the famous chemist giving a talk and not the famous chemist themselves. Example: https://digital.sciencehistory.org/works/tdd0nsg

Current example of what happens when you add a jpeg as a thumbnail to a video work is here: https://staging-digital.sciencehistory.org/works/bh994a0

jrochkind:
b) is _relatively_ easy to implement, a) would take more hours of development (which could be justified depending on how often it will be used and how much time it saves staff!)

jrochkind:
Makes sense -- just will take us a lot more time to implement  choice (a) the in-app frame picker, which could be justified if this is something you plan to do a lot of, but maybe not if it's a very occasional thing! If that makes sense.

PR #3251, opened 2026-01-12
Use representative as poster frame

eddierubeiz:
Ref #3232 

This PR allows you to add an image to a video work and designate it as the representative. This representative image will serve as both:
- the thumbnail in search result lists;
- the poster frame shown on the video work page.

Notes:
- We do not support using one image as the search result thumbnail and _another_ image as the poster frame.
- This PR requires you to set all video works to have the format `moving_image`. If a work doesn't have that format, the video will no longer play. If there is a video work in production that somehow doesn't have `moving image` checked, we won't know that it's a video and we won't show the video player until this is rectified.
  - (Before this PR, the rule for determining whether a work was a video or not used to be: `"If the representative is a video asset, then this must be a video"`.)


jrochkind:
OK, lez do it, thanks!

eddierubeiz:
Screenshot 2026-01-16 at 12 11 46 PM

Issue #3260, opened 2026-01-20
In search results, don't count a video poster frame asset in the total asset count


eddierubeiz:
<img width="732" height="201" alt="Image" src="https://github.com/user-attachments/assets/958cc4c2-f2a4-4df4-8cb5-3792e3f2c3e3" />

If a video has an image representative that's being used as a poster frame (as in https://digital.sciencehistory.org/works/kz5gm8a ),
don't show `2 items` under the thumbnail in search results. Only one of those items is of any interest to the public.

We may just want to suppress the asset count altogether for videos, given that:

- we don't display asset count for works with only 1 asset
- all video works currently have exactly one video asset.

PR #3262, opened 2026-01-21
Use child count of 1 for video assets in search results.

eddierubeiz:
Ref #3260
Ref #3261

We maintain a separate presenter with the sole purpose of calculating and displaying this count, so this somewhat hacky adjustment is pretty well contained.

It's worth noting that we are already treating video thumbnails in search results differently (by showing the extent string, which contains video duration.)

### DIFF
--- a/spec/components/search_result/work_component_spec.rb
+++ b/spec/components/search_result/work_component_spec.rb
@@ -98,8 +98,8 @@ describe SearchResult::WorkComponent, type: :component do
         members: [create(:asset), create(:asset), create(:asset)]
       )
     end
-    it "displays both extents and the asset count under the image" do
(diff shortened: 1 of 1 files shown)


jrochkind (approved):

eddierubeiz:
Correction: the wrong count will never be displayed, since we are hardcoding it to one, and we already have a rule not to display the count unless it's greater than one. If we decide to undo this PR and show child counts on video works, we will have to explicitly change the code.

(When in the future we have video works with more than one asset, the first problem we'll notice will be the duration metadata, which will now be incomplete since it currently displays only the `extent` metadata for the first video it finds.)

We haven't prioritized that work, but I'm sure we can come up with a solution when the time is right.

Issue #3276, opened 2026-01-28
OH AI project: Have answer be more aware of scope, avoid answering as if it looked at everything


jrochkind:
Ref #3178 overall project. came out of 1/27 workshop. 

To consider:

* Avoid saying things like "at least 3 scientists did X" when we may have dozens and dozens it isn't aware of
* Maybe it could explain why it can't give quantitative answers (how many, total, etc) or other things it can't do
* Be more clear when give questions that would require comprehension it can't do
* Avoid saying confusing things like "based on retrieved evidence", users won't know what that means. 

From workshop:

>  I got an “at least three” answer too for women in science when 98 interviews have the subject “women in science” in the DC: https://digital.sciencehistory.org/oral_histories/ask/e9b266ff-2211-4ebb-81e6-c5f1b4bc9378

> It was interesting to see that the bot declines to offer statistical analysis: https://digital.sciencehistory.org/oral_histories/ask/bba8a771-75bc-46c9-8ee4-0a680fbec040

jrochkind:
"List ALL interviewees who were presidents of professional societies?" -- needs a disclaimer.

PR #3282, opened 2026-02-03
Have OH LLM answers be more aware of scope limitations, with instructions and clearer access to OH counts

jrochkind:
Ref #3276

* We want to add real counts of oral histories searched to claude instructions, so the LLM can have an idea of what (small) portion of the total corpus it has analyzed, to be able to properly explain what it can do. 
   * want it to be live and accurate for sub-collection chosen from radio buttons  
   * Will use Rails caching though so we aren't looking it up every time
   * We extract the lookup that was previously used in the counts next to radio buttons, to a CategoryWithChunksCount service object
   * We use this to embed actual count into claude instructions
   
 * We move the count to the "user prompt" (that has the chunks)  rather than the "system prompt" 
   * So system prompt will remain more cacheable should we want to cache it on LLM side for efficiency later. 
   * To be closer to the chunks it's about, might help
   
* We expand instructions to Claude to ask it to provide clarification/disclaimers that it is only providing examples and can't provide exhaustive or quantitative results over the entire collection. 
  * It is instructed to use some judgement on when these instructions are necessary -- they do end up necessary for most of our sample questions, which seems reasonable. 
  * I did ask Claude itself for some advice on how to formulate these instructions; I don't accept it uncritically, I sometimes push back and ask Claude to change, and then adapt it myself when putting into code.  Especially to keep the instructions shorter, the prompt can't get too long or it inhibits the LLM following it, and originally the answer wanted me to add so much!  [See conversation here](https://claude.ai/share/5b2ab913-e1d3-477b-b5be-41f7315d5ab4)

### DIFF
--- a/app/services/oral_history/claude_interactor.rb
+++ b/app/services/oral_history/claude_interactor.rb
@@ -11,7 +11,7 @@ module OralHistory
     # claude sonnet 4.5
     MODEL_ID = "global.anthropic.claude-sonnet-4-5-20250929-v1:0"
 
-    ANSWER_UNAVAILABLE_TEXT = "I am unable to answer this question with the methods and sources available."
+    ANSWER_UNAVAILABLE_TEXT = "This tool was unable to answer this question using the resources available."
 
     # should e threadsafe, and better to re-use for re-used connections maybe
     AWS_BEDROCK_CLIENT = Aws::BedrockRuntime::Client.new(


jrochkind:
You know what, since the number varies, let's move it to user prompt, so we can later add some caching for system prompt.

PR #3287, opened 2026-02-04
LLM user task instruction had gotten out of date after #3280. Update.

jrochkind:
Consulted claude for advice, which included moving it to above chunks instead of below.


jrochkind:
Nevermind, we're going to fold this into #3282 , where we need to do more changes to user prompt.

Issue #3215, opened 2025-12-03
normalize volume in oral history audio derivatives


jrochkind:
Listening ot lots of OH audio, the volumes can be very diverse. 

We are already making a derivative for purpose of listening in browser -- stiching together multiple audio, and in a format we like for streaming listening. 

We can apply some volume normalization when we make this derivative. ffmpeg, which we're already using, has something built in designed for normalizing volume in speech, which is what we have! https://ffmpeg.org/ffmpeg-all.html#speechnorm

jrochkind:
Some of the audio has a lot of static/bg noise -- I suppose we could also consider doing a filter for noise reduction in the derivatives?

PR #3288, opened 2026-02-04
Normalize and compress audio for oral histories

eddierubeiz:
Context for this PR:

We’re doing a round of improvements on the processing for the combined audio derivatives (so e.g. in the [Choudhry interview](https://digital.sciencehistory.org/works/nt2ydj8#prevq=oral+history&tab=ohDownloads) that would be the Complete Interview Audio File as opposed to the 5 Separate Audio Segments. )

We definitely want to _normalize_ the audio (so, e.g. if the interviewer mistakenly records Segments 1 and 2 at a really soft level and the rest of the segments really loud, everything gets brought down or up to a comparable level before we stitch the audio together).

We’re also thinking of _compressing_ the audio as well. This means that if e.g. the interviewer and interviewee are a different distance from the microphone (resulting in one person’s speech being soft and another’s being loud) it will soften the loud bits and louden the soft bits.

We could test this out on an OH that you know to have some big audio-level discrepancies. We could also test the formula on e.g. a taped interview with loud buzzes and bumps due to subpar audio equipment.

Just to see if it makes everything easier to listen to.

Ref #3215

Targets:
- Integrated loudness (I):  -19 LUFS
- True peak (TP):           −1.5 dBTP
- Loudness range (LRA):      ~7 LU

This PR also has `ffmpeg` do a bit of audio compression, ensuring that the loud and soft parts within the file are brought closer to each other (within roughly 7 loudness units), in addition to the entire file being raised / lowered to  about -16 LUFS.

# Results

`ffmpeg -i combined_audio.m4a  -filter:a ebur128=framelog=verbose -f null -`

## Before:
```
Integrated loudness:
I:         -23.1 LUFS
Threshold: -34.4 LUFS

Loudness range:
LRA:        12.5 LU
Threshold: -44.4 LUFS
LRA low:   -32.3 LUFS
LRA high:  -19.7 LUFS
```

## After:
```
Integrated loudness:
  I:         -21.9 LUFS
  Threshold: -32.6 LUFS

Loudness range:
  LRA:         7.2 LU
  Threshold: -42.6 LUFS
  LRA low:   -26.5 LUFS
  LRA high:  -19.3 LUFS
```

### DIFF
--- a/app/services/combined_audio_derivative_creator.rb
+++ b/app/services/combined_audio_derivative_creator.rb
@@ -120,12 +120,10 @@ class CombinedAudioDerivativeCreator
     # see https://www.reddit.com/r/ffmpeg/comments/15kiucp/ffmpeg_how_to_add_dialog_normalization_to_ac3_file/
     # see https://ffmpeg.org/ffmpeg-all.html#speechnorm
 
-    # For podcast-like spoken word, a very standard target is:
-    #  Integrated loudness (I): −16 LUFS (stereo) or −19 LUFS (mono)
-    #  True peak (TP): −1.5 dBTP (conservative, avoids codec overs)
-    #  Loudness range (LRA): ~7 LU (keeps dynamics controlled for speech)
-
-
+    #  Target loudness
+    #  Integrated loudness (I):  −19 LUFS (mono)
+    #  True peak (TP):           −1.5 dBTP (conservative, avoids codec overs)
+    #  Loudness range (LRA):      ~7 LU (keeps dynamics controlled for speech)
     filtergraph = "concat=n=#{components.count}:v=0:a=1, speechnorm, loudnorm=I=-19:TP=-1.5:LRA=7:print_format=summary[aout]"
 
     # Finally, some output options:


jrochkind:
Code looks good! I was not expecting decision to use compression rather than just normalization.  Original ticket did not mention compression. 

Normalization changes the "average" volume of the track to match other tracks, preserving original dynamic range. 

Compression makes the highest and lowest volume parts closer together, reduces dynamic range. (traditionally radio broadcasts use high audio compression).

[Thanks to Eddie for pointing out the difference, and that we've chosen compression here -- I wasn't clear on it when we made the ticket, i will admit!]

I would think that normalization would be more true to the original source, while fixing the problem which I understood as: "I need to turn my volume up or down from usual when listening ot a video, or when I go from one videot ot he text, and some videos are too low to hear even at max volume on my device!" We don't really know for sure what causes that problem, but it could be idiosyncracies in digitization process etc. 

But it could be that normalization alone was not enough to fix that problem?  Or there are other reasons to prefer it?  Perhaps you have already discussed and chosen?  i am fine with whatever @archivistsarah and or oral history team choses, just want to make sure we've intentionally chosen, and documented why! 

Oh, and realizing only now this is for OH, not videos! -- we definitely need to run it by OH team before deploying changes.

eddierubeiz:
I took some time to examine four interviews: Rosenkranz; Battista; Herzberg; and Jones. I expected more dramatic changes to the overall loudness (normalization) and the loundness range (compression). The effects of the editing are actually quite difficult to notice, both to the ear and to `ffmpeg -i`.

```ffmpeg -i old.m4a  -filter:a ebur128=framelog=verbose -f null - && ffmpeg -i new.m4a  -filter:a ebur128=framelog=verbose -f null -```

There are a couple different ways to interpret these results, but first let me post them below.

Issue #3286, opened 2026-02-03
Recreate video derivatives


eddierubeiz:
We recently changed the presets we use (see the list of presets at [aws](https://us-east-1.console.aws.amazon.com/mediaconvert/home?region=us-east-1#/presets/list)) to include normalizing the audio (soundtrack) of the videos derivatives.

We'd like to redo the audio for the 223 videos that are currently in the digital collections. 

See also #3230

eddierubeiz:
"Normalization" in this context means that video with a very soft audio track will be boosted (in the version shown to the public) to a standard loudness level.

eddierubeiz:
# Costs per hour of video
The current cost (as of February 3rd) is 

`01:58:57` of video costs $1.78
1 * 60 * 60 + 58 * 60 + 57 = 7137 seconds of video cost $1.75
1 second of video costs $0.00024520106
One hour costs $0.88

# New cost per hour
$0.88272 x 1.17 = $1.03

PR #3298, opened 2026-02-17
Rake task to recreate HLS derivatives

eddierubeiz:
Ref #https://github.com/sciencehistory/scihist_digicoll/issues/3286 - see there for more information, including estimated cost.

Before doing this, we should:
- [x] come up with a rough estimate of the cost: $125
- [x] make sure we don't want to make any other tweaks to the presets.


jrochkind:
Do we need a new ticket and/or checkbox here to track the remaining work which needs to be done asap?  Completing the missing ones, and re-enabling the status update?

PR #3307, opened 2026-02-23
don't fetch all chunks, too expensive, we only need first, it's okay it is n+1

jrochkind:
oops forgot to push a very important optimization from #3306, necessary to actually run it!


jrochkind:
@eddierubeiz Another self-merge fix

PR #3310, opened 2026-02-26
OH AI, handle more unusual plain text transcripts

jrochkind:
We essentially need to "parse" plain text transcripts as part of chunking them for retrieval. Including having consistent paragraph numbers for citation. 

We had been doing that in a "good enough" way for a proof of concept, but when really digging in realized a bunch were not working well.  Turns out while I captured some consistent formatting features in a bunch of them, there were other subsets that were different, or individual idiosyncratic ones etc. 

Found a lot of these using a "validation" tool that will come in a subsequent PR. Still can't promise this is doing all of them correctly, but it's better than before!  We could consider the pipeline of where these transcripts come from to see if we can get more consistency. 

Some issues we found:

* Some transcripts only have a single newline between paragraphs instead of double -- this is a mess, we try to use recongizing `NAME:` as a splitter too. 

* some have abstracts some don't (none need this, as we have abstracts in metadata). Do a better job at skipping abstracts and other prefatory material. 

* Some lack the marker "[END OF INTERVIEW]" we were using to trim supplementary material (indexes and appendixes).  We now also look for alternate `[END OF TAPE]` that those identified do have instead. 

* Some had embedded timecodes!  Probably newer ones. We strip these from text, and actually keep them as metadata so we can cite to timecode for these. 

* bug in our code, we "remember" speaker label from past paragraph for unlabelled paragraph -- but were only doing that for one paragraph, need to do it for multiple adjacent unlableled. 

Some we found that we have NOT fixed here yet/infeasible to fix. 

* Some transcripts have a blank line at PDF page breaks, splitting a paragraph in two. 
* Some have page number copy/pasted in (they are still there, sorry)
* Schoemaker is a very atypical one which is not formatted like any others, and we really aren't quite treating right, punt on this one for now. `oj07kjf`

## after merge

let's wait until AFTER user tests, in case it goes wrong?

- [x] We're gonna re-create all chunks on prod (which may cost us $50 I dunno, but we gotta do it now)
  -  actually delete all chunks first, to make sure we delete any embargoed. `OralHistoryChunks.delete_all`. 
  -  `heroku run OVERWRITE_CHUNKS=true rake scihist:create_oral_history_chunks -r production` 
  - You will have to temporarily spin up some special_workers. Too many and we get rated limited by the OpenaI embedding endpoint though. Not sure if we can do more than one. 
- [ ] then sync prod to staging again please


jrochkind:
@eddierubeiz self-merging to move on timeline but please review at your convenience

Issue #3320, opened 2026-03-10
Staff user login


jrochkind:
We want to provide a new feature for all staff/people with @sciencehistory.org accounts. 

So we need some changes to app to do that. 

* [ ] new user type for basic internal user with no admin dashboard access
* [ ] with more user accounts, probably need ability to search users on user admin dashboard
* [ ] set up permissions to allow this new user type to do the AI stuff, may need to consult on what this entails (look at other people's questions?)

PR #3326, opened 2026-03-12
allow basic_internal type users access to ai conversations

jrochkind:
Ref #3320

- let basic_internal users create and see individual AI conversations
- simple homepage link to OH AI conversation if user is logged in

re #3399

- [x] Still need to actually fix access_policy access so non-elevated-permission users have access to PDFs that for external users require a request to be made with automatic approval. 

For now feature flag `FEATURE_AI_FOR_BASIC_INTERNAL=true` to enable -- so we can go ahead and merge.


jrochkind:
Added a feature flag to enable this, so we can go ahead and merge it and only set feature flag in staging -- should have done that originally!

eddierubeiz (approved):

PR #3328, opened 2026-03-13
update rubygems dependencies within current requirements

jrochkind:
Basically just `bundle update`. 

- That revealed a slight incompat in `csl` gem, had to add `./` to local paths to load styles, see https://github.com/inukshuk/csl-ruby/issues/13

- Keeping the JS dependency `vite-plugin-ruby` in sync with the ruby gem vite_rails is very confusing to me. Still don't totally understand what it expects. It complained, and just running `yarn upgrade vite-plugin-ruby` to get the most recent version of that, took care of it. 
  - see also https://github.com/ElMassimo/vite_ruby/issues/592

Getting all gem dependencies up to date in prep to then updating to Rails 8.1!


eddierubeiz:
Thanks for all this work.

PR #3329, opened 2026-03-16
Update Devise to 5.x

jrochkind:
To prepare for Blacklight 8.1 upgrade, and just generally stay up to date. 

- update to devise 5.x
- fix spec to look for new default reset password wording


jrochkind:
@eddierubeiz I'm gonna go ahead and self-merge this trivial one, as it turns out it resolves a security alert.

PR #3391, opened 2026-04-09
Create OralHistoryContent::Paragraphs from the structural representation created from ExtractPdfText

jrochkind:
On top of #3390 

The ExtractedPdfTextParagraphSplitter takes the structural JSON representation created by ExtractPdfText, and turns it into standardized `OralHistoryContent::Paragraph`s that we can use for chunking -- and for OHMS-like display in future PR. 

We add cumbersomely named `pdf_logical_page_number` to `OralHistoryContent::Paragraph` -- now we can keep track of PDF page numbers for chunks for citations, and for markers in a HTML display! 

We test with some new test PDFs -- real OH PDFs taken from very old, mid-era, and new, to represent different practices. We try to scrape timestamps from PDFs that have them, for supporting OHMS-like display. Some mid-era PDFs have inline timestamps like `<T: 12 min>`  and very new ones have  inline `[hh:mm:ss]` on every paragraph, reaching new-style OHMS fidelity. 

We only want to make paragraphs from the interview transcript itself, so some of the tricks of the splitter are recognizing the prefatory material (Transcripts always start on page labelled `1`), and post material. 

We don't currently handle footnotes/endnotes at all, ignoring them in this PR. 

There are other edge case tricks that end up being necessary, making the code a bit squirrely:

 * Joins paragraphs split across page bounds

 * Finds page numbers in standard bottom of page and unusual old top-of-page locations,
   and strips and records in metadata

 * Finds internal timestamps in old `<T: N min>` format and new [hh:mm:ss] format,
   and records timestamp metadata.

 * handles timestamps taht reset after `[END OF AUDIO` markers, re-seqencing
  full-transcript timestamps.


jrochkind:
@eddierubeiz feel free to review if you have itme, but don't merge yet, i found another bug!

-----

Currently when it's trying to join paragraphs around page breaks, it's mis-calculating and merging two utterances and missing the speaker change, as in Glusker:

> BOYTIM: Okay. All right. Well, I think we’re, kind of, reaching the end. Are there any other broader reflections that you wanted to share or anything we didn’t talk about? <PAGE-BREAK next='121'></PAGE-BREAK> [00:02:03] GLUSKER: No, I’d say I’d had a lot of people who worked for me, almost all of my practical, structural life, who did good work to, kind of, keep the lab going, make sure we had good X-ray diffraction data, and so on. And all these people who came, I mean, it was really nice how people would come and see me just once in a while, you know. Max Perutz liked to come, but . . . he’s, kind of, an amusing guy to have. I guess he’s dead now, but . . . .

PR #3393, opened 2026-04-10
Google arts and culture download model

eddierubeiz:
Creates a model (with its migration, as it's an ApplicationRecord) to keep track of downloads to GAC.
Some changes to Shrine setup, as we want to store the download objects in our public bucket.

### DIFF
--- a/app/models/google_arts_and_culture_download.rb
+++ b/app/models/google_arts_and_culture_download.rb
@@ -1,22 +1,9 @@
-# Represents one of of our large whole-work derivatives (zip of images, or PDF of images),
-# that are created on-demand.
-#
-# Represents the in_progress/success/error state of creation on demand.
-#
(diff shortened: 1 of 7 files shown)


jrochkind:
Code looks great!

The point of making it "belong" to a user is so you know who to email when it's done?

I am wondering if at this stage it might be better to put who to email when done in a config variable (possibly more than one people), and just have them all effectively be available to all staff, and not keep track of who launched an export. 

Thoughts on that?  

Otherwise, as far as code itself, approved.

eddierubeiz:
I thought I commented on this when merging it; not sure what happened to my initial comment.
I would like to retain the association with the requesting user as a form of records management. These exports can easily get expensive in terms of server time and s3 storage space, so I think it won't hurt to remember to who asked for a report, when and for what purpose.

Issue #3346, opened 2026-03-23
remove CC license from OH works


archivistsarah:
Dave Caruso has let us know that a CC license (any CC license) conflicts with the actual usage restrictions and guidelines for oral histories, often conflicting with information available in the oral histories and their metadata. 

So, for every work that belongs to the Oral History department and has Science History Institute listed as the rightsholder, we want to flip the rights field from a CC license to instead be “In Copyright.”
 
This does not change the fact that we were releasing these files under a CC license for a period of time, so anyone who downloaded or access them before now (or whenever we make the flip) can still distribute them under the terms of the license. But going forward, use of these will be determined solely by copyright law, with permission required from us for all uses that are not Fair Use.

eddierubeiz:
Create rake task; run on staging; examine any edge cases. check with Sarah N..

eddierubeiz:
@archivistsarah  - this is done in staging; I changed 601 oral histories. Compare for instance:

- before: https://digital.sciencehistory.org/works/iv46ri3
- after: https://staging-digital.sciencehistory.org/works/iv46ri3

archivistsarah:
that's great!
what accounts for the 168 oral histories that weren't changed?

eddierubeiz:
Ah, glad you asked:

```
select count(*), json_attributes ->> 'rights_holder' as rights from kithe_models where json_attributes -> 'genre' ?  'Oral histories' group by rights;
 count |                                        rights
-------+---------------------------------------------------------------------------------------
    51 |
     7 | Rice University’s Baker Institute for Public Policy and the Science History Institute
   601 | Science History Institute
   127 | Science History Institute. Additional legal restrictions set by the interviewee.
(4 rows)
```

eddierubeiz:
1) I assume if the rights holder metadata is blank we shouldn't touch anything.

2) what if we share ownership with the interviewee or with some other party? Presumably that's even more of a reason to retract our Creative Commons claim.

@archivistsarah, in that case I can change the recipe to: 

  * If
      * the work is an oral history
      * and the rightsholder is **or includes** "Science History Institute",
      * and the rights field is currently CC:

then set the rights field to “In Copyright.”

archivistsarah:
I didn't see this in your code but just to be sure --  are you filtering for works with the CC license or this is just all of the rightsholders in OH works?

eddierubeiz:
Not sure I understand your question; we can discuss in person if needed.

[...]

PR #3394, opened 2026-04-10
rake task to update OH rights

eddierubeiz:
Ref #3346
If it's an oral history and rightsholder == Science History Institute  then set the rights field to “In Copyright”.
- [x] Run in staging
- [x] Merge
- [x] Run in prod


jrochkind:
Code looks fine, approved, once the outcome is approved by relevant stakeholders.

jrochkind (approved):

jrochkind:
That has become a lot of things to put in the Job class -- I often like to extract that to a Service class, because what if you want to call the functionality not as a bg job?  Although i guess technically you can always do `perform_now`. 

Last time i looked i thought of this, but there were only a few small methods in Jobs class so I just let it go. 

Your call though. Otherwise approved.

eddierubeiz:
[List of works changed in production.txt](https://github.com/user-attachments/files/26898260/Production.-.files.changed.dry.run.txt).  With each work I list its *old* rights statement, just in case.

Issue #2903, opened 2025-02-24
Create exports in standard format for Google Arts & Culture


jrochkind:
If we can export in their standard format, then we can import it ourselves in their website, and do it whenever we want -- this DOES work to update existing imports too!  

So we should talk to our contact (Ilya?) to get what the format looks like, and do this! Before next time we want to update ideally. 

When we get to this -- we can investigate requirements/desires for what staff UI would look like to select items for export/import.

eddierubeiz:
Our contact is
Vilius Klakauskas
vilius@xwf.google.com

eddierubeiz:
Talked to Clare yesterday; what she would find most helpful would be an informal way to export arbitrary works (most often: all works since the last export, but also, a specific set of works that would be helpful for a particular project.)

jrochkind:
> all works since the last export

If our understanding is correct that you can replace existing works with new/updated metadata etc, would it make sense to consider exporting all works, not just since the last export, to update any metadata we may have improved since then, which does happen?

cfhirai:
Hello! 

This is the metadata for a group of the items that were imported through the LSDP program (via Vil). To do a new import the same thing using the Advanced Import tool, a CSV would be set up like these Google Sheets (you have access via your sciencehistory.org Google accounts): 

Advanced Import Sample - Various Items
https://docs.google.com/spreadsheets/d/1U7kWxecRkTe1Idlts_GyDvcTo5LtTjnm__SI2Sq3mt4/edit?usp=sharing

And then it would be loaded into the dashboard (as a Google Sheet) after uploading the image files.

If any metadata were to change, we’d update it on the Google Sheet and import the sheet again. 


Here are more sheet examples. These were created manually for exhibits: 

Varian A-60 
https://docs.google.com/spreadsheets/d/1o4l2OYE3vQ3gukzlMtxm0gkymRyUTR4pJrQgDoLYZZQ/edit?usp=sharing

Monsanto AstroTurf Catalog and others
https://docs.google.com/spreadsheets/d/15yUEUlvSLhr7WROu5efNtHr7BVq4yUbux21WdsJGyJc/edit?usp=sharing

PR #3397, opened 2026-04-14
Google Arts and Culture download creator job, with spec

eddierubeiz:
Ref https://github.com/sciencehistory/scihist_digicoll/issues/2903

A job to create a GAC download object, along with its test. Uses files and metadata from the exporter classes, so this is very simple.

All this job does is zip up the exporter's files, update the download object with progress info so the user can follow along, and upload the resulting zip file to s3 so a user can download it.

So far there's no code to actually run this job.


jrochkind (approved):

jrochkind:
Do you know how long this generally takes as a bg job?  Sadly, if it’s like hours it has a pretty high chance of getting interupted. Like whenever we deploy, and workers get restarted.

Issue #3360, opened 2026-03-27
user generated thumbnail images for video works + video camera icon overlay


archivistsarah:
The thumbnail images of video works (and only the thumbnails, not the poster frames) currently receive a little icon/overlay of a film camera to clearly indicate to users that these thumbnails represent moving images. We recently added the ability for users to create their own thumbnail image files ( #3232 ) but these user generated thumbnails are not displaying with the camera icon when they appear in search results, collection pages, or related works (basically anywhere the thumbnail is used). 

The first work listed below has a user generated thumbnail, the second has a system generated one.

![Image](https://github.com/user-attachments/assets/21f45e90-3e6e-4603-8e6e-0ceb637278e4)

And here's an example of where it shows up on a related work link: https://digital.sciencehistory.org/works/e341hri

PR #3412, opened 2026-04-24
DRY check to determine if a work is a video. Always use work metadata.

eddierubeiz:
Video representatives can be, and often are, an image: the poster frame.

In various parts of the UI, though, we were still using the representative's content type to distinguish whether a work was a video. Hence [bug](https://digital.sciencehistory.org/catalog?f%5Bformat_facet%5D%5B%5D=Moving+Image) ref #3360

Instead let's consistently base the distinction on the work's `format` metadata.

@jrochkind: NB the checks got repetitive enough that it made sense to DRY and encapsulate the check in a new shorthand method `Work.has_genre_moving_image?`, following the pattern of `Work.is_oral_history?`.

See also #3232


jrochkind:
Great idea, thanks!

I was worried it might actually check `members` to see if there's a video asset, which could be expensive if members is not already fetched. But it does not, it seems replacing non-DRY logic that also was not. Cool cool. 

To avoid confusion and keep everything very obvious, should we actually call it `is_genre_moving_image?` (or `genre_moving_image?` the question mark might make `is_` unnecessary), what do you think?

eddierubeiz:
I would argue that it's actually a good thing that the method title abstracts away the basis of the decision.

Here's the idea: in a UI context, a developer doesn't really want to worry about what specific metadata makes the work a video work. (To cite a precedent, I also like that when I use `some_work.is_oral_history?` I don't have to worry about the fact that it's the work's _genre_ that makes it an oral history). Encapsulating the underlying metadata means a bit less cognitive load on the developer.

jrochkind:
@eddierubeiz OK, we can go with it -- my concern though is that genre including `moving pictures` actually _isn't_ the same thing as the work having or being a video...  it could have that genre but NOT have an Asset with "video/" mime-type and vice versa. 

but actually checking for an Asset like that is too expensive to do in this routine way. 

OK if you think ti's best like this go for it.

eddierubeiz:
Had to think about it and look carefully at how the method was being called but I came around to your opinion.

jrochkind:
It's like intentionally breaking the abstraction, reminding us when we call it, yeah this isn't REALLY checking fully, it's just doing what we can afford, so don't forget that and think it's perfect!

eddierubeiz:
I'll even go further than that: in every single place that we call this method, checking the members would actually be a BAD idea. The only place we want to do a "strong" check (i.e. a check that also verifies that we have a video asset to show the user) is in WorkVideoShowComponent, and that's dealt with separately.

PR #3416, opened 2026-04-27
Additional tweaks to edge cases in parsing paragraphs from OH PDFs

jrochkind:
- can handle pages with no text, skipped in orig page numbering
- trim internal header metadata from internal second interview in document
- handle first numbered page starting at 2


app/services/oral_history/extracted_pdf_text_paragraph_splitter.rb
```diff
@@ -107,7 +113,9 @@ def create_paragraphs(extracted_pdf_text_pages)
         # usually on bottom, sometimes on top
         logical_page_number = extract_and_remove_logical_page_number(page_paragraphs)
 
-        if page_index == first_index
+        last_paragraph = paragraphs.last
+
+        if page_index == first_index || (last_paragraph && last_paragraph.text =~ /[END OF INTEVIEW]/)
```
eddierubeiz:
```suggestion
        if page_index == first_index || (last_paragraph && last_paragraph.text =~ /\[END OF INTEVIEW\]/)
```

eddierubeiz: Otherwise it'll match if the paragraph has any of those capital letters in it.

eddierubeiz (dismissed):

app/services/oral_history/extracted_pdf_text_paragraph_splitter.rb
jrochkind: oh man correct! good catch thank you!!  I wonder how this didn't fail any of my tests.

jrochkind: in fact there's no need to use a regex there at all, changing to include?(str)

PR #3418, opened 2026-04-28
OralHistoryContent::Paragraph serializable round-trip to JSON

jrochkind:
Preparing to calculate and cache paragraphs extracted from PDFs, in order to provide HtML view based on them.

Felt like there shoudl have been a less hacky but still concise way to just like "make it serializable", but this is all I got!  ActiveRecord::Serializable was just a pain.


jrochkind:
oops I should prob use attr_json, duh.  Hm, at least if i'm going to store it in db...

jrochkind:
Replacing with a different approach.

PR #3420, opened 2026-05-01
Move the Google Arts and Culture link in the admin menu to be in the "reporting" section.

eddierubeiz:
No issue reference here; trivial.


eddierubeiz:
Going to self-merge (but not deploy). We have consensus about this and it's generally not a big deal.

PR #3421, opened 2026-05-01
Only run DziPackage::ActiveRecordCallbacks callbacks if the asset is an image

### LINKED ISSUE #3313: DZI file location check (run every time the work is saved) is being applied to audio assets
### Background
We have code that checks image assets, every time they are saved to the database, to make sure their derivatives are in the right spot.

### Summary
Our code is incorrectly running image-specific checks on assets that are _not_ images. Specifically, it is looking for DZI tiles on e.g. audio assets.

### Details
Asset.rb has the following callback:
`kithe_earlier_after_commit DziPackage::ActiveRecordCallbacks` which runs on all assets.

The error message looks like this:

`Altering file without a dzi_manifest_file listed, can't find/delete DZI: zbeqi03`

Assets affected recently:

https://digital.sciencehistory.org/admin/asset_files/m7lv0zl
https://digital.sciencehistory.org/admin/asset_files/1myfdk1
https://digital.sciencehistory.org/admin/asset_files/zbeqi03
https://digital.sciencehistory.org/admin/asset_files/9my1hos
https://digital.sciencehistory.org/admin/asset_files/lmp85i9

eddierubeiz:
closes #3313
Includes a regression test.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -40,7 +40,7 @@ class Asset < Kithe::Asset
     Rails.logger.error("AssetPromotionValidation: Asset `#{friendlier_id}` failed ingest: #{promotion_validation_errors.inspect}")
   end
 
-  kithe_earlier_after_commit DziPackage::ActiveRecordCallbacks, only: [:update, :destroy]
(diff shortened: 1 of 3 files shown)


jrochkind:
nice thanks

PR #3431, opened 2026-05-05
ExtractedPdfTextParagraphSplitter can allow_failure_to_sync and just skip extracting timestamps

jrochkind:
When turning PDF text into paragraphs, to extract timestamps sometimes requires arithamtic to turn internal start-over-at-new-audio-file timestamps into the whole-interview timestamps we want.  We need audio file duration info to do this. 

Prior ot this PR, if we lacked sufficient audio file duration info to do this, we just raised. Now, we can optionally just take it in stride, ignore timeestamps that we can't properly place, and generate a warning about it that we store with the storeed paragraphs in OralHistoryContent. 

This allows us to move forward with the generated paragraphs that can still provide value as an HTML transcript with pagination info, and as chunks for AI with pagination info. 

- ExtractedPdfTextParagraphSplitter can allow_failure_to_sync and just skip extracting timestamps
- ParagraphContainer#create can store warnings
- store pdf paragrpah job takes allow_failure_to_sync param
- store pdf paragraphs rake sends allow_failure_to_sync: true

### DIFF
--- a/app/jobs/oral_history_store_extracted_paragraphs_job.rb
+++ b/app/jobs/oral_history_store_extracted_paragraphs_job.rb
@@ -1,7 +1,7 @@
 class OralHistoryStoreExtractedParagraphsJob < ApplicationJob
 
   # TODO extract this to a method in OralHistoryContent, yeah.
-  def perform(oral_history_content)
-    OralHistoryContent::ParagraphContainer.create(oral_history_content: oral_history_content)
+  def perform(oral_history_content, allow_failure_to_sync: false)
+    OralHistoryContent::ParagraphContainer.create(oral_history_content: oral_history_content, allow_failure_to_sync: allow_failure_to_sync)
   end
 end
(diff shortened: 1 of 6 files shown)


jrochkind:
Ran in production `rake scihist:oral_history:store_pdf_paragraphs`

22 works had no text in PDF, so could not be created. 

4 more were missing audio files so could not be fully sync'd. 

I will email Oral History about all of em. 


### pdfs with no text

```
b3bib1g 
nz806091w
9306t0110
6969z1940
rb68xd255
tb09j665k
v692t711k
dn39x231b
bv73c1590
j67314691
hh63sx03z
tq57ns18
00000101
02870x01
wd375x17
x633f189
70795853
73666555
37720d84
sx61dn21
z029p571
joe4nl9
```

### missing audio files

```
pptbo4m
9sluvqs
f160qwq
lutrzki
```

PR #3436, opened 2026-05-07
Get LLM to use page numbers from PDF paragraphs in citations, and link directly to new HTML transcripts

jrochkind:
This includes:

* including page numbers when giving paragrapsh to claude to summaraize
* Instructing claude to use those page numbers in the citations it returns to us
* Using those page numbers in the on-screen citations
* Knowing that we can link directly to the new derived-from-PDF HTML transcripts, for "really freely available" transcripts.

### DIFF
--- a/app/components/oral_history/ai_conversation_citation_component.html.erb
+++ b/app/components/oral_history/ai_conversation_citation_component.html.erb
@@ -11,7 +11,7 @@
 
 
     <span data-shi-slot="footnote-badges">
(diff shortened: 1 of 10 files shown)


jrochkind:
OK, there's a spec in there now....

I don't like how much data we have to set up and how complex it is to get right, at all!  But that's how it's grown... at least there's a test to demo it now. 

I do like how I used FactoryBot `build` instead of `create` to hopefully avoid saving records to DB and make tests faster.

PR #3447, opened 2026-05-11
Highlight that there are two ways to delete HLS derivatives

### LINKED ISSUE #3300: Documentation: better comments on code pertaining to deleting HLS derivatives.
I want to add some explanatory comments in the following classes:

`ActiveEncodeStatus`
`RefreshActiveEncodeStatusJob`
`VideoHlsUploader`

When you create new HLS derivatives, the old ones are not deleted until after the new ones are complete.

This is great, but we can do a better job explaining how the deletion actually comes about, and in particular I want the comments to emphasize that ~~the deletion depends on a regularly run scheduled Heroku task: `rake scihist:active_encode_status:update`.~~

More details:

- The code (at app/uploaders/video_hls_uploader.rb) that deletes the old HLS files is called in a side-effect of ActiveEncodeStatus#update_asset_on_completed .
- That method runs asset.save! which in turn calls DeleteContainingDirectoryShrinePlugin. That in turn deletes the directory containing the HLS playlist.
- See **https://github.com/sciencehistory/scihist_digicoll/pull/3298#issuecomment-3922450940** .

eddierubeiz:
Closes #3300

There are two ways we delete HLS derivatives. Both differ from our regular workflows:

- Normally the deletion is "backgrounded" via a Shrine plugin.
- Orphaned HLS derivatives associated with deleted assets are also removed by `lib/tasks/active_encode_status.rake`, a Rake task called by Heroku (whose main function is to update existing assets with AWS status info re: their new derivatives.)

Next time we need to create hundreds of new HLS derivatives (which of course involves deleting  hundreds of old HLS derivatives) this documentation will come in handy.


jrochkind:
Thank you!

jrochkind (approved):

Issue #3290, opened 2026-02-06
Links to a particular starting point in the audio do not work in mobile browsers


eddierubeiz:
# Use case: 
First friday: [Valentine's day. Themes: wine, roses, chocolate.](https://www.sciencehistory.org/visit/events/wine-roses-and-chocolate-how-romance-and-science-work-together-to-sweeten-the-dark-days-of-february/)

You want to create a QR code that plays an OH transcript from a particular time offset on visitors' cell phone browsers. (More specifically, you want a visitor to follow the link in a QR code from their phone and hear Battista's [chocolate anecdote](https://digital.sciencehistory.org/works/2z10wr24q#p=51&t=1917&tab=ohTranscript) when they push ▶️ .)

 (We do do this for exhibitions programs; it is an important use case)

# Description

- Take the following URL:
  - https://digital.sciencehistory.org/works/2z10wr24q#p=51&t=1917&tab=ohTranscript
- Visit the URL
- Push the play button (do not click on anything else before pushing the play button, as this alters the scenario).

### Expected behavior:
Audio plays, starting 1917 seconds from the beginning of the interview.
### Actual behavior:
As expected.

### Expected behavior:
As above.
### Actual behavior:
Audio does play, but it starts at the beginning of the interview.

eddierubeiz:
In BrowserStack, after reproducing the bug (meaning I pushed the play button)  on latest IOS Safari browser, I can still paste this into the "Safari Web Inspector":
```
var html5Media = $("*[data-role=now-playing-container] audio, .show-video video").get(0);
html5Media.currentTime = 1917;
html5Media.play();
```
The audio does in fact start playing from the correct offset.

eddierubeiz:
~~Noting also that the timecode links work.~~
~~So this is specifically a bug with the code that retrieves the `t=1917` param from the URL.~~
(later edit: it's actually that once the user implicitly allows the player to function by interacting with it the problem goes awah.)

eddierubeiz:
Yesterday I got browserstack working again (using Chrome; don't bother with iOS Safari.) and reproduced the bug.

eddierubeiz:
Analysis of the problem:

[...]

PR #3451, opened 2026-05-12
[WIP] Use media fragments in <source src> urls for deep links into OH audio

eddierubeiz:
# Use case
See Ref #3290

# Context
We want to give the user a convenient way to copy a URL from the user interface and use it to provide a deep link into the OH audio.

# Definitions
For the purposes of this pull request, a URL to an OH is deemed `nice` if:
  - it advances the playhead to a given point in the oral history;
  - it plays from that point when the user clicks the play button;
  - it does so on all browsers, including mobile browsers.
 ## Example
`/works/2z10wr24q?t=1920#tab=ohTranscript` is a `nice` URL.

# Notes
- A `nice` URL does not necessarily make the UI do anything else; it is not obligated, e.g., to open the TOC to the appropriate TOC entry, or scroll to the correct point in the transcript.
- Because Rails **ignores anchors** appended to URLs, the start time in a `nice` URL must be passed to the controller as a parameter, before the # sign, so that the audio can be cued up server-side.

# Goals of this PR
- Change as little as possible;
- Don't break anything that was working before;
- Leave unrelated problems alone, to be fixed in a separate PR;
- Make it possible to manually construct a `nice` URL;
- Wherever appropriate, links to different parts of the OH should be `nice`.

# Controllers changed
- `app/controllers/works_controller.rb` will pass optional parameter `t` to the `WorkOhAudioShowComponent`, if one is provided; the component will then load the page with the audio cued up to the right spot.

# Components changed
- `app/components/work_oh_audio_show_component.rb` takes an optional named parameter, `start_time`, which cues up the audio server side to that start time using a [media fragement](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Media_fragments). 
 - `app/components/oral_history/ohms_index_component.rb` now uses a `nice` URL for `direct_to_segment_link`.
 - `app/components/oral_history/legacy_transcript_component.rb` : we simplify the constructor so it just takes a `work` argument. The transcript and its ID can be easily obtained from the work, and we need the work to construct a `nice` URL.
   - `timecode_content_tag_for_line` now has a `nice` URL.
     - The new `nice` URL is ignored by the JS code that jumps around between timecode links, so that works just as well as it did before. Namely, it works.
   - `app/components/oral_history/downloads_list_component.html.erb` also now has a `nice` URL in links to the start of each segment.

# Presenters changed
 - `app/presenters/combined_audio_derivatives.rb` can be passed an optional start time in its constructor, in which case the `m4a_audio_url` will return a `nice` link.

# Tests changed
 - spec/components/oral_history/legacy_transcript_component_spec.rb
 - spec/system/oral_history_with_audio_spec.rb


jrochkind:
Looks great, thanks!

Still marked draft and WIP, but if it's done it's good with me!

jrochkind:
Does `media fragments` mean a specific thing, and if so are we doing it? If not, maybe just change title of the PR for clarity in the future?

I think the "media fragments" spec is what we were using before, and NO LONGER using? https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Media_fragments

eddierubeiz:
- The phrase `media fragments` does mean a specific thing;
- We have never used that thing before;
- We are using it in this PR.

A media fragment is a URL to a specific file on the Web (the `media file`), which instructs the browser to play or display only a fragment of it, by suffixing the fragment metadata (the `dimension fragments`) to the URL of the media.

So in the line below in the PR, we're directing the browser to play the audio from `t=start_time`.

[ "#{m4a_audio_url_without_start_time}#t=#{start_time}"](https://github.com/sciencehistory/scihist_digicoll/pull/3451/changes#diff-d77635e9f52332896ca3fa0dfed7cc1f183018b94a40077b30649aec36b8c6c1R18)

Referring to the developer.mozilla.org definition:

- The `media file` is the public URL to the combined audio derivative for the OH, namely `{m4a_audio_url_without_start_time}`
- The `temporal dimension fragment` is `#t=#{start_time}`.

jrochkind:
OK thanks!

Whether we were using them before or not, this PR implements query param `?t=start`, rather than anchor fragment `#t=start` which makes the implementation extra not media fragments!

So okay with you if I change the title of this PR from `Use media fragments for deep links into OH audio` (we agree this PR does not do that), to `Switching to query params for deep links into OH audio` (that seems an accurate description?). I will do so, feel free to edit it further!

jrochkind (dismissed):

eddierubeiz:
I think it's important before merging this PR that we agree what it does. :) Happy to discuss in a quick meeting on Slack.

app/presenters/combined_audio_derivatives.rb
```diff
@@ -11,8 +11,16 @@ def initialize(work)
     @work = work
   end
 
-  def m4a_audio_url
-    work&.oral_history_content&.combined_audio_m4a&.url(public:true)
+  def m4a_audio_url(start_time: nil)
```
jrochkind, line 24: I think we need a comment ont his method, explainign it will generate a url directly to the media asset, but with #t= on teh end, whic his Media Fragment spec (we link to url), whihc we count on user-agent taking account of that.

app/components/work_oh_audio_show_component.html.erb
```diff
@@ -28,7 +28,7 @@
         <% if combined_m4a_audio_url.present? && combined_derivatives_up_to_date? %>
           <div class="audio-controls-line">
             <%= render AudioPlayerComponent.new do %>
-                <source src="<%= combined_m4a_audio_url %>" type="audio/mp4">
+              <source src="<%= combined_m4a_audio_url(start_time: start_time) %>" type="audio/mp4">
```
jrochkind, line 45: we need a source comment here saying we're generating a <source> src tag with a #t= media fragment (link to media fragment spec) and we're counting on the BROWSER front-end to auto-advance to this point based on the media fragment in <source src> url. 

We're not entirely sure if our video-js player is helping with this, or if it's purely hte browser -- or if video-js player could break it int he future?  So let's just record that in a comment in case we are debugging later.

jrochkind:
OK approval withdrawn for more changes!

Anotehr one -- I think as of this PR it is _no longer using_ `app/frontend/javascript/audio/timecode_in_anchor.js`, and we should delete this file?  Does that seem right?

Also it woud be good if we had capybara JS test for this behavior -- we may have had it before? Not sure? Can you confirm?  I realize it was only failing on mobile adn we don't test on mobile -- but we still want something testing even on desktop.  Unless it's really not feasible to write that test, which is possible, which crazy audio stuff. Can you look into this, and let us know if we already have such a test (which still passes), or if we can feasibly add one if not?

jrochkind:
@eddierubeiz OK, now that i understand what is going on -- as I said, I was surprised it worked!!

I decided to ask Claude if I should expect it to work -- Claude thinks that it should NOT be expected to work reliably!  Which worries me a bit! 

https://claude.ai/share/f8443d34-fb2b-4f19-808c-af252ffb3d22

Claude, I think, recommends the OLD way we were doing it before this PR, which of course did not reliably work either. 

Curious your thoughts on Claude's answer etc.

jrochkind:
Chat gpt is more positive, but still says it may depend on browser, and it's worried about Firefox. 

https://chatgpt.com/share/6a1730cc-4c40-83ea-ad74-06dca3814723

Have we tested on firefox both mobile and desktop?

jrochkind:
Claude and ChatGPT both seem to recommedn the way we were doing it BEFORE this PR, which we know wasn't reliably working... a bit concerning.

eddierubeiz:
Thanks for the extra research! I strenuously disagree with certain parts of Claude's answer, but absolutely worth documenting which parts and why.

jrochkind:
Both Claude and ChatGPT seem to be saying htis may be unreliable and depend on your browser. 

So at the least, I think we need to make sure we've tested in the full suite of desktop and mobile browsers. 

But I'm also wondering if an alternate approach would be better. Claude is suggesting on mobile iOS, the only thing you can do is way for the user to press "play" themselves, wait for the "play" event, and THEN seek to the time you want.  But that seems like a do-able thing maybe? Thoughts?

```
audio.addEventListener('play', function handler() {
  audio.currentTime = 101;
  audio.removeEventListener('play', handler);
});
```

Let's talk about this again when we have a chance and have thought about it more!

eddierubeiz:
Absolutely.

app/presenters/combined_audio_derivatives.rb
eddierubeiz: Done - good call.

app/components/work_oh_audio_show_component.html.erb
eddierubeiz: Done!

eddierubeiz:
[edited to correct the name of the file I was talking about -- oops!] 


Ah, one clarification I should have made yesterday:

The purpose of this PR is to allow a user to use the "Copy Link Address" feature of the browser to obtain a URL that will work on mobile (as set out in the PR description above.)

This PR does not change JS-powered navigation within an oral history. Granted, `audio/play_at_timecode.js` could use some tweaks, although they do not belong in this PR (see "Goals" above).  This PR does change (much less get rid of) that file, as that behavior works, and always worked, on mobile browsers.

Notes about `audio/play_at_timecode.js`:

When a user clicks on a link with a `*[data-ohms-timestamp-s]` property,
`timecode_in_anchor.js` does two things:
	* advance the playhead to the number of seconds in the js property
	* start the player.

1) In determining where to move the playhead, the javascript code doesn't look at the URL's `href`; that is ignored. Instead it looks at its `data-ohms-timestamp-s`. Since we're not changing any `data-ohms-timestamp-s` info, the JS code is entirely unaffected by our changes in this PR.

2) Because the code is triggered by `$(document).on("click")`, the problem we are trying to solve here (mobile browsers requiring the user to interact with the player before even loading the metadata which allows the playhead to move) does not occur. Because the user makes the first move (clicking).

jrochkind:
Yep, I did understand that!  There is something else we are not connecting on though. Might just have to talk more synchronously.  But I don't believe `audio/timecode_in_anchor.js` implements what you think -- it is only responsible for handling page load for a page with `#t=` in an anchor, implementing the seek to timecode specified in the anchor on page load.  Note the string `ohms-timestamp-s` does not appear in `audio/timecode_in_anchor.js`

The stuff you are talking about is actually implemented in the short and sweet `audio/play_at_timecode.js`! (which does contain the string `ohms-timestamp-s`!)

Comment headings in both files try to explain what they do, but JS file organization has not necessarily been our strong point!

jrochkind:
Oh hey. We actually do already support a version of jump-to-timecode for videos!  You get a url by right-clicking to copy/bookmark on one of the transcript timestamps, to get eg

https://digital.sciencehistory.org/works/4rhsvjv#t=842.38

(That that out, it works!)

So that actually might use the timecode_in_anchor.js code too.

And I guess THAT is probably currently broken in mobile safari too!!! Since it’s the same technique (same code even!) 

I am really thinking we might want to solve this by some of the ways the LLMs suggested -- catching the user pressing play to seek ahead then. I think it's a simpler fix that does not rely on browser seeking based on the media fragment in <source src> url, that I'm not sure I trust all will.

eddierubeiz:
Oh, I'm so sorry - `audio/play_at_timecode.js` was the file I was talking about above, not `audio/timecode_in_anchor.js`! I corrected my comment.

eddierubeiz:
Just commenting on Claude's assertion (see https://claude.ai/share/f8443d34-fb2b-4f19-808c-af252ffb3d22 ) 
Scroll to the bottom and focus in the approach labeled "The reliable alternative", namely:

```  
const audio = document.getElementById('player');
  audio.addEventListener('loadedmetadata', () => {
    audio.currentTime = 101;
  });
```
This approach is also [recommended by ChatGPT
](https://chatgpt.com/share/6a1730cc-4c40-83ea-ad74-06dca3814723).

In my experience, this approach does not work on iOS (Chrome or Safari). I believe the reason is laid out in [this article on developer.apple.com](https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Device-SpecificConsiderations/Device-SpecificConsiderations.html#//apple_ref/doc/uid/TP40009523-CH5-SW1):

"In Safari on iOS (for all devices, including iPad), where the user may be on a cellular network and be charged per data unit, preload and autoplay are disabled. No data is loaded until the user initiates it. This means the JavaScript play() and load() methods are also inactive until the user initiates playback, unless the play() or load() method is triggered by user action. In other words, a user-initiated Play button works, but an onLoad="play()" event does not."

The loading of the metadata that would make it possible to execute `audio.currentTime = 101` cannot happen without a user-initiated action.

It's important to emphasize here that:
* this PR does not change `audio/play_at_timecode.js ` or any other `js` file;
* as far as I can tell, this PR does not or affect the functioning (or malfunctioning) of any front-end code;
* this PR **deliberately leaves in place** the approach recommended above by both ChatGPT and Claude, _even though_  we have demonstrated that it doesn't work on iOS (or at least not before the user interacts with the page) and explained why.

eddierubeiz:
@jrochkind  was able to solve this without using media fragments in #3472 - thanks so much!

PR #3453, opened 2026-05-12
When AI answer links to PDF, specify specific page destination in anchor

### LINKED ISSUE #3443: When linking from AI answer to PDF, link to specific page
Requires storing the metadata on which PDF physical page is logical page 1, for offeset. We sort of get this info in PdfPageSplitter, but need to pay attention to where it's recorded so it can be used again.

jrochkind:
Closes #3443 

Most browsers when displaying a PDF can go to page with `#page=NN` on the end. 

But that's physical PDF page, we need the offset from logical page-number-on-page in citation to translate to link. 

We were already calculating that as a necessary part of splitting the PDF into transcript paragraphs (ignoring front matter and labelling with logical marked page number). But we need to expose it from the PdfParagraphSplitter, and save it in the DB as metadata with the paragraphs (in ParagraphContainer). 

Then we just need to make the AiConversationCitationComponent use that when linking directly to a PDF from a ctiation with a page number. 

Except the link is actually to a redirect from DownloadsController and `#fragments` don't make it through redirects (they aren't sent to back-end with HTTP request at all!), so we supply it to DownloadsController as a query param, and the controller then redirects to it as a fragment. 

## After merge and deploy

We need to run a rake task in production to back-fill the offset metadata

- [ ] `heroku run rake scihist:data_fixes:backfill_pdf_page_offset -r production` should take under 5 minutes to run.

### DIFF
--- a/app/components/oral_history/ai_conversation_citation_component.rb
+++ b/app/components/oral_history/ai_conversation_citation_component.rb
@@ -12,19 +12,23 @@ module OralHistory
         # If OHMS, we link directly to work page with anchor to take to specific paragraph
         if citation_item.can_link_to_html_transcript?
           work_path(citation_item.work.friendlier_id, anchor: "p=#{citation_item.paragraph_start}")
-        elsif citation_item.work.oral_history_content.available_by_request_manual_review?
+        elsif citation_item.oral_history_content.available_by_request_manual_review?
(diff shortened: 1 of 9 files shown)


eddierubeiz (approved):
Awesome.

PR #3455, opened 2026-05-14
store source_fingerprint for OH chunks, and facilities for making sure those in db match current source data correctly

jrochkind:
- factor out 'paragraph_source' that we can use duck-typed to source and freshness info
- Each paragraph source can create a source_fingerprint hash which is saved with each chunk
- ChunkValidator can check source_fingerprints for freshness, using fancy performant SQL provided in scope
- ChunkValidator can use already fetched into memory chunks to validate source_fingerprints too
- ChunkValidator only validate source fingerprints for items expected to have chunks at all
- WIP rake create_oral_history_chunks can create only for invalid; our SQL select scope isn't currently working as desired though

## After merge and deploy

- [ ] Re-create all chunks on production, with latest logic, and fingerprint info!  This will cost us maybe $40. 
  - `heroku run -r production ONLY_INVALID=true OVERWRITE_CHUNKS=true ./bin/rake scihist:create_oral_history_chunks`
  - Spin special_workers up to 1 or 2 workers, until all re-created. 
  - Spin special workers down
  - Sync prod to staging, so staging has up to date chunks

Afterwords, running the rake again should queue ZERO jobs, cause `ONLY_INVALID` and they should all be valid!

### DIFF
--- a/app/models/oral_history_content/ohms_xml/legacy_transcript.rb
+++ b/app/models/oral_history_content/ohms_xml/legacy_transcript.rb
@@ -94,6 +94,13 @@ class OralHistoryContent
         @transcript_paragraphs
       end
 
+      # fingerprint our source data, so we can say if a fingerprint isn't fresh cause source data has changed
+      def source_fingerprint
+        @source_fingerprint ||= {
+          "xml_md5" => Digest::MD5.hexdigest(@nokogiri_xml.to_xml)
+        }
+      end
+
       private
 
       # @return [Array<Integer>] array timecodes included in line number ranges, as numbers of seconds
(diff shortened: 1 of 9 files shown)


jrochkind:
`ONLY_INVALID=true  ./bin/rake scihist:create_oral_history_chunks`

Is trying to use fancy SQL for fetch efficiency, but isn't actually working, needs to be fixed.

jrochkind:
may have solved that but needs more testing

jrochkind:
Tested manually a bit more on staging.  Ready for review and merge!

eddierubeiz (approved):

jrochkind:
Drat, created in production... despite all my testing, not sure if they show up as valid with source fingerprints now... testing further.

jrochkind:
ah, I accidentally had not yet deployed to production when I ran chunk re-creation. :(

PR #3464, opened 2026-05-27
Try to use less memory in batch loading for create chunks

jrochkind:
We are loading a lot, unfortunately. Was using too much RAM on standard-1x dyno on heroku. Fetching smaller batches and manual GC'ing may fix that?


jrochkind:
@eddierubeiz self-merging this trivial one that only touches a rake task, to keep moving.

PR #3465, opened 2026-05-27
try to fix the pre-load correctly for batch fetch for create chunks rake

jrochkind:
Fix to didn't go right in #3464. Will test more this time.


jrochkind:
tested on staging this time, sorry

PR #3471, opened 2026-06-02
rename OralHistoryContent#extracted_pdf_paragraphs to #extracted_paragraph_container

jrochkind:
Maybe somewhat less confusing if longer? Returns an OralHistoryContent::ParagraphContainer object which itself has a #paragraphs accessor, so was confusing me. You know what they say about the hard things in CS.


eddierubeiz (approved):

jrochkind:
OK, this of course is an `attr_json` . 

I forgot to move the old data over to new name, causing it to get "lost". It should be easily recoverable though. (Also could be re-created but we don't want to leave the old data there under the old key name, wasting significant space).

PR #3472, opened 2026-06-03
Autoseek on Play for iOS browsers

### LINKED ISSUE #3290: Links to a particular starting point in the audio do not work in mobile browsers

jrochkind:
Closes #3290 , alternative to #3451 

1. Basic realization (thanks @eddierubeiz and LLM's) is that seeking after a `loadedmetadata` event won't work on iOS
    * because iOS may not load metadata until user presses 'play', and/or because iOS won't let you seek after that non-user-initiated event!
    * But we can wait for a 'play' event and then seek there, which will effectively make the audio start playing where we wanted to seek to when user presses play 
  
2. But we needed to switch from trying to use API directly on html5 media element to using API on our video-js player to make this work. (thanks Claude and ChatGPT)
    * We had been trying to use just html5 direct API to keep this code "decoupled" from our use of video-js, but that didn't work here and was possibly always a bad idea
    * The video-js "play" button isn't the real browser html5 "play" button, which added complications or outright impossibilties to both hooking into the 'play' event and being allowed to seek after it! But video-js has it's own API for events and playing and seeking etc, which worked more reliably -- which makes sense in retrospect. 
    * Also switch to video-js API for _seeking_ was necessary, didn't totally understand why, but it works and makes sense ot be consistent -- we use videoJS api now!

3. To use videoJS api, we needed a reference to videoJS player, which was created in a different JS file async to this. 
    * Figuring out how to get that reference without accidentally creating two videoJS objects was another trick
    * There would have been other solutions available, including ES6 import/export, but this seems good to me. 

In process, changed some names of variables/methods and added comments to reduce confusion in some things I found confusing. 

@eddierubeiz Please test and confirm!  

## current "solved" situation

On desktop browsers, it can SEEK on load, but still can’t PLAY on load on chrome, firefox, or safari desktop. It used to play on load but maybe doesn't anywhere anymore? Oh well, it degrades nicely to seeking on load still.

On mobile iOS  it can’t even seek on load, but I have it working to do so on/before play, while still seeking on load on desktop!

## commits

- basic proof of concept
- change js var name from player to playerDomEl to be clear
- change name of JS function setupTimeSeek to seekAndAutoPlay for clarity
- don't double-create videoJS object, wait for other file to create it
- use videoJS API for all seeking, not HTML5 player directly, more reliable when we are using VideoJS

### DIFF
--- a/app/frontend/javascript/audio/timecode_in_anchor.js
+++ b/app/frontend/javascript/audio/timecode_in_anchor.js
@@ -28,11 +28,9 @@ domready(function() {
     }
 
     var playerDomEl = document.querySelector("*[data-role=now-playing-container] audio, .video-player video");
-    const videoJsPlayer = videojs(playerDomEl);
 
-    if (playerDomEl) {
-
-
+    // Another file should actually be creating the videoJSPlayer obj we need, wait for it if needed.
+    onVideoJSSetupFor(playerDomEl, function(videoJsPlayer) {
       // Try to seek and then auto-play. player might not be in state where it can
       // seek yet, if it is not then try to wait and seek when we can.
       if (playerDomEl.readyState >=  playerDomEl.HAVE_METADATA) {
@@ -76,7 +74,7 @@ domready(function() {
           });
         }
       }
-    }
+    });
   } else if (paragraphNumber && hasOhTabs()) {
       // OH transcript paragraph number, need to make sure we've switched to transcript tab,
       // then scroll to element leaving room for navbar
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Confirmed it works! Thanks so much for looking at this!

eddierubeiz (approved):

PR #3474, opened 2026-06-08
rake scihist:data_fixes:migrate_oh_paragraph_data

jrochkind:
Fix data accidentally orphaned in https://github.com/sciencehistory/scihist_digicoll/pull/3471

Verified on staging -- ChunkValidator validates all OH works after this, no more chunk validation failed!

And our extracted paragraphs are back (could have been fairly easily recreated, this is pure derivative data, but wouldn't want to leave the old copy taking up massive space in DB). 

- [x] `heroku run -r production rake scihist:data_fixes:migrate_oh_paragraph_data`

### DIFF
--- a/lib/tasks/data_fixes/migrate_oh_paragraph_data.rake
+++ b/lib/tasks/data_fixes/migrate_oh_paragraph_data.rake
@@ -1,0 +1,24 @@
+namespace :scihist do
+  namespace :data_fixes do
+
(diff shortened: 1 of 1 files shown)


jrochkind:
@eddierubeiz Self-merging cause you're studying and this is tested on staging and just a data migraiton rake task!

PR #3476, opened 2026-06-09
fix logout button to work for all users

jrochkind:
Our new basic users who can't access staff functions never got logout button.


jrochkind:
@eddierubeiz self-reviewing and merging this simple one to get it in before demo.

PR #3479, opened 2026-06-11
Re-model how OH access/request settings are modelled in DB

### LINKED ISSUE #3253: Re-factor how OH access is modelled in DB, to make it easier to distinguish entirely public from entirely suppressed
It is unfortunate that:

* no request func turned on, with a public PDF, means a public oral history
* request func turned on with non-public PDFs means requestable
* no request func turned on with non-public PDFs means completely private and unavailable

So `available_by_request_mode` == `off` just means we don't have request-funct turned on, but this can be true in two diametrically opposed cases, where getting them confused would be disastrous. 

This gets confusing when trying to fetch from the DB by access level, as in #3252 

Not totally sure the right way to refactor it... We could have another `available_by_request_mode` for "super private", but since for _most_ public UI it would _look_ the same, it would still be very easy to get it wrong without realizing it, with still disastrous consequences. 

Might need more complex OH metadata "validating"? And that never ends well either. 

It's a conundrum. 

We may only have ONE OH (?) that uses `available_by_request_mode` `off` for non-public things...  it's hard to find these, so we may have more than one?    I actually can't find the at present!

See https://github.com/sciencehistory/scihist_digicoll/issues/2916 for original use-case. And linked PR for implementation on display.

jrochkind:
Closes #3253 

**Before**: We had `available_by_request_mode` with values `off`, `automatic`, and `manual_review`. 

If request mode was `off` an OH might be "ordinary" public files, OR might be fully embarboed with _no_ public files. The only way to tell the difference was the `published` status of the work and assets, no way to tell _intent_ other than that.  

This was confusing, possibly error--prone in an area (embargo) it's important to get right, _and_ often a performance challenge (had to fetch assets to figure out what was going on with work). 

**Now**, replaced by `availability_mode` with explicit values `direct`, `automatic_request`, `reviewed_request`, and `embargoed`. 

There is a rake task to translate from old model to new. 

As before, we use rails `enum` feature which among other things automatically provides query methods (like`availability_direct?`) and activerecord fetch scopes (like `OralHistoryContent.availability_reviewed_request.count`). 

There was a LOT of code that had to be changed. Will do some QA in staging too, don't totally trust just the tests. 

## Data migration

This PR leaves the old column in DB there but "hides" it from the app. This is necessary because heroku deploy model means there can be some time when old code (which needs old data) and new code is running simultaneous, overlap. 

### On deploy

- [x] run `heroku rake scihist:data_fixes:migrate_oh_availability_mode -r production`

### After deploy

- [ ] Do another PR to actually fully remove the database column, and the code to "hide" it which is no longer necessary. 

## Commits

- add OralHistoryContent#availability_mode to replace #available_by_request_mode
- rake scihist:data_fixes:migrate_oh_availability_mode
- Switch admin OH dashboard from attribute available_by_request_mode to availability_mode
- Move much use of OralHistoryContent#available_by_request_mode to #availability_mode
- another round of moving OralHistoryContent#available_by_request_mode => availability_mode
- Cover attribute available_by_request_mode to raise, then fix what breaks in tests
- add available_by_request_mode to ActiveRecord ignored_columns, which finds some more breakagest of ix
- refuse to chunk embargoed OH, for safety
- Don't show embargoed message unless oral history is actually tagged embargoed

### DIFF
--- a/app/components/oral_history/admin/available_by_request_component.html.erb
+++ b/app/components/oral_history/admin/available_by_request_component.html.erb
@@ -1,51 +1,54 @@
-form-select<div class="card bg-light mb-3">
-  <h2 class="card-header h3">Un-Published Assets Available By Request</h2>
+<div class="card bg-light mb-3">
+  <h2 class="card-header h3">Availability/Request Mode</h2>
   <div class="card-body">
     <div class="text-muted text-small mb-2">
-      <p>Make <b>marked non-published</b> assets be available anyway by request, after the patron fills out a form?</p>
+      <p>Set <b>marked non-published</b> assets to be available anyway by request, <b>or</b> mark this oral history as embargoed.</p>
     </div>
 
-    <% if allow_change_by_request? %>
-      <%= form_with(model: work.oral_history_content!, url: update_oh_available_by_request_admin_work_path(work), method: :put, local: true) do |f|  %>
+    <%= form_with(model: work.oral_history_content!, url: update_oh_available_by_request_admin_work_path(work), method: :put, local: true) do |f|  %>
 
-        <table class="table table-sm">
-          <tr>
-            <td style="width: 2rem">
-              <%= f.radio_button :available_by_request_mode, "off" %>
-            </td>
-            <td>
-              <label for="oral_history_content_available_by_request_mode_off">No</label>
-              <div class="text-muted">Disabled / Off</div>
-              <div class="text-muted">NOT using 'Available by Request' feature</div>
+      <table class="table table-sm">
+        <tr>
+          <td style="width: 2rem">
+            <%= f.radio_button :availability_mode, "direct" %>
+          </td>
+          <td>
+            <label for="oral_history_content_availability_direct">Direct</label>
+            <div class="text-muted">NOT using 'Available by Request' feature</div>
+            <div class="text-muted">Published items are available to public, non-published items are not.</div>
+          </td>
+        </tr>
+        <tr>
+          <td><%= f.radio_button :availability_mode, "automatic_request" %></td>
+          <td>
+            <label for="oral_history_content_availability_automatic_request">Automatic Request</label>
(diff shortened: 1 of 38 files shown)


jrochkind:
Some manual QA

1. deployed to staging
2. migrate date: heroku run rake scihist:data_fixes:migrate_oh_availability_mode
3. reindex: heroku run rake scihist:solr:reindex
4. Check Availabiltiy facet counts match on prod and staging. 
    * (Did actually notice a problem, which I then fixed with another commit)
    * In fact, staging has 8 extra in "Immediate". All others match. 
    * Confirmed these are all non-published works, which are IMPROVED to be listed as non-published but in immediate facet
5. Look at a few OH's in each category, test out request process for each kind of request. All good. 
6. Test out OH AI search, and specifically that linking to OH transcripts works properly for type, sometimes HTML sometimes PDF sometimes Request form. All seems well.

jrochkind:
Admin dashboard change

## Old
Screenshot 2026-06-11 at 4 54 48 PM


## New
Screenshot 2026-06-11 at 4 54 27 PM

jrochkind:
emailed Dave for sign off

jrochkind:
Given the go-ahead by @dcaruso-scihist

eddierubeiz (approved):

jrochkind:
we have two embargoed OH's recorded in staging. For some reason one of them, `Stephanie Bailey` wasn't properly recorded as embargoed after migration. 

No files were available (since they were all not published), it just briefly lacked the proper embargo message, just had an empty list of files. Now fixed! Not a big deal. 

Confirmed string `This oral history is sealed ` only appears in the two OH's now marked embargoed, so I think we only have 2 published embargoed after all.

After fixing that, counts of availabiltiy facet in production match counts in staging -- which matched previous production before migration. So all is well. 

Did notice that changing availabilty didn't auto-reindex work, which is not good. I see how to fix that, will submit PR to fix.

PR #3483, opened 2026-06-16
Changes to schema.rb file

eddierubeiz:
Ref #3484
Under the recently updated version of PostGreSQL, and under Rails 8.1, we're noting that running `db:migrate` results in changes to the `schema.rb` file:
* Table columns are [reordered alphabetically](https://blog.saeloun.com/2025/12/29/rails-sorts-table-columns-by-name-when-dumping-schema/) as of Rails 8.1
* new line at the top: `enable_extension "pg_stat_statements"`

We should decide if we really want "pg_stat_statements" in this file; a bit more research required.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,12 +13,14 @@
 ActiveRecord::Schema[8.1].define(version: 2026_06_09_143410) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
+  enable_extension "pg_stat_statements"
   enable_extension "vector"
 
   # Custom types defined in this database.
   # Note that some types may not work with other database engines. Be careful if changing database.
-  create_enum "oh_availability_mode_type", ["direct", "automatic_request", "reviewed_request", "embargoed"]
   create_enum "available_by_request_mode_type", ["off", "automatic", "manual_review"]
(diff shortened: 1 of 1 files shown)


eddierubeiz:
The `pg_stat_statements` extension is already running in production:
```
scihist-digicoll-production::DATABASE=> SELECT extname FROM pg_extension;
      extname
--------------------
 plpgsql
 vector
 pg_stat_statements
(3 rows)
```

eddierubeiz:
It looks like our latest postgres upgrade on Heroku may have recreated the DB from a template that had the extension; this is fine as far as I'm concerned. @jrochkind, let me know if you have strong feelings about it.

jrochkind:
I am not sure it's really from heroku...  I think our schema.rb only gets generated in dev from dev database, I don't think we have any workflow that would generate it from heroku.    But it may be the same thing with your pg database in dev?

Regardless, I think it does not bother me, fine to have that extension in our schema.rb recording of our db state, I think?

If we didn't want to, I guess you'd just want to drop the etension from your dev database before creating schema.rb I guess.

eddierubeiz:
Also fine with merging this based on a bit more research.

jrochkind:
I don't think any more research is required, thanks for verifying the extension is already on production -- that's all good then.  This one may be related to new version of pg. Doesn't matter. 

Approved!

jrochkind (approved):

Issue #214, opened 2019-07-08
investigate speeding up spec suite


jrochkind:
#213 was not enough. 

We may be able to make solr_wrapper faster at starting/stopping solr (It seems to be waiting/polling to make sure solr is really up, when I'm not sure it needs to, I think solr's own start command now may not return until solr is really up). 

We also could fix the test running to allow you to leave a test solr up and running continuously, samvera style, on your dev box -- while still letting it be auto started if it isn't. 

We also may have slowness unrelated to the test Solr, that needs to be profiled and investigated to see if we can speed it up.

jrochkind:
Use tools to find the slowest spec examples -- I think we have one or two pathologically slow ones -- and see if they can be optimized or eliminated.

jrochkind:
I feel like our tests have gotten even substantially slower. it's currently taking nearly 3 minutes on my macbook; I think it used to be about one and a half. 

We have made it so you can start up solr separate and leave it running, so that isn't included in  total test time, and allows us to profile specs to find the slowest ones independent of that. 

To find slowest tests, you can just run:

    ./bin/rspec --profile

```
Top 10 slowest examples (67.22 seconds, 38.6% of total time):
  Oral history with audio display When you visit a work with audio assets shows audio assets appropriately
    15.48 seconds ./spec/system/oral_history_with_audio_spec.rb:65
  New Work form save, edit, and re-save new work
    11.91 seconds ./spec/system/work/new_work_spec.rb:19
  Public work show page work with complete metadata smoke tests
    7.16 seconds ./spec/system/work_show_spec.rb:50
  Cart and Batch Edit smoke test
    5.59 seconds ./spec/system/batch_edit_spec.rb:35
  Oral History Access Interviewee bio can add an Interviewee Biography
    5.05 seconds ./spec/system/admin/oral_history_interviewee_bio_spec.rb:5
  DownloadDropdownComponent no derivatives existing renders
    4.95 seconds ./spec/components/download_dropdown_component_spec.rb:15
  CatalogController general smoke test with lots of features loads
    4.73 seconds ./spec/system/catalog_search_spec.rb:26
  Oral History with by-request delivery When you visit an OH with protected by request, automatic delivery assets shows the page without error
    4.72 seconds ./spec/system/oral_history_by_request_spec.rb:20
  Oral history with audio display with combined audio and OHMS can use 'jump to text' feature
    3.89 seconds ./spec/system/oral_history_with_audio_spec.rb:234
[...]

PR #3490, opened 2026-06-22
stub actual PDF extraction when not necessary in specs

jrochkind:
Ref #214

Testing PdfToPageImages, actual shell out to `vips` and `pdftotext` for actual pdf extraction is very slow. We want to test the actual shell out at least a little bit to prove it owrks, but in many tests we can mock it instead. 

Saves ~6 seconds in this test file! 

We had Claude help us do this, starting with prompt:

> So these are slow becaues they are shelling out to CLI process for actual PDF extraction. 
>
> I do want to do that for the example on :70. 
>
> But for the others, maybe we coudl use rspec expectations to mock the CLI call.  but still have a test that gives us confidence the right thing is happening?  Can you suggest how we'd do that?  If you need to change the unit under test to allow dependency injeciton so we can mock, that could be one approach.

(It mocked internal methods instead of changing to allow dependency injection of the CLI call-out, which is fine). 

Saves around 6 seconds!

### DIFF
--- a/spec/services/pdf_to_page_images_spec.rb
+++ b/spec/services/pdf_to_page_images_spec.rb
@@ -189,21 +189,27 @@ describe PdfToPageImages do
     end
   end
 
-  describe "#create_assets_for_pages" do
+  # This one is super slow if we don't mock create_asset_for_page individually, test that
+  # itself please.
+  describe "#create_assets_for_pages, mocked" do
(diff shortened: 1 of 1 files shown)


jrochkind:
Oops, doing one more additional beyond what it suggested!  

In the multi-page `create_assets_for_pages`, let's actually stub the whole singular `create_assets_for_page`, which we test individually.

eddierubeiz (approved):

PR #3497, opened 2026-06-22
Update to latest vite 8.x

jrochkind:
Released in march 2026. Used `./bin/vite upgrade` to upgrade Gems and npm modules for vite and vite_ruby/rails.

Then, using vite 8, it was having trouble building, Claude Code diagnosed slightly different manualChunks config necessary in vite 8 (which now uses rollup everywhere), and made it for me.

Hoping this will help with some of the vulnerability alerts on dependencies used prior to this upgrade, that were a pain to fix without this upgrade.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -357,11 +357,11 @@ GEM
     http-form_data (2.3.0)
     httpclient (2.9.0)
       mutex_m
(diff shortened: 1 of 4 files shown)


jrochkind:
verified deployable on staging

PR #3498, opened 2026-06-22
Tweaks to buggy code alerting user of QA autocomplete errors

eddierubeiz:
Ref #2674 
The QA autocomplete JS was actually calling an error handling method. The code was working OK but the error message was being displayed at the bottom of the page. I adjusted it so it shows up in the right place (see screenshot below). Tested in dev and it now works fine.

### DIFF
--- a/app/frontend/javascript/admin/qa_autocomplete.js
+++ b/app/frontend/javascript/admin/qa_autocomplete.js
@@ -83,15 +83,15 @@ function addAutocomplete(element) {
       showNoSuggestionNotice: true,
       serviceUrl: qa_search_url,
 
-
-
       onSearchError: function (query, jqXHR, textStatus, errorThrown) {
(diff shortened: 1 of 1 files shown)


jrochkind:
Makes sense! 

I guess error handling code that is is called very rarely is highly likely to break without us noting, unless we test  it... testing it maybe seems excessive, I dunno. It would be slow. 

I'm worried that the alert is going to be VERY annoying to staff, and really slow them down when the thing is down/broken.  Before they actually weren't bothered too much just to know it wasn't working, I think they said, this may make it worse? Let's be sure?

Maybe there's osme other less likely to fail way to report an the error?  Not sure. 

Either way, can you ALSO output it to `console.log` for developers with browser dev tools window open? Prob before the alert.

eddierubeiz:
Yeah - I was thinking about it on the way in and I don't like the alert either. I'll think of an alternative today.

eddierubeiz:
OK - I was able to make a few changes to the existing code that make things work well enough.

Screenshot 2026-06-23 at 1 22 42 PM

jrochkind (approved):

PR #3506, opened 2026-06-29
Add provenance to Cart export

### LINKED ISSUE #3501: Report on all Rare Books in DC - Provenance
Jim has recently let me know that purchase information should not be public for rare books. To fix this, I would like a report of all [rare books ](https://digital.sciencehistory.org/catalog?f%5Bdepartment_facet%5D%5B%5D=Library&f%5Bgenre_facet%5D%5B%5D=Rare+books), and the provenance field, in the Digital Collections, so I know which records to update. Basically a cart report, but with the following information: Provenance field, title, and link to record. 

Ideally, next week so I can take down the info ASAP. Thanks!

jrochkind:
Added after 'description' because that's where it is in the metadata form, seemed as good as any. So later columns are all bumped.

Closes #3501

### DIFF
--- a/app/serializers/work_cart_serializer.rb
+++ b/app/serializers/work_cart_serializer.rb
@@ -68,6 +68,7 @@ class WorkCartSerializer
       format:                   'Format',
       genre:                    'Genre',
       description:              'Description',
+      provenance:               'Provenance',
       subject:                  'Subject/s',
       series_arrangement:       'Series Arrangement',
       physical_container:       'Physical Container',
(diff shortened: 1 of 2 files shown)


eddierubeiz:
I think you may be able to simplify this.
If `Work.method_defined? :provenance` evaluates to true, you should not need to define the method:

```
 def provenance(work)
    work.provenance
  end
```
The tests should pass exactly the same way.

jrochkind:
Okeydoke!  Rather more magic in here than we need, but it works!

I opted not to format the provenance in the export, just show them exactly whatis entered (in web page, it actually gets split into two sections, just show them what was in the form input field sez me!)

PR #3509, opened 2026-07-01
migrate from yarn (v1) to pnpm

### LINKED ISSUE #3050: look into updating to latest yarn
We are still using yarn 1.x . 

This is very old and unmaintained -- they are up to yarn 4 maybe -- and we sometimes run into problems with it. (Would ~/Library/Cache/yarn have taken up all my disk space if I were using newer yarn?)

But we were scared to update, cause yarn is a pain, and we don't know what will break or not work right. 

But we should look into it, this is becoming untenable. 

Note current active development of yarn is actually in: https://github.com/yarnpkg/berry

The `yarn` repo is just v1.

jrochkind:
Closes #3050 

@eddierubeiz happy to discuss more about this before we merge, although ended up being very straightforward!

We have been using yarn v1.x for a long time -- which is basically abandonware. There are future versions of yarn that changed how they did things, and did not catch on. 

`pnpm` seems to be the most popular cohice to switch to, and vite which we already use likes it (although vite works with many package managers). 

The state of the JS ecosystem has advanced enough that switcing was really simple!  It just works. Including on heroku. (Tested deploying on staging, all is well). 

## changes

* `brew install pnpm` to get pnpm on your dev workstation -- or `brew bundle` if you want to upgrade all your stuff and install it, as it's been added to our `Brewfile`
* Use `pnpm install` where before you used `yarn install`
* The former `yarn.lock` lock file is replaced by new `pnpm-lock.yaml`.  Normally you don't need to edit (or look) at this file, `pnpm install` etc takes care of it, but sometimes one needs to debug etc.


eddierubeiz:
Thanks for doing the research and making the switch. I can't say I ever became really fond of Yarn. Happy to move over.

Issue #3511, opened 2026-07-01
Investigate and recuperate when child work collections do not match parent work collections


jrochkind:
A child work has independently settable collection membership to it's parent work. 

We thought that child work collections and parent work collections should probably match? 

One consequence of them not matching would be that when you go to the collection the parent is in, and "search within the collection", the child will not show up in the seach if it's not in the collection too. 

But when discussing how to calculate metrics for #3165 and what to do with child works, we realized we shoudl check. In fact, it often doesn't match!

Of **2438** child works in our collection, **1269** (52%) have non-matching collections with their parents!

Of these, most (1141) have no collections on child, while parent has one or more. 

Another 118 have a colleciton on child, but parent has an _additional_ collection. 

And only 10 have a collection on child that is NOT on parent. 

## Complete 10  child works who have a collection  collection that parent does not have

Child ID|Child Title|Child Collections|Parent Title|Parent Collections|
--------|-----------|-----------------|------------|------------------|
hm50ts358|Plate No. 1: Medicinal Plants|BOLD|The Book of Health|Modern Books and Journals
736665662|Principal Office and Laboratories, Indianapolis, U.S.A.|Permanent Exhibition|Hand Book of Pharmacy and Therapeutics|Modern Books and Journals; Animals and the Making of Human Vaccines; Downstream; Sherlock Holmes and the Case of the Poisonous Plants
bk128c17q|Microscopic view of a flea|Roy G. Neville Historical Chemical Library; Permanent Exhibition|Micrographia : or, Some physiological descriptions of minute bodies made by magnifying glasses. With observations and inquiries thereupon|Roy G. Neville Historical Chemical Library
f4752h45x|Hercules Thanite + DDT Toxicant, For Better Household Sprays|Records of Hercules Incorporated; Permanent Exhibition|1946 Hercules Advertisements|Records of Hercules Incorporated
cc08hg88v|Crystals of Fats, Butter, Beef & Lard, Representing Butter, Oleomargarine & Butterine|Fats, Butter, Beef, and Lard|Dr. Thomas Taylor's reply to "Science," relating to the crystals of butter, animal fats, and oleomargarine|Modern Books and Journals
bn9997519|Alchemical master and disciple|Roy G. Neville Historical Chemical Library; Permanent Exhibition|Theatrum Chemicum Britannicum|Roy G. Neville Historical Chemical Library
[...]

PR #3512, opened 2026-07-02
Admin dashboard Asset count for a Collection

### LINKED ISSUE #3165: Adding (per-collection) File Count to the Dashboard of the DC
I was asked to provide the number of published and created files in the HMBC collection. This could be a useful feature to have on the dashboard for the DC, especially as I move towards large-scale digitization. 

Ideally, we would be able to filter published vs not published, and if they are in a collection (such as HBMC). I no longer create childworks, so that should not be an issue for future collection, but could be an issue for some records that have been made in the past. 

Eddie suggested " it already says '209 public items, 312 total'; we could have it say something like '209 public works, 309 total; X public files, Y total”

jrochkind:
Closes #3165

Adds method Collection#contained_asset_count, that will do a count. Can do it as all, or as just published. 

* Note it does count ONE level of child work. That's enough for now, I don't think we have anything more than that, keeps the SQL under control. See also #3511 

* If you want both all and published, as we do, that's two invocations of the method, two separate SQL queries. Seems okay performance wise, one query for both would be better, but would be much squirrelier SQL, this is fine. Also if we wanted to fetch for a whole list of Collections at once without an inefficient query per collection, would need fancier SQL. This is fine!

Add a previously absent Collection dashboard admin 'show' page, that only has these statistics on it at present.

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -44,10 +44,11 @@ class Admin::CollectionsController < AdminController
   end
   helper_method :sort_link_maker
 
+  # ADMIN dashboard show
   # GET /collections/1
   # GET /collections/1.json
   def show
-    authorize! :read, @collection
+    authorize! :update, @collection
   end
 
   # GET /collections/new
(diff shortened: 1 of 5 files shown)


jrochkind:
Screenshot 2026-07-02 at 1 12 15 PM

jrochkind:
meh, let's add work count to dashboard too as long as we're at it, to have it all in one place.

jrochkind:
eh, that one doesn't include child works at present lol. Nevermind I'm tired, let's deliver what was asked for for now.

PR #3522, opened 2026-07-09
instrument RAM use in copy_database_to_s3 task

jrochkind:
Code from Claude Code. Our nightly task is routinely going over standard-1x RAM sizes on heroku. It does not seem like we should need so much RAM just to download a file from heroku and upload it to s3. But Claude Code suggests it is behind the scenes multi-threaded in upload_file, and could be doing it... but suggests this instrumentation to confirm.


jrochkind:
@eddierubeiz wanna take a look at this claude-generated one?  Would like to deploy today so I can use it to try to figure out where the RAM is going in the rake task.

PR #3528, opened 2026-07-20
Admin tool to generate QR code for collection url

### LINKED ISSUE #3527: QR Code for Box and Folder Search Results
QR codes that link to specific box and folder 

use case: QR codes for Photo 51 records that are not available for researchers. Instead we hope to give a QR code to researchers to view the items. 

https://digital.sciencehistory.org/collections/1x51leb?q=&sort=&box_id=10&folder_id=15
https://digital.sciencehistory.org/collections/1x51leb?q=&sort=&box_id=10&folder_id=16
https://digital.sciencehistory.org/collections/1x51leb?q=&sort=&box_id=10&folder_id=17

in the future this could help researchers see what has been digitized in a specific folder.

To find the SHI logos go to P:\Operations\Templates & Logos\1_Brand Toolkit

jrochkind:
Closes #3527 potentially

Does not include custom branding yet, I haven't been able to find an example or style guide.  But still includes logic to add in the icon, which does significantly increase complexity, as it requires image compositing! 

* Note that the common way to put a logo in a QR Code seems to be just to turn up error tolerance very high then just smack it on top of the QR code occluding some data. Weird, but there seems to be no good way to do otherwise in ruby or python, and this seems to be what peopel do, as far as I can tell. https://stackoverflow.com/questions/45481990/how-to-insert-logo-in-the-center-of-qrcode-in-python

This code was pretty much all Claude Code generated, including tests! Particularly useful for figuring out ChunkyPNG api for compositing -- that would have taken me forever otherwise. 

I did review code and sometimes tweak it manually (or in furtehr instructions to Claude). 

 

- **add QrCodeCreator service object, can superimpose an icon in middle**
- **admin/collections_controller#generate_qr_code**
- **admin ui for generating collection url qr code**
- **only show box and folder for QR code gen when archives**

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -110,6 +110,11 @@ gem "attr_json", "~> 2.3"
 
 gem 'kithe', "~> 2.19"
 
+# QR code generation (see QrCodeCreator service). chunky_png is already
+# a dependency of rqrcode, but we also use it directly ourselves.
+gem "rqrcode", "~> 3.0"
+gem "chunky_png"
+
 gem "traject", ">= 3.5" # to include support for HTTP basic auth in Solr url
 
 gem 'simple_form', "~> 5.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -191,6 +191,7 @@ GEM
       launchy
     childprocess (5.1.0)
       logger (~> 1.5)
+    chunky_png (1.4.0)
     citeproc (1.1.0)
       date
       forwardable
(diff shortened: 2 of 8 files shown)


jrochkind:
OK we don't need branding -- I may actually remove the code and tests for an icon in the middle of the QR code, to keep things slim!

Claude can always generate it back as easy as it did this time.

jrochkind:
Eh, changed my mind, I'll just use the unused "put an icon in the middle" code in there and tested for now.

PR #3541, opened 2026-07-23
more attempts to keep copy_database_to_s3 RAM use down

jrochkind:
To fit within heroku standard-1x. This is all in conversation with claude, the claude-created instrumentation to measure RAM use is also in there still. Will be removed if it works, or if we give up.

The previous attempt to reduce threads did not actually result in reduced RAM consumption. If this attempt doesn't work either, we'll just give up and increase heroku dyno size

### DIFF
--- a/lib/tasks/copy_database_to_s3.rake
+++ b/lib/tasks/copy_database_to_s3.rake
@@ -48,8 +48,16 @@ namespace :scihist do
       $stdout.puts(msg); $stdout.flush
       Rails.logger.info(msg)
     end
+
+    # AWS upload of very large file allocates so many objects that we are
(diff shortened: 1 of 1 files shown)


eddierubeiz (approved):
Approved - just fix the conflict before merging. Thanks!

jrochkind:
Oh wait, we never deployed this!!!  That actually means I'm not sure if my last attempt worked or not!

I got confused and thought I had deployed it and it hadn't worked. 

Nevertheless, not worth anymore of my time, we'll just raise the dyno size.

jrochkind:
nevermind, i guess giving up on this.

PR #3543, opened 2026-07-27
remove all RAM monitoring instrumentation from rake copy_database_to_s3

jrochkind:
Nothing we/claude could come up with could keep this task within heroku standard-1x RAM limit. 

Copying a large (2.5G) database backup file to S3 in ruby just takes a lot of time. 

MAYBE we could use less RAM installing AWS cli to heroku and using that instead. 

But we'll just bump to heroku-2X for this nightly scheduled task instead.

### DIFF
--- a/lib/tasks/copy_database_to_s3.rake
+++ b/lib/tasks/copy_database_to_s3.rake
@@ -18,42 +18,6 @@ namespace :scihist do
 
   """
   task :copy_database_to_s3 => :environment do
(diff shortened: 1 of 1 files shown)


jrochkind:
Turns out we never actually deployed #3541, so that MIGHT have worked. 

But it's not worth spending any more time on, just increase dyno to standard-2x at this point.

PR #3544, opened 2026-07-29
WorkVideoShowComponent can support work with multiple video files

### LINKED ISSUE #3550: Allow Video show page to handle multiple attached video assets
With a list, and click to switch. Needs to handle transcripts, captions, and bookmarked timecode links taking account of multi file. 

A first step to using the video page for audio too and expected audio multi-file use cases, in #3457

jrochkind:
This ended up much more complicated than anticipated, mostly in the Javascript.  Closes #3550 

Loading new video, captions track, transcript, keeping them all sync'd -- ends up being a lot to keep track of, and video.js ends up having a lot of async race conditions here. Coudl nota have gotten through it without Claude Code, which found and suggested fixes and tested many race conditions. 

Almost all fo this code is by Claude Code, with lots of back and forth with me, tried to keep the code well-organized and readable, did my best. But it's a lot of JS for us. 

Sorry for such a large PR, the individual commits are mostly good readable units if you want to look at them. 

- **first step, can list multiple video files**
- **refactor a bit to prepare for multiple files**
- **make sure to order by position**
- **refactored with much help from claude for future dynamic track switching"**
- **JS to load segments in for multi-segment AV**
- **add "now playing" highlighting to row in list, using same highlight color as in transcript**
- **now playing bar under player**
- **VttTranscriptComponent can take friendlier_id to link to specific timecode in specific segment for multi-segment videos**
- **audio/timecode_in_anchor now for OH audio only, but factor out utility methods to re-use for video**
- **refactor to have loadMedia always set highlighting**
- **handle video anchor links with segment assetID as well as timecode**
- **fix seeking on load after loading different segment**
- **link to separate transcript page with good timecode links**
- **fix :asset_with_faked_file :video factory**
- **video.js don't use aspect ratio if we don't have one**
- **swapping video media swaps transcripts too**
- **correct aspect ratio on media swap**
- **preserve user selected player settings on media file change**

### DIFF
--- a/app/components/oral_history/paragraph_transcript_component.rb
+++ b/app/components/oral_history/paragraph_transcript_component.rb
@@ -60,12 +60,19 @@ module OralHistory
     def render_timestamp_marker(start_seconds)
       content_tag("a",
                   format_ohms_timestamp(start_seconds),
-                  href: "#{base_link}#t=#{start_seconds}",
+                  href: "#{base_link}##{timestamp_anchor(start_seconds)}",
                   class: "ohms-transcript-timestamp default-link-style",
                     # must be formatted exactly the same in JS transcript highlighter
                     # code that searches for it.
                   data: { "ohms_timestamp_s" => "%.3f" % start_seconds.round(3)}
       )
+    end
+
+    # Fragment identifier for a timestamp link. VttTranscriptComponent extends this
+    # to also name the asset (`&a=<id>`), for switching segments of a multi-segment
+    # video work; plain audio oral histories (single combined derivative) don't.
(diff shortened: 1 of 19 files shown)


jrochkind:
Oh also involved a bunch of factory improvement/fixing, around some weird things like what `storage` a factory created shrine file is in. Phew!

eddierubeiz:
Really impressive! Looks great at first glance; will approve for now so you can self-merge at the right time (will be out of town for the rest of the week).

eddierubeiz (approved):

jrochkind:
@eddierubeiz self-merging.

PR #3546, opened 2026-07-29
Vips use: avoid malicious file triggering unintended unsafe loaders

jrochkind:
Recent [rails vulnerability announcement](https://github.com/rails/rails/security/advisories/GHSA-xr9x-r78c-5hrm) drew our attention to a vips vulnerability we could be safer with, against malicious files as inputs to vips. 

All our files are uploaded by staff, so our risk is less, but defense in depth is still wise. 

Vips by default will auto-select a file type specific 'loader' by the _contents_ of the file. Some loaders are from third party libraries and often have insecurities, the vips author trusts them less than vips own loaders. A malicious file could appear disguise itself as a safe file, but then trigger processing by a vulnerable loader, and trigger reveal of file system or even remote code execution. 

(And indeed one of the loaders DID have a vulnerability, which Rails default handling was then vulnerable to, in the recent Rails CVE). 

[Vips provided](https://www.libvips.org/2022/05/28/What's-new-in-8.13.html) a `VIPS_BLOCK_UNTRUSTED` env flag which prevents use of any of these less trusted loaders.  

In this PR, we now use that in many cases where we are processing images. 

However, we also process pdfs, and the pdf loader is considered untrusted.  We have to use pdf loader to process pdf's, we just hope that poppler  (the library which provides the pdf loader) patches it's vulnerabilities. However, in that case, we rewrite our vips commands to _force_ use of pdf loader, so at least a file can't trick it into using some _other_ loader with vulnerabities we do not need to be exposed to. 

Rewriting vips CLI commands was a bit tricky, thanks Claude for help. Claude Code authored most of this code, in dialog with me to try to ensure it was simple and readable code.  

- **use VIPS_BLOCK_UNTRUSTED to prevent malicious file from triggering vulnerable loader that's not for intended file type**
- **read PDFs with vips cli 'pdfload' to prevent auto-selection of unintended vulnerable loader for file**

### DIFF
--- a/app/derivative_transformers/kithe/vips_cli_pdf_to_jpeg.rb
+++ b/app/derivative_transformers/kithe/vips_cli_pdf_to_jpeg.rb
@@ -29,6 +29,11 @@ module Kithe
     class_attribute :vips_thumbnail_command, default: "vipsthumbnail"
     class_attribute :vips_command, default: "vips"
     class_attribute :vips_header_command, default: "vipsheader"
+    class_attribute :pdfinfo_command, default: "pdfinfo"
+
+    # libvips default render resolution for pdfload, used as a ceiling so we never
+    # try to render at higher-than-native resolution.
+    DEFAULT_PDF_DPI = 72.0
 
     attr_reader :max_width, :jpeg_q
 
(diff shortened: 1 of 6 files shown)


jrochkind:
Really vips CLI ought to have an option for "insist upon trusting file extension, raise if sniffed file type does not match filename extension"! That would solve a lot of this.

eddierubeiz (approved):

PR #3569, opened 2026-08-28
service class, python wrapper for asking Gemini for HTR transcripts

eddierubeiz:
No user interface for now.

Calls some new python code, `python_script/gemini_htr.py`, which is a thin wrapper around the actual call to Gemini.

To aid experimentation in dev, this version downloads all the assets before sending them over to Gemini. Under the right conditions, we could of course consider providing Gemini with URLS to the files, so Gemini can obtain the files itself, but I'm not at all convinced this is a good idea in practice.

Code produced with extensive help from both ChatGPT and Google Gemini .

Where the metadata goes: a comparison with other image transcripts in the DC. Note that jsonb keys for this PR are a one-line change.

  | Metadata | Ephemeral? (machine-generated) | Current/proposed home (model + DB column) | Additional jsonb key | Location in code |
  |---|---|---|---|---|
  | Human-written transcript | **No** | `Asset` → `kithe_models.json_attributes` (attr_json's default container — no `container_attribute:` given) | `"transcription"` | `app/models/asset.rb` |
  | HOCR transcript (Tesseract OCR) | **Yes** | `Asset` → `kithe_models.derived_metadata_jsonb` | `"hocr"` | `app/models/asset.rb` |
  | Gemini handwriting transcript (new) | **Yes** | `Asset` → `kithe_models.derived_metadata_jsonb` | `"htr_transcript"` | `app/models/asset.rb` (constant `Asset::HTR_TRANSCRIPT_ATTRIBUTE`) |

Note: we're not discussing audio/video transcripts or oral/history transcripts.


jrochkind:
Overall design is looking good!  And a good break-point for a PR unit, thanks! 

As far as storage in the Asset model -- can you examine (or ask Claude code) where/how OCR and whisper transcripts are being stored, to see if you can make this parallel, and report back?  I think we are getting very hetereogenous and it would be good to work against the forces of entropy and get some consistency going, if we can make this consistent with one of the existing. 
* Although I do see you it's parallel to bredig transcription, maybe that's good enough
* Comment "# see also gemini_htr_transcript below" is wrong method name

The `gemini_handwriting_transcription_service.rb` file is obviously AI-written, and I think over-complicated. Can you spend some time with it simplifying it making it match conventions we already use?  Some things I noted on an initial pass (not necessarily comprehensive):
* That error hiearchy is over-complicated, if you are going to have so many different error classes they should at least all have a common superclass GeminiTranscriptionError or something. We don't need so many prob, but perhaps doesn't hurt if they have a common superclass at least. 
* Things like ` HTR_TRANSCRIPT_ASSET_ATTRIBUTE = :htr_transcript` seem way over-complicated. 
* We don't use `Open3.capture3` for running shell commands, we use TTY::Command
* In general, for AI-authored code I think we should maybe set the expectation for now that that the developer who engaged the agent first reviews as if they were reviewing another developer's code. I think this one can probably be reviewed with an eye for "how can it be simplified", I think the agent has over-complicated in some places. 
* If you were using Claude Desktop, it might be easier to tell the agent this, give it instructions even including some of the ones above and have it do it's thing further. But it's important you review the results of course!

It's hard for me to review the one python shell script, I think it would be good with at least a few comments to guide us not being pythonists. When I had Claude write a python shell wrappers, I made sure to write teh comments myself manually, to make sure I understood the code that was output enough to comment it.

jrochkind:
The `db_log_*` stuff is confusing me. Is there a simpler way to organize the code for same outcome? I think the LLMs may have over-DRY'd.

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,488 @@
...
+      MESSAGE
+      db_log_error(msg)
+      raise InvalidResponseError, msg
+    end
+
+    pages
+  end
+
+  # Attach the transcript of each page to its asset
+  def attach_transcripts!(pages, staged_images:)
```
jrochkind, line 237: i think maybe you can inline this whole thing instead of bhaving it extracted to method

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,488 @@
...
+    asset.update!(Asset::HTR_TRANSCRIPT_ATTRIBUTE => transcript)
+  end
+
+  # The model will often provide notes about the transcription process.
+  # Put these notes in the rails log so we can look at them in production, as needed.
+  def log_model_feedback(data)
+    if data["general_feedback"].present?
+      Rails.logger.info(
+        "Gemini HTR general feedback for work #{work.friendlier_id}: " \
+        "#{data['general_feedback']}"
```
jrochkind, line 267: does this belong in the database with that "request status" stuff so the admin can see it in the dashboard? Or do we still need it at all? or maybe only in debug?  or if we don't ask for it, we will be using less tokens.

jrochkind: if we do want it, I'd give it a name like "gemini_self_explanation" or something.

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,488 @@
...
+    Rack::Mime::MIME_TYPES.key(image_derivative.mime_type) ||
+      raise(
+        UnsupportedImageTypeError,
+        "Unknown MIME type: #{image_derivative.mime_type}"
+      )
+  end
+
+  def db_log_status(status)
+    db_log_save!('status' => status)
+  end
```
jrochkind, line 411: this indirection with these `db_log_` methods and then `work_transcript_requests` is very hard for me to follow, for something that is actually pretty simple?

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,488 @@
...
+  def work_transcript_requests
+    work.public_send(Work::HTR_TRANSCRIPT_REQUEST_ATTRIBUTE) ||
+      work.public_send(:"#{Work::HTR_TRANSCRIPT_REQUEST_ATTRIBUTE}=", {})
+  end
+
+  def db_log
+    @db_log ||= { 'errors' => [], 'status' => "", 'start_time' => nil }
+  end
+
+  def transcript_request_id
```
jrochkind, line 435: if this does stay, add a comment saying what this is and what it's used for

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,325 @@
...
+  def work_eligibility_problems
+    problems = []
+    if eligible_assets.empty?
+      problems << "no usable images were found"
+    end
+    if eligible_assets.count > MAX_FILES_TO_TRANSCRIBE
+      problems  << "we are limiting the number of requested pages to transcribe to #{MAX_FILES_TO_TRANSCRIBE}"
+    end
+    unless work.published?
+      problems  << "this work is not published"
```
jrochkind, line 64: think we talked about removing this, so people can transcribe things before publishing them?

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,325 @@
...
+    end
+    unless public_domain?
+      problems  << "this work is not in the public domain"
```
jrochkind, line 67: sounds like Jahna may be on board with removing this restriction in code for now?  Staff can still use whatever policies they have for deciding when to HTR something.

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,325 @@
...
+  def gemini_client
+    @gemini_client ||= HTTP
+      .headers("x-goog-api-key" => ScihistDigicoll::Env.lookup("gemini_api_key"))
+      .timeout(GEMINI_HTTP_TIMEOUT)
+  end
+
+  def gemini_generate_content_url
+    model = ScihistDigicoll::Env.lookup("gemini_model")
+    "#{GEMINI_API_BASE_URL}/models/#{model}:generateContent"
+  end
```
jrochkind, line 116: this method above is only used once, and I don't think improves readability by making it a separate method instead of just inline? What do you think?

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,325 @@
...
+  # We only keep the current request's log -- not a history of past attempts.
+  def db_log_save!(fields)
+    db_log.merge!(fields)
+    work.htr_transcript_status = db_log
+    work.save!
+  end
+
+  def db_log
+    @db_log ||= { 'errors' => [], 'status' => "", 'start_time' => nil }
+  end
```
jrochkind, line 324: I still think this db_log stuff is over-complicated indirection. Can we simplify it, what do you think? It's confusing to me to figure out what's going on from it, but maybe that's just me.

config/locales/gemini_transcript_prompt.en.yml
```diff
@@ -0,0 +1,30 @@
+en:
```
jrochkind, line 1: I am not sure putting this stuff in I18n is the right way to go. Curious if that was your idea or LLMs idea?  

It doesn't seem like I18n to me, I don't think it's something we would ever have a Spanish or French version or whatever... although honeslty we prob aren't going to translate our app generally, but still use I18n for UI stuff... but this isn't really UI stuff either? Although i do see why someoen would think it was convenient to put it here.   

When itneracting with LLM for OH AI, I put this in an .erb file.  Do we think using consistent patterns here is important for our overall code legibiltiy/maintainability? What do you think?  Do you like putting it in I198n

app/services/gemini_handwriting_transcription_service.rb
```diff
@@ -0,0 +1,325 @@
...
+
+  def eligible_asset?(asset)
+    representative = asset.leaf_representative
+    return false unless representative&.content_type&.start_with?("image/")
+
+    derivatives = representative.file_derivatives
+
+    derivatives[:download_large].present? ||
+      derivatives[:download_full].present?
+  end
```
jrochkind, line 295: I would collapse this into `eligible_assets?` above?  As it's only used there, and isn't going to be used anywhere else, and I feel like it would be more readable liek that?

The LLMs really like having lots of methods. But what do you think, do you find this more readable, or is it just LLM choice?

I think with this LLM generated code, it really is incumbent on us, at least at this stage of our use, to actually review and consider if we actually like what it wrote and find it readable and maintainable, not just to take what it gives us.

PR #3573, opened 2026-08-31
Fix media player timestamp seek bug, found by sarah

jrochkind:
For reasons somewhat mysterious our 'audio' tag changes to 'video' tag after we switch segments, not going to worry about it, but since our seek to timecode code was only looking for 'video' it actually didn't work UNTIL switching segments.

Fixed, with specs for regression, claude helped with specs.


eddierubeiz:
Thank you!

PR #3574, opened 2026-09-01
run 'rake assets:precompile' in Github CI

### LINKED ISSUE #3332: Add a check that `rake assets:precompile` works to CI
Don't necessarily want to add this to our `spec` run, to make it slower every time -- but do want CI doing it?

Upgrading some of our gems just broke this for some reason, and I didn't realize until after merged to master. 

Haven't yet diagnosed.

jrochkind:
Resolves #3332 for now -- not totally convinced this is enough to catch all failures, as we are still running it in test env not production? But one step for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,6 @@ jobs:
             RAILS_ENV=test bundle exec rails db:test:prepare
             pnpm install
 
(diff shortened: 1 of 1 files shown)


jrochkind:
Well, failure is INTERESTING

PR #3579, opened 2026-09-04
Refactor python command adapter to return an array instead of a string

eddierubeiz:
Two small tweaks suggested by chat_gpt, in a side note, while I was working on a separate project:

* Refactors the utility method that prefixes 'uv' to a python command.
  * Before it returned a string;
  * Now it returns an array that can be converted back to method arguments using the splat operator.
Rationale: "Preserves argument boundaries and avoids shell quoting/injection problems", said ChatGPT. Fair enough.

* Changes the call to `which` to fix a small bug.

### DIFF
--- a/app/lib/scihist_digicoll/util.rb
+++ b/app/lib/scihist_digicoll/util.rb
@@ -78,9 +78,9 @@ module ScihistDigicoll
     # it uses uv to install dependencies, it installs them into PATH directly and you
     # don't/can't use `uv run`, a bit inconvenient.
     #
-    # command_array = ScihistDigicoll::Util.python_exec_command("img2pdf")
+    # command_array = ScihistDigicoll::Util.prefix_python_exec_command("img2pdf")
     # TTY::Command.new(printer: :null).run( *command_array, [other args...] )
-    def self.python_exec_command(bare_command)
+    def self.prefix_python_exec_command(bare_command)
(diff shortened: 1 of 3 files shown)


jrochkind:
I think this is an example of the LLMs coming up with things to fix that don't probably matter, but are theoretically good ideas. 

> an array that can be converted back to method arguments using the splat operator.

Do we need to find the uses of this method and convert with splat operator?  Maybe this is already covered by tests so not necessary, but I feel nervous to check we've confirmed any uses are still working.

jrochkind:
As `bare_command` is still being passed in as a string instead of an array, I'm not sure it actually fixes what it means to fix, does it suggest a false confidence? we could go back to callers and fix that, but is this a yak we need to shave?

eddierubeiz:
> Do we need to find the uses of this method and convert with splat operator? Maybe this is already covered by tests so not necessary, but I feel nervous to check we've confirmed any uses are still working.

The method is only called twice in the codebase. I converted both to use the spat operator in this PR. There are no other calls to this method.

jrochkind:
I am pretty confident this is entirely unnecessary deck chair shuffling, the vulnerability it reports is not realistic as all the values we are passing in are fixed string literals and likely to remain so.  But ok! 

Maybe a good lesson that we have to use our judgement with what Claude suggests to "fix" as well.

PR #3585, opened 2026-09-14
When we updated to Rails 8.1, we forgot to update framework defaults. Tests still all pass with 8.1 defaults, and they all seem reasonable, so no problem.

jrochkind:
you can see all the new defaults at https://github.com/rails/rails/blob/3989ebf3473d71e4ceca28154b0b57b5bf22db24/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_1.rb.tt


jrochkind:
We updated to 8.1 back in March btw. https://github.com/sciencehistory/scihist_digicoll/pull/3331

